Issue #606, opened 2014-08-08
add convenience script for running tests locally


jairideout:
Now that the unit/doctest command [is getting longer](https://github.com/biocore/scikit-bio/blob/master/.travis.yml#L22), and we have pep8/flake8/checklist.py checking, it'd be great to have a simple top-level script to run these tests locally (with an option to build the documentation too).

jairideout:
Fixed in #808.

PR #800, opened 2014-12-26
Add task runner bash script

anderspitman:
Provides an abstraction for runnings tests, pep8, etc.

Possible solution for #606. I don't know how I feel about using bash for this. Is there a de facto task runner like gulp or grunt for python?


anderspitman:
@jairideout I have some preliminary feedback on this. As I see it, we basically have 2 goals:

1) Have ability to run everything Travis does locally, ie run all tests across Python environments before pushing.

2) Be able to quickly run tests local without having to remember ``-I DONOTIGNORE`` etc.

After playing with tox a bit, I think it would be a possible solution for 2, but maybe not for 1. The problem is that tox uses pip to install dependencies, and it can take a very long time the first time it runs. I believe this is because pip builds packages like numpy from source. Once the tox virtualenvs are up and running it runs much faster. However, since Travis makes a fresh environment each time, the Travis runs would be very slow. I'm currently trying to find a way to get tox to use conda, which I'm pretty sure install numpy from binaries, but no luck so far.

One thing we could do would be to create a wrapper script (something more rudimentary than what's in this PR) to simplify running our tests, and then just invoke that script from Travis. That would cover 2 pretty well. Then, if we want 1, we could look into actually running Travis locally ([see this issue](https://github.com/travis-ci/travis-ci/issues/2114)). That way we're simulating the CI environment more closely.

Personally, I would prefer to get everything working under tox and use Travis as a light wrapper, but it doesn't look promising so far. I'll keep working on that for the time being though.

Thoughts?

anderspitman:
Upon further investigation, I'm even more convinced that tox is the wrong tool for the job. Fundamentally it's built around pip/virtualenv, and we're using conda (which is far superior IMO) for our environments.

Jorge-C:
For point number 2 I'd probably write a short makefile so that we could run `make test`, etc. Some projects have `runtests.py` files that likely do the same thing.

I agree with the tox shortcomings identified by @anderspitman, but a "conda based tox" would solve our problems right? That's an interesting idea, and google points to [`ctox`](https://github.com/hayd/ctox), "a really rough hack to replicate tox using conda". We can also write an ad hoc script that looks like this:
```
for PY in [2.7, 3.4]
  for NPY in [1.7, 1.9]
    conda create -n hopefully_env_name_doesnt_clash python=PY numpy=NPY...
    pip install scikit-bio
    nosetests
    rm ./build (?)
    conda env --remove -n hope...
```
What do you think would work better (even if it's something else!)? This feature will come quite handy during development :)

anderspitman:
@Jorge-C so I've actually been working on this for a couple days. I just pushed a rough draft for you to look at. Honestly from glancing through the ctox source it looks like basically the same thing: poor man's tox using conda instead of pip/virtualenv.

My script attempts to achieve the 2 goals above while focusing on the following:
* When running under Travis, let Travis handle environments.
* When not running under Travis, handle environments manually (I'm pulling the env info directly out of the .travis.yaml file).
* Simulate separate Travis environments as closely as possible, while requiring reasonable amounts of time and disk space (ie don't download everything fresh or reinstall environments every time unless explicitly desired).
* Allow cached runs if desired (running tests locally after small changes should be fast).

In order to accomplish these things I've done the following:
* A separate conda env is used for each travis env
* A completely separate copy of the skbio source is used for each travis env. This is necessary to avoid conflicts with numpy version when building extensions.
* conda package caching is used.

It's still a pretty rough implementation. A lot of details will probably change, but I think the overall concept and design has potential. I'd like to get feedback on this approach. I'll also take a closer look at ctox. If it does everything we need and seems more stable we should just use that.

jairideout:
Thanks for your work on this @anderspitman, and sorry for the delay in following up. I like the idea of using ctox if possible, though it looks a little rough.

In the meantime, it'd be great to get 2) implemented so that it's easier to run the full test suite locally (albeit in a single environment). I think we'd want a `setup.cfg` file to simplify the nosetests command, and then a simple script (bash, makefile, python script, I don't have strong feelings) to run the unit/doc tests, flake8, checklist.py, etc.

Would you be interested in working on 2) in a separate pull request? That way we could leave this one open for discussion / further review.

ebolyen:
:+1: I agree with @jairideout. Let's continue to explore our options for an easy to deploy local CI system, but for now a script for a single environment is still very useful.

Jorge-C:
@anderspitman I had a quick look at your script and I really like it. If I understand correctly, this gives a way to easily run the same set of tests in local machines and travis (and thus flake8, pep8, etc). Neat! 

One quick question, when you say that "conda package caching is used" do you mean the caching that conda does on its own or are you doing something else on top? Thanks.

anderspitman:
@jairideout I think this is a good way to go. Really fleshing out a fully "local CI" solution such as the last script I pushed has quite a few risks and pitfalls and I think should be approached carefully. Honestly it might even be overkill. I think for local testing we can just have a script to run the relevant commands and let the developer handle the environments. I'll get a basic script put together ASAP.

@Jorge-C Yeah I'm referring to conda's automatic caching where it stores bz2's in the ``pkgs`` directory.

jairideout:
Sounds good, thanks @anderspitman!

anderspitman:
Came across [this really useful thread](https://groups.google.com/a/continuum.io/forum/#!topic/conda/a5fmhLFhaJs) which discussions the options currently available. Should be a useful reference for when we're ready to look at a more full featured environment testing tool.

anderspitman:
@jairideout @ebolyen This could possibly be closed? Unless it's something we wanted to look into sooner rather than later.

jairideout:
Good idea, let's close this since we have another solution in place (#808) and there are higher priority issues to work on for beta.

Issue #760, opened 2014-11-07
Required and recommended dependencies


dansondergaard:
Would it be possible to make scipy and matplotlib non-mandatory dependencies? There are a huge bunch of use cases where none of them are necessary (reading and writing various file formats, e.g. FASTA) and in those cases it seems wasteful to have to install/compile scipy and matplotlib in e.g. Docker images.

There could be required and recommended dependencies, and the packages using an unavailable recommended dependency could just raise an exception.

Thank you for building scikit-bio!
- Dan

dansondergaard:
Unfortunately I don't have the time to look at this. Busy with teaching/work, so if anybody else wants it, it's up for grabs :)

jvkersch:
Just wanted to chime in on this: we're currently using skbio on a headless box and getting matplotlib to run means installing a bunch of otherwise unneeded X libraries. I saw PR #886 and this would help our usecase along quite a bit --- is there anything I can do to help this get in?

dansondergaard:
This is no longer relevant to me so I'll close the issue.

PR #802, opened 2015-01-08
Optional dependencies

dansondergaard:
This PR is a first attempt at fixing #760.

For convenience I have added the optional packages to `extras_require`. I have updated the README to specify that some packages are not installed as dependencies, but should be installed separately.

So far, an `ImportError` is raised, but we could use a custom exception to include more information for the user.

This is my first contribution, so comments are very welcome! :)


Jorge-C:
If we want to keep as much functionality as possible working when not having the extra dependencies installed, a few more changes are necessary.
For instance, in the base.py file for the ordination methods there are top level imports for IPython and matplotlib, and if those aren't installed you'd lose *all* functionality, not just plotting and notebook integration, and I think that's a deal breaker.

So, how do we solve that? Other projects import extra dependencies inside the functions/methods that use them (eg. [pandas and matplotlib](https://github.com/pydata/pandas/blob/53682e3c802f38ef56d69683e0573c8ffae8ecaf/pandas/tools/plotting.py#L2801)). It looks a little bit annoying, but seems cleaner than putting the top level import inside a try block and set a `NO_MATPLOTLIB` variable and check that inside each method/function...

ElDeveloper:
:+1: for avoiding the "NO_MATPLOTLIB" pattern.

dansondergaard:
:+1: for avoiding NO_MATPLOTLIB. I'll have a look at the patch in the weekend.

jairideout:
:+1: to putting imports inside the functions/methods where they're used, and avoiding `NO_MATPLOTLIB`.

It'd be really nice to have a custom exception that states it's an optional dependency, and describes an easy way to obtain all optional dependencies, e.g., `pip install scikit-bio[all]`. Maybe this could subclass `ImportError`?

I think we'd also want the respective unit tests to only be run if the optional dependency is available ([pandas does this](https://github.com/pydata/pandas/blob/d5963dcf31c9fe8529b704efd72306a34eb77953/pandas/util/testing.py#L196)).

ElDeveloper:
Yeah, that makes sense, we also might be able to use decorators?
Something like `@requires_matplotlib`. Maybe not? I'm not sure.

jairideout:
:+1:, that'd be really elegant. Maybe we could have a version of the decorator for skipping unit tests too!

Something like:

```python
from skbio import requires

@requires('matplotlib')
def plot(...):
    ...
```

and 

```python
from skbio.util import skip_if_no

@skip_if_no('matplotlib')
def test_plot(self):
    ...
```

ebolyen:
I really like the decorator idea. As for the testing decorator, that should probably just go on the class.
I think it will lead to better test organization as well.

I think it might even be valuable to inject the import from the decorator itself. 
What it might ultimately look like:
```
@requires('matplotlib.pyplot')
def some_method_on_a_class(plt, self, other, things):
    plt.plot(...)
```
This would prevent the runtime cost of having a try, except block, because the actual content of the function would be determined on load.

ebolyen:
Basically the decorator could act as an `#ifdef`

Jorge-C:
That looks cool!

Can we however keep this PR to implement extra dependencies (ie move the imports to the functions/methods that require them)?

--

A simple decorator like 
```
class wrap(object):
    def __init__(self, wrapped):
        self.w = wrapped
    def __call__(self, *args, **kwargs):
        global plt
        import matplotlib.pyplot as plt
        return self.w(*args, **kwargs)
```
would work but it comes with its sets of issues (namely we lose introspection features, the arguments of the original functions in autocomplete, etc). There are solutions to those problems but they add more complexity so overall I'd prefer that Dan could get in the functionality he needs and then we can discuss the tangential concerns in a new issue. Those would be 1) using a decorator vs repeating imports, which IMO is a stylistical issue; and 2) decorating tests that stress non-required dependencies.

Jorge-C:
Just read @ebolyen's comment and I'm not sure I'm getting it, but I feel it may belong to the new issue I was mentioning (ie, how would the decorator for concert number 1 would actually be defined).

ebolyen:
:+1: for separate issue, we can refactor how we handle imports later. Lets use this PR to identify them.

jairideout:
:+1: for getting the imports moved into their respective functions/methods for now.

ebolyen:
Note: It looks like my initial presumption of a try/except block was incorrect. It might be much simpler to simply have a decorator like `@optional` which just waits for an import error and raises an acceptable error itself.

ebolyen:
Myself and @jairideout just talked about this a bit more. We think a simple importer function may be much better both cognitively and implementation-wise. We can also just lint our codebase with checklist to ensure that no local imports are occurring and to suggest the skbio import function.

It would look like:

```
def some_function(foo):
    plt = skbio.optional_import('matplotlib.pyplot')
    plt.plot(foo)
```

If matplotlib isn't around, the import will raise a helpful exception, otherwise it just imports it as expected.

dansondergaard:
Great to see so much feedback on this!

I like the idea of the importer function and wouldn't mind to work on that later. I agree with @Jorge-C that this PR should just be for moving the imports into the functions :)

Jorge-C:
Any recent progress here @dansondergaard? It would need rebasing on top of master to solve conflicts, and move the then extra imports to the functions that need them. Thank you!

dansondergaard:
@Jorge-C, I got exceedingly busy during the last few weeks of a course I'm teaching, but I'll rebase and move the imports during next week! Sorry for the delay.

dansondergaard:
Apologies for the wait, @Jorge-C. I have now rebased and moved the imports into their respective functions. There are more details in the commit messages.

I am not able to run tests locally since I get the following error (no useful Google results to solve this):

    No module named _ssw_wrapper

However, Travis seems happy :)

anderspitman:
@dansondergaard, for testing locally are you using ``pip install -e .`` to install skbio?

dansondergaard:
@anderspitman, nope, but that helped. Thanks!

jairideout:
Closing in favor of #886, which contains all of the commits in this pull request.

PR #808, opened 2015-01-19
Add task_runner script

anderspitman:
This is a slightly simpler script than originally pushed in #800. I believe this covers all the basic requirements mentioned in item 2) of https://github.com/biocore/scikit-bio/pull/800#issuecomment-69440574

I think it also covers #606.

If this looks good to everyone I'll update the documentation to match.

This does not use the setuptools setup.cfg/nosetests integration procedure mentioned by @corburn and described [here](http://nose.readthedocs.org/en/latest/api/commands.html). This doesn't seem to work inside a conda environment. I've opened a [nose issue](https://github.com/nose-devs/nose/issues/873) to try and resolve this, but I don't think we necessarily need it for now.


task_runner
```diff
@@ -0,0 +1,37 @@
...
+    set -x
+    nosetests ${TARGET} --with-doctest --ignore DO_NOT_IGNORE_ANYTHING
+    pep8 ${TARGET}
+    flake8 ${TARGET}
+    ./checklist.py
+    ;;
+  "docs" )
+    set -x
+    pushd doc
+    make html
```
ebolyen: Once in a while, we need to generate the sphinx source rsts from scratch. Would you mind adding something like a "docs --clean" which would run `make clean` before `make html`?

task_runner
```diff
@@ -0,0 +1,37 @@
...
+  "tests" )
+    set -x
+    nosetests ${TARGET} --with-doctest --ignore DO_NOT_IGNORE_ANYTHING
+    pep8 ${TARGET}
+    flake8 ${TARGET}
+    ./checklist.py
+    ;;
+  "docs" )
+    set -x
+    pushd doc
```
jairideout: This section can be simplified to:

```
make -C doc html
```

task_runner
```diff
@@ -0,0 +1,37 @@
...
+  TARGET=${2}
+fi
+
+COMMAND=${1}
+
+case "${COMMAND}" in 
+  "tests" )
+    set -x
+    nosetests ${TARGET} --with-doctest --ignore DO_NOT_IGNORE_ANYTHING
+    pep8 ${TARGET}
```
jairideout: Can you update the `pep8` and `flake8` commands to also check `setup.py` and `checklist.py`? This check is performed in .travis.yml.

task_runner
```diff
@@ -0,0 +1,37 @@
...
+if [ ${2} ]; then
+  TARGET=${2}
+fi
+
+COMMAND=${1}
+
+case "${COMMAND}" in 
+  "tests" )
+    set -x
+    nosetests ${TARGET} --with-doctest --ignore DO_NOT_IGNORE_ANYTHING
```
jairideout: I think we'll need a way to turn on/off `--with-doctest` and `--with-coverage` (assuming that Travis will be using this). doctests are only run with Python 2.x. Maybe default to running both of these?

task_runner
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
jairideout: This should be `#!/usr/bin/env bash` for portability (see this [SO answer](http://stackoverflow.com/a/10383546/3776794) for details).

jairideout:
Thanks for working on this @anderspitman! I added a few inline comments.

It'd be nice to give the script a more descriptive name than `task_runner`. I don't have any great ideas right now but ideally we could have something that's short and descriptive. Maybe remove the doc-building functionality from this script and rename it to `runtests`? This would also simplify usage slightly: instead of `./task_runner tests`, it'd be `./runtests`.

In #606 I said we'd want a way to build the docs with this task runner, but it's a one-line make command (`make -C doc html`). Maybe it isn't worth having that functionality in this script?

anderspitman:
I implemented most of the requested changes, with a couple deviations:

1) What do you think about ``--with-coverage`` defaulting to off, but ``--with-doctest`` defaulting to on? This seems like the most common setup for developers. Or do we need to run coverage every time locally?

2) I used environment variables as the interface for controlling doctest and coverage. That seems natural for travis and developers shouldn't need to deviate from the defaults often. Does that seems reasonable?

jairideout:
Sorry for the slow reply, been tied up with the upcoming QIIME 1.9.0 release.

> 1) What do you think about --with-coverage defaulting to off, but --with-doctest defaulting to on? This seems like the most common setup for developers. Or do we need to run coverage every time locally?

:+1: for having doctests on by default and coverage off by default.

> 2) I used environment variables as the interface for controlling doctest and coverage. That seems natural for travis and developers shouldn't need to deviate from the defaults often. Does that seems reasonable?

I think that makes sense. It seems like this script could be replaced with a makefile, which can be controlled in a similar way with environment variables. [scikit-learn has a makefile](https://github.com/scikit-learn/scikit-learn/blob/master/Makefile) for this sort of thing. In the future, it'd be easy to add targets for things other than testing, such as doc-building, building the project, release management, etc. This could also be advantageous because developers are generally familiar with Make and we wouldn't have to maintain a custom bash script / command-line interface. Sorry I didn't think of this before -- what are your thoughts on using a makefile here?

cc @ebolyen

anderspitman:
Agree 100%. I was on the fence before about whether to use bash or makefile for #800. I ended up using python after things got more involved handling environments and whatnot. Now that we're going with a simple solution to start with make should be perfect.

jairideout:
Sounds good, thanks for your work on this!

ebolyen:
:+1: for `make`

Jorge-C:
Agree that Python would be best for the more involved work in #800, but a makefile similar to the one from scikit-learn that @jairideout pointed out would be perfect here.

anderspitman:
Finally converted this over to a makefile. The functionality should be the same as the previous script, but implemented in make. The only difference is that if you want to pass a target (ie only run tests in skbio/sequence if that's where you're working), you have to set ``TARGET`` explicitly on the command line:

``make TARGET=skbio/sequence test``

``TARGET`` is not an environment variable. If you wanted to set an env variable as well as target it would be:

``WITH_DOCTEST=FALSE make TARGET=skbio/sequence``

Slightly confusing but I think it's fine since most people will hopefully not need to set the environment variables.

ebolyen:
Thanks @anderspitman! test failures were a result of a pep8 update. I've fixed it and have rerun your tests, so they should pass shortly.

jairideout:
This looks great @anderspitman! Can you please update .travis.yml and the docs to describe how to run the tests?

anderspitman:
Close/reopen to trigger a build, since I'm confused that it failed.

ElDeveloper:
@anderspitman, I restarted the build. However I think the problem is coming from some changes that went into the .travis.yml last week, maybe try pulling from upstream/master?

anderspitman:
I think it worked that time actually. I haven't merged any recent changes from master yet.

anderspitman:
Hm upon closer inspection the docs aren't correct as currently pushed. #844 explains the problem in more detail.

anderspitman:
Hm. I have a branch successfully merged with master, but linting complains about some missing init files. Is that just my branch or a current known issue?

jairideout:
We removed a few subpackages recently, but git will keep these directories around because they contain .pyc files (ignored by git). checklist.py complains about these because they look like subpackages that are missing `__init__.py` files. Can you try deleting those directories and rerunning the linter?

anderspitman:
Sure thing. Also, I know sometimes projects prefer to rebase when merging after master has moved ahead a significant amount. Do we have a strong position on that?

Jorge-C:
I personally prefer rebasing, as my impression is that it leads to cleaner git trees, but we don't have a policy and that hasn't caused any issues so far.

jairideout:
I don't think we've formalized this process yet, though it's something we should definitely start thinking about. We've usually been having contributors rebase their open pull requests if there are merge conflicts and they are easy to fix. Can you try fixing the merge conflicts? I'm happy to help if there are any issues along the way. Since this PR spans many files, we can prioritize merging this to avoid repeated merge conflicts.

ebolyen:
:+1: For whatever is easiest for you. Our git tree isn't exactly a work of art, so I don't think we need to enforce any particular style on contributors.

ElDeveloper:
Yeah that's gonna be a tough one to enforce.

anderspitman:
Ok almost got it ready to go. I went ahead and did a rebase. However, it's complaing about ``skbio/parse`` not having an init file. I looked through that directory and there doesn't seem to be much there, just a ``sequences`` subdirectory with a ``__init__.py``, which is broken because a bunch of the stuff it imports doesn't exists. Is ``skbio/parse`` supposed to exist at all anymore?

jairideout:
`skbio.parse` and `skbio.format` were recently removed since all code within those subpackages had been deprecated for awhile. You can just remove those directories, git shouldn't be tracking anything within them.

anderspitman:
Ah I failed to ``git rm`` those during the rebase. Redid and all is well now.

anderspitman:
This is now fully rebased off master. I also added a couple extra changes:

1) coverage can be specified like ``python -c "import skbio; skbio.test(coverage=True)``. Defaults to false.

2) doctests always run for python 2. python 3 is autodetected and doctests do not run (see #844).

3) The Makefile now uses ``python -c`` internally. This gives the ``skbio.util.TestRunner`` class complete control over how the tests are run. So now the makefile is basically just a very thin wrapper that runs the tests with ``python -c`` and then lints.

anderspitman:
Also please carefully verify my changes to ``.travis.yml``. Don't want to open up any testing holes there.

anderspitman:
Just kidding it didn't push yet...

anderspitman:
Wow that did not do good things to coverage. I'll look in to that...

anderspitman:
Is there an easy way to test coverage locally?

ElDeveloper:
After running the test suite you should be able to run `coverage report -m` and that should give you the coverage report. But I haven't tested this in my system to verify it works with your PR.

anderspitman:
So apparently running ``nose`` via the API using ``python -c "import skbio; skbio.test(coverage=True)", which is what the latest commit does, does strange things with coverage. For example, import lines are not considered as being run.

anderspitman:
I changed back to using the ``nosetests`` command from the Makefile in order to get around the coverage weirdness when using ``python -c``. I think we should look into that issue more, but it doesn't seem to affect the normal tests and doctests (except it might not be running those in the topmost directory, not sure). So since normal users shouldn't need to check coverage, ``python -c`` should work fine.

If this is acceptable I believe this is ready for review.

skbio/tree/Untitled0.ipynb
```diff
@@ -0,0 +1,70 @@
...
+      "print(g)"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": [
+      {
+       "output_type": "stream",
+       "stream": "stdout",
+       "text": [
+        "Zachary's Karate Club\n"
```
ebolyen: How do I join this club? :open_hands:

ElDeveloper: Relevant: http://networkkarate.tumblr.com

anderspitman: It's super elite. You sure you wanna try?

Makefile
```diff
@@ -0,0 +1,25 @@
...
+endif
+
+ifeq ($(WITH_DOCTEST), FALSE)
+	DOCTEST =
+else
+	DOCTEST = --with-doctest
+endif
+
+test:
+	nosetests -v $(TARGET) $(DOCTEST) $(COVERAGE) --ignore DO_NOT_IGNORE_ANYTHING
```
jairideout: Can you remove `-v`?

ebolyen:
It is strange that coverage is decreasing.

jairideout:
Yeah, not sure what's going on with coverage, playing around with this locally. Coveralls is only reporting the two Py2 runs (no Py3). Coverage could have also decreased because `skbio.util.TestRunner` isn't being executed by the test suite.

Makefile
anderspitman: Sure, I originally added it as a lazy way to keep [this comment](https://github.com/biocore/scikit-bio/blame/30357156ec94e74012d653cb566eeff6c3b9af42/doc/source/development/coding_guidelines.rst#L272) accurate. We could add a ``VERBOSE=TRUE`` option to the makefile or edit the comment to not include an example.

anderspitman:
I would probably implicate TestRunner for coverage decreasing, although I'm concerned that the Py3 coverage isn't being reported.

Makefile
jairideout: That makes sense -- I don't have strong feelings either way on this one, so your call.

jairideout:
I think coverage isn't running correctly with `python -c "import skbio; skbio.test(coverage=True)"` because parts of scikit-bio are already imported before coverage is run. Since things like class/method/function definitions are only executed on import, coverage doesn't see those lines as being executed (see this [SO post](http://stackoverflow.com/a/8637093/3776794) for details).

@ebolyen and I chatted about this and had a few potential solutions:

1. As described in the SO post above, find some way to avoid importing scikit-bio functionality until coverage is run (or `coverage.start()` is called). Not sure how feasible this is.
2. Make `skbio/__init__.py` runnable by adding `if __name__ == '__main__':` to the bottom of the file. If executed as a "script", `test` will be called to invoke the unit tests. Coverage can then simply execute the module to run the tests (`coverage run -m skbio.__init__`). I tested this locally and the coverage report looked good, though I haven't verified if this will work with coveralls. One downside here is that the unit tests will be run twice on Travis: once for the actual unit testing, once for coverage. Perhaps this could be split out into a separate build so that things could be run in parallel, though I don't think it's necessary to do that in this PR (we're planning some other CI improvements in the near future, so it could happen then).

    To summarize, `skbio/__init__.py` would have the following at the bottom of the file:
    
    ```python
    test = TestRunner(__file__).test
    if __name__ == '__main__':
        test()
    ```

    And `make test` would run `coverage run -m skbio.__init__` if coverage was enabled.

3. Add a simple Python script under a `scripts` directory that gets installed when scikit-bio is installed (e.g., `test-scikit-bio` or `scikit-bio-tests`). This Python script would basically do what `TestRunner` currently does by running nose/coverage via API calls. That way users will be able to run the tests without requiring the source code, the makefile could call this script as part of `make test`, and Travis can simply run `make test`.

I'm leaning toward option 2 since we won't have to add the notion of scripts to scikit-bio, it's easier to implement, and closest to the implementation and documentation updates you've made in this PR. What do you think?

anderspitman:
@jairideout somehow missed that comment until just now. Sorry about that.

Is it really necessary for coverage to run under ``python -c import skbio;skbio.test()``? That interface is primarily for end users to make sure all the functionality works. Developers can still use the makefile which invokes ``nosetests`` directly. I agree it would be nice to use the ``python -c`` in the makefile, especially since down the road we might be able to run linting via APIs and get rid of the makefile completely, which would be nice for potential windows developers.

That said, I like option 2 except for the tests running twice. Makes it feel like a pretty ugly hack to me. I had considered option 3 before but I agree it would be best to avoid extra scripts.

Makefile
anderspitman: Went ahead and added a verbose option.

jairideout:
> Is it really necessary for coverage to run under `python -c import skbio;skbio.test()`?

It isn't strictly necessary but I'd like to avoid having two ways of discovering and executing tests, since we've run into several issues with getting nose to find and run all of the unit tests and doctests. I'm worried that having two different paths (one in the Makefile for developers/Travis and the other for end users testing a scikit-bio install) may get out-of-sync without us noticing.

I think running coverage as part of `python -c import skbio;skbio.test()` is only necessary if we go with option 3. If we go with option 2, we'd remove the coverage functionality from `skbio.test()` and instead run `coverage run -m skbio.__init__` from within the Makefile (since this only needs to be run during development, but not during end-user installation checking).

I agree that having to run the tests twice (option 2) is less than ideal, though we were already doing this in the past and we may be able to split it into a separate Travis job to reduce impact on test runtimes. 

@ebolyen what are your thoughts on these options?

ebolyen:
I agree with @jairideout that we probably don't want two testing paths. One testing path has already screwed us a couple times, we will never notice if they get out of sync. I actually like option 2. We should rework how travis runs the tests anyways. (e.g. only one instance should make the docs in 2.7, another instance would run coverage in 3, and then linting and testing would happen separately for 2.7 and 3 [that's not needed for this PR]) Our tests would end up a lot faster this way too.

anderspitman:
I agree completely. 2 testing paths is a bad idea. I'll proceed with option 2.

anderspitman:
@jairideout I'm a little confused about why you say the tests will run twice. It seems like if coverage is specified we run ``coverage run -m skbio/__init__``, which runs all the tests and does coverage. Otherwise we run ``python -c "import skbio; skbio.test()"``, which only does the tests. What am I missing here?

jairideout:
That's a really good point! Does running the tests and coverage via `coverage run -m skbio/__init__` fail with a non-zero exit code if any of the unit tests or doctests fail? If it looks like running the tests this way will catch any errors/failures then we should be able to do it in one pass.

anderspitman:
I'm pretty sure it does. I will confirm

anderspitman:
Yep, I double checked that both unit tests and doctests are caught by ``coverage run``. So I think the current commit is ready to go.

skbio/util/_testing.py
```diff
@@ -8,6 +8,53 @@
...
+        self._filename = filename
+        self._test_dir = os.path.dirname(filename)
+
+    def test(self):
+        """Performs the actual running of the tests.
+        """
+        # NOTE: it doesn't seem to matter what the first element of the argv
+        # list is, there just needs to be something there.
+        argv = [self._filename, '-I DO_NOT_IGNORE_ANYTHING']
+        if not self.is_py3():
```
Jorge-C: Can we remove the `is_py3` method and just test `if not PY3:`?

Jorge-C:
I had a minor suggestion, this is great @anderspitman!

skbio/util/_testing.py
anderspitman: We could. I wanted to keep it abstracted since there were a couple different methods of determining the Python version put forward. I don't have strong feelings either way though.

jairideout:
Thanks @anderspitman! I'll review now.

skbio/util/_testing.py
jairideout: No strong feelings on this.

jairideout:
A non-zero exit code isn't reported for either the `coverage run ...` or `python -c` commands in the makefile, so subsequent commands (e.g., `pep8`) are run regardless of failure and Travis won't report actual failures because an exit code of 0 is always returned.

CONTRIBUTING.md
```diff
@@ -128,8 +128,8 @@ Tests can be executed using [nose](https://nose.readthedocs.org/en/latest/) by r
 Note that this is possible because the lines below are added at the end of each `__init__.py` file in the package, so if you add a new module, be sure to include these lines in its `__init__.py`:
 
 ```python
-from numpy.testing import Tester
-test = Tester().test
+from skbio.util import TestRunner
+test = Tester(__file__).test
```
jairideout: `Tester` -> `TestRunner`

Makefile
```diff
@@ -0,0 +1,19 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2013--, scikit-bio development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+ifeq ($(WITH_COVERAGE), TRUE)
+	TEST_COMMAND = coverage run -m skbio/__init__
```
jairideout: This command fails locally for me:

```
$ WITH_COVERAGE=TRUE make test
coverage run -m skbio/__init__
No module named skbio/__init__
make: *** [test] Error 1
```

Changing it to `TEST_COMMAND = coverage run -m skbio.__init__` fixed it for me.

README.rst
```diff
@@ -64,11 +64,14 @@ If you're interested in working with the latest development release of scikit-bi
...
-After this completes, you can run the scikit-bio unit tests as follows. You must first ``cd`` out of the ``scikit-bio`` directory for the tests to pass (here we ``cd`` to the home directory).
-::
+After this completes, you can run the scikit-bio unit tests from with a Python or IPython session as follows::
 
-    cd
-    nosetests --with-doctest skbio
+    import skbio
+    skbio.test()
+
+To only run tests for a specific scikit-bio module, for example sequence, you can do::
```
jairideout: "for example sequence" -> "for example `skbio.sequence`"

doc/source/development/coding_guidelines.rst
```diff
@@ -269,11 +269,11 @@ Some pointers
...
 
-- *Docstrings for testing methods should be considered optional*, instead the description of what the method does should be included in the name itself, therefore the name should be descriptive enough such that when running ``nose -v`` you can immediately see the file and test method that's failing.
+- *Docstrings for testing methods should be considered optional*, instead the description of what the method does should be included in the name itself, therefore the name should be descriptive enough such that when running the tests in verbose mode you can immediately see the file and test method that's failing.
 
 .. code-block:: none
 
-    $ nosetests -v
+    $ VERBOSE=TRUE make test
```
jairideout: Doesn't look like the makefile supports verbose mode.

doc/source/development/new_module.rst
```diff
@@ -24,8 +24,8 @@ like this::
 
   from __future__ import absolute_import, division, print_function
 
-  from numpy.testing import Tester
-  test = Tester().test
+  from skbio.util import TestRunner
+  test = Tester(__file__).test
```
jairideout: Tester -> TestRunner

skbio/diversity/alpha/tests/__init__.py
```diff
@@ -5,3 +5,6 @@
 #
 # The full license is in the file COPYING.txt, distributed with this software.
 # ----------------------------------------------------------------------------
+
+from skbio import TestRunner
```
jairideout: This isn't necessary in test init files, can you please remove?

skbio/alignment/tests/__init__.py
```diff
@@ -5,3 +5,6 @@
 #
 # The full license is in the file COPYING.txt, distributed with this software.
 # ----------------------------------------------------------------------------
+
+from skbio import TestRunner
```
jairideout: This isn't necessary in test init files, can you please remove?

anderspitman:
@jairideout Oh I misread your question from earlier. Didn't realize the exit code was the key piece for Travis, but that makes sense. Sounds like we're back to running the tests twice?

anderspitman:
The rest of the changes shouldn't be a problem.

jairideout:
The tests don't report the correct exit code in either case (via `coverage run` or `python -c`). We might get away with only running the tests once as long as the exit code is reported correctly (not sure if this is possible or not offhand).

skbio/__init__.py
```diff
@@ -79,3 +77,8 @@
     __doc__ = title + art
 else:
     __doc__ = title + art + __doc__
+
+test = TestRunner(__file__).test
+
+if __name__ == '__main__':
+    test()
```
jairideout: I think exiting here with `sys.exit` will take care of the exit code issue. For example:

```python
sys.exit(test())
```

And then `test()` would need to return the exit code from invoking the tests. checklist.py takes a similar approach.

skbio/util/__init__.py
```diff
@@ -60,17 +60,15 @@
...
 # -----------------------------------------------------------------------------
 
-from numpy.testing import Tester
-
 from ._warning import EfficiencyWarning
 from ._exception import TestingUtilError
 from ._misc import (cardinal_to_ordinal, create_dir, find_duplicates, flatten,
                     is_casava_v180_or_later, remove_files, safe_md5)
-from ._testing import get_data_path
+from ._testing import get_data_path, TestRunner
```
jairideout: Since `TestRunner` is a public member, can you please add it to the docstring in this file (above)? I think putting it under `Testing functionality` makes sense.

skbio/util/_testing.py
```diff
@@ -8,6 +8,53 @@
...
+from future.utils import PY3
+
+
+@nottest
+class TestRunner(object):
+    """Simple wrapper class around nosetests functionality.
+
+    Parameters
+    ----------
+    filename : __file__ attribute passed in from the caller. This tells the
```
jairideout: The correct docstring format is to have the type listed after the colon, and the description on an indented line below that. Can you please update?

skbio/util/_testing.py
```diff
@@ -8,6 +8,53 @@
...
+        tester where to start looking for tests.
+
+    Attributes
```
jairideout: These attributes don't exist, I think this section can be removed.

jairideout:
@anderspitman I'm done reviewing. Can you also add a changelog mention about this new functionality? Probably under `Miscellaneous`. Important things to note are the addition of `make test`, `python -c "import skbio; skbio.test()"`, and `skbio.util.TestRunner`.

skbio/__init__.py
anderspitman: Turns out nose conveniently has a ``run_exit`` method which does what you would expect.

Makefile
anderspitman: That's odd. I went ahead and changed it. Either way works in m bash 4.3. Are you on Mac?

jairideout: Yeah

skbio/__init__.py
anderspitman: nose conveniently provides a ``run_exit`` method which does what you would expect. Using it internally in ``TestRunner`` seems to do what we want.

jairideout: :+1:

skbio/util/_testing.py
anderspitman: argv no longer exists, but self._test_dir and self._filename are both used. They are both "private" though. Should I still remove the section?

anderspitman:
Ok I think I've got it all in there. One addition change for review is that I remove the ``is_py3`` method from ``TestRunner`` per @ElDeveloper's recommendation. Seemed like obsessive abstraction disorder in retrospect.

skbio/util/_testing.py
```diff
@@ -8,6 +8,45 @@
...
+        self._filename = filename
+        self._test_dir = os.path.dirname(filename)
+
+    def test(self, verbose=False):
+        """Performs the actual running of the tests.
+        """
+        # NOTE: it doesn't seem to matter what the first element of the argv
+        # list is, there just needs to be something there.
+        argv = [self._filename, '-I DO_NOT_IGNORE_ANYTHING']
+        if not PY3:
```
Jorge-C: Thanks, this looks cleaner :+1:

skbio/util/_testing.py
jairideout: Not sure where to put this comment. In response to your question about documenting private member variables, we generally leave these out and only document things that are public (so this section can be removed).

anderspitman: Yeah sorry that comment was spurious. I was confusing class attributes.

anderspitman:
Is coverage failing just because it went down by 0.02% or is something else wrong?

ElDeveloper:
Yeah, coveralls changed some settings and if there's any decrease in coverage the build will fail.

Yoshiki Vázquez-Baeza

jairideout:
Thanks @anderspitman, I tested locally and everything works great! The coverage decrease is fine because it is only the `-v` line in `TestRunner` that doesn't get executed.

Thanks for all of your work on this -- it's exciting to have an easy, unified way of running the test suite for users, devs, and Travis!

PR #779, opened 2014-11-25
Added heatmap plotting method to Alignment (and tests).

Kleptobismol:
Fixes #765

Based on @gregcaporaso's scikit-bio cookbook recipe:

http://nbviewer.ipython.org/github/biocore/scikit-bio-cookbook/blob/master/Alignment%20viewing%20and%20filtering.ipynb

PR #816, opened 2015-02-11
ENH: add Alignment.heatmap method

### LINKED ISSUE #765: add TabularMSA.heatmap method
Add `heatmap` method to `skbio.alignment.TabularMSA`. @gregcaporaso wrote a function in [this scikit-bio cookbook recipe](http://nbviewer.jupyter.org/github/biocore/scikit-bio-cookbook/blob/master/Alignment%20viewing%20and%20filtering.ipynb) called `msa_to_heatmap`, which can largely be reused here.

Note that this won't be the default `TabularMSA` plotting method. `TabularMSA`'s default plot method will be a [sequence logo](http://schneider.ncifcrf.gov/sequencelogo.html) (#805).

jairideout:
This pull request includes @Kleptobismol's commits in #779. I added a few commits with documentation updates, a few bug fixes, some refactoring, and more unit tests.

Fixes #765.

@gregcaporaso @ebolyen can you please review? @gregcaporaso I'll follow up with a comment on a specific part of the code that I'd like your feedback on.


skbio/alignment/_alignment.py
```diff
@@ -1662,6 +1665,189 @@ def to_phylip(self, map_labels=False, label_prefix=""):
...
+        every third sequence ID are displayed).
+
+        Parameters
+        ----------
+        value_map : dict, collections.defaultdict
+            Dictionary mapping characters in the alignment to numeric values.
+            ``KeyErrors`` are not caught, so all possible values should be in
+            `value_map`, or it should be a ``collections.defaultdict`` which
+            can, for example, default to ``nan``.
+        legend_labels : iterable, optional
```
jairideout: @gregcaporaso does the behavior described for `legend_labels` make sense?

In your original cookbook implementation, all of the values in `value_map` were used to compute the minimum, median, and maximum. This could produce strange behavior if a value in `value_map` wasn't used in the heatmap. For example, if `value_map` had an unused mapping to a maximum value, the "maximum" label wouldn't show up in the legend because a different (smaller) maximum value was used in the heatmap/colorbar.

Also, if a `defaultdict` was supplied, mappings using the default value weren't being considered when computing min/median/max. For example, suppose the `defaultdict` had a default value that happened to be the true maximum, the "maximum" label would be placed in the wrong spot.

jairideout:
@gregcaporaso @ebolyen this is ready for review -- tests are passing now and I manually verified that coverage is at 100% for the code I added (coveralls isn't rerunning for some reason).

ebolyen:
@jairideout Merge conflict.

jairideout:
Fixed, thanks!

skbio/alignment/_alignment.py
gregcaporaso: @jairideout, that makes sense and I see that it has to be that way now that you bring it up. But, I wonder if we should include the values in the legend label as well in this case. So, next to the label for the minimum, you'd always include the min value in parenthesis, and same for mean and max. Otherwise it could be misleading if your labels were *hydrophilic", "medium", and "hydrophobic", and then there were no "hydrophobic" residues in the alignment (I know, very unlikely, but just working from the cookbook example).

gregcaporaso:
Looks good, thanks @jairideout! I just added one suggestion based on your question - if that doesn't make sense we can discuss tomorrow.

skbio/alignment/_alignment.py
```diff
@@ -1405,6 +1406,194 @@ def sequence_length(self):
...
+            mapping are included when computing the minimum, median, and
+            maximum. ``nan`` values are ignored.
+        fig_size : tuple, optional
+            Size of the figure in inches. If ``None``, defaults to figure size
+            specified in the user's matplotlibrc file.
+        cmap : matplotlib colormap, optional
+            See here for choices:
+            http://matplotlib.org/examples/color/colormaps_reference.html
+            If ``None``, defaults to the colormap specified in the user's
+            matplotlibrc file.
```
gregcaporaso: What's the default if it's not specified in the user's matplotlibrc?

gregcaporaso: With any luck, we'll have a better default in the future as mpl is discussing a change to the default color map (matplotlib/matplotlib#875). Anyone want to weigh in on that discussion?

skbio/alignment/_alignment.py
jairideout: That's a good idea, I'll add that in. Since these labels are pretty specialized, I think a better default is to *not* include these labels on the legend at all (i.e., `legend_labels=None`). If someone wants to mark the min/median/max, then they have the option to, but then we're not forcing users to label the legend this way. Does that make sense?

skbio/alignment/_alignment.py
jairideout: This will fall back to whatever matplotlib's default is (there is always a "base" config file included in a matplotlib install, similar to how QIIME config files work). I don't think listing this here is appropriate because matplotlib's defaults may change and we won't want to update these docs when that happens.

jairideout:
Thanks for reviewing @gregcaporaso! I had another question (see inline). In the meantime I'll work on the change you suggested.

skbio/alignment/_alignment.py
gregcaporaso: Yep, agree.

skbio/alignment/_alignment.py
gregcaporaso: Yep, good idea.

jairideout:
@gregcaporaso I made the changes we discussed. Can you and @ebolyen review?

skbio/alignment/tests/test_alignment.py
```diff
@@ -753,6 +754,136 @@ def test_sequence_length(self):
...
+        self.check_heatmap_sanity(
+            fig, ['A', 'A', 'C', 'C', 'G', 'G', 'T'], ['seq4', 'seq2'],
+            ['a (0.0500)', 'b (0.3400)', 'c (1.0700)'])
+        self.assertEqual(fig.get_figwidth(), 42.0)
+        self.assertEqual(fig.get_figheight(), 22.0)
+
+    def test_heatmap_invalid_sequence_order(self):
+        # duplicate ids
+        with self.assertRaises(ValueError):
+            self.a1.heatmap({}, sequence_order=['d1', 'd2', 'd1'])
```
ebolyen: Can you assert some basic verbs in the error message exist to ensure that the correct error is being raised?

skbio/alignment/tests/test_alignment.py
```diff
@@ -753,6 +754,136 @@ def test_sequence_length(self):
...
+
+        # provided set of ids doesn't match alignment's set of ids
+        with self.assertRaises(ValueError):
+            self.a1.heatmap({}, sequence_order=['d2', 'd3', 'd0'])
```
ebolyen: Same as above

skbio/alignment/tests/test_alignment.py
jairideout: Excellent idea!

skbio/alignment/_alignment.py
```diff
@@ -1405,6 +1407,200 @@ def sequence_length(self):
...
+        #     http://matplotlib.org/examples/pylab_examples/
+        #     colorbar_tick_labelling_demo.html
+        fig, ax = plt.subplots()
+        if fig_size is not None:
+            fig.set_size_inches(fig_size)
+
+        cax = ax.imshow(mtx, interpolation='nearest', cmap=cmap)
+
+        sequence_count = self.sequence_count()
+        y_ticks = [0] + list(range(3, sequence_count - 3, 3))
```
ebolyen: This can be changed to `list(range(0, sequence_count - 3, 3))`. Also can you explain why we subtract three from the count with a step of 3? It is not clear *why* we do this.

skbio/alignment/_alignment.py
```diff
@@ -1405,6 +1407,200 @@ def sequence_length(self):
...
+        if sequence_count > 1:
+            # we have more than one sequence, so add the index of the last
+            # sequence
+            y_ticks += [sequence_count - 1]
```
ebolyen: In regard to the above, this will exclude the last "step-able" sequence and then add only the last sequence. i.e.
`sequence_count = 15`
`y_ticks = [0, 3, 6, 9, 14]` not `y_ticks = [0, 3, 6, 9, 12, 14]` which may have been the intended behaviour.

skbio/alignment/tests/test_alignment.py
```diff
@@ -753,6 +754,136 @@ def test_sequence_length(self):
...
+
+    def test_heatmap_missing_character_in_value_map(self):
+        with self.assertRaises(KeyError):
+            self.a1.heatmap({})
+
+    def test_heatmap_invalid_legend_labels(self):
+        value_map = {'A': 0.61, 'C': 1.07, 'T': 0.05, 'G': 0.07, '.': np.nan,
+                     '-': np.nan}
+        with self.assertRaises(ValueError):
```
ebolyen: Same as above for testing basic verbage of the error message.

skbio/alignment/tests/test_alignment.py
```diff
@@ -753,6 +754,136 @@ def test_sequence_length(self):
...
+        # brittle test that may fail depending on matplotlib versions
+        if exp_legend_labels is not None:
+            legend_labels = [e.get_text() for e in axc.get_xticklabels()]
+            self.assertEqual(legend_labels, exp_legend_labels)
+
+    def test_heatmap_empty(self):
+        # no seqs
+        aln = Alignment([])
+        value_map = {'A': 0.1, 'C': 1.5, 'U': -0.42, 'G': 0.55, 'T': 99}
+        with self.assertRaises(AlignmentError):
```
ebolyen: Test the message

skbio/alignment/tests/test_alignment.py
```diff
@@ -753,6 +754,136 @@ def test_sequence_length(self):
...
+            aln.heatmap(value_map)
+
+        # no positions
+        aln = Alignment([DNA(''), RNA('', id="s2")])
+        with self.assertRaises(AlignmentError):
```
ebolyen: Test the message

skbio/alignment/tests/test_alignment.py
```diff
@@ -753,6 +754,136 @@ def test_sequence_length(self):
...
+        # duplicate ids
+        with self.assertRaises(ValueError):
+            self.a1.heatmap({}, sequence_order=['d1', 'd2', 'd1'])
+
+        # provided set of ids doesn't match alignment's set of ids
+        with self.assertRaises(ValueError):
+            self.a1.heatmap({}, sequence_order=['d2', 'd3', 'd0'])
+
+    def test_heatmap_missing_character_in_value_map(self):
+        with self.assertRaises(KeyError):
```
ebolyen: Test the message

jairideout:
Thanks for reviewing @ebolyen! I'm going to hold off on addressing your comments until after the alignment object is refactored (#823) , then I'll update the changes here to work with the new API. I'm marking this as "do not merge" for now.

PR #819, opened 2015-02-12
STY: Fix new pep8 errors from pep8's update

### LINKED ISSUE #818: pep8 has been updated
The carnage is terrible. We need to change a lot of imports to remain compliant now.

ebolyen:
fixes #818

### DIFF
--- a/skbio/alignment/_alignment.py
+++ b/skbio/alignment/_alignment.py
@@ -627,7 +627,9 @@ class SequenceCollection(SkbioObject):
                                           "or fn is provided.")
 
         if ids is not None:
-            fn = lambda _: ids
+            def fn(_):
+                return ids
+
         elif fn is None:
             def fn(_):
                 new_ids = []
--- a/skbio/draw/_distributions.py
+++ b/skbio/draw/_distributions.py
@@ -14,12 +14,12 @@ from itertools import cycle
 import warnings
 
 import numpy as np
-
 from matplotlib import use
-use('Agg', warn=False)
 import matplotlib.pyplot as plt
 from matplotlib.lines import Line2D
 from matplotlib.patches import Polygon, Rectangle
+
+use('Agg', warn=False)
 
 
 def boxplots(distributions, x_values=None, x_tick_labels=None, title=None,
--- a/skbio/draw/_distributions.py
+++ b/skbio/draw/_distributions.py
@@ -14,12 +14,9 @@ from itertools import cycle
 import warnings
 
 import numpy as np
-from matplotlib import use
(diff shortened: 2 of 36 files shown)


skbio/alignment/_alignment.py
```diff
@@ -627,7 +627,9 @@ def update_ids(self, ids=None, fn=None, prefix=""):
                                           "or fn is provided.")
 
         if ids is not None:
-            fn = lambda _: ids
```
wasade: this is frowned upon now?

ebolyen: It would seem so.

jairideout: Yes, pep8 now disallows storing lambdas in variables.

wasade: That's unfortunate.

ElDeveloper: It's ridiculous ...

Jorge-C: :cry:

However, from the PEP8 example, maybe we can keep those super simple functions in one line

> Always use a def statement instead of an assignment statement that binds a lambda expression directly to an identifier. 
Yes:
`def f(x): return 2*x`.

ElDeveloper: That's a really good suggestion!

skbio/draw/_distributions.py
```diff
@@ -14,9 +14,6 @@
 import warnings
 
 import numpy as np
-
-from matplotlib import use
-use('Agg', warn=False)
 import matplotlib.pyplot as plt
```
Jorge-C: I believe the use command *must* run before importing pyplot...

skbio/alignment/_alignment.py
ebolyen: I did not know that was possible. Will update once I get travis back into a working state.

skbio/draw/_distributions.py
ebolyen: Working on a way around this issue now...

ebolyen:
I've built the docs locally and can confirm that the images are still rendered correctly.

skbio/alignment/_alignment.py
jairideout: > However, from the PEP8 example, maybe we can keep those super simple functions in one line

I'm going to merge this so we can get things passing again. Using one line functions in the future sounds like a good idea but I don't think it's too critical at this point. If someone's available to work on this in a separate PR that'd be great!

Jorge-C:
Thank you guys! :100:

PR #856, opened 2015-02-19
MAINT: remove shebang lines from codebase

jairideout:
These aren't necessary and were already partially removed.

### DIFF
--- a/skbio/io/tests/test_clustal.py
+++ b/skbio/io/tests/test_clustal.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 # -----------------------------------------------------------------------------
 # Copyright (c) 2013--, scikit-bio development team.
 #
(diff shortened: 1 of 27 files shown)


jairideout:
Before merging we should verify that the number of tests matches current master.

ebolyen:
:+1: Looks good otherwise, ready for merge when we confirm.

skbio/io/tests/test_clustal.py
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
```
Jorge-C: Actually, the shebang line is useful in any file with a `if __name__ = '__main__'`. It's a matter of taste but it allows running them with `$ foo.py` instead of `$ python foo.py`.

jairideout: nose doesn't like test files to be executable, so even having the shebang won't allows us to do either of these:

```
$ ./foo.py
$ foo.py
```

ebolyen:
Number of tests matched locally.

Issue #835, opened 2015-02-13
Adding in Compositional Data analysis methods


mortonjt:
Composition Data analysis is an alternative technique to analyzing compositional data (aka data involving percentages and proportions).

Most of the OTU data that we are dealing with are inherently compositional, and applying some of these techniques would really ease analysis/interpretation of this sort of data.

As far as I can tell, these techniques have not been implemented anywhere in the Python world.
I've started drafting some of these methods [here](https://github.com/mortonjt/ANCOM/blob/master/ancom/linalg/composition.py).

Also, a bunch of nice visualization tools can be used within this framework, such as [compositional biplots](http://en.wikipedia.org/wiki/Biplot) and [ternary plots](http://en.wikipedia.org/wiki/Ternary_plot)

I realize that scikit-bio is undergoing some infrastructure changes.  Perhaps once the [Contingency Table](https://github.com/biocore/scikit-bio/issues/821) has been added, I can start porting over some of these methods.

Jorge-C:
What are the differences between compositional biplots and plain biplots (ie plotting samples as dots and continuous variables as arrows?

The ordination methods used to have (very barebones) 2D biplots and triplots (when plotting samples and sites as dots, plus variables as arrows) through matplotlib, though they ended up [removed](https://github.com/biocore/scikit-bio/commit/03e11a33b02ad2dd219634a8576314dabfeb0664) and we just added a simple scatter plot later.

I agree that recovering (but mainly improving) bi/triplots would be fantastic, and ternary plots look cool.

:+1:

mortonjt:
Technically they are exactly the same.

The main difference is that the biplots that I'm proposing have a particular log transformation (called centre log transformation) applied to the OTU percent abundance data.   A standard SVD can be applied to this data to obtain PC coordinates, however, a [more correct SVD](http://www.sis2014.it/proceedings/allpapers/2863.pdf) for compositional data for potentially more accurate results.

Jorge-C:
I see, thanks! So they'd be useful for contingency table that have relative abundances, and you'd take the element wise logarithm of the table, and then center it. Is there a standard way to deal with null abundances?

mortonjt:
Exactly!

So zeros are still a problem with this area - there are a couple of common ways about it.
1) [Multiplicative replacement](http://www.leg.ufpr.br/lib/exe/fetch.php/pessoais:abtmartins:dealingwithzerosandmissingvalues.pdf).
[...]

Issue #829, opened 2015-02-12
Convert Ordination Classes to Functions


ebolyen:

Jorge-C:
Just to provide the rationale on why they're classes:
The whole computation consists of two parts, an expensive one that's always necessary and you don't want to keep recomputing, and a cheap one depending on what aspect of the data you want to focus on when visualizing (ie, the scaling you want).

jairideout:
Thanks @Jorge-C, sorry that we missed the `scaling` parameter to the `scores` method. Here's a couple potential solutions (which aren't necessarily mutually exclusive) that might make the API more intuitive for our target users:

- Keep the OO interface so that users interested in different scalings can efficiently compute these. Provide high-level procedural interfaces that simply use the OO interface internally.
- Make the ordination classes callable (`__call__`) instead of a `scores` method. This would make the classes a little more "procedural".

Do you anticipate more methods being added to the ordination classes in the future?

mortonjt:
In the future, I'm planning on adding in some SVD methods.  But they would most likely be tied with the contingency table.  Do you think they would belong in the Ordination classes, or a separate PCA class?

ebolyen:
I'm looking at some of the source code for what scaling does. In some cases it looks trivial, and in others its just a scalar (go figure). I'm not entirely sure how yet, but I feel like we may be able to sidestep this concern entirely with the correct factoring.

jairideout:
Fixed in #925.

PR #860, opened 2015-02-23
Composition Analysis methods

mortonjt:
This is continuing the discussion from https://github.com/biocore/scikit-bio/issues/835

This methods are the barebones required to construct a compositional biplot

The math is a bit funky, so don't hesitate to ask questions - I know that it took me a while to understand it.

@Jorge-C @ElDeveloper @antgonza are you still willing to look over this PR?


skbio/stats/composition.py
```diff
@@ -0,0 +1,375 @@
...
+
+.. currentmodule:: skbio.stats.composition
+
+This module provides functions for compositional data analysis.
+
+Many 'omics datasets are inheriently compositional - meaning that they are best
+interpreted as proportions or percentages rather than absolute counts.
+
+Formally, :math:`x` is a composition if :math:`\sum_{i=0}^D x_{i} = c` and
+:math:`x_{i} > 0`, :math:`1 \leq i \leq D`  and :math:`c` is a real valued
```
Jorge-C: Is $x_i > 0$ really a requirement? Not allowing null abundances seems restrictive.

ElDeveloper:
@mortonjt, I just had a look at the code and it looks good, there's a few minor things that could be improved, but before getting into those changes, what do you think about adding an `Examples` section in the main docstring where the functionality you are adding is being exercised in a real use-case. As a reference see what the ones in the [beta diversity module](http://scikit-bio.org/docs/latest/generated/skbio.diversity.beta.html#module-skbio.diversity.beta) and the [power module](http://scikit-bio.org/docs/latest/generated/skbio.stats.power.html#module-skbio.stats.power).

mortonjt:
Thanks for looking at it @ElDeveloper !

I like the the idea of having an `Examples` section in the main docstring.  The question is should this be added now or later?  These are pretty barebones features and I think the `Examples` section will be difficult to visualize until the composition biplot, ternary plot and some of the other compositional methods (e.g. ilr transform) have implemented.

But it may not be a bad idea to get started.  I can add a PCA plot of the clr transformed sample data to give a taste of how to manipulate this sort of data.  What do you think?

ElDeveloper:
I see what you are saying. If this will cause duplication of effort i.e.
creation of an example that will not ultimately be used, then I don't
think it's a good idea, else it makes sense! I guess I suggested this
because concrete use-cases give me a better idea of where the API could
be improved or where specific problems could arise.

Jorge-C:
I also have a few minor / stylistic comments (will go through them when I've got some time) but I like the code in general, thanks!

On a more broad note, I wonder how much overlap exists between this and ordination. I.e., ordination techniques can potentially benefit from these transformations when working from compositional data ([paper here](http://www.esajournals.org/doi/pdf/10.1890/0012-9658(1997)078%5B0929%3ACDICET%5D2.0.CO%3B2), I haven't gone through it but might be interesting), so my impression is that it makes sense to move composition analysis there, but I'd like to know your opinion.

mortonjt:
Nice find @Jorge-C !  That's actually the strategy that I'm currently using to construct the biplots.
However, this approach does have flaws, the eigenvectors are not guaranteed to sum to zero, which means that these PC reduced values cannot be transformed back into the simplex space.  [This paper](http://www.sis2014.it/proceedings/allpapers/2863.pdf) goes into more detail about that.  However, I think it is reasonable to stick with the standard SVD to start with - these errors can be addressed later.

I did actually raise that question [here](https://github.com/biocore/scikit-bio/issues/829).  However, I'm not sure how well it would currently fit, since these Ordination methods only work on distance matrices.  

To give an idea about the direction this API is headed (the bullet points are sorted with respect to time)

__Near Future__
* A `clr_inv(mat)` method will be added to transform points from the real space back to the simplex space
* A `ilr(mat)` and a corresponding `ilr_inv(mat)` method will be added.  This will enable actual statistical manipulation of compositional data, because it is an isometry and an isomorphism into the real space.  The `clr()` method has some problems because it is not an isomorphism, so applying standard statistical techniques can get funky.
* The Aitchison distance metric will probably be added.  This can serve as an alternative to our existing distance metrics (aka. Unifrac, Bray-Curtis ...)
* Aitchison variance estimators might be added
* The Aitchison inner product might be added.  
* A `alr(mat)` and a corresponding `alr_inv(mat)` will probably be added.  This is required for the following EM algorithm.

__Distant Future__
* A EM method method will likely be added as an alternative to the `multiplicative_replacement()`.
* A compositional SVD will be added at some point ( to reduce errors from the `clr()` transform )
* A whole slew of compositional statistical tests, regressions and time series techniques may be added.  Feel free to glance over Chapters 6, 7 and 8 [here](http://www.sediment.uni-goettingen.de/staff/tolosana/extra/CoDa.pdf) to get an idea.

mortonjt:
I realize that this is quite a bit of information to absorb - the math is a bit dense.

@Jorge-C , you are still in Boulder right?  Do you think it would be beneficial to meet up and step through some of the math related to this PR?

@ElDeveloper , I thought about the `Examples` section. Perhaps it is better to add this after some plotting utilities have been added?

mortonjt:
I come up with a really simple example to illustrate how these techniques can be used in compositional studies, so scratch my previous comment.

Jorge-C:
@mortonjt that's right! Let's coordinate over email
```
python -c "import numpy as np; print(''.join(chr(int(round(i))) for i in np.polyval(np.fromstring('\xe4\xf7\x93\xd2ss ?\x82@\xaa\x19:\xf1}\xbf\xf1\xdd\x85\x11\xdb\xa2\xc7?\x1dRPU\xa6:\x05\xc0)8\x99K\n\xea7@\x04U\xa4\xab4\x8ea\xc0l\xc7\xc3x+\xe6\x80@m(\x92\xfc:\xe5\x94\xc0\xado2\xde(w\x9f@p\xf7#\x9fY\xd0\x99\xc0\xa9\xc7<\n\xb8v\x81@\xc4L\xe5\xff\xff\x7fZ@'), range(12))))"
```
at gmail dot com hehe.

mortonjt:
Hahaha.  Nice!  I just sent an email

wasade:
:+1:

You of course know about `import this`, right?

ebolyen:
@Jorge-C Unfortunately for you, I have a web crawler that executes arbitrary python it finds on the internet looking for emails.

Also @mortonjt, would you mind adding some more unit tests, ideally we would prefer 100% coverage for new functionality added to scikit-bio. You can check the coverage report to find which lines weren't tested. Thanks!

ElDeveloper:
@Jorge-C*slow clap*

@mortonjt Thanks for adding the examples, I'll have a look!

Jorge-C:
Hahah, @ebolyen then my only hope is that your script runs in a big-endian machine and the array gets scrambled!

Jorge-C:
Hey, @mortonjt and I met today to go through the math (spoiler: it's sweet. It has very rich structure as the operations defined here can make the simplex into a Hilbert space, you can even define derivatives and integration along a continuous variable like time...) so I'll review this PR soon :)

mortonjt:
Thanks @Jorge-C !

skbio/stats/tests/test_composition.py
```diff
@@ -0,0 +1,117 @@
...
+                                     clr, centralize)
+
+
+class CompositionTests(TestCase):
+
+    def setUp(self):
+        self.data1 = np.vstack((
+            np.array([2, 2, 6]),
+            np.array([4, 4, 2])))
+        self.data2 = np.array([2, 2, 6])
```
Jorge-C: The array can be directly created in a cleaner way as 
```
np.array([[2, 2, 6],
          [4, 4, 2]])
```
instead of
```
np.vstack((np.array([2, 2, 6]),
           np.array([4, 4, 2])))
```
(there are several instances of that)

skbio/stats/tests/test_composition.py
```diff
@@ -0,0 +1,117 @@
...
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+from unittest import TestCase, main
+
+import numpy as np
+from numpy.testing import assert_array_almost_equal
```
Jorge-C: Please use `import numpy.testing as npt` for consistency.

When starting a module it sometimes looks cleaner to import just the function you're using, but then modules grow and end up with infinite lists of names imported.

ElDeveloper: :+1:, although this is kind of a gray area, I like to do
`np.testing.assert....`, just sayin'.

skbio/stats/tests/test_composition.py
```diff
@@ -0,0 +1,117 @@
...
+        self.data2 = np.array([2, 2, 6])
+
+        self.data3 = np.vstack((
+            np.array([1, 2, 3, 0, 5]),
+            np.array([1, 0, 0, 4, 5]),
+            np.array(range(1, 6))))
+        self.data4 = np.array([1, 2, 3, 0, 5])
+
+    def test_closure(self):
+        assert_array_almost_equal(_closure(self.data1),
```
Jorge-C: Use `npt.assert_allclose` instead (there are several instances of this).

From the docstring of the function you're using:
```
.. note:: It is recommended to use one of `assert_allclose`,
          `assert_array_almost_equal_nulp` or `assert_array_max_ulp`
          instead of this function for more consistent floating point
          comparisons.
```

skbio/stats/tests/test_composition.py
Jorge-C: Thanks for chiming in! Actually, `npt` is straight from [our docs](http://scikit-bio.org/docs/latest/development/coding_guidelines.html#how-do-i-organize-my-modules-source-files) but it's kind of hidden :)

Jorge-C:
Another general comment, @mortonjt: public functions should expect `array_like` arguments instead of ndarrays, and call `np.asarray` to convert them (that's cheap if they're already numpy arrays). There are some examples eg in the [alpha diversity module](https://github.com/biocore/scikit-bio/blob/e5a9f73351b8c979810625c3686da994649e1e7d/skbio/diversity/alpha/_base.py#L681).

skbio/stats/tests/test_composition.py
ElDeveloper: *slow :clap:*

mortonjt:
:+1:

Jorge-C:
Thanks! It seems [one note](https://github.com/biocore/scikit-bio/pull/860#discussion_r25803453) was hidden too early (using `assert_allclose` instead of `assert_array_almost_equal`).

skbio/stats/composition.py
```diff
@@ -0,0 +1,419 @@
...
+
+import numpy as np
+
+import scipy.stats as ss
+
+
+def _closure(mat):
+    """
+    Performs closure to ensure that all elements add up to 1
+    mat : array-like
```
Jorge-C: DOC: input args should be described inside a section
```
Parameters
----------
```

skbio/stats/composition.py
```diff
@@ -0,0 +1,419 @@
...
+       a matrix of proportions where
+       rows = compositions
+       columns = components
+
+    Returns
+    -------
+    mat_ : numpy.ndarray, np.float64
```
Jorge-C: More doc: when the name of the output variable is not informative, it can be omitted.

mortonjt: Ok.  But keep the datatype right?

skbio/stats/composition.py
```diff
@@ -0,0 +1,419 @@
...
+    :math:`x \ominus y = C[x_1 y_1^{-1}, ..., x_D y_D^{-1}]`
+
+    :math:`C[x]` is the closure operation defined as
+    :math:`C[x] = [\frac{x_1}{\sum x},...,\frac{x_D}{\sum x}]`
+    for some :math:`D` dimensional real vector :math:`x` and
+    :math:`D` is the number of components for every composition.
+
+    Parameters
+    ----------
+    x : numpy.ndarray
```
Jorge-C: Should be `array_like` too (also, note the underscore, not hyphen). BTW, the reference for docstrings formatting is [in numpy's repo](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#docstring-standard)

skbio/stats/composition.py
Jorge-C: Actually, I don't know how that shows in the html docs. It is probably fine, though I personally don't find the dtype that important, particularly because it's the most common one.

skbio/stats/composition.py
```diff
@@ -0,0 +1,419 @@
...
+    Returns
+    -------
+    mat_ : numpy.ndarray, np.float64
+       A matrix of proportions where all of the values
+       are nonzero and each composition (row) adds up to 1
+
+    """
+    mat = np.asarray(mat)
+    if not isinstance(mat.dtype, float):
+        mat = mat.astype(np.float64)
```
Jorge-C: These 3 lines can be condensed into `mat = np.asarray(mat, dtype=np.float64)`

skbio/stats/composition.py
```diff
@@ -0,0 +1,419 @@
...
+    >>> import numpy as np
+    >>> from skbio.stats.composition import multiplicative_replacement
+    >>> X = np.array([[.2,.4,.4, 0],[0,.5,.5,0]])
+    >>> multiplicative_replacement(X)
+    array([[ 0.1875,  0.375 ,  0.375 ,  0.0625],
+           [ 0.0625,  0.4375,  0.4375,  0.0625]])
+
+    """
+    mat = np.asarray(mat)
+    z_mat = (mat == 0).astype(np.float32)
```
Jorge-C: Why is it casted to float32?

skbio/stats/composition.py
```diff
@@ -0,0 +1,419 @@
...
+
+    if len(mat.shape) == 1:
```
Jorge-C: No need to call len on the shape tuple, `mat.ndim` is already what you want :)

skbio/stats/composition.py
mortonjt: Adding that in.  However, right now that method is being treated as private, so there won't be any docs generated for it.

skbio/stats/composition.py
mortonjt: Cause I was being dumb :(

skbio/stats/composition.py
mortonjt: nice!

mortonjt:
@Jorge-C - looked over your comments carefully and I think I addressed them.  Thanks!

skbio/stats/composition.py
```diff
@@ -0,0 +1,416 @@
...
+array([ 0.25,  0.25,  0.5 ])
+
+Reference
+---------
+.. [1] V. Pawlowsky-Glahn. "Lecture Notes on Compositional Data Analysis"
+.. [2] J. A. Martin-Fernandez. "Dealing With Zeros and Missing Values in
+       Compositional Data Sets Using Nonparametric Imputation"
+
+"""
+from __future__ import absolute_import, division, print_function
```
vivekiitkgp: Can you please update the module header according to new guidelines? 
http://scikit-bio.org/docs/latest/development/new_module.html

vivekiitkgp:
@mortonjt Just a minor note from my side. Thanks!

skbio/stats/composition.py
mortonjt: This is to knock off an extra layer of nesting correct?

I'm a bit on the edge on that - because I think if the `skbio.stats` was used as an import rather than `skbio.stats.composition`, it may be confusing exactly what sort of statistical tools are being used.
(e.g If I were to use the `perturb` function, it would make no sense in the `skbio.stats` context by itself, but it would make more sense in the `skbio.stats.composition` import since it is a compositional data analysis tool)

vivekiitkgp: No.

You just need to move the `from future ..` line from its current position to the next of copyright header.

mortonjt: Ohhh...

Derp.  Makes sense.  That'll be corrected in the upcoming commit.  Thanks!

Jorge-C:
Hey @mortonjt, I pushed some changes to my repo on top of your commits (mostly simplifying it using operators, removing special casing of 1D/2D cases using numpy functions, some speed improvements). You can pull those changes into your copy using something like:
```
git remote add jorge git@github.com:Jorge-C/bipy.git
git pull --ff-only jorge stylistic
```
and then push them to github so that they show up here  (I could not figure out how to make a PR to your forked repo b/c it didn't let me fork it). Let me know if you have any questions about those!

Also, `centralize` should test for the requirements stated in the Notes docs (positivity, rows add up to about one). Do you want to add that?

After that, this should be ready for merge. Thanks Jamie!

Jorge-C:
I finally figured out how to open a PR to your repo @mortonjt.

mortonjt:
Awesome!! Thanks @Jorge-C !

Let me squash some of my commits and I'll merge your PR right into my fork.
Question: When you mean test for the requirements stated in the Notes, do you mean adding in an assert statement into the functions, or adding in a unit test?

Adding in an assert statement may not be a bad idea, but I'll probably add in another parameter to enable /disable checking.

Jorge-C:
The former, but w/out assert, ie
```python
if np.any(mat <= 0):
    raise ValueError("
```
The reason is that we should never error out silently, and so I don't think we should let those sanity checks be disabled.

mortonjt:
:+1:

skbio/stats/composition.py
Jorge-C: This still needs to be fixed, including in a few other docstrings.

mortonjt:
Just renamed closure function.  In case this function is used outside of this script

skbio/stats/composition.py
```diff
@@ -0,0 +1,410 @@
...
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+from __future__ import absolute_import, division, print_function
+import numpy as np
+import scipy.stats as ss
+
+
+def closure(mat):
```
ebolyen: Are any of the operations which follow unclosed? It doesn't appear so. In that case should this method be private?

mortonjt: That's what I thought when I first pushed this method.

However, when dealing with subcompositions, it is necessary to re-close the data in order to maintain that everything adds up to one.

For example, if there are 7 phyla you are interested in, but 3 of them are rare - you can exclude them and close the remaining 4 phyla so that their proportions add up to 1.  This a legitimate operation because it maintains [subcompositional coherence](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.324.1621&rep=rep1&type=pdf) (aka the ratios are preserved)

I'm already planning on using this technique when making tetrahedron plots for Emperor in the near future

mortonjt:
What else needs to be accomplished in this PR before merging?

In addition, I'm planning on adding in the following methods in the very near future

`clr_inv` - inverse clr transform
`ilr` - isometric log ratio transform
`ilr_inv` - inverse isometric log ratio transform
`inner` - Aitchison inner product (tests for orthogonality and is required in `ilr`)

Not only would these function aide statistical analysis, but they can enhance phylogenetic tree visualization.  Should these methods go into this PR, or an upcoming PR?

skbio/stats/composition.py
```diff
@@ -0,0 +1,410 @@
...
+
+    Returns
+    -------
+    numpy.ndarray, np.float64
+       A matrix of proportions where all of the values
+       are nonzero and each composition (row) adds up to 1
+
+    Notes
+    -----
+    - All of the values in x and y must be greater than zero
```
Jorge-C: Either the docstring or the checks are not correct regarding dealing with zeros, here and in other functions.

In fact, validation is the same for all functions, so it should be extracted to a helper function. Either improving `closure` and calling that for each input array or having `_validate` that just converts to arrays and checks signs and raises the appropriate error.

I like the 1st option because it seems that this module would error out in a more user friendly way if someone tries to use this with non-compositional data by accident.

mortonjt: :+1:

Jorge-C:
Thanks @mortonjt! I proposed some minor docstring changes in your repo, but removing repetition from arg checking is the last important thing missing. Great job!

mortonjt:
Thanks Jorge!  I decided to keep the non-negative parameter checking as is, since it is necessary to use the closure operation on compositions with zeros in them.

Jorge-C:
That sounds fine. Then the `Notes` section in the docstrings should just say non-negative instead of greater than zero, right?

Jorge-C:
Great, thanks Jamie! I'll merge this tomorrow unless someone else wants to review and wants an extension hehe.

mortonjt:
Thanks Jorge!!

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ described here [Making a flat list out of lists of lists](http://stackoverflow.c
 ## Version 0.2.3 (2015-02-13)
 
 ### Features
+* Added `skbio.stats.composition` for analyzing data made up of proportions
```
jairideout: Can you please move this under the Features heading in 0.2.3-dev?

skbio/stats/composition.py
```diff
@@ -0,0 +1,384 @@
...
+   centralize
+
+Examples
+--------
+
+>>> import numpy as np
+
+Consider a very simple environment with only 3 species. The species
+in the environment are equally distributed and their proportions are
+equilvalent:
```
jairideout: equilvalent -> equivalent

skbio/stats/composition.py
```diff
@@ -0,0 +1,384 @@
...
+    --------
+    >>> import numpy as np
+    >>> from skbio.stats.composition import multiplicative_replacement
+    >>> X = np.array([[.2,.4,.4, 0],[0,.5,.5,0]])
+    >>> multiplicative_replacement(X)
+    array([[ 0.1875,  0.375 ,  0.375 ,  0.0625],
+           [ 0.0625,  0.4375,  0.4375,  0.0625]])
+
+    """
+    mat = np.atleast_2d(mat)
```
jairideout: Some of the checks here are the same as in `closure`. It'd be nice to move these checks into a private function that `closure` and `multiplicative_replacement` call to reduce duplication.

jairideout:
Thanks @mortonjt and @Jorge-C for your work on this! I'm taking a quick pass through and will follow up when I'm done.

skbio/stats/tests/test_composition.py
```diff
@@ -0,0 +1,195 @@
...
+        self.data4 = np.array([1, 2, 3, 0, 5])
+        self.data5 = [[2, 2, 6], [4, 4, 2]]
+        self.data6 = [[1, 2, 3, 0, 5],
+                      [1, 0, 0, 4, 5],
+                      [1, 2, 3, 4, 5]]
+        # Bad datasets
+        self.bad1 = np.array([1, 2, -1])
+        self.bad2 = np.array([[[1, 2, 3, 0, 5]]])
+
+    def testclosure(self):
```
jairideout: testclosure -> test_closure

skbio/stats/tests/test_composition.py
```diff
@@ -0,0 +1,195 @@
...
+        pmat = perturb_inv(closure(self.data5),
+                           closure([.1, .1, .1]))
+        imat = perturb(closure(self.data1), closure([10, 10, 10]))
+        npt.assert_allclose(pmat, imat)
+
+        with self.assertRaises(ValueError):
+            perturb_inv(closure(self.data1), self.bad1)
+
+    def test_multiplicative_replacement(self):
+        np.set_printoptions(precision=10)
```
jairideout: Can this line be removed?

skbio/stats/composition.py
```diff
@@ -0,0 +1,384 @@
...
+    if np.any(mat < 0):
+        raise ValueError("Cannot have negative proportions")
+    if mat.ndim > 2:
+        raise ValueError("Input matrix can only have two dimensions or less")
+    mat = mat / mat.sum(axis=1, keepdims=True)
+    return mat.squeeze()
+
+
+def multiplicative_replacement(mat, delta=None):
```
jairideout: Can you please document `delta` in the docstring?

skbio/stats/composition.py
```diff
@@ -0,0 +1,384 @@
...
+    """Replace all zeros with small non-zero values.
+
+    It uses the multiplicative replacement strategy, replacing zeros
```
jairideout: It'd be helpful to add a `References` section to this function that includes the multiplicative replacement strategy reference.

skbio/stats/composition.py
```diff
@@ -0,0 +1,384 @@
...
+
+    Returns
+    -------
+    numpy.ndarray, np.float64
+       A matrix of proportions where all of the values
+       are nonzero and each composition (row) adds up to 1
+
+    Notes
+    -----
+    - All of the values in x and y must be non negative
```
jairideout: This renders as a bullet point, probably clearer to remove the hyphen.

jairideout: Or just remove the Notes section altogether, as I think every function requires the values to be nonnegative.

skbio/stats/composition.py
```diff
@@ -0,0 +1,384 @@
...
```
jairideout: Same comment as above (in `perturb`).

skbio/stats/composition.py
mortonjt: I had that reference in the main docs on at the top of the file.  But maybe it'll be a good idea to also add it to this function

skbio/stats/composition.py
```diff
@@ -0,0 +1,384 @@
...
+    >>> y = np.array([1./6,1./6,1./3,1./3])
+    >>> perturb(x,y)
+    array([ 0.0625,  0.1875,  0.5   ,  0.25  ])
+
+    """
+    x, y = closure(x), closure(y)
+    return closure(x * y)
+
+
+def perturb_inv(x, y):
```
jairideout: What do you think about merging `perturb_inv` into `perturb`, so that users can perform the inverse by specifying an optional parameter, e.g., `inverse=True`? The docstrings and signature are essentially the same between the two functions, with the only difference being the first latex equation. The implementations are also very similar except for using `/` instead of `*`.

skbio/stats/composition.py
```diff
@@ -0,0 +1,359 @@
...
+    a : float
+        a scalar float
+
+    Returns
+    -------
+    numpy.ndarray, np.float64
+       A matrix of proportions where all of the values
+       are nonzero and each composition (row) adds up to 1
+
+    >>> import numpy as np
```
jairideout: add an Examples section above this code

skbio/stats/composition.py
```diff
@@ -0,0 +1,359 @@
...
+       a matrix of proportions where
+       rows = compositions and
+       columns = components
+
+    Returns
+    -------
+    numpy.ndarray
+         clr transformed matrix
+
+    >>> import numpy as np
```
jairideout: add an Examples section above this code

skbio/stats/composition.py
```diff
@@ -0,0 +1,359 @@
...
+       a matrix of proportions where
+       rows = compositions and
+       columns = components
+
+    Returns
+    -------
+    numpy.ndarray
+         centered composition matrix
+
+    >>> import numpy as np
```
jairideout: add an Examples section above this code

skbio/stats/composition.py
jairideout: Thanks! Fine to leave in module-level docstring too. I thought it'd be helpful to have it listed on the function in case users read the function's docs without seeing the module-level docs.

skbio/stats/composition.py
mortonjt: I am a little on the edge about that, here's why.

If you are stacking many of these operations together, adding in additional parameters to specify inverses can be messy really quickly.  Its a similar reason why there is a `np.multiply` and a `np.divide` function.  If there was only a single `np.multiply` function with a parameter specifying division, you can imagine that matrix manipulations can become very messy.

I contemplated having a `Composition` object and overloading the `+`, `-`, `*` and `\` operators.
But I'm still in favor of having this scheme, because the operator overloading can get confusing and the data structures are a bit more transparent.

skbio/stats/composition.py
```diff
@@ -0,0 +1,359 @@
...
+    that the compositions still add up to 1
+
+    Parameters
+    ----------
+    mat: array_like
+       a matrix of proportions where
+       rows = compositions and
+       columns = components
+    delta: float, optional
+       a small number to be used to replace zeros
```
jairideout: Can you describe what delta is used if it is `None` (the default)?

skbio/stats/composition.py
```diff
@@ -0,0 +1,359 @@
...
+       A matrix of proportions where all of the values
+       are nonzero and each composition (row) adds up to 1
+
+    """
+    mat = np.atleast_2d(mat)
+    if np.any(mat < 0):
+        raise ValueError("Cannot have negative proportions")
+    if mat.ndim > 2:
+        raise ValueError("Input matrix can only have two dimensions or less")
+    mat = mat.astype(np.float64)
```
jairideout: Can pass `copy=False` here for cases where `mat` is already the correct `dtype`.

skbio/stats/tests/test_composition.py
```diff
@@ -0,0 +1,192 @@
...
+
+from unittest import TestCase, main
+import numpy as np
+import numpy.testing as npt
+from skbio.stats.composition import (closure, multiplicative_replacement,
+                                     perturb, perturb_inv, power,
+                                     clr, centralize)
+
+
+class CompositionTests(TestCase):
```
jairideout: Another test case that'd be great to add is a check to ensure that the input array(s) aren't modified in-place by the function. In a lot of cases, this won't happen because a copy of the array will be made during execution (e.g., in `np.atleast_2d` or `astype`). However, if the input array is a numpy array of the correct `dtype` already, there's a chance that the manipulations are applied to the input instead of a copy.

I don't think any of the functions modify the input in-place, but it'd be good to test this to ensure this behavior doesn't change in the future.

skbio/stats/composition.py
jairideout: Sounds good, that makes sense. Thanks for explaining!

skbio/stats/composition.py
```diff
@@ -0,0 +1,359 @@
...
+    for some :math:`D` dimensional real vector :math:`x` and
+    :math:`D` is the number of components for every composition.
+
+    Parameters
+    ----------
+    x : array_like
+        a matrix of proportions where
+        rows = compositions and
+        columns = components
+    y : array_like
```
jairideout: add below this:

a matrix of proportions where

skbio/stats/composition.py
mortonjt: Good call.

But what do you think about having a check like

```
if mat.dtype != np.float64:
    mat = mat.asdtype(np.float64)
```

That way, it would eliminate the need to pass in another parameter.

skbio/stats/composition.py
```diff
@@ -0,0 +1,359 @@
...
+    --------
+    >>> import numpy as np
+    >>> from skbio.stats.composition import multiplicative_replacement
+    >>> X = np.array([[.2,.4,.4, 0],[0,.5,.5,0]])
+    >>> multiplicative_replacement(X)
+    array([[ 0.1875,  0.375 ,  0.375 ,  0.0625],
+           [ 0.0625,  0.4375,  0.4375,  0.0625]])
+
+
+    References
```
jairideout: add a section underline here and move References above Examples (order is specified this way in numpydoc spec)

skbio/stats/composition.py
jairideout: Cite `[1]_` here

skbio/stats/composition.py
```diff
@@ -0,0 +1,359 @@
...
+perturbation vector would be as follows
+
+>>> antibiotic = np.array([1./2, 1./2, 1])
+
+And the resulting perturbation would be
+
+>>> perturb(otus, antibiotic)
+array([ 0.25,  0.25,  0.5 ])
+
+Reference
```
jairideout: Reference -> References

Move this above Examples

skbio/stats/composition.py
jairideout: It'd be simpler to just call `mat.astype(np.float64, copy=False)` instead of using an if statement. That way we can let numpy determine whether to make a copy or not.

skbio/stats/composition.py
mortonjt: I'll change the remaining references.  Thanks!

jairideout:
@mortonjt @Jorge-C done reviewing, had some inline comments, most of them are minor. The code and docstrings are really clean, thanks for your work on this!

Let me know when you're ready for me to take a final pass. Also note that there's a merge conflict.

mortonjt:
Thanks @jairideout !  Just added the changes that you suggested

skbio/stats/tests/test_composition.py
mortonjt: :+1:

skbio/stats/composition.py
```diff
@@ -0,0 +1,367 @@
...
+    if mat.ndim > 2:
+        raise ValueError("Input matrix can only have two dimensions or less")
+    if mat.dtype != np.float64:
+        mat = mat.astype(np.float64)
+    mat = mat / mat.sum(axis=1, keepdims=True)
+    return mat.squeeze()
+
+
+def multiplicative_replacement(mat, delta=None):
+    r"""Replace all zeros with small non-zero values. [1]_
```
jairideout: Citations should occur in the body of the docstring or in subsequent sections (e.g., Notes). Can you move this down to follow "multiplicative replacement strategy"?

jairideout:
Sounds good, thanks! Looks like there are some remaining comments. Let me know when those are done and I'll review again.

mortonjt:
Okay.  I believe this addresses all of the suggestions.  Thanks for reviewing this @jairideout !

jairideout:
Thanks @mortonjt! This is almost ready for merge, just two comments:

- [ ] add in-place modification unit tests for `test_clr` and `test_centralize` (similar to what the other tests have)
- [ ] in `closure`, I noticed the `astype(np.float64)` call was removed. Is this type casting necessary for e.g., integer arrays? Just want to make sure I understand why it isn't needed anymore.

mortonjt:
Nice catch for the `test_clr` and `test_centralize` tests.  I shouldn't have missed those.

I talked to @Jorge-C about this offline.  The reason why the casting is necessary is because `__future__.division` automatically casts division output as a float.

jairideout:
Looks good, thanks again @mortonjt and @Jorge-C for all of the work on this!

mortonjt:
Whoo!!!

Jorge-C:
Thank you guys!

ElDeveloper:
:fireworks:

PR #862, opened 2015-03-02
Add copyright header validator to checklist.py

### LINKED ISSUE #855: Checklist.py should verify the copyright headers on each file.
The current standard for scikit-bio's copyright headers is defined [here](http://scikit-bio.org/docs/latest/development/new_module.html). What checklist should be able to validate:

If there is a module level docstring:
 1. The docstring appears at the top of the file. Variant isn't important (can be `r"""....` or plain `"""....` and any style of quotation).
 2. This docstring is immediately followed by a new line.
 3. After this new line, our copyright header appears in full.
 4. Finally another new line.

If there is **no** module level docstring:
 1. The copyright header appears at the top of the file in full.
 2. Followed by a new line.

I would suggest using the [tokenize](https://docs.python.org/3/library/tokenize.html) module in the standard library to accomplish this as it can identify comments while the AST module has already stripped them.

vivekiitkgp:
Closes #855.

Signed-off-by: Vivek Rai <vivekraiiitkgp@gmail.com>

### DIFF
--- a/checklist.py
+++ b/checklist.py
@@ -16,6 +16,7 @@ import os.path
 import subprocess
 import sys
 import ast
+import tokenize
 
 import dateutil.parser
 
@@ -36,8 +37,9 @@ def main():
 
     """
     root = 'skbio'
-    validators = [InitValidator(), ExecPermissionValidator(),
-                  GeneratedCythonValidator(), APIRegressionValidator()]
+    validators = [InitValidator(), CopyrightHeadersValidator(),
+                  ExecPermissionValidator(), GeneratedCythonValidator(),
+                  APIRegressionValidator()]
 
     return_code = 0
     for validator in validators:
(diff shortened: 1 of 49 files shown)


vivekiitkgp:
@ebolyen Should I add another commit *fixing* those files, so that the tests pass?

ebolyen:
Thanks @vivekiitkgp, I'll review this shortly! Leave the failure for a little bit, we will want to fix them before we merge, but I would like to review them. Thanks!

ebolyen:
This is very robust! Thank you @vivekiitkgp! Only a couple of very minor comments regarding readability and this is good to go. Let's make those changes, let the tests fail again, and then fix the tests in this PR.

checklist.py
```diff
@@ -144,6 +146,67 @@ def _system_call(self, cmd):
...
+                    map(lambda x: x[1], tokens[pos:pos + 14]))
+            if cheader != self.COPYRIGHT_HEADER or tokens[pos + 14][0] != 54:
+                invalid_files.append(f.name)
+            f.close()
+
+        return invalid_files
+
+    @staticmethod
+    def is_shebang(line):
+        return line == '#!/usr/bin/env python'
```
Jorge-C: Shebang lines can have spaces, ie
```
#! /usr/bin/env python
```
is a valid one that would fail the test. In fact, the space is used by the [python docs](https://docs.python.org/3/using/windows.html#shebang-lines).

A more robust approach would be to check that it's the first line of the script and that it `.startswith('#!')`.

vivekiitkgp:
@ebolyen I fixed the warnings and updated the code as suggested. Weirdly though, the first task on travis build failed.

ebolyen:
@vivekiitkgp That is strange, there seems to be something up with running it in Python 3. Looks like the token numbers are different between Python 2.7 and Python 3. Looks like you can use the constants from `tokenize` to fix this ( `tokenize.STRING`, `tokenize.NL`, and `tokenize.COMMENT`).

vivekiitkgp:
@ebolyen Thanks. Done.

checklist.py
```diff
@@ -144,6 +146,71 @@ def _system_call(self, cmd):
...
+    Parameters
+    ----------
+    skip_dirs : iterable of str, optional
+        Directory names to skip during validation. Defaults to skipping any
+        directories named ``'data'`` or ``'__pycache__'`` (and anything
+        contained within them).
+
+    """
+
+    reason = "Files non-conforming to standard copyright headers"
```
jairideout: It could be helpful to point to the docs that describe this format:

http://scikit-bio.org/docs/latest/development/new_module.html

jairideout:
Thanks @vivekiitkgp, this looks good! I had a minor inline comment. Can you also add a mention to the changelog under the Miscellaneous section?

vivekiitkgp:
@jairideout I have already added link to that doc in the class docstring. (L153)

jairideout:
I saw that but thought it would be helpful to have the link in the error message itself because developers may not know to look at the docstring in checklist.py for this information.

checklist.py
```diff
@@ -144,6 +146,71 @@ def _system_call(self, cmd):
...
+                continue
+            pos = 0
+            f = open(os.path.join(root, _file))
+            tokens = list(tokenize.generate_tokens(f.readline))
+
+            # A shebang line or module docstring is described using just two
+            # tokens: the main string and a terminating newline. By convention,
+            # however, these lines are always followed by a newline and hence
+            # we must advance by 3 positions to get to next logical line.
+            if tokens[pos][1].startswith('#!'):
```
wasade: Sorry for not commenting earlier on this. I do think we want to be more aggressive, and only allow either `#!/usr/bin/env python` or `#! /usr/bin/env python`. The reason for this is that it is not uncommon to see something like `#!/path/to/very/specific/python/binary`, which isn't good, and the present check will allow for that situation.

jairideout: Can we explicitly disallow shebang lines in scikit-bio library code? We currently don't have any under the `skbio/` directory (only checklist.py and setup.py have them).

wasade: That also works :) They are only advisable for scripts being executed on
the command line, and only necessary if executing without an explicit call
to the interpreter, IIRC

gregcaporaso: I agree, I think that's a good solution.

jairideout:
@vivekiitkgp there are a couple of remaining comments (shebang lines and adding the URL to the error message). Once those are addressed this is ready for merge!

vivekiitkgp:
@jairideout Done.

jairideout:
Thanks @vivekiitkgp! There are merge conflicts (probably related to the changelog). Can you please fix those? This is ready for merge after that.

vivekiitkgp:
@jairideout Yes. Done.

jairideout:
Thanks! Not sure why coverage decreased, your changes shouldn't affect coverage (I think coveralls might be confused).

wasade:
@jairideout, I noticed the samething with coverage the other day on BIOM. I think a new version was released that changed how it was calculated (which is weird...) as I looked at the log and there was no actual change in any of the files.

jairideout:
Ah thanks for the details, that makes sense!

PR #863, opened 2015-03-05
Update cython sources

Jorge-C:
Rebuilding .c sources with latest cython, v-0.22.

I came accross an issue doing this: the ssw wrapper .c file was updated fine, but the generated c code for subsampling was not updated till I first manually removed it... Has someone else seen something similar?


ebolyen:
I believe it waits for the .pyx file to change, not necessarily the version of cython.

Jorge-C:
That makes sense, though I don't understand why the ssw wrapper was updated automatically (it hasn't changed since the last .c rebuild). Thanks @ebolyen!

jairideout:
That's strange, I wonder if Cython skipped the subsampling module because the filename starts with two underscores, where SSW starts with only one underscore?

Jorge-C:
Yeah, that was my hunch too. Maybe we shouldn't mess with too many underscores hehe.

jairideout:
Yeah, underscores have caused problems with nose too... even though it's a headache, it helps distinguish public and private modules which makes it worth it IMO.

We have double underscores for subsample because there's already a `_subsample` module. Maybe there's a better way to restructure this to avoid double underscores?

ebolyen:
:+1: for restructuring

Jorge-C:
Aha, though personally I don't think it's worth it: it brings issues to the table that we wouldn't have otherwise and I don't perceive any real benefit from "hiding" the actual modules (I guess I'm missing some other benefit).

For accelerated modules, we keep the cython implementation with the same module name prefixed by an underscore (ie `_subsample.pyx` was the fast, not compulsory module, and `subsample.py` had the pure python implementation). This is the convention followed in CPython's stdlib (see [PEP399](http://legacy.python.org/dev/peps/pep-0399/) for details) and I think it's good to stick to upstream's conventions. So, if you want to refactor that, I suggest reverting the extra underscores everywhere for increased consistency and decreased random issues :white_check_mark:

jairideout:
The reason we name the modules with a leading underscore is to mark them as part of the "private" API. This makes it immediately clear whether a subpackage/module is private or public to both users and developers.

For users, this helps prevent importing functionality from an import path that we meant to keep private. For developers, this gives us the freedom to refactor private subpackages/modules at will without worrying about breaking backwards-compatibility.

Definitely agree it's less than ideal to not follow how upstream Python handles this, and it's a pain to deal with tools that ignore underscored files, but I think the benefits of having a clear delineation between public and private functionality outweigh the costs.

jairideout:
I upgraded to Cython 0.22, ran `pip install -e .`, and it updated both SSW and subsample. Not sure why it didn't work for you but the underscore filenames don't seem to be the issue. Also, checklist.py ensures that generated .c files are newer than the .pyx files so it should catch any issues in the future.

Jorge-C:
Thanks Jai!

jairideout:
No problem, thanks for updating the sources!

PR #864, opened 2015-03-06
ENH: numpy no longer needed prior to running setup.py

ebolyen:
Thanks to http://stackoverflow.com/a/21621689/579416
we can now `pip install scikit-bio` without numpy! It would be great if others could confirm this works on their system on a fresh venv. 
This is very similar to the approach matplotlib uses, just not quite as heavyweight.


anderspitman:
:+1:

jairideout:
:godmode: 

Can you add a changelog mention for this Very Exciting Feature :tm: ?

PR #865, opened 2015-03-06
Logo change

johnchase:


ElDeveloper:
There's a mention to the ASCII tree logo at the bottom of the file, this should perhaps(?) be removed.

johnchase:
@ElDeveloper the ASCII logo will still be in the `skbio.__init__` so I was thinking we should reference it somewhere. Perhaps the README.rst isn't the best place though?

ElDeveloper:
That's true.

README.rst
```diff
@@ -1,31 +1,8 @@
...
-                      /
-                     *
-                    / \
-                   /   \
-        Proteobacteria  \
-                       Cyanobacteria
-
-|Build Status| |Coverage Status|
+.. image:: /assets/logo.png
+  :alt: Scikit-Bio Logo
```
jairideout: Can you make this alt text all lowercase? Just for consistency with the project name.

README.rst
```diff
@@ -1,31 +1,8 @@
...
-                       /
-                      /
-                     *
-                    / \
-                   /   \
-        Proteobacteria  \
-                       Cyanobacteria
-
-|Build Status| |Coverage Status|
+.. image:: /assets/logo.png
```
jairideout: I'm not sure if this will render on pypi or the scikits website (not sure how to easily verify this hunch though). It might be safer to provide a full URL to the image file.

jairideout:
> Perhaps the README.rst isn't the best place though?

How about adding the ASCII logo/text credits as a comment in `skbio._init__` since that's the only place this art is included now?

ElDeveloper:
Makes sense!

Yoshiki Vázquez-Baeza

README.rst
ebolyen: Definitely want to use the full URL.

ebolyen: Might also consider using the `svg` copy as it is half the size (in bits).

johnchase:
@ebolyen I was basing my decision for the relative path based on this [Relative links in READMEs](https://help.github.com/articles/relative-links-in-readmes/). I wasn't able to find if it is guaranteed to work in scikits or pypi (or that the alternative is always guaranteed to work).

ebolyen:
GitHub can handle relative links because it knows our package structure, 
but PyPi has no way of actually knowing as it is only looking in our 
manifest etc... and will render the RST itself. I wouldn't expect it to 
work, but it might. Either way, an absolute link would be better.

johnchase:
@ebolyen thanks I made those changes.

README.rst
```diff
@@ -1,31 +1,9 @@
...
-                      /
-                     *
-                    / \
-                   /   \
-        Proteobacteria  \
-                       Cyanobacteria
-
-|Build Status| |Coverage Status|
+.. image:: http://scikit-bio.org/assets/logo.svg
+   :target: http://scikit-bio.org/assets/
```
jairideout: This points to an invalid URL (404). Can you either update to point to http://scikit-bio.org or remove the target? I don't have strong feelings either way.

skbio/__init__.py
```diff
@@ -4,6 +4,8 @@
 # Distributed under the terms of the Modified BSD License.
 #
 # The full license is in the file COPYING.txt, distributed with this software.
+#
+#scikit-bio's ASCII art tree was created by `@gregcaporaso`
```
jairideout: The build is failing because this isn't pep8 compliant (checklist.py will also complain b/c the copyright header needs to stay intact across all source files). Can you please move this comment directly above the `art` variable in this file?

There's also credits in the readme about the ASCII art title. Please move that directly above the `title` variable in this file.

jairideout:
Looks good, thanks @johnchase!

Issue #867, opened 2015-03-07
update/fix outdated user/dev docs


jairideout:
Lots of our user/dev (non-API) docs are outdated due to changes over the past several months (e.g., different testing strategies, documentation guidelines, issue labeling, etc.). I'll take a pass through and fix any issues. This should make things clearer for users and devs getting started with scikit-bio contributions.

Will address #850 as part of this.

PR #868, opened 2015-03-07
Fix test instructions in README

anderspitman:
Somehow this slipped through the cracks in #808


anderspitman:
Probably falls under #867

Issue #871, opened 2015-03-11
tests can't be run in interactive session


anderspitman:
After the changes I made to the testing procedures, particularly using nose's ``run_exit`` method, running tests within an interactive python session causes the session to exit, and running from within ``IPython`` throws the following exception:

```
Ran 177 tests in 0.070s

OK
An exception has occurred, use %tb to see the full traceback.

SystemExit: False

To exit: use 'exit', 'quit', or Ctrl-D.
```

Looks like I'll need to detect whether we're running interactively.

anderspitman:
Hm this might be a little tricky, since ``python -c "import skbio; skbio.test()"`` is technically an interactive session, as far as I can tell. Unless someone else knows a way to differentiate. The easiest way to solve this would be to add an ``exit`` parameter to ``TestRunner.test`` and update ours docs to instruct the users to do the following when running interactively:

``` python
>>> import skbio
>>> skbio.test(exit=False)
```

An alternative would be to not support running tests interactively and remove those instructions from our docs. I like option 1 better.

Jorge-C:
That's too bad... If we need an argument, I feel it might be better to make it necessary for the `python -c "import ...` version rather than in a more "interative" usage (it seems more likely that users are copy/pasting the first command when following the installation instructions, whereas if they stumble upon the `skbio.test` they won't realize they need an extra argument).

jairideout:
I think we can work around this by changing the user-level instructions to be:

```
$ python -m skbio.__init__
```

instead of 

```
python -c "import skbio; skbio.test()"
```

Then have `TestRunner` return the exit code instead of using `run_exit` and update `skbio.__init__` to pass this value to `sys.exit` within `if __name__ == '__main__':`. That way the exit code will be retained when executing the tests non-interactively, and simply returned in an interactive context.

jairideout:
@anderspitman can I assign this to you?

anderspitman:
@jairideout yep I got it. I like this solution.

anderspitman:
Seems to work. @jairideout what do you think about adding a ``run_tests.py`` file in ``skbio`` rather than having this live in ``__init__.py``?

``python -m skbio.run_tests`` 

seems more clear than

``python -m skbio.__init__``

Not sure if it's worth adding another file though.

anderspitman:
[...]

PR #872, opened 2015-03-11
Interactive tests

anderspitman:
For #871

### DIFF
--- a/skbio/test.py
+++ b/skbio/test.py
@@ -6,14 +6,13 @@
 # The full license is in the file COPYING.txt, distributed with this software.
 # ----------------------------------------------------------------------------
 
-ifeq ($(WITH_COVERAGE), TRUE)
-	TEST_COMMAND = coverage run -m skbio.test
-else
-	TEST_COMMAND = python -m skbio.test
-endif
+import sys
+from skbio.util import TestRunner
 
-test:
-	$(TEST_COMMAND)
-	pep8 skbio setup.py checklist.py
-	flake8 skbio setup.py checklist.py
-	./checklist.py
+test = TestRunner(__file__).test
+
+if __name__ == '__main__':
+    if test():
+        sys.exit(0)
+    else:
+        sys.exit(1)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ described here [Making a flat list out of lists of lists](http://stackoverflow.c
 ### Miscellaneous
 * Changed testing procedures
     - Developers should now use `make test`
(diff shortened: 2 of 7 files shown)


jairideout:
@ebolyen can you please review this?

ebolyen:
@jairideout yup!

skbio/test.py
```diff
@@ -0,0 +1,18 @@
...
+import sys
+from skbio.util import TestRunner
+
+test = TestRunner(__file__).test
+
+if __name__ == '__main__':
+    if test():
+        sys.exit(0)
+    else:
+        sys.exit(1)
```
ebolyen: This can technically be compressed to: `sys.exit(not test())` But I don't know if that is awesome for readability.

Jorge-C: I like the short version better.

ebolyen:
@anderspitman Tested this locally and it is working great! Way easier to type too! Only one comment which is pretty discretionary, so let me know if you are ready for merge.

skbio/test.py
ebolyen: The downside is it abuses the notion that "True" is 1 and "False" is 0, which have opposite semantics for return codes.

anderspitman: Yeah that's what I started with then based of of [this](http://stackoverflow.com/questions/2764017/is-false-0-and-true-1-in-python-an-implementation-detail-or-is-it-guarante) decided to play it safe. 
```python
sys.exit(0) if test() else sys.exit(1)
``` 
is also an option. Or we could handle it internally in ``test`` and return an exit code rather than a success bool. I do agree the current implementation is a bit clunky.

jairideout: Why not just:

```python
sys.exit(test())
```

I don't think it hurts to keep the original return code intact.

ebolyen: Nose doesn't return a code, just a boolean, hence the `not` which flips the semantics. I dislike my own suggestion at this point because of the semantic inversion and less because of True or False getting redefined in Py2.x.

jairideout: Ah sorry, missed that in the nose docs

ebolyen: This is probably just splitting hairs at this point, so it's up to you @anderspitman. This PR is good to go either way.

anderspitman: Agreed, worrying about redefinition of ``True``/``False`` is a bit paranoid, but ``sys.exit(not test())`` rubs me wrong too. I vote to merge it as is.

ebolyen: Done! Thanks @anderspitman!

PR #875, opened 2015-03-16
Allows mutli-p powers; changes calculation order

JWDebelius:
* Updates the power calculation to allow tests which return a p value or a 1-dimensional array of p-values. This hopefully will provide support for regressions and timeseries, and GLMs.
* Switches the order of power calculation so counts are drawn before runs. This hopefully represents a more random sample at each point. It is also likely necessary for parallelization. It slightly increases the run time, in `subsample_paired_power` since more subsamples must be drawn. In a test data set, the change did not significantly alter the results.
* Updates documentation values to reflect changes associated with random sampling
* Removes functions not being used.


skbio/stats/power.py
```diff
@@ -821,23 +770,30 @@ def _calculate_power(p_values, alpha=0.05):
 
     """
 
-    w = (p_values < float(alpha)).sum()/float(p_values.shape[0])
+    if len(p_values.shape) == 1:
```
mortonjt: Swap this out with p_values.ndim==1 ?

skbio/stats/tests/test_power.py
```diff
@@ -134,6 +142,24 @@ def test_subsample_power(self):
...
+        test_p, test_c = subsample_power(self.f,
+                                         samples=self.pop,
+                                         num_iter=10,
+                                         num_runs=5,
+                                         draw_mode="matched")
+        self.assertEqual(test_p.shape, (5, 4))
+        npt.assert_array_equal(np.array([10, 20, 30, 40]), test_c)
+
+    def test_subsample_power_multi_p(self):
+        f = lambda x: np.array([0.5, 0.5])
```
mortonjt: If I remember correctly, pep8 hates the old lambda definition.  Use def f(x): return np.array([0.5, 0.5])

Here's the [pep8 reference](https://www.python.org/dev/peps/pep-0008/)

skbio/stats/tests/test_power.py
```diff
@@ -264,77 +329,19 @@ def test__compare_distributions_matched_mode(self):
...
-        test_mean, test_bound = bootstrap_power_curve(self.f,
-                                                      self.pop,
-                                                      self.num_samps,
-                                                      num_iter=100)
-        # Checks the function returned sanely
-        npt.assert_allclose(test_mean, known_mean, rtol=0.05, atol=0.05)
-        npt.assert_allclose(test_bound, known_bound, rtol=0.1, atol=0.01)
+    def test__compare_distributions_multiple_returns(self):
+        known = np.array([[1, 1, 1], [2, 2, 2], [3, 3, 3]])
+        f = lambda x: np.array([1, 2, 3])
```
mortonjt: Same here

skbio/stats/power.py
JWDebelius: Thanks. I didn't know that was possible.

Jorge-C: It's great to see how effectively code reviews spread knowledge across developers :)

JWDebelius:
I believe the decreased coverage has to do with the way coverage is calculated. When I run nosetests with coverage, I get 100% coverage of my code. I think @ElDeveloper and @josenavas ran into the same issue on Qiita.

jairideout:
coverage looks confused again, fine to ignore that "failure"

skbio/stats/power.py
```diff
@@ -565,104 +612,6 @@ def confidence_bound(vec, alpha=0.05, df=None, axis=None):
     return bound
 
 
-def bootstrap_power_curve(test, samples, sample_counts, ratio=None,
```
jairideout: Can you please add this function back and deprecate it? At a minimum we need to give users warning a full release in advance before something is removed. See `skbio.util.flatten` for an example of a deprecated function.

jairideout:
Can you please update the changelog? Important things to note are the deprecation of `bootstrap_power_curve` and any API changes that will affect users.

JWDebelius:
I updated the change log, and replaced `bootstrap_power_curve`. When I check coverage in my desktop it's at 100%.

skbio/stats/power.py
```diff
@@ -49,7 +49,6 @@
     subsample_power
     subsample_paired_power
     confidence_bound
-    bootstrap_power_curve
```
jairideout: Can you add this back? It'll get removed when the deprecated function is removed in 0.3.1.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 ### Deprecated functionality
 * Deprecated `skbio.util.flatten`. This function will be removed in scikit-bio 0.3.1. Please use standard python library functionality
 described here [Making a flat list out of lists of lists](http://stackoverflow.com/a/952952/3639023), [Flattening a shallow list](http://stackoverflow.com/a/406199/3639023) ([#833](https://github.com/biocore/scikit-bio/issues/833))
+* Deprecated `skbio.stats.power.bootstrap_power_curve` will be removed in scikit-bio 0.3.1. It is depreciated in favor of using ``subsample_power`` or ``sample_paired_power`` to calculate a power matrix, and then the use of ``confidence_bounds`` to calculate the average and confidence intervals.
```
jairideout: depreciated -> deprecated

skbio/stats/power.py
```diff
@@ -569,6 +618,12 @@ def bootstrap_power_curve(test, samples, sample_counts, ratio=None,
                           alpha=0.05, mode='ind', num_iter=500, num_runs=10):
     r"""Repeatedly calculates the power curve for a specified alpha level
 
+    .. note:: Deprecated in scikit-bio 0.2.3-dev
+    ``bootstrap_power_curve`` will be removed in scikit-bio 0.3.1. It is
+    depreciated in favor of using ``subsample_power`` or
```
jairideout: depreciated -> deprecated

The text underneath the `.. note::` line needs to be lined up with the `n` in `note` (required by numpydoc/Sphinx).

CHANGELOG.md
jairideout: Can you please add details about other user-facing changes made in this PR? Important things to note are any API changes/additions (e.g., the addition of keyword arguments), bug fixes, and changes in behavior. Please put these descriptions under the appropriate heading(s) in the changelog (e.g., `Features`, `Bug fixes`, `Backward-incompatible changes`, etc.).

jairideout:
Thanks @JWDebelius! Had a few minor inline comments.

skbio/stats/power.py
jairideout: What I mean here is adding this line back so that the docs still link to the deprecated function. Sorry if this wasn't clear before.

CHANGELOG.md
```diff
@@ -38,6 +39,7 @@ described here [Making a flat list out of lists of lists](http://stackoverflow.c
     - Autodetect Python version and disable doctests for Python 3.
 * `numpy` is no longer required to be installed before installing scikit-bio!
 * Upgraded checklist.py to check source files non-conforming to [new header style](http://scikit-bio.org/docs/latest/development/new_module.html). ([#855](https://github.com/biocore/scikit-bio/issues/855))
+* The method of subsampling was changed for ``skbio.stats.power.subsample_paired_power``. Rather than drawing a paired sample for the run and then subsampling for each count, the subsample is now drawn for each sample and each run. In test data, this did not significantly alter the power results.
```
jairideout: Thanks for adding this description! Can you please also describe the new keyword arguments (`ratio`, `num_p`) under the `Features` heading, and what you mean by "multi-p powers"?

skbio/stats/power.py
```diff
@@ -173,6 +174,15 @@ def subsample_power(test, samples, draw_mode='ind', alpha_pwr=0.05,
         "ind" mode should be used.
     alpha_pwr : float, optional
         The critical value used to calculate the power.
+    ratio : 1-D array, optional
+        The fraction of the sample counts which should be
+        assigned to each group. If this is a 1-D array, it must be the same
+        length as `samples`. If no value is supplied (`ratio` is None),
+        then an equal number of observations will be drawn for each sample.
+    num_p : unsigned int, optional
```
jairideout: This parameter isn't available in the function signature.

skbio/stats/power.py
```diff
@@ -311,22 +338,28 @@ def subsample_power(test, samples, draw_mode='ind', alpha_pwr=0.05,
...
+
+    if num_p == 1:
+        power = power[:, :, 0]
 
     return power, sample_counts
 
 
 def subsample_paired_power(test, meta, cat, control_cats, order=None,
-                           strict_match=True, alpha_pwr=0.05,
+                           strict_match=True, alpha_pwr=0.05, ratio=None,
```
jairideout: Please document `ratio` in the docstring.

skbio/stats/power.py
JWDebelius: Sorry. I updated the function so `num_p` is calculated automatically (lines 317-324) and then forgot to remove the description from the doc string.

CHANGELOG.md
```diff
@@ -38,6 +39,8 @@ described here [Making a flat list out of lists of lists](http://stackoverflow.c
     - Autodetect Python version and disable doctests for Python 3.
 * `numpy` is no longer required to be installed before installing scikit-bio!
 * Upgraded checklist.py to check source files non-conforming to [new header style](http://scikit-bio.org/docs/latest/development/new_module.html). ([#855](https://github.com/biocore/scikit-bio/issues/855))
+* The method of subsampling was changed for ``skbio.stats.power.subsample_paired_power``. Rather than drawing a paired sample for the run and then subsampling for each count, the subsample is now drawn for each sample and each run. In test data, this did not significantly alter the power results. 
+* Power calculations (`skbio.stats.power.subsample_paired_power` and `skbio.stats.power.subsample_power`) now accept tests which return a 1-dimensional array of p-values. For example, the method could be applied to a linear model which returns a p value for each variable being tested.<br/>The tests can also calculate power for uneven group sizes (for example, twice as many elements from group A than from Group B). Before this change, the number of elements drawn from Group A and Group B were equal.
```
jairideout: Move this under `Features` and please describe the new keyword argument(s) you added to the public-facing functions.

skbio/stats/tests/test_power.py
jairideout: :+1:

jairideout:
Thanks @JWDebelius! Note that there are a couple of remaining comments from myself and @mortonjt that need to be addressed.

Is anyone available to review the actual code? I focused my review on project/organizational stuff (changelog, docstrings, deprecation).

JWDebelius:
@Jorge-C, would you be able to review the logic/code?

Jorge-C:
Sure, I will review it! There's a merge conflic and tests failures, can you please check those?

JWDebelius:
I had a brain-wave over the weekend for something that might significantly speed up some of the code.

Is it better to try to implement that, now, and make the pull request slightly larger, or get this merged, and then add the functionality?

mortonjt:
Awesome!

Would you like to create another branch, implement your speedup and post
your benchmarks?

Normally it's easier to review smaller PRs, so maybe it may be appropriate
to push another PR implement this speedup once this PR is merged.

JWDebelius:
Thanks. @mortonjt. I will create that branch after this is merged. @Jorge-C, I think the issues have been fixed.

Thank you both!

skbio/stats/power.py
```diff
@@ -295,9 +307,22 @@ def subsample_power(test, samples, draw_mode='ind', alpha_pwr=0.05,
...
                                  'length in "matched" draw_mode.')
 
-    # Checks there are enough samples to subsample
-    if num_ids <= min_observations:
-        raise ValueError('There are not enough samples for subsampling.')
+    # Checks the ratio
+    num_groups = len(samples)
+    if ratio is None:
+        ratio = np.ones((num_groups))
+    ratio = np.asarray(ratio)
```
Jorge-C: Can you please move this casting to an `else` clause?

For the error checking that follows it, I don't mind if you move it or not. It's not necessary when ratio is None, but it's cheap/doesn't hurt to do.

skbio/stats/power.py
```diff
@@ -295,9 +307,22 @@ def subsample_power(test, samples, draw_mode='ind', alpha_pwr=0.05,
...
+        raise ValueError('There must be a ratio for each group.')
+
+    # Determines the number of p-values generated by the statistical test
+    p_return = test(samples)
+    if isinstance(p_return, float):
+        num_p = 1
+    elif isinstance(p_return, np.ndarray) and len(p_return.shape) == 1:
+        num_p = p_return.shape[0]
+    else:
+        raise TypeError('test must return a float or one-dimensional array.')
```
Jorge-C: Not a fan of returning different types depending on the input to `test`, but I don't have a suggestion right now (wanted to add a note to keep track of this :)

skbio/stats/power.py
```diff
@@ -483,6 +519,23 @@ def subsample_paired_power(test, meta, cat, control_cats, order=None,
...
+
+    # Determines the number of p values returned by the test
+    p_return = test(sub_ids)
+    if isinstance(p_return, float):
+        num_p = 1
+    elif isinstance(p_return, np.ndarray) and len(p_return.shape) == 1:
+        num_p = p_return.shape[0]
+    else:
+        raise TypeError('test must return a float or one-dimensional array.')
+
```
Jorge-C: This code appears one more time, so it's probably worth refactoring into a helper function (will make it easier to keep updated/in synch).

skbio/stats/power.py
```diff
@@ -850,7 +919,7 @@ def _compare_distributions(test, samples, counts=5, mode="ind", num_iter=1000):
         this may be useful when working with regression data where
         :math:`x_{1}, x_{2}, ..., x_{n}` maps to :math:`y_{1}, y_{2}, ... ,
         y_{n}`.
-    num_iter : int, optional
+    num_iter : unsigned int, optional
```
Jorge-C: I think we shouldn't use `unsigned int` in our docs talking about scalars. We should instead say `positive` or `non negative int` depending on what we want. `unsigned int` is incorrect because it doesn't exist as a type in python, and seems hard for users to understand what it means unless they've programmed in C/...

If the input is an array, and the `dtype` is `uint` it's fine to specify that.

skbio/stats/power.py
```diff
@@ -942,22 +1014,17 @@ def _calculate_power_curve(test, samples, sample_counts, ratio=None,
     num_iter : int
         The default is 1000. The number of p-values to generate for each point
         on the curve.
-
```
Jorge-C: @jairideout doc whisperer, is it OK to remove empty lines between sections in the docstrings?

jairideout: I'm not sure if Sphinx/numpydoc cares, but probably good to add these lines back for readability (e.g., if a user is reading the help in an ipython session).

Jorge-C: Thanks Jai!

skbio/stats/power.py
```diff
@@ -821,23 +883,30 @@ def _calculate_power(p_values, alpha=0.05):
 
     """
 
-    w = (p_values < float(alpha)).sum()/float(p_values.shape[0])
+    if p_values.ndim == 1:
+        p_values = p_values * np.array([[1]])
+
+    w = (p_values < float(alpha)).sum(1)/float(p_values.shape[1])
```
mortonjt: Looks like axis 1 is being summed here correct?

May make it more clear if `sum(axis=1)` was used instead.

skbio/stats/power.py
JWDebelius: The return depends on how `test` is structured, not the input. 
`test` can be any statistical test which returns a float or a 1-dimensional array of floats.

JWDebelius:
Getting a documentation failure in the python 3 build.

Jorge-C:
I've restarted the build, it was a connectivity issue to the biom format
website.

JWDebelius:
@Jorge-C, just checking to see if you've had a chance to look over the new changes?

Thank you!

Jorge-C:
No, sorry, I'll try to review it in the next few days.

skbio/stats/power.py
```diff
@@ -821,24 +830,31 @@ def _calculate_power(p_values, alpha=0.05):
 
     """
 
-    w = (p_values < float(alpha)).sum()/float(p_values.shape[0])
+    if p_values.ndim == 1:
+        p_values = p_values * np.array([[1]])
```
mortonjt: Would it be preferable to do `p_values = np.atleast_2d(p_values)` instead?

skbio/stats/power.py
```diff
@@ -821,24 +830,31 @@ def _calculate_power(p_values, alpha=0.05):
...
+def _compare_distributions(test, samples, num_p, counts=5, mode="ind",
+                           num_iter=1000):
     r"""Compares two distribution arrays iteratively
 
     Parameters
     ----------
     test : function
         The statistical test which accepts an array_like of sample ids
-        (list of lists) and returns a p-value.
+        (list of lists) and returns a p-value. This can be a one-dimesnional
```
mortonjt: Should be `one-dimensional`

skbio/stats/power.py
```diff
@@ -821,24 +830,31 @@ def _calculate_power(p_values, alpha=0.05):
 
     """
 
-    w = (p_values < float(alpha)).sum()/float(p_values.shape[0])
+    if p_values.ndim == 1:
+        p_values = p_values * np.array([[1]])
+
+    w = (p_values < float(alpha)).sum(axis=1)/float(p_values.shape[1])
```
mortonjt: This isn't critical.

But this is completely equivalent to  `w = (p_values < alpha).sum(axis=1)/p_values.shape[1]`
since Python 3 division is being imported through `from __future__ import division`
So the `float` conversion is unnecessary

skbio/stats/power.py
```diff
@@ -908,11 +924,131 @@ def _compare_distributions(test, samples, counts=5, mode="ind", num_iter=1000):
             subs = [np.random.choice(np.array(pop), counts[i], replace=False)
                     for i, pop in enumerate(samples)]
 
-        p_values[idx] = test(subs)
+        p_values[:, idx] = test(subs)
+
+    if num_p == 1:
+        p_values = np.hstack(p_values)
```
mortonjt: Another way to do this operation is to do `p_values.squeeze()`

skbio/stats/power.py
```diff
@@ -908,11 +924,131 @@ def _compare_distributions(test, samples, counts=5, mode="ind", num_iter=1000):
...
 
     return p_values
```
mortonjt: Noticed that the doc-string hasn't been updated to reflect the change to `p_values`.
It would be nice to mention this change (i.e. the returned `p_values` array has dimensions `num_p` and `num_iter`)

skbio/stats/power.py
```diff
@@ -908,11 +924,131 @@ def _compare_distributions(test, samples, counts=5, mode="ind", num_iter=1000):
...
+        The fraction of the sample counts which should be
+        assigned to each group. If this is a 1-D array, it must be the same
+        length as `samples`. If no value is supplied (`ratio` is None),
+        then an equal number of observations will be drawn for each sample. In
+        `matched` mode, this will be set to one.
+    min_observations : positive int, optional
+        The minimum number of observations in any sample to perform power
+        analysis. Note that this is not the same as the minimum number of
+        samples drawn per group.
+        max_counts : positive int, optional
```
mortonjt: Don't forget to correct this indentation

skbio/stats/power.py
```diff
@@ -476,44 +485,34 @@ def subsample_paired_power(test, meta, cat, control_cats, order=None,
...
+                                        num_p=num_p,
+                                        counts=count,
+                                        num_iter=num_iter,
+                                        mode="matched")
+            power[id1, id2, :] = _calculate_power(ps, alpha_pwr)
+
+    if num_p == 1:
+        power = power[:, :, 0]
 
     return power, sample_counts
```
mortonjt: Same thing here. It would be nice to specify the dimensions of `power` and `sample_counts` in the doc-string

JWDebelius: There's not a fixed length for either of these, and the length is dependent on four other variables. It depends on the minimum group size, `min_counts`, `max_counts`, and `counts_interval`.

mortonjt: Exactly.  But especially looking at `power` it isn't immediately obvious which variables are responsible for which dimension until I look at the source code.

skbio/stats/power.py
```diff
@@ -476,44 +485,34 @@ def subsample_paired_power(test, meta, cat, control_cats, order=None,
...
-        # Calculates the power curve
-        power[id1, :] = _calculate_power_curve(test,
-                                               sub_ids,
-                                               sample_counts,
-                                               num_iter=num_iter,
-                                               alpha=alpha_pwr,
-                                               mode="matched")
+    power = np.zeros((num_runs, len(sample_counts), num_p))
+
+    # Calculates power instances
```
Jorge-C: This code is almost duplicated in `subsample_power`. Can it be refactored to avoid duplication?

skbio/stats/power.py
Jorge-C: :+1:

skbio/stats/power.py
JWDebelius: The relationship for length isn't clean. Power could be defined in terms of `sample_counts`, but ultimately, there's not much you're going to get out of the dimensions that will be useful.

mortonjt: Maybe I'm not being clear ...

Whenever I have implemented functions with complicated outputs (e.g. multidimensional arrays and dictionaries), I have always regretted not documenting them better.

Looks like this is a public method, so users could be potentially tampering with this method.  The output of this method could really confuse some users.

If you can state that
- the first dimension of `power` is the number of runs
- the second dimension of `power` is the number or sample counts
- the third dimension of `power` is number of p-values returned from a specific test

This function will become so much easier to use

Jorge-C: I agree this would be fantastic to make it easier to understand and work with.

skbio/stats/power.py
JWDebelius: Not with major refactoring of the way `paired_subsamples` returns samples.

Under the current set up, the critical difference is that a paired subsample is drawn for ever iteration. This increases the number of Monte Carlo simulations, and helps to address a bug in the current code.
`paired_subsamples` could be refactored to return all the subsampled pairs, and _calculate_power_curve's draw could be slightly refactored to support the correct drawing, and this could be set up as boiler plate.

JWDebelius:
Working with the code, I found a major bug (see #910). I'd like to close this PR as it requires extensive changes to the code base, and then use this as a spring board for fixing the issue.

Issue #830, opened 2015-02-12
What to do about skbio.spatial?


ebolyen:
`skbio.spatial` currently only has a single function `procrustes`. Scipy has a partial implementation of this (the general case), so should we adapt it and move our `procrustes` into ordination as that is how it is used in scikit-bio?

gregcaporaso:
related to scipy/scipy#3786

If we could eventually drop ours and use SciPy's, that'd be ideal.

ElDeveloper:
It would be nice if we can figure out how to use that function from our procrustes implementation. When I was trying to do that, I encountered the following problem:

https://github.com/scipy/scipy/issues/3994

Which prevented me (at that time) from moving the code out of skbio and into scipy.

jairideout:
> If we could eventually drop ours and use SciPy's, that'd be ideal.

:+1: 

@ElDeveloper are you interested in taking the lead on this issue, since you've already started working on it? If so, I'll assign to you.

ElDeveloper:
I am! However, I couldn't find a good way to use their function without
special-casing for our needs (which in general seems like a bad
practice), but that's where I got stuck at. I'll take another round at
this piece of code this Friday or if not earlier and report back.

jairideout:
Awesome, thanks @ElDeveloper!

ElDeveloper:
PR in https://github.com/scipy/scipy/pull/4540

jairideout:
Thanks! Let's leave this issue open so that we can deprecate scikit-bio's `spatial` subpackage once `procrustes` is in a scipy release.

ElDeveloper:
Sounds good.

jairideout:
Noticed that scipy/scipy#4540 was merged, thanks for working on this @ElDeveloper!

ElDeveloper:
Not a problem, the code review helped improve the code a lot (mainly by making better use of the numpy API).

ElDeveloper:
https://github.com/scipy/scipy/releases/tag/v0.16.0 woot! :beers:

jairideout:
:shipit:

Issue #825, opened 2015-02-12
refactor skbio.stats.ordination subpackage


jairideout:
`A user executes ordination techniques using a simple procedural API.`
- [x] convert objects to functions (#829)
- [ ] columns of all zeros (#292)
- [ ] Fix OrdinationResults (#521, #665)

Depends on:
- [ ] contingency table (#821)

mortonjt:
I believe this is done

PR #879, opened 2015-03-20
Additional decomposition methods and approximate PCoA

wasade:
~~**DO NOT MERGE**~~

~~**NOT READY FOR REVIEW**~~

Laying out the initial proposed changed to the PCoA object to support changing the eigen decomposition used as well as supporting approximate PCoA methods. The approximate methods do not actually do a eigen decomposition, and there isn't (AFAIK) and analog to eigenvalues and proportion explained hence the need to be able to redefine the `scores` method. However, the scores method is potentially dependent on the decomposition being done, so the present structure may not be sufficient 

Specific examples of the methods being ported can be found [here](https://github.com/pycogent/pycogent/compare/master...antgonza:principal_coordinates_new_methods). 
cc @antgonza @Jorge-C @mortonjt  

To clarify, much of the documentation and methods being merged in were originally implemented by @antgonza and others. The primary source for all this can be found [here](https://github.com/pycogent/pycogent/compare/master...antgonza:principal_coordinates_new_methods).


skbio/stats/ordination/_principal_coordinate_analysis.py
```diff
@@ -24,7 +25,94 @@
 #   so, so I'm not doing that.
 
 
-class PCoA(Ordination):
+def _eigh(F_matrix, **ignored):
+    """Wrapper around np.linalg.eigh
```
mortonjt: Is this **ignored keyword necessary?

Jorge-C:
Tip, originally from @ElDeveloper: for doc changes, we can save some builds adding `[skip ci]` to the commit message.

skbio/stats/ordination/_principal_coordinate_analysis.py
wasade: we need to support `**kwargs` for the methods, and in this method they are ignored

mortonjt:
I've been taking a closer look at the ssvd method - its pretty freakin cool!

One thing that I think would be really awesome is if we could separate this decomposition methods from the PCoA methods.  In pycogent, the entire ssvd functionality is never exploited, but having all of the functionality of an svd could be immensely helpful (i.e. sample/feature ordination for giant datasets)

Perhaps, we could create a separate file, like `skbio.linalg`, to place all of these decomposition methods in.  Then we could let the PCoA class(es) be used for summary results.

wasade:
@ebolyen @jairideout, when you get a second, could you comment on where you think the decomp methods should go? I like the idea of them being accessible in `skbio.linalg`, however that does suggest we may want to push them back into scipy or something. On the other hand, we'd like to get these methods available as soon as possible for the purposes of benchmarking

jairideout:
> could you comment on where you think the decomp methods should go?

I'm not in favor of adding a `linalg` subpackage to scikit-bio because ideally that functionality should be added to a lower-level/general package (e.g., scipy) that scikit-bio depends on. We're trying to keep scikit-bio's public API focused on bioinformatics, biology, ecology, etc. as much as possible and have started moving this sort of functionality to other packages (e.g., moving `procrustes` to scipy in #830).

If scipy (or another package) isn't an option, does it make sense to put the decomposition functionality in the new `skbio.stats.composition` subpackage in #860?

> we'd like to get these methods available as soon as possible for the purposes of benchmarking

It might be worth waiting until benchmarks have been performed before adding this functionality to scikit-bio. Once it is included in the codebase, additional developer time must be spent on package maintenance and support, so ideally we'd like this functionality to have been benchmarked before inclusion (so that there is a clear reason why it is included in scikit-bio, when users should use these methods, etc.).

wasade:
I should step back, benchmarks have been performed against cogent. These
were pending merge there. I agree about lower level packages, but at the
same time, we have a clear and specific bioinformatic use for these already.

At the minimum, we need the objects here to be revisited as they do not
support swapping out the decomp method or approximate methods. We can do
the implementations of the methods separately but it will be painful if
skbio isn't reasonably adapted first.

antgonza:
These methods are basically relying on methods already implemented in
scipy but adapted for our biological usage. However, these methods can
be used in different ordination methods. With this in mind, I think
the best will be to add them in another subpackage (like
stats.eigen_decomposition).

Now, all these methods have been tested and are either the classic
approach, really good for performance or to teach
eigenvalue/eigenvector approximations (their pros/cons).

antgonza:
Sorry, forgot to add: We could have a call to discuss this so we avoid
multiple emails ...

jairideout:
@wasade sounds good, thanks for clarifying what you meant by benchmarking these methods. For now, I recommend focusing on the high-level public API that users will interact with to perform ordination, since we already know this needs to exist and makes sense living in scikit-bio. We can then investigate where the lower-level decomposition methods should live (in scipy, scikit-bio, or elsewhere). This can happen in parallel with this PR, or even after merge. Does this sound like a reasonable plan to move forward?

@antgonza I'm not sure I understand your comment about these methods already being in scipy. If they are in scipy or another package that scikit-bio relies on, we should be using them here and not redefine/duplicate them. Apologies if I misunderstood your comment.

antgonza:
Sorry for not being clear. There are two general cases of methods we
want to add. (1) Well know methods to calculate PCoA. Examples will be
eigh and svd. They exists in scipy.linalg but we need to call them in
a specific way so they can be used. (2) Methods which are
approximations, like Nystrom or SCMDS, which are basically "tricks" to
have fast PCoAs, which also rely on on methods of (1). Basically, we
were thinking in something like alignments:
http://scikit-bio.org/docs/latest/alignment.html.

mortonjt:
That makes sense.

Keep in mind, the original intent for creating a separate package for these decomposition methods is to make these accessible across all of the skbio.  PCoA isn't the only package that could benefit from these decomposition methods - `canonical_correspondence_analysis`, `correspondence_analysis` and `redundancy_analysis` classes can all benefit from these methods since they all use SVD.

I have a vibe that scipy wouldn't want these decomposition methods in their package - their are several different decomposition methods out there and they are only supporting 2 different decomposition methods.   

Looks like scikit-learn had the same problem - they placed all of their decomposition methods into the `sklearn.decomposition` package - it may fit in scikit-learn better than in scipy.  The other option is to do something similar and have all of these functions directly accessible from the `skbio.stats.ordination` package and place the functions within a file such as `skbio/stats/ordination/decomposition.py`.

gregcaporaso:
I agree - I think the best thing to do will be have high level ordination API go in skbio.ordination, and lower level stuff be private and maybe only temporarily in skbio, ultimately moving to a lower-level package where it could more widely benefit the python community (sklearn is a great thought for that @mortonjt, and scipy would also of course be good if they're interested). 

Thanks for the work on this @wasade!

wasade:
@mortonjt, I'm going to pull out SSVD and drop it into a benchmark-specific IPython notebook, which I'll place in a separate branch. I'll also pull out the `APCoA` class.

@jairideout and @gregcaporaso, would you prefer this PR then stick with essentially what is in place (minus the bits I'm pulling out) or would you like it to scope wider? It is possible the `OrdinationResults` object will be insufficient but we won't know until we proceed further.

wasade:
As an update, I've created a new repo and stubbed out the benchmark [notebook](http://nbviewer.ipython.org/github/wasade/miniature-octo-dubstep/blob/master/ipynb/PCoA%20method%20benchmarking.ipynb).

gregcaporaso:
Awesome, thanks @wasade. Looking forward to seeing the results.

I'll let @jairideout answer your question about the scope of this PR as he's been more closely involved with it than I have.

jairideout:
@wasade could you elaborate on what you mean by making this PR/code have a wider scope?

That's totally fine if `OrdinationResults` needs changes to accommodate the new methods (we are planning on revisiting the ordination API in #825).

wasade:
@jairideout, specific to `_principal_coordinates_analysis` or not

wasade:
I think this is good for review

antgonza:
Looks good but could you add some tests? Perhaps for the wrappers you could borrow the ones from scipy.

wasade:
Not sure what needs to be tested? I disagree with testing these wrappers given their present complexity -- coverage in this case doesn't add anything meaningful. This is just a shuffling around of logic at this point and uses the existing test code that is in place. The only area that really has new logic is the PCoABase constructor, which I could add a test for I guess? It isn't clear to me what the tangible benefit is though.

skbio/stats/ordination/__init__.py
```diff
@@ -113,12 +113,13 @@
 from ._correspondence_analysis import CA
 from ._redundancy_analysis import RDA
 from ._canonical_correspondence_analysis import CCA
-from ._principal_coordinate_analysis import PCoA
+from ._principal_coordinate_analysis import PCoA, PCoABase
```
Jorge-C: I'm not sure we should import the base class here (nor export it in `*` imports).

ebolyen: :+1: Base classes probably don't make sense to export generally, unless you would expect users to extend said base classes.

skbio/stats/ordination/_principal_coordinate_analysis.py
```diff
@@ -24,7 +25,58 @@
...
+        The resulting k eigenvalues.
+    array
+        The resulting k eigenvectors.
+
+    References
+    ----------
+    .. [1] http://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.eigsh.html  # noqa
+    """
+    val, vec = sp.sparse.linalg.eigsh(F_matrix, k=k, **kwargs)
+    return val, vec.T
```
Jorge-C: Is this transpose correct? It makes `eigsh` inconsistent with `eigh` and so it'll break the rest of the code if used.

mortonjt: Very good point - completely missed that.

skbio/stats/ordination/__init__.py
wasade: It is necessary for our benchmarking as were adding in a derived class. Can
import direct but that is not assured to be stable

ebolyen: That seems reasonable, another option is to bump `_principle_coordinate_analysis` to public so that it may be imported there. This feels like a better solution, though I am not sure why. Any thoughts @jairideout?

mortonjt: If this is only necessary for benchmarking, can we just remove this import just before the PR is merged?
The benchmarking will be complete by then correct?

ebolyen: I had assumed the goal was a publishable benchmark, is this correct?

wasade: Yes

Jorge-C:
Do you guys think the private thin wrappers `_eigh` and `_eigsh` are necessary? Some methods will end up needing actual functions to do stuff, but it seems overkill for these two.

wasade:
they're being passed `**kwargs`. doesn't need to be wrappers, but i think
the other options would be more complex?

Jorge-C:
@wasade made a [PR that removes](https://github.com/wasade/scikit-bio/pull/4) thin wrappers to compare both approaches.

Jorge-C:
:+1:

jairideout:
@wasade I reviewed the code and have some comments about the overall design, with some recommendations about how this could be simplified to fit the existing ordination code.

I don't think we need a `PCoABase` class for the functionality you're adding here. If I understand correctly, the current `PCoA` class uses `np.linalg.eigh` internally, and you are expanding on this to allow users to specify a different eigen function.

To support this functionality, `PCoA` can have two optional parameters added to the constructor (currently called `eig_f` and `eig_kwargs`). `eig_f` can default to the current eigen function (numpy's), and we can support other eigen functions either as string names or a function that meets some signature spec. Then, `PCoA._pcoa` can use the specified eigen function.

Does this design make sense? I think it is just as flexible as the current code and should be simpler to implement and maintain.

wasade:
`PCoABase` is in place because the `scores` method will differ for some of the approximate methods and we need an easy way to define what scoring method is in use. The same type of indirection used with the eigen methods can also be used here I believe based on what is being done in the processing notebook, as long as there is a 1-1 correspondence between the eigen methods and the scoring methods.

jairideout:
Are there plans to add the approximate methods to this PR? If not, I recommend using the simplified design I proposed and refactoring when the new methods are ready in a separate PR.

wasade:
The [decision](https://github.com/biocore/scikit-bio/pull/879#issuecomment-85740474) was to not put the approximate methods in here at this time, but to wait for a manuscript to be released. We can of course change this, and put the methods in this PR. 

If we refactor as you suggest then we may as well just close out this PR. That refactor would require that we redevelop the objects in the notebook anyway, and that we move forward independently of the PCoA module in skbio.

gregcaporaso:
@wasade and I just discussed offline (and I discussed with @jairideout and @ebolyen during that discussion). Here's what we decided:

For now, we're going to close this pull request, thereby not adding the new ``PCoABase`` class. Instead, for @wasade's benchmarks, the plan should be one of two things:
 1. copy the code and tests into the benchmarking repo and modify as necessary (e.g., just using the current code in this PR), or 
 2. define that repo to depend specifically on scikit-bio 0.2.3, in which case there are no practical problems with overridding the ``PCoA`` class, or just calling ``PCoA._E_matrix``/``PCoA._F_matrix``, which are static. 

Either of these would be easy to achieve, and it will reduce workload for scikit-bio by not creating a new class which would almost immediately be deprecated (see #825 for context on that). 

Again, sorry for the moving target that scikit-bio is right now. We know that it is inconvenient, but we're in alpha and a lot is changing before we get to beta (as laid out in [the roadmap to beta](https://github.com/biocore/scikit-bio/wiki/beta-release-roadmap)).

wasade:
Sounds good, thanks @gregcaporaso, @jairideout and @ebolyen!

ebolyen:
Excited to see how your benchmarks turn out @wasade!

jairideout:
Thanks @wasade!

PR #882, opened 2015-03-24
ENH: TreeNode.tip_tip_distance altered to warn with RepresentationWarnin...

### LINKED ISSUE #791: TreeNode.tip_tip_distances doesn't work with GG 13_8 trees
I'm not sure if this is an issue with the ``TreeNode`` or the tree itself, but I get this error with all of the different trees in this build of Greengenes. @wasade, do you know what's going on here? I confirmed that the md5 of the tree that I'm working with locally is the same as the md5 of the tree on ftp://greengenes.microbio.me.  

```python
In [1]: from skbio import TreeNode

In [2]: t = TreeNode.read('/Users/caporaso/data/gg_13_8_otus/trees/70_otus.tree')

In [3]: t.tip_tip_distances()
NoLengthError                             Traceback (most recent call last)
<ipython-input-3-0105b92aef7c> in <module>()
----> 1 t.tip_tip_distances()

/Users/caporaso/.virtualenvs/qiime/lib/python2.7/site-packages/skbio/tree/_tree.pyc in tip_tip_distances(self, endpoints)
   2740                 if child.length is None:
   2741                     raise NoLengthError("Node with name '%s' doesn't have a "
-> 2742                                         "length." % child.name)
   2743
   2744                 distances[child.__start:child.__stop] += child.length

NoLengthError: Node with name 'k__Bacteria' doesn't have a length.

In [4]: t = TreeNode.read('/Users/caporaso/data/gg_13_8_otus/trees/70_otus_unannotated.tree')

In [5]: t.tip_tip_distances()
NoLengthError                             Traceback (most recent call last)
<ipython-input-5-0105b92aef7c> in <module>()
----> 1 t.tip_tip_distances()

/Users/caporaso/.virtualenvs/qiime/lib/python2.7/site-packages/skbio/tree/_tree.pyc in tip_tip_distances(self, endpoints)
   2740                 if child.length is None:
   2741                     raise NoLengthError("Node with name '%s' doesn't have a "
-> 2742                                         "length." % child.name)
   2743
   2744                 distances[child.__start:child.__stop] += child.length

NoLengthError: Node with name 'None' doesn't have a length.
```

MikeDeberg:
TreeNode.tip_tip_distance altered to warn with RepresentationWarning when substituting None lengths with 0(zero).

fixes #791

### DIFF
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -8,6 +8,7 @@
 
 from __future__ import absolute_import, division, print_function
 
+import warnings
 from operator import or_
 from copy import deepcopy
 from itertools import combinations
@@ -23,6 +24,7 @@ from skbio._base import SkbioObject
 from skbio.stats.distance import DistanceMatrix
 from ._exception import (NoLengthError, DuplicateNodeError, NoParentError,
                          MissingNodeError, TreeError)
+from skbio.util import RepresentationWarning
 
 
 def distance_from_r(m1, m2):
@@ -2430,10 +2432,14 @@ class TreeNode(SkbioObject):
             starts, stops = [], []  # to calc ._start and ._stop for curr node
             for child in node.children:
                 if child.length is None:
-                    raise NoLengthError("Node with name '%s' doesn't have a "
-                                        "length." % child.name)
+                    warnings.warn("Node with name '%s' doesn't have a "
+                                  "length." % child.name,
+                                  RepresentationWarning)
 
-                distances[child.__start:child.__stop] += child.length
+                    distances[child.__start:child.__stop] += 0
+
+                else:
+                    distances[child.__start:child.__stop] += child.length
 
                 starts.append(child.__start)
                 stops.append(child.__stop)
--- a/skbio/tree/tests/test_tree.py
+++ b/skbio/tree/tests/test_tree.py
@@ -9,6 +9,7 @@
 from __future__ import absolute_import, division, print_function
 
 from unittest import TestCase, main
+import warnings
 
 import numpy as np
 import numpy.testing as nptest
@@ -18,6 +19,7 @@ from six import StringIO
 from skbio import DistanceMatrix, TreeNode
 from skbio.tree import (DuplicateNodeError, NoLengthError,
                         TreeError, MissingNodeError, NoParentError)
+from skbio.util import RepresentationWarning
 
 
 class TreeTests(TestCase):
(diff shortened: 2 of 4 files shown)


ebolyen:
@jairideout Can you review? Should be a quick one.

skbio/util/_warning.py
```diff
@@ -20,3 +20,14 @@ class EfficiencyWarning(Warning):
 
     """
     pass
+
+
+class RepresentationWarning(Warning):
```
gregcaporaso: :+1: this is a great idea

skbio/tree/_tree.py
```diff
@@ -2430,10 +2432,14 @@ def update_result():
             starts, stops = [], []  # to calc ._start and ._stop for curr node
             for child in node.children:
                 if child.length is None:
-                    raise NoLengthError("Node with name '%s' doesn't have a "
-                                        "length." % child.name)
+                    warnings.warn("Node with name '%s' doesn't have a "
```
jairideout: Can you please update this warning message to explain that a length of zero will be used?

jairideout: It'd also be helpful to include the name of the class and method that the warning applies to (`TreeNode.tip_tip_distances`) because it may not be obvious to users where this warning is being raised from.

skbio/tree/_tree.py
```diff
@@ -2430,10 +2432,14 @@ def update_result():
...
+                                  "length." % child.name,
+                                  RepresentationWarning)
 
-                distances[child.__start:child.__stop] += child.length
+                    distances[child.__start:child.__stop] += 0
```
jairideout: Use `0.0` since lengths are internally stored as floats. The current implementation will work but I prefer sticking with a single data type for consistency.

skbio/tree/_tree.py
```diff
@@ -2430,10 +2432,14 @@ def update_result():
...
+
+                else:
+                    distances[child.__start:child.__stop] += child.length
```
jairideout: Minor stylistic comment: there's a small amount of duplication in the if/else branches (`distances[child.__start:child.__stop] += ...`). It'd be cleaner to store the length in a local variable within the if/else and then have a single statement after the if statement that sets the value.

skbio/tree/tests/test_tree.py
```diff
@@ -598,8 +600,22 @@ def test_tip_tip_distances_non_tip_endpoints(self):
 
     def test_tip_tip_distances_no_length(self):
         t = TreeNode.read(StringIO(u"((a,b)c,(d,e)f);"))
-        with self.assertRaises(NoLengthError):
-            t.tip_tip_distances()
+        with warnings.catch_warnings(record=True):
+            warnings.simplefilter("error")
+            with self.assertRaises(RepresentationWarning):
+                t.tip_tip_distances()
```
jairideout: Can you please assert the return value here matches what is expected?

ebolyen: @MikeDeberg tried this at first, but execution of the block ends as soon as the `RepresentationWarning` is sent, so variables never get set, etc... You'll notice that this is checked in the test below because we just ignore the warning. (Do you know a way around this?)

skbio/util/_warning.py
```diff
@@ -20,3 +20,14 @@ class EfficiencyWarning(Warning):
...
+    """Warn about substitions, assumptions, or particular alterations
```
jairideout: substitions -> substitutions

The first sentence of the docstring needs to be on a single line that is <= 79 characters, otherwise it gets truncated.

skbio/tree/tests/test_tree.py
jairideout: Interesting! I recommend using `numpy.testing.assert_warns`, which simultaneously asserts that the warning is raised and allows a return value to be passed through. See `skbio/io/tests/test_base.py` for an example of how this is used. This also avoids having to use `warnings` to catch and filter warnings.

jairideout:
Thanks @MikeDeberg! I'm done reviewing, please see my inline comments.

Can you also update `CHANGELOG.md` to note the change in behavior of `TreeNode.tip_tip_distances`? This could go under either the `Bug fixes` or `Features` heading. I'm leaning towards `Features` because this really wasn't a bug (the appropriate error message was being raised before). We're just making the method more flexible.

Also, please note the addition of `skbio.util.RepresentationWarning` under `Miscellaneous`.

jairideout:
Closing in favor of #1028, which contains the commits here and updates based on code review.

PR #883, opened 2015-03-24
STY: Add `#  noqa` to __init__.py line

ElDeveloper:
Use `#  noqa` to supress an error instead of adding a "spurious
statement" to silence flake8.


ebolyen:
Awesome! Thanks @ElDeveloper!

PR #886, opened 2015-03-31
Optional dependencies

### LINKED ISSUE #760: Required and recommended dependencies

Jorge-C:
This includes all commits from @dansondergaard in #802, and finishes the changes necessary to be able to run the test suite with no matplotlib installed, including adding a build to travis and doc changes (ie, most of the work necessary to close #760).

I'm using @jairideout's suggestion, found in pandas, to skip tests that need matplotlib.


skbio/util/_testing.py
```diff
@@ -94,3 +94,12 @@ def get_data_path(fn, subfolder='data'):
...
     return data_path
+
+
+def _skip_if_no_matplotlib():
+    """Skip rest of a test if matplotlib is not available."""
+    try:
+        import matplotlib  # noqa
+    except ImportError:
+        import nose
+        raise nose.SkipTest("matplotlib not installed")
```
ebolyen: It might be better to raise `SkipTest` from the [stdlib](https://docs.python.org/2/library/unittest.html#unittest.SkipTest) instead of Nose. Unless there was a specific reason you are using nose's skiptest.

skbio/util/_testing.py
```diff
@@ -94,3 +94,12 @@ def get_data_path(fn, subfolder='data'):
     path = os.path.dirname(os.path.abspath(callers_filename))
     data_path = os.path.join(path, subfolder, fn)
     return data_path
+
+
+def _skip_if_no_matplotlib():
```
ebolyen: This would be incredibly awesome if it was a decorator, then it acts as a bit of a "living comment". It could even be generalized to something like:

```
@skip_if_missing("matplotlib")
def test_something(self):
    pass
```
You can use [import_module](https://docs.python.org/3/library/importlib.html#importlib.import_module) to import via a string.

skbio/draw/tests/test_distributions.py
```diff
@@ -12,7 +12,10 @@
 
 import numpy as np
 import numpy.testing as npt
-import matplotlib.pyplot as plt
+try:
```
ebolyen: Would it be worthwhile to create a utility context-manager with would handle the exception and skip it?

skbio/util/_testing.py
ElDeveloper: There's also Numpy's skipif decorator.
http://docs.scipy.org/doc/numpy/reference/generated/numpy.testing.decorators.skipif.html

ebolyen:
@Jorge-C These updates to @dansondergaard PR look really good! Had a couple of comments in regards to the utility functions.

dansondergaard:
Looks absolutely great! I didn't want to mess with the testing setup since, honestly, I'm not very experienced with TravisCI :) Thanks for doing this, @Jorge-C!

skbio/util/_testing.py
Jorge-C: I've pushed some changes to use @ElDeveloper's suggestion to use numpy's decorator.

Though, to answer @ebolyen's comment, I was using the nose's version for consistency (it's exactly the same as unittest's `SkipTest`, but it's easier to import `nose` and know that everything is there than to remember whether something is in unittest or not).

skbio/draw/tests/test_distributions.py
Jorge-C: I added a comment for future reference! Not sure if hiding it behind some sort of context manager could make it simpler or more maintainable.

OTOH, this made me think that this approach of trying to import optional dependencies on top, and passing in the except clause looks much neater and easier to maintain than repeating imports in every single method/function that needs them... Thoughts?

README.rst
```diff
@@ -21,7 +21,17 @@ To install the latest release of scikit-bio::
 
     pip install scikit-bio
 
-Equivalently, you can use the ``conda`` package manager available in `Anaconda <http://continuum.io/downloads>`_ or `miniconda <http://conda.pydata.org/miniconda.html>`_ to install scikit-bio and its dependencies without having to compile them::
+Some plotting features require extra dependencies, and an
+``ImportError`` will be raised if you try to use them without them.
```
squirrelo: This wording is a bit confusing. Maybe something along the lines of 
``ImportError`` will be raised if you try to use packages without these extra dependencies installed.

Jorge-C: Good catch, using `them` with two meanings one word apart is horrible style, I'll fix it. Thanks!

dansondergaard: Must have been tired when I wrote that. The whole sentence sounds weird.

Jorge-C: Oh, it's in fact my fault! I wrote that without realizing it was nonsensical. My :coffee: to English process failed ;)

mortonjt:
What is the status of this @Jorge-C ?  This is a critical component in easing the installation process for scikit-bio - mainly due to matplotlib.

jairideout:
This will take some work to get in place. Can you use `conda` to install scikit-bio in the meantime? That should clear up the matplotlib install issues.

mortonjt:
That means we'll need to resolve the conda installation issues for linux ...

jairideout:
It should have been resolved for the latest release (0.5.0), use the Anaconda.org channel (default channel) instead of biocore:

```python
conda install scikit-bio
```

Does that work for you?

wasade:
Closing as this appears to have gone stale, please reopen if this should be finished and merged

PR #890, opened 2015-04-02
[master] BUG: fixed TreeNode.extend when given tree.children; fixes #889

### LINKED ISSUE #889: [0.2.3] TreeNode.extend seems to mangle the results
@JTFouquier was working with the `TreeNode` object and was unable to make extend produce the expected results.

Here we create two subtrees and `append` them to a specific node on a scaffold.
http://hastebin.com/ewihulekib.py

When we change this to `extend` on the children, we would expect the root to be substituted by the scaffold node.
http://hastebin.com/qevonatuha.py
However, certain children are skipped in the extend resulting in two mangled trees.

This also occurs when using [`from_newick`](https://github.com/biocore/scikit-bio/blob/0.2.3/skbio/tree/_tree.py#L1945) instead of the new IO:
http://hastebin.com/ibuhicayaq.py

I believe this will require a `0.2.4` release.

ebolyen:
This fixes an issue where doing `tree1.extend(tree2.children)` or `TreeNode(.... children=tree2.children)` would leave both trees broken.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Bug fixes
 * Changed `BiologicalSequence.distance` to raise an error any time two sequences are passed of different lengths regardless of the `distance_fn` being passed. [(#514)](https://github.com/biocore/scikit-bio/issues/514)
+* Fixed issue with ``TreeNode.extend`` where if given the children of another ``TreeNode`` object (``tree.children``), both trees would be left in an incorrect and unpredictable state. ([#889](https://github.com/biocore/scikit-bio/issues/889))
 
 ### Deprecated functionality
 * Deprecated `skbio.util.flatten`. This function will be removed in scikit-bio 0.3.1. Please use standard python library functionality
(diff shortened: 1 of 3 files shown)


ElDeveloper:
Not super familiar with the code, but this is likely missing a note in the ChangeLog. :dragon_face:

wasade:
:+1:

ebolyen:
@ElDeveloper Yes, currently working on a backport for 0.2.4, will update this soon.

ebolyen:
Related to #891

ElDeveloper:
Got it, thanks!

Jorge-C:
Should we change version strings in the backported change?

ebolyen:
@Jorge-C We will when there is a release. (Which will be soon)

Jorge-C:
@ebolyen Cool, it makes sense if we're just backporting one or two things, but if we ended up having more regular bug fix releases it would be more clear to call this 0.2.4.dev -> make changes -> rename to 0.2.4 when ready to publish next bug fix release, instead of working on 0.2.4.

PR #891, opened 2015-04-02
[0.2.4] BUG: fixed TreeNode.extend when given tree.children; fixes #889

### LINKED ISSUE #889: [0.2.3] TreeNode.extend seems to mangle the results

ebolyen:
**This is not a duplicate**
This is a PR for 0.2.4, this is a new thing for us so hold on to your pants!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # scikit-bio changelog
 
+## Version 0.2.3-dev
+
+### Bug fixes
+* Fixed issue with ``TreeNode.extend`` where if given the children of another ``TreeNode`` object (``tree.children``), both trees would be left in an incorrect and unpredictable state. ([#889](https://github.com/biocore/scikit-bio/issues/889))
+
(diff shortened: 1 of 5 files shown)


ebolyen:
Please put all discussion on #890.

wasade:
:hand: :jeans:

PR #892, opened 2015-04-04
More Composition analysis methods

mortonjt:
This PR does the following 

* Adds 4 new methods `inner`, `ilr`, `ilr_inv` and `clr_inv`
* Clarifies some mathematical details from previous PR
* Cleans up some of the mathjax equations 

This will be the last major PR realized for visualizations.  Like I said before, these methods will complement phylogenetic visualization very nicely.  In addition, these methods are a stepping stone for compositional statistical analysis.

### DIFF
--- a/skbio/stats/composition.py
+++ b/skbio/stats/composition.py
@@ -18,20 +18,20 @@ analyzed using Aitchison geometry. [1]_
 
 However, in this framework, standard real Euclidean operations such as
 addition and multiplication no longer apply. Only operations such as
-perturbation and power can be used to manipulate this data. [1]_
+perturbation and power can be used to manipulate this data.
 
 This module allows two styles of manipulation of compositional data.
 Compositional data can be analyzed using perturbation and power
 operations, which can be useful for simulation studies. The
 alternative strategy is to transform compositional data into the real
-space.  Right now, the centre log ratio transform (clr) [1]_ can be
-used to accomplish this.  This transform can be useful for performing
-standard statistical tools such as parametric hypothesis testing,
-regressions and more.
+space.  Right now, the centre log ratio transform (clr) and
+the isometric log ratio transform (ilr) [2]_ can be used to accomplish
+this. This transform can be useful for performing standard statistical
+tools such as parametric hypothesis testing, regressions and more.
 
 The major caveat of using this framework is dealing with zeros.  In
 the Aitchison geometry, only compositions with nonzero components can
-be considered. The multiplicative replacement technique [2]_ can be
+be considered. The multiplicative replacement technique [3]_ can be
 used to substitute these zeros with small pseudocounts without
 introducing major distortions to the data.
 
@@ -46,14 +46,22 @@ Functions
    perturb
    perturb_inv
    power
+   inner
    clr
+   clr_inv
+   ilr
+   ilr_inv
    centralize
 
 References
 ----------
 .. [1] V. Pawlowsky-Glahn. "Lecture Notes on Compositional Data Analysis"
-.. [2] J. A. Martin-Fernandez. "Dealing With Zeros and Missing Values in
-       Compositional Data Sets Using Nonparametric Imputation"
+
+.. [2] J. J. Egozcue. "Isometric Logratio Transformations for
+   Compositional Data Analysis"
+
+.. [3] J. A. Martin-Fernandez. "Dealing With Zeros and Missing Values in
+   Compositional Data Sets Using Nonparametric Imputation"
 
 
 Examples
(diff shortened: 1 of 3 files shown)


Jorge-C:
Cool! There's a test that doesn't pass in all builds (seems to be an accuracy thing) and a doctest too.

mortonjt:
I'm right now benchmarking these methods, so hold on a bit so I can more rigorously test some of these methods.

mortonjt:
Looks like the last run time out ... Mind if we could restart the Travis build?

Once this passes, @Jorge-C , would you mind to take a look over this PR once you have time?
I'm confident that all of the bugs have been weeded out.

ebolyen:
@mortonjt @Jorge-C Besides, the merge conflict, what is the status of this PR?

Jorge-C:
I believe it was ready when I removed the do not merge tag, not sure if Jamie wanted to add anything else here.

ElDeveloper:
Just FYI, @mortonjt is out of town until Monday.

mortonjt:
Alright.  I fixed a potential issue with multiplicative replacement.

Basically, it was possible for this replacement strategy to generate negative proportions with the default parameters before.  Just fixed the default parameter in this last commit.

@Jorge-C , do you mind if you could take a look over this?

Jorge-C:
In the new description of `delta`:
> :math:`\delta = \frac{1}{N M}` where :math:`N`
is the number of components and :math:`M` is the
number of individuals found in the largest sample

but from looking at the code, M seems to be the maximum number of zero proportions in a sample. Are both equivalent?

mortonjt:
Not quite.  M is the largest sequencing depth.

Here's the details why I changed the default parameter for this method.

In an edge case, there are N-1 zeros in the sample, and there is only 1 read in the remaining taxa, it is impossible for the replaced zeros to be larger than that taxa with the current default parameter.  In addition, if `(N-1) * delta > 1`, then the scaling factor will become negative and that remaining taxa is now a negative proportion.

Jorge-C:
Ok, I just saw why docs and code seemed inconsistent to me, although I don't understand how it happens  :confused:

When looking at the changes from your [new commit](https://github.com/mortonjt/scikit-bio/commit/25875c6dc4b960ae2a2c3f32513fe0a8f1e36021#diff-704d4acf77c05c8963c80a29372d7343L189), even before the merge from master, delta is computed in a different way. However, when checking the full diff between your PR and master, the way it's computed is consistent with the new docs but the only commit in between is a merge from master that doesn't modify `compositions.py` (?)

I'll add a minor comment inline, but I'm also interested to know how did that line change after `25875c6` :smile:

skbio/stats/composition.py
```diff
@@ -150,8 +158,9 @@ def multiplicative_replacement(mat, delta=None):
     delta: float, optional
        a small number to be used to replace zeros
        If delta is not specified, then the default delta is
-       :math:`\delta = \frac{1}{N^2}` where :math:`N`
-       is the number of components
+       :math:`\delta = \frac{1}{N M}` where :math:`N`
+       is the number of components and :math:`M` is the
+       number of individuals found in the largest sample
```
Jorge-C: This description doesn't seem too clear in the module's context, maybe it should use composition instead of sample, and expand on what number of individuals means in a matrix of proportions.

mortonjt:
That was my fault.  It should have been `mat.sum(axis=-1).max()`, not `tot.max()`

I know that the math is a bit dense.  If you'd like, we can use Google Hangouts and maybe Groupboard to sketch out some of the proofs behind these methods.

Jorge-C:
Thanks! The math is fine, this is just a git question :) `25875c6` uses `tot.max()`, the following commit `d00db53` doesn't modify that line, but somehow the combination no longer uses `tot.max()`. Maybe it's a github glitch?

mortonjt:
I actually corrected this in `d00db53` right before the merge.  So the `delta` in `25875c6` is wrong.

Jorge-C:
Ah, I see. It could be useful to move that change from `d00db53` to `25875c6` (and force push, new commits will be created) so that the problem is fixed in the commit that says so in the log instead of in the merge, but that's some added git magic just for tidiness, so it's your call.

mortonjt:
I'm probably going to squash these commits, so that problem should disappear.  Sound good?

Jorge-C:
:+1:

mortonjt:
I'll start squashing right before merge.

Is that the only problem?

Jorge-C:
The minor doc inline comment too, maybe a native speaker could double check it?

skbio/stats/composition.py
mortonjt: Oh ....

Crap.  That was poor oversight.  I remember why I used `delta=N^2` before.  This was appropriate for proportions.  The `delta=1/NM` is appropriate for counts.  I'm going to revert ...

mortonjt:
Alright, removed that docstring and default parameter.  It was back to its original status.

Also added those experimental decorators.  Mind you could take another look @Jorge-C?

Jorge-C:
:+1: do you mind squashing if possible?

mortonjt:
Consider it squashed :)

Jorge-C:
Wonderful, thank you Jamie!

mortonjt:
Thanks!!

Issue #781, opened 2014-11-27
add option to FASTA reader to ignore blank/whitespace-only lines


jairideout:
Would be useful to add an option to the FASTA reader to ignore blank/whitespace-only lines.

jairideout:
This change should happen with the FASTQ parser too.

ebolyen:
Some unit tests that definitely should exist for this functionality:

 * The case where an empty line occurs in the middle of a record should fail
 * The case where an empty line occurs instead of the sequence should fail
 * The case where an empty line occurs between records should succeed

ebolyen:
@anderspitman Would you be up for tackling this issue? Should be relatively straight-forward and will introduce you to IO. I am happy to explain anything in regards to IO as well.

Jorge-C:
Just curious, is there a standard for these files? Does it say anything
about blank lines?

ElDeveloper:
I haven't checked in a while but can we also add that it should report
what was the last record it successfully parsed? Usually useful to
figure out what's wrong with your file.

jairideout:
From the file format descriptions I could find for FASTA/Q, blank/whitespace-only lines aren't part of the format. However, we've had users with files that had a trailing blank line at the end of the file, or with records separated by newlines (maybe for readability?).

I think we should support ignoring these lines only if they appear *between* records. If they appear in the middle of a record it's hard to tell if a user incorrectly catted a file together, if there was an empty sequence (not allowed), etc.

I'm on the fence about whether this behavior should be the default or not. Thoughts?

Jorge-C:
That's very reasonable.

I don't have any thoughts regarding whether it should be possible to error
or not on blank lines.

2015-03-06 13:30 GMT-07:00 Jai Ram Rideout <notifications@github.com>:

anderspitman:
@ebolyen Somehow I didn't notice you tagged me until @jairideout suggested this issue today. Sorry about that. Happy to work on this.

anderspitman:
@jairideout we want to implement this as an option and not default behavior, correct? The current documentation explicitly says blank spaces aren't valid.

@ebolyen Looking through the potential test cases you mention. I know in the interest of performance we don't currently validate records unless/until the user calls ``is_valid``, correct? In that case do we want to do whitespace validation every time the blank space feature is enabled, or roll whitespace validation into ``is_valid``? I'm assuming the latter.

jairideout:
> we want to implement this as an option and not default behavior, correct?

[...]

PR #893, opened 2015-04-07
Allow newlines when reading fasta files

anderspitman:
For #781. Still need to do fastq and docs, but wanted to get preliminary feedback on the approach.

My algorithm was basically that newlines are fine unless they immediately precede a sequence line. That seems to match all the test cases @ebolyen specified, and I've added or modified the unit tests accordingly.


anderspitman:
Added FASTQ. Still need to update docs and fix linting. Also, there's currently some duplication in the FASTA/FASTQ tests, which I just made worse... It would probably be a good idea to do some refactoring there. Not sure if that should be part of this PR or not.

anderspitman:
I'm planning one more push on this in the next hour or so then it should be ready for review

anderspitman:
Ready for review

skbio/io/fastq.py
```diff
@@ -480,21 +480,30 @@ def _parse_sequence_data(fh):
 def _parse_quality_scores(fh, seq_len, variant, phred_offset):
     phred_scores = []
     qual_len = 0
-    for chunk in _line_generator(fh):
-        if chunk.startswith('@') and qual_len == seq_len:
-            return phred_scores, chunk
+    prev_was_blank = False
+    for chunk in fh:
+        chunk = chunk.strip()
```
jairideout: Is it possible to factor this new code to use `_line_generator`? That way there's only a single place where line preprocessing (i.e., whitespace stripping) takes place. Adding a flag to `_line_generator` allowing/disallowing blank lines should do the trick.

skbio/io/fastq.py
```diff
@@ -449,9 +449,9 @@ def _alignment_to_fastq(obj, fh, variant=None, phred_offset=None,
 
 def _line_generator(fh):
     for line in fh:
-        line = line.rstrip('\n')
+        line = line.strip()
```
jairideout: Can you update the fastq docs to match this new behavior? The current docs state that lines aren't stripped.

skbio/io/fastq.py
```diff
@@ -449,9 +449,9 @@ def _alignment_to_fastq(obj, fh, variant=None, phred_offset=None,
...
         if not line:
-            raise FASTQFormatError("Found blank line in FASTQ-formatted file.")
+            raise FASTQFormatError("Found blank line within FASTQ record.")
```
jairideout: Can you update this error message to indicate blank or whitespace-only (similar to FASTA error message)? It only mentioned blank before because `rstrip('\n')` was being used, but now all leading/trailing whitespace is stripped.

skbio/io/fastq.py
```diff
@@ -480,21 +480,30 @@ def _parse_sequence_data(fh):
...
-            return phred_scores, chunk
+    prev_was_blank = False
+    for chunk in fh:
+        chunk = chunk.strip()
+        if chunk:
+            if chunk.startswith('@') and qual_len == seq_len:
+                return phred_scores, chunk
+            else:
+                if prev_was_blank:
+                    raise FASTQFormatError("Found blank line within FASTQ "
```
jairideout: Same comment as above, copied here:

Can you update this error message to indicate blank or whitespace-only (similar to FASTA error message)? It only mentioned blank before because `rstrip('\n')` was being used, but now all leading/trailing whitespace is stripped.

It'd be nice if these two error paths could become one again... but I can't think of a good way to do this (in `_line_generator` or elsewhere) since you have to error out *after* grabbing the next line.

skbio/io/tests/test_fasta.py
```diff
@@ -63,12 +64,16 @@ def setUp(self):
             'empty',
             'whitespace_only',
             'fasta_invalid_missing_header',
-            'fasta_invalid_blank_line',
```
jairideout: You renamed this file to `fasta_invalid_blank_line_in_record` but it doesn't look like it's used in the tests. It's also misnamed because the file should be valid (the blank line occurs between records).

skbio/io/tests/test_fasta.py
```diff
@@ -216,18 +229,46 @@ def setUp(self):
              {'qual': get_data_path('qual_invalid_missing_header')},
              FASTAFormatError, 'without a header.*QUAL'),
 
-            # fasta and qual with blank line
-            ('fasta_invalid_blank_line', {}, FASTAFormatError,
+            # fasta and qual with blank line after header
```
jairideout: This comment should indicate that the blank line is within the sequence.

skbio/io/tests/data/fasta_ws_line_between_records
```diff
@@ -0,0 +1,15 @@
+>seq1 desc1
```
jairideout: Can you add this file to the list of positively-sniffed files in `SnifferTests`?

skbio/io/tests/data/qual_blank_line_between_records
```diff
@@ -0,0 +1,15 @@
+>seq1 desc1
```
jairideout: Can you add this file to the list of positively-sniffed files in `SnifferTests`?

jairideout: Sorry, this should go in the list of negatives, not positives.

skbio/io/tests/test_fasta.py
```diff
@@ -216,18 +229,46 @@ def setUp(self):
...
+             FASTAFormatError, 'without quality scores'),
+
+            # fasta and qual with whitespace only sequence
+            ('fasta_invalid_whitespace_only_sequence', {}, FASTAFormatError,
+             'without sequence data'),
+            ('fasta_3_seqs_defaults',
+             {'qual': get_data_path('qual_invalid_whitespace_only_sequence')},
+             FASTAFormatError, 'without quality scores'),
+
+            # fasta and qual with blank line within sequence
```
jairideout: This comment should indicate that the blank line is after the header.

skbio/io/tests/data/qual_ws_line_between_records
```diff
@@ -0,0 +1,15 @@
+>seq1 desc1
```
jairideout: Can you add this file to the list of negatively-sniffed files in `SnifferTests`?

jairideout:
Thanks for working on this @anderspitman! I'm done reviewing; in addition to my inline comments, can you add the following test cases:

- [ ] Test that repeated blank/whitespace-only lines occurring between records parses correctly (for FASTA, QUAL, and FASTQ). Current tests are testing single occurrences of "valid" blank lines.

- [ ] Test that repeated blank/whitespace-only lines occurring at end of file are ignored (for FASTA, QUAL, and FASTQ).

- [ ] Test that a FASTQ file with various occurrences of leading/trailing whitespace throughout parses correctly. Previous behavior was to error out in this case since only `rstrip('\n')` was called. Since FASTA parsers already worked this way (calling `strip`), those tests might be helpful to adapt for FASTQ.

I'm curious what the behavior should be at the beginning of the file. I think the changes in this PR (and previous behavior) disallows blank/whitespace-only lines that occur before any record. @ebolyen @anderspitman how do you think this should behave? I'm on the fence: it makes sense to allow (ignore) blank/whitespace-only lines at the beginning of the file for consistency with trailing blank line behavior, but it also means that an entire file of whitespace-only lines would need to be processed before failing at the end (this might not be a big deal in practice though).

> Also, there's currently some duplication in the FASTA/FASTQ tests, which I just made worse... It would probably be a good idea to do some refactoring there. Not sure if that should be part of this PR or not.

Do you mean the duplicated tests for FASTA and QUAL files? Agree a refactoring would be helpful, but not necessary in this PR (would actually prefer not to have those changes in this PR so it's easier to review).

ebolyen:
@jairideout I think whitespace at the beginning is probably the only whitespace a user is in a good position to easily fix. However, I don't see why it should matter either. Would we be comfortable with some limit on the sniffer (say 5 blank lines) and then have the actual parser not really care? I think that is the greatest performance risk. By the time the parser has been invoked, success or failure is really the only outcome, so why fail if you could succeed? A limitation on the sniffer like this seems reasonable, if you really have that many blank lines, you have a horrific fasta file and should probably do something about it (say explicitly telling skbio that it is fasta and re-encoding it).

jairideout:
Thanks @ebolyen, those are all really good points! Let's go with supporting blank/whitespace-only lines at the beginning of the file and have the sniffers modified to only sniff so far before failing.

@anderspitman can you also incorporate these changes in this PR?

anderspitman:
@jairideout sounds good.

anderspitman:
@jairideout re FASTQ whitespace stripping: is there a reason we weren't stripping before? I'm assuming so since it's explicitly mentioned in the docs. I can change it back to the old way if that's preferable.

anderspitman:
@jairideout the duplication which we discussed is really bothering me. I think I have a solution. What do you think about refactoring `line_generator` to handle the line state like so:

```python
def _line_generator(fh):
    prev = 'needs_nonblank_init_value'
    for line in fh:
        line = line.strip()
        if line and not prev and not line.startswith('@'):
            raise FASTQFormatError("Found blank line within FASTQ record.")
        prev = line
        yield line
```

This seems to work. There are 3 tests which fail:

```python
Test:
('whitespace_only', FASTQFormatError, 'blank line.*FASTQ'),

Error:
AssertionError: "blank line.*FASTQ" does not match "Expected sequence (@) header line at start of file: ''"
```
I think this one is fine to change the expected value.

```python
Test:
 ('error_trunc_at_plus.fastq', FASTQFormatError,
'blank line.*FASTQ'),

Error:
AssertionError: "blank line.*FASTQ" does not match "Found incomplete/truncated FASTQ record at end of file."
```
That one is actually more correct

```python
Test:
 ('error_no_qual.fastq', FASTQFormatError,
 'blank line.*FASTQ'),

Error:
AssertionError: "blank line.*FASTQ" does not match "Found more quality score characters than sequence characters. Extra quality score characters: '1_397_389'"
```
This one I'm somewhat more concerned about. It's treating the blank line as part of the quality scores, and the error is more vague. If we can stomach that I think overall this solution is much cleaner. Almost no changes to the original.

adamrp:
@anderspitman @jairideout I can't think of a reason that whitespace was not stripped from FASTQ previously. At first I thought that some whitespace characters could be valid quality characters, but that is not the case (space is the highest at 32, and the ASCII offset for qual scores is almost always 33 or 64)

anderspitman:
@adamrp good to know. It sounds like we can probably keep the changes I made.

anderspitman:
There's actually a 4th test that fails after my changes.

```python
Test:
 ('error_trunc_at_seq.fastq', FASTQFormatError,
'blank line.*FASTQ'),

Error:
AssertionError: "blank line.*FASTQ" does not match "Found incomplete/truncated FASTQ record at end of file."
```

Again I think this error is more accurate

anderspitman:
Also, the `error_no_qual.fastq` error is more serious than I thought at first. Definitely the wrong exception. Working on that now.

skbio/io/fastq.py
anderspitman: Fixed. see below

skbio/io/fastq.py
anderspitman: Fixed.

skbio/io/fastq.py
anderspitman: Fixed.

skbio/io/fastq.py
anderspitman: Fixed, see below for `_line_generator`

skbio/io/tests/test_fasta.py
anderspitman: Removed

skbio/io/tests/test_fasta.py
anderspitman: Fixed.

skbio/io/tests/data/fasta_ws_line_between_records
anderspitman: Fixed.

skbio/io/tests/data/qual_blank_line_between_records
anderspitman: Added to negatives

skbio/io/tests/test_fasta.py
anderspitman: Fixed

anderspitman:
@jairideout everything up to the new tests you requested should be done. I can add those and implement the blank/whitespace-at-beginning logic soon.

Most importantly I pushed some changes to the `_line_generator` stuff. After working on this for a while, I was almost able to get the 2 paths back together by having `_line_generator` keep track of the blank lines (see above), but I had that one case that was failing. From digging into that, I realized maybe we should go the other way. The paths were different because there were really different errors happening in each case, I was just giving them the same name. So what I did with this push was add more specificity to the error messages, as well as handling more cases. I also added line numbers to the error printouts. `_line_generator`'s only job now is stripping and counting lines. I like the way it ended up. What do you think?

jairideout:
@anderspitman regarding whitespace-stripping of FASTQ, I originally didn't implement this because the format itself doesn't describe this behavior and says that whitespace of any kind is invalid. However, as @adamrp noted, I think this behavior is safe because the quality scores will never contain whitespace. Also, the FASTA reader strips whitespace while parsing, so this change is probably a good thing because FASTA/Q will handle whitespace in the same way.

Thanks for working on refactoring `_line_generator`. From a quick look, those new error messages look good because they're more specific than before. I agree that the one error you noted could be problematic (though it sounds like your new approach fixes this?). I'll take a closer look soon and follow up. Thanks!

anderspitman:
@jairideout excellent. Yeah the 4th error should be fixed. 2 of them are more accurate now (now indicate truncated sequence), and the whitespace-only one complains about missing '@' at beginning instead of blank, but that one will change when I implement the changes to that logic anyway, so I think we're in pretty good shape from a design point of view, assuming I'm not missing something.

anderspitman:
Just realized the line reporting isn't going to work in it's current state, since `_line_generator` is called multiple times per file. Hmm I'll see if I can find another way.

jairideout:
Sounds good, just let me know when you're ready for me to review again or if you need input. It's okay if there isn't a great way to factor this logic into `_line_generator`.

anderspitman:
Will do. Might need to just remove it, but I want to take a look at it first and see if there's a simple way.

anderspitman:
Everything except blanks at the beginning of file should be done. Working on that now...

anderspitman:
I went ahead and removed the line reporting. It would be more involved and is outside the scope of this PR IMO

anderspitman:
@jairideout @ebolyen I've come up against an interesting case. First a question: why is `skbio/io/tests/data/empty` considered a valid file? It doesn't make sense to me that it would pass the tests. At the very least it should be failing in the current master because the file doesn't start with a `@`. So I started looking into it and I'm seeing behavior that confuses me. Maybe you can help me understand. If you attempt the following snippet:

```python
from skbio.io.fastq import _line_generator
with open("skbio/io/tests/data/empty", 'U') as fh:
    line = next(_line_generator(fh))
```

A `StopIteration` exception is raised, which makes sense. However, if you attempt the following:

```python
from skbio.io.fastq import _fastq_to_generator
list(_fastq_to_generator("skbio/io/tests/data/empty", variant='sanger'))
```

No exception is raised. As near as I can tell, `_line_generator` is being invoked exactly the same in these two cases. But there's even more to it than that. Not only is no error being raised in the second case, it appears execution actually stalls on this line within `_fastq_to_generator`:

```python
seq_header = next(_line_generator(fh))
```

Which is why the empty file is passing all the validity tests. It's not actually passing, the code after this point is just not executing.

I can sort of see why it would hang on the call to `next`, since `_line_generator` will return but never yield anything, but why the different behavior in each case? The only guess I've come up with so far is that this has something to do with the fact that we have a generator within a generator, but that's just a guess. Anyone know what's actually happening?

anderspitman:
BTW this is easy to fix, assuming that an empty file should indeed be considered invalid:

```python
    seq_header = ''
    for line in _line_generator(fh):
        seq_header = line
        if seq_header:
            break
    if not seq_header:
        raise FASTQFormatError("Blank or whitespace-only file")
```

I was just hoping to understand what's going on and verify we don't have an underlying bug in our platform.

jairideout:
@ebolyen and I originally decided that empty files could be read, but not positively sniffed, for the sequence file formats that contain records (FASTA, FASTQ). This allows you to read an empty file using `skbio.read` and obtain a generator that doesn't yield anything. This also allows you to read an empty file into a structure that can be empty, such as `SequenceCollection`.

I think files that are empty or only contain whitespace should continue to operate this way, since conceptually they are FASTA/FASTQ files with zero records.

anderspitman:
Gotcha. Are there cases where reading from an empty file is what the user intends?

ebolyen:
@anderspitman Rarely, which is why the `<empty_file>` sniffer will win that battle and complain. You have to explicitly say the format, and even then the sniffer will warn unless you suppress.

anderspitman:
@jairideout currently whitespace_only is invalid but empty is valid. You're saying we should change it so both are valid?

anderspitman:
Also, why does the sniffer attempt to read 10 records? What extra information is given if it can read 1 vs reading more than 1?

anderspitman:
Alrighty. I believe everything we've discussed up to this point has been implemented. Re: my last question: I went ahead and changed whitespace-only files to be valid but issue sniffer warning.

I'm a little unhappy with my solution for having the sniffer check the first 5 lines, because I'm doing a `fh.seek(0)`. Not sure how I feel about that.

jairideout:
@anderspitman re: whitespace-only file failing but empty file passing:

I think the new behavior should treat both empty and whitespace-only files the same way, i.e., as files with zero FASTA/FASTQ records. I think the current behavior is a result of the parsing logic not allowing blank/whitespace-only lines.

re: sniffing more than 1 record:

The sniffer sniffs *up to* 10 records (10 is pretty arbitrary) just to avoid positively sniffing a file that happened to look like it was FASTA/FASTQ format at the beginning, but actually isn't. We're taking a small runtime hit to check more of the file, which should make the sniffer's decision a little more reliable.

I'll take another pass through the code now.

skbio/io/fasta.py
```diff
@@ -557,6 +563,10 @@
 
 @register_sniffer('fasta')
 def _fasta_sniffer(fh):
+
+    if _count_blank_lines(fh) > 5:
```
jairideout: The current implementation of `_count_blank_lines` will read an entire whitespace-only file since it only stops reading if it finds a non-whitespace line or reaches the end of the file. Can you modify the way the sniffer works so that it'll bail as soon as it reaches 5 blank lines?

anderspitman: Ah you're right. I refactored it when I pushed it up to `_base.py` without thinking about it. Technically we have duplicated logic here between `fasta.py` and `fastq.py` because of the `> 5`. What do you think about calling the funtion `_has_too_many_blanks` and moving the `> 5` inside?

jairideout: There's also some duplication between FASTA and FASTQ readers related to handling of blank/whitespace-only lines at the beginning of the file that could be refactored into `_line_generator`, and I think we can push the duplicated sniffer logic into `_line_generator` as well.

How about removing `_count_blank_lines` and updating `_line_generator` to accept a keyword argument `skip_blanks=False`? If `False` (by default, the safest behavior), lines will be stripped and yielded (this is the current implementation). If `True`, only non-blank lines will be yielded. This should allow the duplicate logic to be removed from the FASTA and FASTQ readers.

For sniffing, we could add another keyword argument to indicate the maximum number of consecutive blank lines to skip if `skip_blanks=True`. The default behavior would place no restrictions on this number (necessary for the readers), and the sniffers could pass 5. If this number is exceeded, the function could either raise an error (which the sniffers could catch and handle), or it could just start yielding lines at this point and the sniffers would check the next line to determine whether to keep sniffing.

I'm not too concerned about having this number (5) duplicated across the two sniffers since we could conceivably want a different number to be passed for different formats (though 5 seems reasonable for both FASTA and FASTQ).

Does this factoring make sense?

skbio/io/fastq.py
```diff
@@ -447,18 +465,19 @@ def _alignment_to_fastq(obj, fh, variant=None, phred_offset=None,
...
-        yield line
+def _blank_error(unique_text):
+    error_string = ("Found blank or whitespace-only line {} in "
+                    "FASTQ file").format(unique_text)
+    raise FASTQFormatError(error_string)
 
 
 def _parse_sequence_data(fh):
     seq_chunks = []
+    prev = '@'
```
jairideout: It might be clearer (and safer, in case the caller's logic changes in the future) to update `_parse_sequence_data` and `_parse_quality_scores` to accept `prev` as a positional argument. Then the callers would supply their actual previous line when calling this function instead of faking a previous line.

jairideout:
@anderspitman just finished another pass through. I only looked at the functional code and ignored docs and tests for now. I think there are a few things to refactor that'll simplify the code, reduce duplication between FASTA and FASTQ implementations, and fix the sniffer performance issue (see my inline comments). Let me know when you're ready for another review and thanks for your work on this!

skbio/io/fastq.py
anderspitman: :+1:

skbio/io/fasta.py
anderspitman: Agree we should be able to push line skipping into `_line_generator`. So the way it would work is if `skip_blanks` is `True` it will consume (up to `max_blanks` if set, raising an exception if exceeded) blanks at the beginning of the file. Once the first nonblank line is read, blanks of ignored by `_line_generator` and handled by the caller. Seem reasonable?

anderspitman: @jairideout I don't think there's a simple way to move the `max_blank` logic up to `_line_generator`. The problem is that the sniffer doesn't use `_line_generator` directly, so there's no where to capture a return value and indicate the file was not identified by the sniffer. So the only way I thought to do it was have `_line_generator` raise an exception and have the sniffer catch it. But then the files always except, not just when being sniffed. I haven't been able to come up with a way to branch that logic.

ebolyen:
@anderspitman Just a heads up, #908 is going to conflict with this PR. Depending on which is merged first.

anderspitman:
Yup I figured. I'm happy to do the rebase if you want to push that through. Not sure if we're done here.

skbio/io/fasta.py
jairideout: That makes sense, thanks for looking into this more. I think the way to go is have a simple utility function used by the FASTA and FASTQ sniffers that handles the blank line logic at the beginning of the file, similar to what you have now. Then `_line_generator` can be updated to have a `skip_blanks` parameter that the readers can use accordingly, depending on whether they are at the beginning of the file or in the middle of parsing records. Does that sound good to you?

jairideout:
I followed up on the inline comment, let me know when this is ready for review. Thanks!

anderspitman:
Based on your line comment, I think it's ready for review. I added `skip_blanks` to `_line_generator` but left `_count_blank_lines`.

skbio/io/_base.py
```diff
@@ -185,3 +186,21 @@ def _format_fasta_like_records(generator, id_whitespace_replacement,
...
+
+
+def _line_generator(fh, skip_blanks=False):
+    for line in fh:
+        line = line.strip()
+        if line or not skip_blanks:
+            yield line
+
+
+def _count_blank_lines(fh):
```
jairideout: This still can take a full pass through the file, which we don't want to happen when sniffing.

anderspitman: Quite right, forgot to fix that...

anderspitman: Should be good now. I went with `_too_man_blanks` to avoid repeating the 5.

skbio/io/_base.py
```diff
@@ -185,3 +186,25 @@ def _format_fasta_like_records(generator, id_whitespace_replacement,
...
+    too_many = False
+    for line in _line_generator(fh):
+        if line:
+            break
+        else:
+            count += 1
+            if count > max_blanks:
+                too_many = True
+                break
+    fh.seek(0)
```
jairideout: Sweet, this looks good! Minor comment: is this `seek` necessary? If there aren't too many blank lines at the beginning, I think the sniffers can just keep reading from the file handle's current position.

anderspitman: The problem was when the file is valid, this function will consume the
first valid line then the file is sniffed incorrectly. There may well be a
way around this. I'd prefer to get rid of the seek.

jairideout: @ebolyen suggested having this function return the first non-blank line (if there aren't too many blanks), then the sniffers can use this line (and the rest of the contents of the file handle) to continue sniffing. Having a `seek(0)` call is unsafe in sniffers because the user may supply a file handle that has already been `seek`ed to an offset.

Related to this, I noticed that the current FASTA sniffer is already calling `seek(0)` before trying sniff as QUAL. This bug would be great to fix in this PR since we're already changing the logic of the sniffer somewhat. @ebolyen suggested a new sniffing strategy:

1. try to read up to 10 FASTA records
2. for each record, check the contents of the sequence to see if they look like quality scores. if they are, bail since we don't want to sniff a FASTA file as QUAL

anderspitman: I like that idea, but I don't think it will work because the sniffers don't handle the lines directly; they invoke `_fasta_to_generator` and `_fastq_to_generator` with the file handle. We could add a parameter to the generators if we really want to avoid `seek`.

Barring that my recommendation would be checking if the user has `seek`ed to an offset, saving the offset, sniffing, then resetting to the user's value.

anderspitman: Just found [this comment](https://github.com/anderspitman/scikit-bio/blob/3d4db981a17ea0a9d783bbabfa880c5641f30caf/skbio/io/_registry.py#L98). That does present an interesting challenge.

anderspitman: Also just upvoted @ebolyen 's [SO question](http://stackoverflow.com/q/26367423/943814) before realizing he was the OP, ha.

anderspitman: Correct me if I'm wrong but it seems like using `seek` within the sniffer isn't a problem from the user's perspective because we're using a copy of the file handle, which shouldn't affect the user's copy, right? It appears that another consequence of this that if the user `seek`s the file, the sniffer still starts at the beginning of the file, but the generator will respect the seek. Am I understanding that correctly?

jairideout: Good catch! We forgot about the filehandle copy that happens before sniffing. Let's leave the implementation as-is since `seek`/`tell` is a larger issue that can be dealt with later on. I totally agree with you that it's odd to sniff from the beginning of the file and read from some offset.

I'll take a full pass through the PR later today or tomorrow morning.

skbio/io/tests/test_fasta.py
```diff
@@ -154,8 +178,19 @@ def setUp(self):
                  description='detailed description \t\twith  new  lines',
                  quality=[42, 42, 442, 442, 42, 42, 42, 42, 42, 43])],
             {},
-            list(map(get_data_path, ['fasta_multi_seq', 'fasta_max_width_5'])),
-            list(map(get_data_path, ['qual_multi_seq', 'qual_max_width_5']))
+            list(map(get_data_path, ['fasta_multi_seq', 'fasta_max_width_5',
+                                     'fasta_blank_lines_between_records',
+                                     'fasta_ws_lines_between_records'])),
```
jairideout: Can you also add the corresponding fasta test files here like you did for qual?

jairideout:
Thanks for all of the work on this PR @anderspitman! This is almost ready for merge, just the following things are left (all should be easy to resolve):

- [ ] resolve merge conflicts
- [ ] one more code review comment related to tests (inline comment [here](https://github.com/biocore/scikit-bio/pull/893/files#r29685767))
- [ ] add changelog mention, be sure to note the change in fastq reading regarding stripping of lines while parsing (as well as the blank/ws support for fasta/fastq)

jairideout:
@anderspitman I'm going to issue a new PR shortly with your commits here and the remaining to-do items I mentioned above. @ebolyen and I are close to submitting a PR for the Sequence object refactor, which affects some pieces of `skbio.io`. It'll be easier to have these changes merged to master first so that we can pull and fix any issues on our branch.

Thanks again for your work on this!

jairideout:
Closing in favor of #916.

skbio/io/tests/test_fasta.py
jairideout: Nevermind, I see that the fasta files are different than the expected results here. Having the qual files exercising these cases should be sufficient since fasta and qual readers use the same underlying parsing logic.

anderspitman: @jairideout are you sure you don't want me to add the fasta files? There's a comment in the code [here](https://github.com/anderspitman/scikit-bio/blob/3d4db981a17ea0a9d783bbabfa880c5641f30caf/skbio/io/tests/test_fasta.py#L232) which mentions that some of the test logic is intentionally duplicated even though FASTA and QUAL share the same parsing data, just in case they diverge at some point.

anderspitman:
@jairideout sounds good! Just to be sure I understand, are you going to finish the last couple items or did you want me to issue a PR to your branch?

jairideout:
I have an open PR (#916) where I'm addressing these changes.

skbio/io/tests/test_fasta.py
jairideout: Good catch, I'll update the corresponding fasta files to match the expected results here so they can be used in the sniffer and reader tests

anderspitman:
Ok cool

PR #903, opened 2015-04-25
ENH: Add support for URL loading in open_file(s)

ElDeveloper:
This PR adds the necessary code so open_file and open_files can load data from a URL, cache is supported through [CacheControl](https://cachecontrol.readthedocs.org/en/latest/). This module is MIT licensed, built directly on-top of python-requests and recommended on [the python-requests website](http://docs.python-requests.org/en/latest/community/recommended/).  

There are two remaining bits:

- Where should this behaviour be documented, currently the only docstring that's complete is the one for open_file, but IIRC this function is going to be deprecated soon. The docstring for open_files is empty, so not sure what I should do.

- A mention in the changelog is pending from the previous point.


wasade:
That's awesome!!! Can't comment inline right now, but there are PyCogent
mentions

ElDeveloper:
@wasade good catch!

jairideout:
@ebolyen can you review?

ebolyen:
@ElDeveloper The logic of this code looks really straight forward. Tested it locally with Wireshark and it is doing everything you would expect! Regarding the documentation. I actually am not super sure either. Some description of this capability should probably be in the module level docs for IO and the input params for `read` `write` and `sniff` could be updated (including the metaprogrammed methods). @jairideout Thoughts on the documentation side of things?

ElDeveloper:
Good to hear that. Sounds good, just let me know where I should be
adding the notes and I'll do that. Although this is a bit unrelated,
what's the reasoning to deprecate open_file? I find that function to be
rather useful. Also, will the docstring in open_files be added once
open_file is deprecated?

jairideout:
@ebolyen regarding the docs, I think the spots you mentioned are sufficient.

ebolyen:
@ElDeveloper Actually it looks like `with_open` is used for the sniffers, so it likely won't be deprecated. As far as the docstring goes for `with_opens` I think that was just an oversight. If you would like to add them, please feel free! For the rest of the documentation, a note about this feature where it makes sense in the `io/__init__.py` would be good. The dosctring templates in `_registry.py` can be updated (you will find them at the bottom of the file). Finally `read`, `write`, and `sniff` should be updated in the same way.

skbio/io/util.py
```diff
@@ -42,7 +47,17 @@ def _get_filehandle(filepath_or, *args, **kwargs):
...
-        fh, own_fh = open(filepath_or, *args, **kwargs), True
+        if requests.compat.urlparse(filepath_or).scheme in {'http', 'https'}:
+            sess = CacheControl(requests.Session(),
+                                cache=FileCache(gettempdir()))
+            req = sess.get(filepath_or, **kwargs)
+
+            # if the response is not 200, an exception will be raised
+            req.raise_for_status()
+
+            fh, own_fh = BytesIO(req.content), True
```
ebolyen: Does this load the request entirely into memory?

ElDeveloper: Yes that is the case. IIRC it is already in memory i.e. the content
attribute from the request.

ElDeveloper:
@ebolyen this should be ready for review/merge.

skbio/io/util.py
ElDeveloper: I think you can also pass `stream=True` and requests will in turn stream the file. 

http://docs.python-requests.org/en/latest/user/advanced/#streaming-requests

ebolyen:
Thanks @ElDeveloper I will have some time tomorrow to look at this!

skbio/io/util.py
ebolyen: Any chance this option can be the default mechanism? Will it work with the caching?

ElDeveloper: Although I understand your motivation for making it the default, I don't think it's a good idea, we should probably add a note in the docstring. It should work ... but I haven't tested that yet.

jairideout: Is there a reason we wouldn't want this to be the default?

ElDeveloper: Come to think about it, we should make it the default but still allow
for users to have control of this. So modifying kwargs to set
stream=True if None is provided, that way if someone needs to do
stream=False, they can still change it. That sounds good?

ebolyen: That sounds great @ElDeveloper, we should make this a first class kwarg in the registry system. (Just like `mode` is at the moment, though it will be removed soonish)

wasade: It may be necessary to pull down the full contents for things like xml or some binary like hdf5. I don't know if `h5py.File` will unroll a generator for instance, which I'm assuming would be the result of using `stream=True`? However, I think the hook in the registry like @ebolyen suggests should manage this.

ElDeveloper: Turns out I was wrong, even if you pass `stream=True`, all the data will be read into memory [because we were accessing the `.contents` property](http://docs.python-requests.org/en/latest/user/advanced/#blocking-or-non-blocking), the only way we can retrieve the data chunk by chunk is if we [use the `iter_lines` method](http://docs.python-requests.org/en/latest/user/advanced/#streaming-requests). The downside to this is that we cannot (I don't know of a way to do this) put this iterator into a file-like object, unless we roll our own wrapper.

The other thing I tried was to (1) generate the request, (2) read all the contents into `_`, (3) let CacheControl write the data into disk and (4) return a regular file handle to the cached file. Unfortunately this doesn't seem to work very well, I'm still trying to figure out why this happens, I've opened an issue on their repo https://github.com/ionrock/cachecontrol/issues/77

ebolyen: I like your second option the best, it is a shame it isn't behaving as expected. Let's open a new issue for this as an improvement to URL loading, since streaming seems to be non-trivial.

ebolyen:
Merged, thanks @ElDeveloper. Let's create a new issue to improve this feature for streaming URLs.

Issue #836, opened 2015-02-13
error when using gzip filehandle with I/O sniffer


jairideout:
When trying to read a gzipped file using scikit-bio's I/O system, the following error is raised while sniffing:

```python
In [1]: from skbio import DistanceMatrix

In [2]: import gzip

In [3]: f = gzip.open('dm.txt.gz', 'rb')

In [4]: DistanceMatrix.read(f)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-0454e36d1310> in <module>()
----> 1 DistanceMatrix.read(f)

/Users/jairideout/dev/scikit-bio/skbio/io/_registry.pyc in read(cls, fp, format, **kwargs)
    698         @classmethod
    699         def read(cls, fp, format=None, **kwargs):
--> 700             return skbio_io_read(fp, into=cls, format=format, **kwargs)
    701
    702         read.__func__.__doc__ = _read_docstring % (

/Users/jairideout/dev/scikit-bio/skbio/io/_registry.pyc in read(fp, format, into, verify, mode, **kwargs)
    599
    600     if format is None:
--> 601         format, fmt_kwargs = sniff(fp, cls=into, mode=mode)
    602         kwargs = _override_kwargs(kwargs, fmt_kwargs, verify)
    603     elif verify:

/Users/jairideout/dev/scikit-bio/skbio/io/_registry.pyc in sniff(fp, cls, mode)
    522             continue
    523         format_sniffer = _sniffers[fmt]
--> 524         is_format, fmt_kwargs = format_sniffer(fp, mode=mode)
    525         if is_format:
    526             possibles.append(fmt)

/Users/jairideout/dev/scikit-bio/skbio/io/_registry.pyc in wrapped_sniffer(fp, mode, **kwargs)
    106                 # true offset of the iterator.
    107                 if hasattr(fh, 'name'):
--> 108                     cfh = open(fh.name, fh.mode)
    109                 else:
    110                     cfh = copy.copy(fh)

TypeError: file() argument 2 must be string, not int
```

If I suppress sniffing by supplying a format and turning off verification, the file is successfully read:

```python
In [1]: from skbio import DistanceMatrix

In [2]: import gzip

In [3]: f = gzip.open('dm.txt.gz', 'rb')

In [4]: DistanceMatrix.read(f, format='lsmat', verify=False)
Out[4]: <skbio.stats.distance._base.DistanceMatrix at 0x10bb13f10>
```

[...]

Issue #690, opened 2014-09-25
io should support reading and writing gzip for all supported file formats


gregcaporaso:

ebolyen:
Add URL support too!

jrderuiter:
Any eta on this?

ebolyen:
@jrderuiter Not a solid one at this time, but it is a very simple addition to make. This can be achieved very similarly to #903. (Waiting for #903 to be merged would be a good idea). I am sure one of use will get around to this soon, but if you are interested in contributing this change, we would be happy to help you get started!

jrderuiter:
@ebolyen I think I have working version now (need to test it some more). I will wait on the #903 being merged first and then open a pull request if all is working.

ebolyen:
That's really exciting @jrderuiter! Hopefully #903 will get merge very soon (just waiting on some doc updates).

jairideout:
Fixed in #963.

PR #909, opened 2015-04-30
ENH: Support for gzip in io.

jrderuiter:
Adding support for gzip reading in the io module, addressing issue #690. Also made some adjustments to get the sniffer working properly with gzip. This might not be needed though, as @ebolyen mentions in issue #836 that filemode preservation will no longer be needed in the future. In that case we can remove the get_filemode function that I added to util.

I haven't touched the docs yet, but will update them if pointed out what needs changing.


jrderuiter:
Any comments or remarks? What about changes w.r.t. the filemode preservation?

gregcaporaso:
@jrderuiter, we'll get back to you tomorrow on this. Thanks for the PR!

skbio/io/util.py
```diff
@@ -37,17 +39,53 @@ def _is_string_or_bytes(s):
...
 
 
-def _get_filehandle(filepath_or, *args, **kwargs):
+def get_filehandle(filepath_or, *args, **kwargs):
     """Open file if `filepath_or` looks like a string/unicode/bytes, else
     pass through.
     """
     if _is_string_or_bytes(filepath_or):
-        fh, own_fh = open(filepath_or, *args, **kwargs), True
+        if filepath_or.endswith('.gz'):
```
ebolyen: There isn't a strong guarantee of a filename actually existing. Additionally it would be better if it checked for the [magic number](http://en.wikipedia.org/wiki/Gzip#File_format) (`1f 8b`) of a gzip file.

ebolyen:
We just merged #903 So there are merge conflicts for this PR, sorry about that, but it was inevitable. Filemode preservation looks good (we should even be able to reuse the logic in the future when the registered IO methods make the decisions about filemode instead of the user). Your mechanism for detecting gzip files should be changed to inspect the files contents rather than trusting the filename. 

Sorry for the delay in getting back to you, but thanks so much for tackling this @jrderuiter!

skbio/io/util.py
gregcaporaso: @ElDeveloper wrote [this](https://github.com/biocore/qiime/blob/master/qiime/util.py#L1421) for QIIME, could be adapted here.

wasade: :+1:

ElDeveloper: @gregcaporaso, that was @adamrp's work :goat:

gregcaporaso: Oops, sorry for the misattribution @adamrp :flushed:

adamrp: :laughing: Not a problem!

jrderuiter:
The magic number check was on my todo list, so I'll definitely add that. I'll try to make the adjustments tonight!

jrderuiter:
This should address most issues. I rebased the pull-request to account for #903. The current implementation does not allow loading of gzip files from URLs. Is this something we would want to do as well? That would require wrapping the BytesIO instance with Gzipfile but should be possible.

ebolyen:
Thanks @jrderuiter! I think supporting gzip with URLs is definitely something we would want to support. (Most services provide *only* gzipped files)

jrderuiter:
Do you have an example to test? I'll look into the implementation.

ElDeveloper:
> That would require wrapping the BytesIO instance with Gzipfile but should be possible

Maybe a crazy idea but after the BytesIO object is returned, we can probably call open_file again and it should be able to determine that it's a gziped file right? If it isn't it will just return the object untouched? ... maybe I'm missing some logic here, but that may just work ...

ebolyen:
@ElDeveloper That should work. @jrderuiter You can find some example files [here](http://greengenes.secondgenome.com/downloads/database/13_5).

jrderuiter:
Do we want to write gzipped files as well? (I think that would be nice.) The problem there is that the _is_gzip function will not work in this case, as we are opening an empty file and there is no magic number to read. The .gz extension detection would be an easy way to detect if a gzipped file is desired. Otherwise we need to add a gzip flag to the writer decorator that indicates whether a gzipped file should be opened for writing or not.

adamrp:
IMHO gzip flag on decorator is more explicit, so I would do that.

ebolyen:
@jrderuiter I think writing gzip would be awesome. I expect the easiest way to accomplish this is adding a flag to the registry `write(..., gzip=True)`. That can probably be done in a second PR so that we can get the reading merged in at least, unless you would really like to tackle it in this one.

wasade:
Just a note, we may want to disallow gzip for some file types in the
future. For instance, HDF5 natively supports gzip internally on datasets so
writing the file as gzip (which I don't think you could do anyway...)
wouldn't make sense

ebolyen:
@wasade Great point. Perhaps we need a way for registered IO to override things `mode` is a similar situation.

jrderuiter:
I just added a new commit with some changes. The new get_filehandle code now supports gzip for both files and urls.

I also made some changes to the get_filehandle arguments, essentially adding 'mode', 'binary' and 'gzip' arguments to the function. Mode is currently used to indicate whether a file should be opened in read or write mode, binary specifies whether the returned fh is text or binary and gzip specifies if a file should be opened as gzip or not (for either reading or writing). By default gzip is None, in this case the function _is_gzip is used to decide if a file should be opened as gzip (when reading files). For writing files gzip then defaults to False.

I haven't tested any writing yet, but reading seems to work fine in the cases I tested. Higher level functions will of course need to be adjusted to the new signature if you guys approve!

jrderuiter:
First attempt at changing the read/write API to include binary and gzip options. Let me know what you think! 

P.S. Can anyone tell me what the mutate_fh argument in the reader decorator is for?

jairideout:
Thanks for your work on this @jrderuiter, and I apologize for the slow reply. We are at a scikit-bio code sprint this week and are a bit behind on our responses. @ebolyen or I will have a chance to look at your PR and answer your questions next week.

skbio/io/_registry.py
```diff
@@ -206,7 +188,9 @@ def wrapped_reader(fp, mode='U', mutate_fh=False, **kwargs):
                     else:
                         kwargs[file_arg] = None
 
-                with open_files(files, mode) as fhs:
+                # Do we want binary argument to come from the user
+                # (wrapped_reader) or from the reader (register_reader)?
```
ebolyen: Almost certainly the reader, the user providing that information is actually actively unhelpful for the readers. This was a fundamental design flaw that I missed when I wrote the registry originally.

skbio/io/_registry.py
```diff
@@ -41,7 +40,7 @@ def _override_kwargs(kw, fmt_kw, warn_user):
     return fmt_kw
 
 
-def register_sniffer(format):
+def register_sniffer(format, binary=False):
```
ebolyen: :+1:

skbio/io/_registry.py
```diff
@@ -90,28 +89,13 @@ def register_sniffer(format):
...
-                # true offset of the iterator.
-                if hasattr(fh, 'name'):
-                    cfh = open(fh.name, fh.mode)
-                else:
-                    cfh = copy.copy(fh)
-                    cfh.seek(0)
+            raise DuplicateRegistrationError(
+                msg="'%s' already has a sniffer." % format)
+
+        def wrapped_sniffer(fp, mode='r', binary=binary, **kwargs):
```
ebolyen: We should be able to drop `mode` as a registry-facing kwarg, and I don't think we actually want users to be able to override `binary` either, so you should be able to just leave that information in the closure.

skbio/io/_registry.py
```diff
@@ -195,7 +177,7 @@ def decorator(reader):
         # We wrap the reader so that basic file handling can be managed
         # externally from the business logic.
         if cls is None:
-            def wrapped_reader(fp, mode='U', mutate_fh=False, **kwargs):
+            def wrapped_reader(fp, mode='r', mutate_fh=False, **kwargs):
```
ebolyen: You should be able to drop mode here as well. (This might cause some test failures, but these tests can be reframed in the context of a registered operation rather than user input)

skbio/io/_registry.py
```diff
@@ -233,7 +217,7 @@ def wrapped_reader(fp, mode='U', mutate_fh=False, **kwargs):
         else:
             # When an object is instantiated we don't need to worry about the
             # original position at every step, only at the end.
-            def wrapped_reader(fp, mode='U', mutate_fh=False, **kwargs):
+            def wrapped_reader(fp, mode='r', mutate_fh=False, **kwargs):
```
ebolyen: same as above

skbio/io/_registry.py
```diff
@@ -324,7 +308,8 @@ def decorator(writer):
 
         # We wrap the writer so that basic file handling can be managed
         # externally from the business logic.
-        def wrapped_writer(obj, fp, mode='w', **kwargs):
+        def wrapped_writer(obj, fp, mode='w', binary=binary,
```
ebolyen: mode and binary can be dropped from the userspace

skbio/io/_registry.py
```diff
@@ -485,7 +471,7 @@ def _rw_getter(name, fmt, cls):
     return None
 
 
-def sniff(fp, cls=None, mode='U'):
+def sniff(fp, cls=None, mode='r'):
```
ebolyen: mode still shouldn't be user facing, but now we have the potential for sniffers to disagree about what mode the file should be. This is problematic, I think there may be ways around this entirely though. (looking into them)

ebolyen: My bad, this is already handled fine by `reopen_file`

ebolyen:
@jrderuiter I am super excited about this PR! Would you be amenable to my forking of it? I think I know a way to change IO slightly such that we can generalize the kwargs that registered readers/writers/sniffers provide. (This is just an idea, so I would feel bad putting you on the hook for it should it not pan out.) 

Generally, I think we should be able to get away with just an `encoding` and `newline` kwarg. The removal of `'U'` mode also makes me nervous, but that is only because we are using the wrong kind of `open` if we were to change it to `io.open` then Py2 and Py3 will behave the same way! `encoding` can take either `'binary'` or any of the python text codecs, we can default to `"ascii"`. Then with the newlines, registered readers can handle that how they need, by default it can be `None` which is equivalent to Python 2's `File` `"U"` mode in an `io.File`.

ebolyen:
Sorry missed your earlier comment, `mutate_fh` is a defunct kwarg it should be removed. Once upon a time we though `tell` would give use the offset from python's perspective rather than the c-lib underneath. So it seemed reasonable and even useful to be able to read a filehandle without modifying it (by seeking back to the original position each time).

jrderuiter:
@ebolyen I made some adjustments that incorporate the encoding/newline kwargs that you mentioned. Everything seems to be working ok judging from the current test suite and some personal read/write tests. Of course, we should add some new tests to to test the new functionality (I haven't done this yet). We should probably test some files with different encodings/newlines. 

Let me know what you think of the changes. Note I haven't changed any of the documentation, we should do so when we decide on the API.

Feel free to fork this pull request to make any further adjustments!

jrderuiter:
Any updates on your end? :)

ebolyen:
Hey @jrderuiter, sorry I have been neck deep in the Sequence API. I'll have some time to work with this early next week (I hope). I'll fork and try and bring it back up to speed with master since we made a lot of changes and the merge will probably be a little touchy.

jrderuiter:
No problem. I would like to get this functionality into skbio sometime soon though :)

ebolyen:
@jrderuiter We plan on having this no later than Beta, which is  in a couple weeks, so you should be able to work with it really soon! Unfortunately I did need to basically rewrite how IO works because of Py2/Py3 concerns, so I wasn't able to really use much of this PR (that in no way diminishes the importance of the work done here!). 

Preview of things to come:
  -  `skbio.io.open` - works just like python's `open` except it is py2/3 compatible and understands compression, encoding, and any source that we can read from (like URLs (hopefully this list grows))!
  - Modular `IOSource` extensions so that we can continue enhancing IO without needing to fuss with the core logic too much each time.
  - When a file is text-based, readers and writers will speak in python Unicode objects; when binary, Byte objects. This removes all of the ambiguity between py2/py3 and allows us to actually support encoding correctly. 

Thanks again for your initial work here, we are super excited to have people using this library and getting feedback! It definitely helps us target our development efforts.

I should have a PR in with this soon, I'll ping you when it's up.

Issue #910, opened 2015-05-01
subsample_paired_power draws subsamples in the wrong place


JWDebelius:
After simulation, and some consideration, it looks like the current implementation of [`subsample_paired_power`](https://github.com/biocore/scikit-bio/blob/master/skbio/stats/power.py#L328) needs to draw subsamples before each test to fully explore the space and collapse the number of permutations needed.

This addresses some of the issues in #880.

gregcaporaso:
@JWDebelius, do you have an ETA for a fix on this? We're hoping to release 0.3.0 next Thursday, before the sprint, but we don't want to do that release with a known bug.

JWDebelius:
I think I have a fix. It worked for test data, and I'm running on the American Gut now to confirm. I should have a  new PR in by tomorrow.

gregcaporaso:
Fantastic, thank you!

jairideout:
Fixed in #911.

Issue #880, opened 2015-03-21
Make Power Code faster


JWDebelius:
I'd like to improve the run time on the power calculation. Depending on the computer and the test, the run time can be quite long. Running permutations on the American Gut dataset can take upwards of an hour.

This typically involves 10 runs x 5 - 10 points/run x 500 iterations / point  = 25,000 tests - 50,000 tests. When I profiled about 85% of the time was spent on running tests. I'm sure with better profiling, the method could be sped-up more.

There are a couple of steps I have in mind to improve the code, although I'm open to suggestions.

- [ ] Switch the way `subsample_paired_power` draws samples so a new sample is drawn for each round, removing a bias.
- [ ] Optimize the code so to limit the number of points and runs needed, or try to optimize the apparent signal-to-noise so an appropriate number of points are drawn.
- [ ] Parallelize the code-base on permutations. I tried using `multiprocessing.Pool`, but this had several draw backs.

mortonjt:
Awesome!

This may be a good place for my fast permutation test.

josenavas:
How do we want to handle parallelization on scikit-bio? I've seen [joblib](https://pythonhosted.org/joblib/) which is being used, e.g., by scikit-learn. I did not play with it but probably is something that we want to give a shot? What others think?

ebolyen:
Joblib looks like it would make a lot of sense in the context of scikit-learn where the focus is on the algorithms (not the structures) and creating a lightweight pipeline DSL. I'm not sure how well that will work with our current understanding of scikit-bio.
 
I'm not sure how we want to deal with it. It seems a little beyond our scope, unless we could provide some relatively transparent way to support different parallelization schemes. Maybe we can define a superclass that is injected? Then users can extend that superclass with their own parallel drivers.

Alternatively, we can ignore this for now and see if anything interesting comes up in the near future.

wasade:
@JWDebelius, can you post the profile results (just top 10 functions) sorted by cumulative time and by total time please?

JWDebelius:
@wasade, I used line profiler, so it may not be the most accurate representation of actual time. I didn't profile the test, but it's defined as

```
def test_alpha(x):
    vals = [map_.loc[i, 'PD_whole_tree_mean'].values for i in x]
    return scipy.stats.kruskal(*vals)[1]
```

Here are my results:
```
Sorted by total time
--------------------------------------------------------------------------------------------------------------------------------------------
[...]

PR #911, opened 2015-05-02
Fixes subsample power

JWDebelius:
* Addresses the issue raised in #910. 
* Includes features and commentary from #880


JWDebelius:
I keep getting a `The command "PYTHONWARNINGS=ignore WITH_COVERAGE=TRUE make test" exited with 2.` error. I can't find it on google and I don't know what it means. Could someone please help?

JWDebelius:
@wdwvt1, @Jorge-C, @mortonjt: Could any of you review this in the next day or so?

Thank you!

skbio/stats/power.py
```diff
@@ -664,12 +672,11 @@ def bootstrap_power_curve(test, samples, sample_counts, ratio=None,
 
 
 def paired_subsamples(meta, cat, control_cats, order=None, strict_match=True):
-    r"""Gets a set of samples to serve as controls
-
+    r"""Draws a list of samples varied by `cat` and matched for `control_cats`
+    """ """
```
mortonjt: I think this may be accidental ...

skbio/stats/power.py
```diff
@@ -870,49 +866,276 @@ def _compare_distributions(test, samples, counts=5, mode="ind", num_iter=1000):
...
+    i1 = 0
+    i2 = 0
+
+    # Groups the data by the control groups
+    ctrl_groups = meta.groupby(control_cats).groups
+    # Identifies the samples that satisfy the control pairs
+    for (g, ids) in viewitems(ctrl_groups):
+        if not _check_nans(g, switch=True) and strict_match:
+            continue
+        try:
```
mortonjt: Just out of curiosity, what sort of errors do you anticipate catching?

If possible, may want to replace with an if/else statement.  It could make debugging in the future less painful.

skbio/stats/power.py
```diff
@@ -870,49 +866,276 @@ def _compare_distributions(test, samples, counts=5, mode="ind", num_iter=1000):
...
+    index : list
+        A list of numpy arrays describing the positions of samples to be drawn.
+        The first array is an index array. The second gives an integer
+        corresponding to the `control_cat`-group, and the third lists the
+        position of the reference group sample in the list of samples.
+
+    Returns
+    -------
+    ids : list
+        A set of randomnly selected ids groups from each group.
```
mortonjt: Should be "randomly"

skbio/stats/power.py
```diff
@@ -870,49 +866,276 @@ def _compare_distributions(test, samples, counts=5, mode="ind", num_iter=1000):
...
+
+
+def _draw_paired_samples(meta_pairs, index, num_samps):
+    """Draws a random set of ids from a matched list
+
+    Parameters
+    ----------
+    meta_pairs : dict
+        Describes the categories matched for metadata. The
+        `control_cat`-grouped samples are numbered, correspoinding to the
```
mortonjt: Should be "corresponding"

skbio/stats/power.py
```diff
@@ -735,69 +743,49 @@ def paired_subsamples(meta, cat, control_cats, order=None, strict_match=True):
...
 
 
-def _check_strs(x):
-    r"""Returns False if x is a nan and True is x is a string or number"""
+def _get_min_size(meta, cat, control_cats, order, strict_match):
+    """Determines the smallest group represented"""
+    if strict_match:
+        all_cats = copy.deepcopy(control_cats)
+        all_cats.append(cat)
+        meta[all_cats]
```
mortonjt: This was probably left over from debugging.  Don't forget to snip this out

skbio/stats/power.py
```diff
@@ -870,49 +866,276 @@ def _compare_distributions(test, samples, counts=5, mode="ind", num_iter=1000):
...
+        elif mode == "paired":
+            subs = _draw_paired_samples(*samples)
         else:
             subs = [np.random.choice(np.array(pop), counts[i], replace=False)
                     for i, pop in enumerate(samples)]
 
-        p_values[idx] = test(subs)
+        p_values[:, idx] = test(subs)
+
+    if num_p == 1:
```
mortonjt: I think you can get rid of this statement - its a little redundant.  `squeeze()` won't do anything, unless `num_p==1`

mortonjt:
It's looking better!  Good work

skbio/stats/power.py
JWDebelius: This catches the error python throws when it can't locate a sample in the look up. I'd have to do the look up twice to catch the location, which seems more expensive than just building it with try-except. I can specify the exception at the end, which is a `KeyError` exception.

skbio/stats/power.py
JWDebelius: This gets called twice.

mortonjt: ?  

Do you need `meta[all_cats]` ?

JWDebelius: Sorry... I thought you meant the function. Yes, that's left over from debugging.

mortonjt: :)

skbio/stats/power.py
mortonjt: You're referring to lookups in `meta` and `meta_pairs` correct?

Perhaps you can do
```
if ids in meta.index and i1 in meta_pairs:
    ...
else
    ...
```
try, except statements can convolute logic and may make it confusing to read for others, including your future self.
I personally try to avoid them when they aren't necessary.

wasade: @JWDebelius, exceptions are expensive when they happen, so it is only cheaper to do it this way if the exception is unlikely. That being said, unless you've profiled and shown that this actually is a bottleneck, it is better to err on the side of readability.

jairideout: If the `try`/`except` statement is kept, please specify what exception types are to be caught in the `except` block. It's generally not a good idea to have an `except` statement that catches everything. For example, if the user tries to kill their program with ctrl-c, a `KeyboardInterrupt` exception will be caught and ignored.

JWDebelius: The if statement, as best I can define it would be:
```python 
if (ids in meta.index.values and np.array([o in meta.loc[ids].groupby(cat).groups for o in order]).all()):
```

@jairideout: I'm sorry. I forgot that I could do that. But, I'm happy to set it to `except KeyError` if its okay to keep the try/except logic.

wasade: That requires a list lookup and a separate list traversal.

Just to make sure I understand what's going on:

1) in this forloop, groups of samples are being pulled out
2) it is possible (rarely? frequently?) that some IDs do not exist in some groups
3) an efficient and explicit check is desired to determine if these IDs exist in the groups

If I read this correctly, the current place where a `KeyError` is thrown is either line 1074 or 1075. Is that correct?

JWDebelius: The list look ups is why I used a try/except statement. It seemed cleaner.

You are understanding the problem. The `KeyError` can be thrown at 1074 or 1075 (but, it's most likely to be 1075).

wasade: Okay, thank you. 

From playing in an interpreter, I'm unable to get a command like what is on 1074 to throw an exception:

```python
19:58:46 (mcdonadt@Daniels-MacBook-Pro-3):/Users/mcdonadt/ResearchWork/software/American-Gut/data/AG
> md = pd.read_csv('AG.txt', sep='\t')

19:59:56 (mcdonadt@Daniels-MacBook-Pro-3):/Users/mcdonadt/ResearchWork/software/American-Gut/data/AG
> md.loc[[0, 1]].groupby('SEQ_RUN').groups
Out[20]: {'no_data': [0, 1]}

20:00:03 (mcdonadt@Daniels-MacBook-Pro-3):/Users/mcdonadt/ResearchWork/software/American-Gut/data/AG
> md.loc[['foo', 'bar']].groupby('SEQ_RUN').groups
Out[21]: {}
```

With respect to line 1075, it should be possible to do `id_vecs = [m_ids[o] for o in order if o in m_ids]`

What this then means is that it should be possible to do something like the following without try/except:

```python
m_ids = meta.loc[ids].groupby(cat).groups
id_vecs = [m_ids[o] for o in order]
if id_vecs:
    # ... all that other stuff
else:
    index1.append(np.array([]))
    index2.append(np.array([]))
```

Does that make sense and seem like it'll work?

skbio/stats/power.py
```diff
@@ -870,49 +866,276 @@ def _compare_distributions(test, samples, counts=5, mode="ind", num_iter=1000):
...
+            index1.append((loc_vec + i2))
+            index2.append(np.zeros(loc_vec.shape) + i1)
+            i1 = i1 + 1
+            i2 = i2 + len(loc_vec)
+        except:
+            index1.append(np.array([]))
+            index2.append(np.array([]))
+
+    index = [np.hstack(index1), np.hstack(index2)]
+    if meta_pairs == {}:
```
wasade: `if not meta_pairs`

JWDebelius:
@mortonjt, @wasade: Thank you for the review!

JWDebelius:
@jairideout, @mortonjt, @wasade: Are there any more changes that need to be made so this can be merged?

skbio/stats/power.py
```diff
@@ -173,28 +177,34 @@ def subsample_power(test, samples, draw_mode='ind', alpha_pwr=0.05,
...
         on the curve.
-    num_runs : unsigned int, optional
+    num_runs : positive int, optional
         The number of times to calculate each curve.
 
     Returns
     -------
     power : array
-        The power calculated for each subsample at each count.
+        The power calculated for each subsample at each count. The array is has
```
mortonjt: I'm guessing you meant `The array has` rather than `The array is has`

skbio/stats/power.py
```diff
@@ -360,48 +363,50 @@ def subsample_paired_power(test, meta, cat, control_cats, order=None,
...
         The number of p-values to generate for each point on the curve.
-    num_runs : unsigned int, optional
+    num_runs : positive int, optional
         The number of times to calculate each curve.
 
     Returns
     -------
     power : array
-        The power calculated for each subsample at each count.
+        The power calculated for each subsample at each count. The array is has
```
mortonjt: Same here

skbio/stats/power.py
```diff
@@ -209,6 +219,10 @@ def subsample_power(test, samples, draw_mode='ind', alpha_pwr=0.05,
     ValueError
         If the `counts_interval` is greater than the difference between the
         sample start and the max value, the function raises a ValueError.
+    ValueError
+        There are not an equal number of groups in `samples` and in `ratios`.
+    TypeError
+        `test` does not return a float or a 1-dimensional numpy array.
 
 
     Examples
```
mortonjt: Do you think it would be worth it to put another example with the new ratio parameter?

JWDebelius: I expanded the example to include one that addressed the ratio parameter.

JWDebelius:
@mortonjt, @Jorge-C, @ElDeveloper, @ebolyen, @jairideout: Could someone review this again to see if its ready for merge?

Thank you!

JWDebelius:
Bumping this again? Please? @mortonjt, @Jorge-C, @ElDeveloper, @wasade, @ebolyen, @jairideout

I'd really like to get this merged with the bug fix for #910. It has all the fixes from #875.

It would be awesome if there could be a release with this for American Gut.

Thanks!

skbio/stats/power.py
mortonjt: Don't forget to address this comment.

mortonjt:
Are there any references for these power calculations?  Might want to add them to this PR.
But even if we don't include them, I'd be interested in following up on the math behind the scenes.

Also, don't forget to resolve merge conflicts.

mortonjt:
Otherwise, I think this PR is getting close to merge.  Does anyone else want to take another glance at this?

skbio/stats/power.py
JWDebelius: Yes, but this more readable.

JWDebelius:
The paper is next on my list after American Gut goes out. So, no reference until it gets written. Basically, this and the American Gut power notebook are the two documents which currently contain the formal math.

gregcaporaso:
@JWDebelius, can you merge from upstream master - there are currently merge conflicts. We'll get back to you on this later today. 

Re: doing a release, our next planned one (0.3.0) is going to break backward compatibility, so updating to use this for American Gut may be a little bit of work. That release is planned for early July, just before SciPy. What is your timing for publication on this? If this timing won't work let's discuss some alternatives.

JWDebelius:
@gregcaporaso. Thank you. Sorry it took so long to updated.

In my ideal world, I could have the updated notebook when we submit the AG manuscript. We're trying to finalize it for submission in the next week or so.

jairideout:
@JWDebelius I'll take a final pass through today.

skbio/stats/power.py
```diff
@@ -173,28 +177,34 @@ def subsample_power(test, samples, draw_mode='ind', alpha_pwr=0.05,
         "ind" mode should be used.
     alpha_pwr : float, optional
         The critical value used to calculate the power.
-    min_observations : unsigned int, optional
```
jairideout: Please note the removal of this parameter (including the reason why it's removed) under "Backward-incompatible changes".

CHANGELOG.md
```diff
@@ -8,18 +8,22 @@
...
 
 ### Performance enhancements
 * The speed of quality score decoding has been significantly improved (~2x) when reading `fastq` files.
 * The speed of `NucleotideSequence.reverse_complement` has been improved (~6x).
 
+
 ### Bug fixes
 * Changed `BiologicalSequence.distance` to raise an error any time two sequences are passed of different lengths regardless of the `distance_fn` being passed. [(#514)](https://github.com/biocore/scikit-bio/issues/514)
 * Fixed issue with ``TreeNode.extend`` where if given the children of another ``TreeNode`` object (``tree.children``), both trees would be left in an incorrect and unpredictable state. ([#889](https://github.com/biocore/scikit-bio/issues/889))
+* Changed the way power was calculated in `subsample_paired_power` to move the subsample selection before the test is performed, as happens in `subsample_power`. This improves the accuracy of the returned estimate.
```
jairideout: Can you include a link to the relevant issue (#910)? See the other changelog entries for examples of how this should be formatted.

skbio/stats/power.py
```diff
@@ -253,82 +264,124 @@ def subsample_power(test, samples, draw_mode='ind', alpha_pwr=0.05,
...
+            power[id1, id2, :] = _calculate_power(ps, alpha_pwr)
+
+    power = power.squeeze()
 
     return power, sample_counts
 
 
 def subsample_paired_power(test, meta, cat, control_cats, order=None,
                            strict_match=True, alpha_pwr=0.05,
-                           min_observations=20, max_counts=50,
```
jairideout: Same comment as above about removal of `min_observations` parameter.

skbio/stats/power.py
```diff
@@ -569,6 +615,12 @@ def bootstrap_power_curve(test, samples, sample_counts, ratio=None,
                           alpha=0.05, mode='ind', num_iter=500, num_runs=10):
     r"""Repeatedly calculates the power curve for a specified alpha level
 
+    .. note:: Deprecated in scikit-bio 0.2.3-dev
+       ``bootstrap_power_curve`` will be removed in scikit-bio 0.3.1. It is
```
jairideout: 0.3.1 -> 0.4.1

skbio/stats/power.py
```diff
@@ -870,49 +914,277 @@ def _compare_distributions(test, samples, counts=5, mode="ind", num_iter=1000):
...
-
-    # Prealocates the pvalue matrix
-    p_values = np.zeros((num_iter))
-
     for idx in range(num_iter):
         if mode == "matched":
             pos = np.random.choice(np.arange(0, samp_lens[0]), counts[0],
                                    replace=False)
             subs = [sample[pos] for sample in samples]
+        elif mode == "paired":
```
jairideout: Can you add a unit test for this branch?

skbio/stats/power.py
```diff
@@ -870,49 +914,277 @@ def _compare_distributions(test, samples, counts=5, mode="ind", num_iter=1000):
...
+        sample start and the max value, the function raises a ValueError.
+    ValueError
+        There are not an equal number of groups in `samples` and in `ratios`.
+    TypeError
+        `test` does not return a float or a 1-dimensional numpy array.
+
+    """
+
+    if draw_mode not in {'ind', 'matched'}:
+        raise ValueError('mode must be "matched" or "ind".')
```
jairideout: Can you add a unit test for this branch?

CHANGELOG.md
jairideout: Can you also expand the description of how and why the previous output was wrong? i.e. the severity of the bug, if the previous output should be completely mistrusted, etc.

skbio/stats/power.py
```diff
@@ -145,20 +142,27 @@
 from future.utils import viewitems
 from future.builtins import range
 
+from warnings import warn
```
jairideout: Order these new imports as:

```python
...
from future.builtins import range

import collections
import copy
import warnings

import numpy as np
...
```

jairideout:
@JWDebelius I took a quick pass through and had some comments, most should be pretty easy to resolve. Let me know when they're addressed and this should be ready for merge!

Note that I didn't perform a detailed review of the actual power code itself; I'm assuming the previous reviewers have this covered (in this PR and the reviews originally in #875).

JWDebelius:
@jairideout, I've finished review. 

The failures I'm getting are due to Permanova, anosim, and gradient test failures. I haven't touched any of the files that are causing the error. It'm not sure if its a stochastic travis failure, and Im not sure how to solve the failures.

Thanks!

jairideout:
Thanks @JWDebelius! The failures are unrelated to your changes -- I have a fix for those in another PR that's open right now.

PR #915, opened 2015-05-05
Added ``weblogo`` method to ``skbio.Alignment``.

### LINKED ISSUE #805: make Alignment plottable as a sequence logo
@gregcaporaso, @Kleptobismol, and I chatted awhile ago about adding a plotting method to `Alignment` to create a [sequence logo](http://schneider.ncifcrf.gov/sequencelogo.html). This could be a really cool visualization and might be a good default plot type for `Alignment`.

Kleptobismol:
This pull request adds a method which plots an ``Alignment`` object as a [weblogo](http://genome.cshlp.org/content/14/6/1188.full) (sometimes called sequence logo).

This code is based on code originally written by @widmannj. Thank you Jeremy for letting us use this.

Here's an example:
justplainexample

I have a question about the aesthetics of the weblogo. This is an example of a weblogo (not created with my program) 
weblogo example 
And here is a picture of one created using my program 
exampleweblogo
 The X and Y axis lines that connect the edges of the box in the second picture (indicated by the red brackets) in my opinion look a little out of place. It's not a big deal, honestly, in my opinion, whether or not they are there, but I'd like to know what the community thinks: if it matters, if they should be removed, etc.

Fixes #805.


gregcaporaso:
Thanks @Kleptobismol! 

I agree those axis lines look a little out of place and could be removed.

skbio/alignment/_alignment.py
```diff
@@ -1405,6 +1408,84 @@ def sequence_length(self):
         else:
             return len(self._data[0])
 
+    def weblogo(self, reverse=False):
+        """Plots alignment as a weblogo.
+
```
wasade: @Kleptobismol, would it be possible to add in a `Notes` section that gives a little more detail on what a weblogo is? It may be useful to also include a `References` section that links out to either [here](http://weblogo.berkeley.edu/) or the papers described in the link

skbio/alignment/_alignment.py
```diff
@@ -1405,6 +1408,84 @@ def sequence_length(self):
...
+        if self.sequence_count() == 0:
+            raise AlignmentError("You have passed an empty Alignment.")
+
+        # acquires amount of each nucleotide at a given point
+        pos_freq = self.position_frequencies()
+
+        # sets up graph with x and y ticks and labels
+        fig, ax = plt.subplots()
+        fig.set_size_inches([(len(pos_freq)*1.3), 6])
+        ax.set_xticks(np.array([i+0.35 for i in range(len(pos_freq))]))
```
wasade: Possible to leverage numpy further here? For instance:

```python
ax.set_ticks(np.arange(0.35, len(pos_freq)))
```

skbio/alignment/_alignment.py
```diff
@@ -1405,6 +1408,84 @@ def sequence_length(self):
...
+        ax.set_xlim(right=len(pos_freq))
+        ax.set_xticklabels(range(1, self.sequence_length()+1, 1))
```
wasade: The `step` is not necessary here as it will default to 1

skbio/alignment/_alignment.py
```diff
@@ -1405,6 +1408,84 @@ def sequence_length(self):
...
+        ax.set_ylim(top=1.0, bottom=0.0)
+        ax.set_yticks([0.0, 0.25, 0.5, 0.75, 1.0])
+        ax.set_yticklabels(['0.0', '', '', '', '1.0'])
+
+        # runs through every sequence and graphs it
+        for i in range(len(pos_freq)):
```
wasade: Iterating over `pos_freq` would reduce the lookup overhead in the inner forloop. For instance:

```python
for idx, pos in enumerate(pos_freq):
    ...
```

Since `pos` is a `dict`, the `del` operation will operate in place on it.

skbio/alignment/_alignment.py
```diff
@@ -1405,6 +1408,84 @@ def sequence_length(self):
...
+            for gap in BiologicalSequence.gap_alphabet():
+                if gap in pos_freq[i]:
+                    # instead of drawing an empty box for gaps, we delete gap
+                    # positions
+                    del pos_freq[i][gap]
+            self._draw_position(ax, i, pos_freq[i], reverse)
+
+        return fig
+
+    def _draw_position(self, ax, position, freqs, reverse):
```
wasade: This is minor, but would it be possible for this private member method name to explicitly indicate it is associated with `weblogo`? Something like `_draw_weblogo_position`?

skbio/alignment/_alignment.py
```diff
@@ -11,11 +11,14 @@
...
 from six import StringIO
 
+import operator
 from collections import Counter, defaultdict, OrderedDict
 
 import numpy as np
+import matplotlib.pyplot as plt
 from scipy.stats import entropy
 
+from skbio.alignment._text import add_letter
```
wasade: I don't think this was included in the PR?

wasade:
@Kleptobismol, just a few minor comments. I think this looks great though, and I really like how the resulting figure is tested as well.

jairideout:
Thanks for reviewing @wasade!

@Kleptobismol there are merge conflicts so the Travis tests didn't run. Also `skbio.alignment._text` wasn't included in the PR. We can work on resolving the merge conflict tomorrow when we meet if you need help with that.

I think the method should be renamed to `sequence_logo` since that's the type of plot that's generated, and all occurrences of "weblogo" in the text should become "sequence logo". [WebLogo](http://weblogo.berkeley.edu/logo.cgi) is a website for creating sequence logos, so we shouldn't use that name here.

The y-axis is in relative frequencies but [according to wikipedia](http://en.wikipedia.org/wiki/Sequence_logo) the y-axis in a sequence logo displays information content in bits. We can chat more about this when we meet but I think we'll want to use the equations in the paper you linked to (also on the wikipedia page).

Thanks for your work on this!

skbio/alignment/_alignment.py
RNAer: :+1:  more information would be great. I remember original paper uses information content rather than frequency in seq logo (?). would it be useful to provide optional parameter to calculate either of them?

jairideout:
> I agree those axis lines look a little out of place and could be removed.

I agree with this change too. @Kleptobismol you can remove the upper and right spines following [these instructions](http://stackoverflow.com/a/9751369/3776794).

skbio/alignment/_alignment.py
Kleptobismol: Fixed.

skbio/alignment/_alignment.py
Kleptobismol: Fixed.

skbio/alignment/_alignment.py
Kleptobismol: Fixed.

skbio/alignment/_alignment.py
Kleptobismol: Fixed.

jairideout:
@Kleptobismol the test failures are unrelated to your changes. I created a separate issue to get this fixed (#930).

skbio/alignment/_alignment.py
Kleptobismol: Fixed.

Kleptobismol:
All the errors except 2 appear to be from test_gradient code, which I had no part in, does anyone know whose code this is, and if it has been fixed? Also, any help fixing these two errors in 3.4 would be much appreciated. I'm very confused as to what's going on with them.

skbio/alignment/_alignment.py
Kleptobismol: Fixed.

jairideout:
@Kleptobismol I'm not sure offhand what the Python 3 errors are, let's look at this when we meet today or tomorrow.

RNAer:
wondering what is the status of this PR? this is certainly very useful.

jairideout:
This is old code applied to `Alignment`, which is now `TabularMSA`, and is not being actively developed. There are some issues to sort out in order to match the output from WebLogo, and I think the y-axis needs to be entropy instead of conservation. There are also better ways to accomplish the matplotlib letter generation which will simplify the code quite a bit. See discussion in #805. If anyone wants to work on this we would appreciate the help. I recommend using this code as a reference (if it is helpful) but start from scratch with a new PR.

PR #924, opened 2015-05-12
PCoA to pcoa

wasade:
Reduces the `PCoA` object to a first-class function.


jairideout:
I pulled these commits into #925.

PR #931, opened 2015-05-23
[DO NOT MERGE] Replace Sequence.quality with ranged_metadata

anderspitman:
There are several items to discuss before finalizing how this will work. At least:

- [ ] Currently removed the test that quality is not copied when initializing a sequence. How do we want this to work?

- [ ] Should we try to make quality read-only? It was before but isn't in this commit.

- [ ] Certain quality-specific logic maybe should be removed in favor of more generic metadata logic. For example there are places where `_has_quality` should probably be replaced with `_has_metadata`, etc.

- [ ] Not entirely sure I'm happy the the overall cleanliness of these changes. Open to suggestions.

- [ ] We likely will need additional new tests now that we're using a Pandas DataFrame to hold the quality data.

- [ ] How will metadata fit in to sequence equality comparisons?

EDIT: Added metadata equality item.

### DIFF
--- a/skbio/io/_base.py
+++ b/skbio/io/_base.py
@@ -178,12 +178,15 @@ def _format_fasta_like_records(generator, id_whitespace_replacement,
         else:
             header = id_
 
-        if require_qual and seq.quality is None:
+        if require_qual and 'quality' not in seq.ranged_metadata:
             raise ValueError(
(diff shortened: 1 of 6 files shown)


anderspitman:
Closing in favor of #938

PR #944, opened 2015-06-10
REL: __version__ now exists in one place

### LINKED ISSUE #939: define scikit-bio version in a single place
Currently, scikit-bio's version is stored in `skbio.__version__` and `setup.py`. It'd be nice to only store this info once.

johnchase:
version in `setup.py` is now defined by the `__version__` that is present in `__init__.py`
fixes #939


jairideout:
I haven't reviewed this in depth yet, but can you add Flask's license to `licenses/`? This is important because you're copying code from another project.

johnchase:
@jairideout I added the flask license to the licenses directory. Are there changes that need to be made elsewhere that I am missing regarding licensing?

setup.py
```diff
@@ -13,6 +13,8 @@
 from setuptools import find_packages, setup
 from setuptools.extension import Extension
 from setuptools.command.build_ext import build_ext as _build_ext
+import re
```
jairideout: move these to the top with the other package-level imports (`os`, `platform`)

jairideout:
The licensing stuff looks good, thanks for adding it. One other comment: can you update RELEASE.md so that it is consistent with this new versioning scheme?

Jorge-C:
I'd also suggest squashing the commits to remove those unrelated to the PR for clarity (the three merges, biopy addition and removal). Thanks!

Issue #923, opened 2015-05-11
Stable, experimental and deprecated functionality


wasade:
A user should know:

- Whether functionality is stable, experimental or deprecated
- If the functionality is deprecated, the user should have an expectation about when it will be removed, and why
- If the functionality is experimental, the user should know when the functionality was added

A developer can specify the status of functionality. Experimental functionality is to be experimental for <= 2 releases. If > 2 releases, the release manager must explicitly mark the functionality as overdue for the given release. 

All experimental code is to be placed under `skbio.experimental`. Any import from from `skbio.experimental` will raise a warning.

A developer can learn about how to mark methods through `CONTRIBUTING.md`

A state diagram for functionality:
 
![hackpad com_2fyrtr6iinw_p 327785_1431380076978_slide1](https://cloud.githubusercontent.com/assets/474290/7576014/0449259e-f7f7-11e4-8056-811c92f9c6af.jpg)

gregcaporaso:
This was addressed in #945.

PR #945, opened 2015-06-11
adding stability state decorators

gregcaporaso:
This contains a first pass at defining stability states of code in scikit-bio (#923). Our initial discussions of this happened at the UCSD skbio sprint in May, and @ebolyen and I spent some more time discussing after that resulting in [this doc](https://hackpad.com/skbio-stableexperimentaldeprecated-ideas-mRBeikR4JXX). I'm going to be offline for a few days, so wanted to get my work on this visible, but there is still more to do before merge. After merging, in a separate pull request, we'll work through the code base and define the stability state of all public functionality. 

I updated the state diagram to indicate the three possible states more clearly: experimental, stable, and deprecated. 

image

The docstring of the base class for these decorators describe the theory behind these. That will be ported to an easily accessible document before merge. 

We ultimately decided to use decorators instead of splitting the code base into ``skbio/`` and ``skbio/experimental`` as we can achieve what we want to with these decorators with a lot less complexity for new developers. (For example, instead of having to tell people to move code to different places, define new experimental methods of stable classes as functions under ``skbio/experimental``, and so on, we can just decorate functionality.) 

To-do items following code review with @ebolyen, @jairideout and @anderspitman. 

Next steps:
 - [x] How does numpydoc format deprecation warnings?
 - [x] ``UserWarning`` to ``DeprecationWarning``
 - [x] experimental ``until`` to ``as-of``
 - [x] deprecated should take ``as-of`` in addition to ``until``
 - [x] force naming of kwargs for readability
 - [x] update ``deprecated``'s ``wrapped_f.__name__``, etc (see [link from Evan](http://stackoverflow.com/questions/147816/preserving-signatures-of-decorated-functions))
 - [x] better handling of missing keys in the constructors
 - [x] move high-level doc to new rst (maybe under *User Documentation*)
 - [x] add unit tests

Longer term:
 - ``deprecated_parameter`` (#958)
 - plan for deprecating full modules and subpackages (#959)


wasade:
Seen update_wrapper by chance?

https://docs.python.org/2/library/functools.html

gregcaporaso:
I hadn't, thanks for the pointer, I'll check into it when I get back to this next week.

ebolyen:
Unfortunately neither update_wrapper nor wraps will copy the function
signature, which is the hard part to copy (takes introspection and screwing
with the function object) the decorator package can take care of that
nastiness.

wasade:
That's annoying `update_wrapper` doesn't do it. Thanks for the heads up with the decorators package, that does look pretty awesome. Took a quick stroll through the code as well. One question though, what's the reason for needing to copy the function signature? Other packages, such as Click, which make extensive use of decorators, do not go that deep. From the PR, it looks like the motivation is for "update deprecated's wrapped_f.__name__", but `update_wrapper` will do that.

ebolyen:
Click can really get away with throwing away your original function, it just wraps for politeness (it also actually modifies the signature is many cases, rendering the old one useless). Similarly our IO registry is actually augmenting our signature so keeping the old one is still unhelpful. However in this case, we need Sphinx to be able to document experimental, and deprecated functionality so we need the signature copied. Greg just wasn't that specific in his OP.

wasade:
Read more about the package this afternoon, and agree with the specific
utility of doc. Should be awesome

skbio/util/__init__.py
```diff
@@ -72,7 +72,8 @@
 
 from ._warning import EfficiencyWarning
 from ._exception import TestingUtilError
-from ._decorator import classproperty, overrides
+from ._decorator import (classproperty, overrides, stable, experimental,
```
jairideout: don't export here

doc/source/index.rst
```diff
@@ -21,6 +21,16 @@ API Reference
    io
    util
 
+User Documentation
+------------------
+
+The user documentation contains high-level information for users of scikit-bio.
```
gregcaporaso: This section of the documentation (*User* and *Developer*) is a little awkward. I think we should reorganize this a bit, but not sure if it can happen before beta.

gregcaporaso:
@jairideout, @anderspitman, @ebolyen, this is ready for final review.

doc/source/index.rst
jairideout: :+1: let's revisit during the website sprint next week.

doc/source/user/api_stability.rst
```diff
@@ -0,0 +1,57 @@
+API Stability
+=============
+
+All public functionality in scikit-bio has a defined stability state.
+These states inform users and developers to what extent they can rely on
+different APIs in the package.
+
+You can find out the stability state of public functionality by looking at its
```
jairideout: Can you add a link to numpydoc here and note that our docstrings follow this format? Then your references to short/extended summaries will be clearer.

skbio/util/_decorator.py
```diff
@@ -7,6 +7,9 @@
 # ----------------------------------------------------------------------------
 
 from __future__ import absolute_import, division, print_function
+from warnings import warn
```
jairideout: Can you import these packages instead of importing pieces from them? In general we're trying to avoid these types of imports because it breaks the package namespace and makes it less clear where a function came from.

```python
import textwrap
import warnings

import decorator # this is 3rd party so put in its own section
```

skbio/util/_decorator.py
```diff
@@ -82,3 +85,207 @@ def __get__(self, cls, owner):
 
     def __set__(self, obj, value):
         raise AttributeError("can't set attribute")
+
+
+class _state_decorator(object):
+    """ Base class for decorators of all public functionality.
+    """
+
+    _line_prefix = '\n        '
```
jairideout: Can you add a comment describing the specific number of trailing spaces here? Or build `_line_prefix` in such a way that it's self-documenting.

skbio/util/_decorator.py
```diff
@@ -82,3 +85,207 @@ def __get__(self, cls, owner):
...
+        # state_desc_prefix, while the others should start with spaces to align
+        # the text in this section. This is for consistency with numpydoc
+        # formatting of deprecation notices, which are done using the note
+        # Sphinx directive.
+        state_desc_lines[0] = '%s%s' % (state_desc_prefix, state_desc_lines[0])
+        header_spaces = ' ' * len_state_desc_prefix
+        for i, line in enumerate(state_desc_lines[1:], 1):
+            state_desc_lines[i] = '%s%s' % (header_spaces, line)
+
+        doc_lines.insert(
```
jairideout: This `insert` could be expensive for large docstrings O(n). Since this will be applied to every docstring in scikit-bio it's probably worth optimizing to avoid inserts at the beginning of the list.

skbio/util/_decorator.py
```diff
@@ -82,3 +85,207 @@ def __get__(self, cls, owner):
...
     def __set__(self, obj, value):
         raise AttributeError("can't set attribute")
+
+
+class _state_decorator(object):
+    """ Base class for decorators of all public functionality.
+    """
+
+    _line_prefix = '\n        '
+    _required_kwargs = []
```
jairideout: I recommend making this a tuple to avoid bugs where this is mutated and it affects all instances since it is a class variable.

skbio/util/tests/test_decorator.py
```diff
@@ -72,5 +75,146 @@ def foo(cls):
...
+    def test_function_output(self):
+        f = self._get_f('0.1.0', until='0.1.4',
+                        reason='You should now use skbio.g().')
+        self.assertEqual(f(1), 43)
+
+    def test_deprecation_warning(self):
+        f = self._get_f('0.1.0', until='0.1.4',
+                        reason='You should now use skbio.g().')
+        # adapted from SO example here: http://stackoverflow.com/a/3892301
+        with warnings.catch_warnings(record=True) as w:
```
jairideout: use `npt.assert_warns` (grep codebase to find examples)

jairideout: Nevermind, that won't allow you to test the deprecation message, just that the correct type of warning is raised.

jairideout:
@gregcaporaso this looks great! I had a few comments, most should be easy to resolve.

skbio/util/_decorator.py
gregcaporaso: Skimmed over this before, but it was a bit of a magic number. Thanks for catching, fixed now.

doc/source/user/api_stability.rst
gregcaporaso: Done.

gregcaporaso:
@jairideout, this is ready for re-review.

skbio/util/_decorator.py
```diff
@@ -82,3 +86,252 @@ def __get__(self, cls, owner):
...
+                          state_desc_prefix='State: '):
+        docstring_content_indentation = \
+            self._get_indentation_level(docstring)
+        # Hande the case of no initial docstring
+        if docstring is None:
+            return "%s%s%s" % (
+                ' ' * docstring_content_indentation, state_desc_prefix,
+                state_desc)
+
+        doc_lines = docstring.split('\n')
```
jairideout: You're splitting the docstring twice, here and in `_get_indentation_level`. Can you refactor to avoid performing this operation twice? I'm mainly concerned about efficiency, not so much code duplication.

jairideout:
Thanks @gregcaporaso, had one more comment regarding efficiency (should be easy to fix). Once that's done this is ready for merge!

gregcaporaso:
Done, good call.

jairideout:
Thanks!

Issue #949, opened 2015-06-12
next steps for sequence objects


jairideout:
Next steps for sequence objects:

- [x] make sure we're shallow copying `metadata` and `positional_metadata` when appropriate to avoid cascading effects if a user updates metadata on a "parent" object (@jairideout)

- [x] add setters/deleters for `metadata`/`positional_metadata` with same guards as constructor (@jairideout)

- [x] ensure `positional_metadata` has uniform index set (@jairideout)

- [x] lowercase API improvements (mainly for I/O) (@anderspitman)

- [x] fix `six.b` bug in `_munge_to_bytestring` (@anderspitman)

- [x] add `Sequence.copy` method, remove usage of `_to` outside of `Sequence` class in favor of `copy` (@jairideout)

- [x] update `Sequence.__repr__` to be more informative and human-readable. ensure doctests are using new repr instead of printing pieces of the sequence object  (@jairideout)

- [x] `StockholmAlignment` removal (@gregcaporaso)

- [x] refactor to make `NucleotideSequence` a mix-in (@anderspitman)

- [x] add `Sequence.__nonzero__` (@anderspitman)

~~- [ ] add HTML repr for `Sequence` (@anderspitman)~~

- [x] add `NucleotideSequence.gc_frequency(relative=False)` and `NucleotideSequence.gc_content` (@anderspitman)

- [x] mark stable API (@gregcaporaso)

- [x] final pass through sequence documentation to make sure everything is up-to-date with latest changes (e.g., regarding mutability) (unassigned)

ebolyen:
@jairideout with regards to `six.b` that usage is correct, in py2 you could write `bytes("string here")` and it would work fine, but py3 requires an encoding as an argument. Py2 naturally does not support an encoding argument as it is actually an alias for `str`. `six.b` just mediates that. `six.u` is also useful and does an analogous thing but for unicode.

PR #950, opened 2015-06-13
ENH: add Sequence.metadata property setter

jairideout:
Addresses part of #949.

@anderspitman or @ebolyen can you please review?


ebolyen:
Looks good!

PR #952, opened 2015-06-18
DOC: Update the documentation for lladser_pe

mortonjt:
This is adds a note that clarifies that `skbio.diversity.alpha.lladser_pe` will not return probabilities less than 1 if the coverages is too low.


jairideout:
Thanks @mortonjt!

Jorge-C:
Is there an issue for the tests that fail because of some `pdt.assert_series_equal` problem?

mortonjt:
idk what's going on there - there's only 1 line difference

jairideout:
Master is currently failing due to a new version of pandas that has stricter testing utility requirements (e.g., `assert_series_equal`). I have a fix for this in #954.

PR #953, opened 2015-06-18
BUG: Fixed creation of non-symmetric matrices from `randdm`.

### LINKED ISSUE #943: random distance matrices aren't always symmetric
[This line](https://github.com/biocore/scikit-bio/blob/master/skbio/stats/distance/_base.py#L772) does not create symmetric matrices (since it's modifying it in place, it depends on how numpy will traverse the memory to make the updates).

Fortunately, this is caught by the `DistanceMatrix` constructor.

shiffer1:
Fixes #943 Pair programmed with @johnchase 
Added test to `test_base.py` for detection of whether large distance matrices were symmetric.
Modified `_base.py` to stop creation of non-symmetric distance matrices and to pass test.
Updated `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 ### Bug fixes
 * Changed `BiologicalSequence.distance` to raise an error any time two sequences are passed of different lengths regardless of the `distance_fn` being passed. [(#514)](https://github.com/biocore/scikit-bio/issues/514)
 * Fixed issue with ``TreeNode.extend`` where if given the children of another ``TreeNode`` object (``tree.children``), both trees would be left in an incorrect and unpredictable state. ([#889](https://github.com/biocore/scikit-bio/issues/889))
-* Fixed issue where non-symmetric distance matrices were being created by `randdm`.
+* Fixed issue where `randdm` was attempting to create asymmetric distance matrices.This was causing an error to be raised by the `DistanceMatrix` constructor inside of the `randdm` function, so that `randdm` would fail when attempting to create large distance matrices.
 ([#943](https://github.com/biocore/scikit-bio/issues/943))
 
 ### Deprecated functionality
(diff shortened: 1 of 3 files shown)


skbio/stats/distance/tests/test_base.py
```diff
@@ -542,6 +542,9 @@ def test_default_usage(self):
 
         self.assertTrue(found_diff)
 
+        obs3 = randdm(100)
+        self.assertTrue(obs3 == obs3.T)
```
Jorge-C: Maybe `self.assertEqual(obs3, obs3.T)` would be more clear.

Also, it should live in its own method with a descriptive name (say, `def test_large_matrix_for_symmetry`) so that when it fails we know exactly what failed, as opposed to big test methods where a failure doesn't really tell you what exactly failed.

Jorge-C:
A small comment, but looks good otherwise!

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@
 ### Bug fixes
 * Changed `BiologicalSequence.distance` to raise an error any time two sequences are passed of different lengths regardless of the `distance_fn` being passed. [(#514)](https://github.com/biocore/scikit-bio/issues/514)
 * Fixed issue with ``TreeNode.extend`` where if given the children of another ``TreeNode`` object (``tree.children``), both trees would be left in an incorrect and unpredictable state. ([#889](https://github.com/biocore/scikit-bio/issues/889))
+* Fixed issue where non-symmetric distance matrices were being created by `randdm`.
```
jairideout: This is minor but important: the current wording implies that `randdm` actually created asymmetric distance matrices. The function *tried* to create asymmetric distance matrices but an error was raised by the `DistanceMatrix` constructor when this happened. Can you reword so that this is clearer? The current wording makes it sound like invalid data was being created by `randdm`, which is far worse of a bug than having an error raised.

jairideout:
Thanks @shiffer1 and @johnchase!

PR #954, opened 2015-06-18
ENH: add setters/deleters for Sequence.metadata/positional_metadata

### LINKED ISSUE #948: memory optimization for Sequence metadata
`Sequence.positional_metadata` is currently initialized to an empty `DataFrame` with 0 columns and an index the size of the sequence if positional metadata isn't provided. To save on memory when positional metadata isn't present, the `Sequence.positional_metadata` property could be smarter and only create the `DataFrame` when the user asks for it (e.g., if they wanted to add positional metadata after instantiation). Internal to the `Sequence` class, a private attribute will be used instead of the `positional_metadata` property.

This same strategy can be used for `Sequence.metadata` (which is a `dict`).

jairideout:
Other changes:
  - Added shallow copying of metadata/positional metadata and docs describing this behavior.
  - Positional metadata index is now always set to ``np.arange(len(self))``.
  - Added ``skbio.util.assert_data_frame_almost_equal`` and updated unit tests to use it.

Address part of #949. Fixes #948.


jairideout:
@ebolyen @anderspitman can you both review this? I think this one's worth having a couple set of eyes on.

skbio/sequence/_sequence.py
```diff
@@ -106,20 +99,181 @@ class Sequence(collections.Sequence, SkbioObject):
...
+    ['Alice', 'Bob', 'Carol']
+    >>> metadata['authors']
+    ['Alice', 'Bob', 'Carol']
+
+    The behavior for updating positional metadata is similar. Let's create a
+    new sequence with positional metadata that is already stored in a
+    ``pd.DataFrame``:
+
+    >>> positional_metadata = pd.DataFrame(
+    ...     {'qual': [3, 3, 4, 10], 'list': [[], [], [], []]})
```
ebolyen: Change this to `quality` since that is what `io` will be producing

skbio/sequence/_sequence.py
```diff
@@ -106,20 +99,181 @@ class Sequence(collections.Sequence, SkbioObject):
...
+
+    Next let's create a sequence that has been derived from another sequence:
+
+    >>> subseq = seq[1:3]
+    >>> subseq
+    Sequence('CG', length=2, has_metadata=False, has_positional_metadata=True)
+    >>> subseq.positional_metadata
+      list  qual   gaps
+    0   []     3  False
+    1   []     4  False
```
ebolyen: I think there should be an example of setting a non-mutable piece of metadata to make it clear that only mutable objects have this behavior, simple things like ints and strings are fine.

skbio/sequence/_sequence.py
```diff
@@ -106,20 +99,181 @@ class Sequence(collections.Sequence, SkbioObject):
...
+         list  qual
+    0      []     3
+    1  [item]     3
+    2      []     4
+    3      []    10
+
     """
     default_write_format = 'fasta'
-    __hash__ = None  # TODO revisit hashability when all properties are present
+    __hash__ = None
```

skbio/sequence/_sequence.py
```diff
@@ -187,30 +351,107 @@ def metadata(self):
     def metadata(self, metadata):
         if not isinstance(metadata, dict):
             raise TypeError("metadata must be a dict")
-        self._metadata = metadata
+        # shallow copy
+        self._metadata = metadata.copy()
+
+    @metadata.deleter
+    def metadata(self):
+        if hasattr(self, '_metadata'):
```
ebolyen: Why is this check necessary? Can't you just reassign instead?

skbio/sequence/_sequence.py
```diff
@@ -187,30 +351,107 @@ def metadata(self):
...
+        >>> s.positional_metadata
+           exons  quality
+        0   True        3
+        1   True        3
+        2  False       20
+        3   True       11
+
+        Update positional metadata:
+
+        >>> s.positional_metadata['gaps'] = [False, False, False, False]
```
ebolyen: Would it be worth using `DNA` for this example so that you could use `s.positional_metadata['gaps'] = s.gaps()`?

skbio/sequence/_sequence.py
```diff
@@ -187,30 +351,107 @@ def metadata(self):
...
+                "number of characters in the sequence (%d)." %
+                (num_rows, len(self)))
+
+        positional_metadata.reset_index(drop=True, inplace=True)
+        self._positional_metadata = positional_metadata
+
+    @positional_metadata.deleter
+    def positional_metadata(self):
+        if hasattr(self, '_positional_metadata'):
+            del self._positional_metadata
```
ebolyen: Same question as above, is this needed?

skbio/sequence/_sequence.py
```diff
@@ -187,30 +351,107 @@ def metadata(self):
...
+        self.positional_metadata = pd.DataFrame(index=np.arange(len(self)))
```
ebolyen: Any chance this can be optimized to be `self._positional_metadata = None` and the getter can just build a dataframe on the fly? You shouldn't have to change any tests

skbio/sequence/_sequence.py
```diff
@@ -223,13 +464,19 @@ def __init__(self, sequence, metadata=None,
...
-            if metadata is None:
-                metadata = {}
 
         self._set_sequence(sequence)
-        self.metadata = metadata
-        self._set_positional_metadata(positional_metadata)
+
+        if metadata is None:
+            # reset to default empty metadata representation
+            del self.metadata
```
ebolyen: I know this is being optimistic, but you could also do `self._metadata = None` and modify the getter slightly

skbio/sequence/_sequence.py
```diff
@@ -223,13 +464,19 @@ def __init__(self, sequence, metadata=None,
...
+        else:
+            self.metadata = metadata
+
+        if positional_metadata is None:
+            del self.positional_metadata
```
ebolyen: Same here

ebolyen:
Looks good @jairideout, just a couple minor comments in the deleters and a request for a memory optimization when metadata isn't needed (should be simple and won't impact tests).

skbio/sequence/_sequence.py
jairideout: done

skbio/sequence/_sequence.py
jairideout: I have examples of this above where I add a new columns (`gaps`) and change one of the quality scores to 999.

skbio/sequence/_sequence.py
jairideout: `del` gets called from the constructor where `self._metadata` hasn't been assigned yet. I did it this way to have only a single representation of "missing" metadata (defined by the deleter). I have a comment about this in the constructor but I'll also add one here for clarity.

skbio/sequence/_sequence.py
ebolyen: Right, sorry, I apparently can't ready RST without my eyes glazing over.

skbio/sequence/_sequence.py
```diff
@@ -187,30 +351,107 @@ def metadata(self):
...
+    @metadata.deleter
+    def metadata(self):
+        if hasattr(self, '_metadata'):
+            del self._metadata
+        self.metadata = {}
 
     @property
     def positional_metadata(self):
-        """Pandas DataFrame containing metadata on a per-character basis.
+        """pandas.DataFrame containing metadata on a per-character basis.
```
anderspitman: [Here](https://github.com/biocore/scikit-bio/pull/954/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR13) you use pd.DataFrame instead of pandas.DataFrame. Should probably be consistent.

skbio/sequence/_sequence.py
jairideout: done

skbio/sequence/_sequence.py
jairideout: done

skbio/sequence/_sequence.py
ebolyen: I like `pd.DataFrame`

skbio/sequence/_sequence.py
jairideout: I know the feeling

skbio/sequence/_sequence.py
jairideout: good idea, updated to use `.gaps()` and `.degenerates()`

skbio/sequence/tests/test_sequence.py
```diff
@@ -82,8 +88,15 @@ def test_init_empty_sequence(self):
             self.assertEqual(seq.values.dtype, '|S1')
             self.assertEqual(seq.values.shape, (0, ))
             npt.assert_equal(seq.values, np.array('', dtype='c'))
-            self.assertEqual('', str(seq))
-            self.assertEqual(0, len(seq))
+            self.assertEqual(str(seq), '')
```
anderspitman: Is this order (observed, expected) our convention in general or did you just change it to be consistent with the neighboring code?

ebolyen: generally yes, on both accounts.

skbio/sequence/_sequence.py
jairideout: :+1: I'll go with `pd.DataFrame` since we're generally using `np.something` in the sequence docs

jairideout: done

skbio/sequence/tests/test_sequence.py
jairideout: yes

anderspitman:
Done with review. Looks really good overall.

skbio/sequence/_sequence.py
anderspitman: :+1:

skbio/sequence/_sequence.py
jairideout: With efficient handling of missing metadata this code was refactored. It's much clearer now, both here and in the constructor.

skbio/sequence/_sequence.py
jairideout: same comment as above

skbio/sequence/_sequence.py
jairideout: you got it

skbio/sequence/_sequence.py
jairideout: done

skbio/sequence/_sequence.py
jairideout: done

jairideout:
@ebolyen I added the memory optimization you requested for `metadata` and `positional_metadata` (#948). I updated all internal usage of these properties to also be memory efficient. Finally we have tests that should catch regression in the future. Affected methods were `__eq__`, `__getitem__`, `has_metadata`, `has_positional_metadata`, `__init__`, and `_to`. The getter/setter/deleter code is actually cleaner and easier to follow now!

@ebolyen @anderspitman this is ready for another review (there were substantial changes in my latest commit).

skbio/io/tests/test_fasta.py
```diff
@@ -411,8 +411,10 @@ def test_fasta_to_generator_valid_files(self):
                 obs = list(_fasta_to_generator(fasta_fp, **kwargs))
                 self.assertEqual(len(obs), len(exp))
                 for o, e in zip(obs, exp):
-                    self.assertTrue(o.equals(e,
-                                             ignore=['positional_metadata']))
+                    # ignore positional metadata by creating a copy of the
+                    # expected sequence without positional metadata
+                    e = e._to(positional_metadata=None)
```
ebolyen: Can you add a TODO saying to use `copy` and a setter in the future?

ebolyen: Actually I see this is used a lot, maybe it's not worth it to worry about it now.

skbio/sequence/_sequence.py
```diff
@@ -180,37 +334,128 @@ def metadata(self):
...
+                "Number of positional metadata values (%d) must match the "
+                "number of characters in the sequence (%d)." %
+                (num_rows, len(self)))
+
+        positional_metadata.reset_index(drop=True, inplace=True)
+        self._positional_metadata = positional_metadata
+
+    @positional_metadata.deleter
+    def positional_metadata(self):
+        del self._positional_metadata
```
ebolyen: Is this still needed? The GC should be all we need now.

skbio/sequence/_sequence.py
```diff
@@ -180,37 +334,128 @@ def metadata(self):
...
     def metadata(self, metadata):
         if not isinstance(metadata, dict):
             raise TypeError("metadata must be a dict")
-        self._metadata = metadata
+        # shallow copy
+        self._metadata = metadata.copy()
+
+    @metadata.deleter
+    def metadata(self):
+        del self._metadata
```
ebolyen: Is this still needed?

ebolyen:
@jairideout finished with review, just two questions, then I think this is good to go!

skbio/io/tests/test_fasta.py
jairideout: Added a TODO to this occurrence. When we have `copy` we can search for all usages of `_to` and replace with `copy` and a setter.

skbio/sequence/_sequence.py
jairideout: fixed

skbio/sequence/_sequence.py
jairideout: fixed

jairideout:
Thanks for reviewing @ebolyen! I addressed your comments.

ebolyen:
Thanks @jairideout!

PR #956, opened 2015-06-22
Lowercase

anderspitman:
Not quite finished (need a few more tests) but we should be able to start reviewing.


skbio/sequence/_iupac_sequence.py
```diff
@@ -164,23 +171,36 @@ def _motifs(self):
...
+        return self._bytes > self._ascii_lowercase_boundary
+
+    def lowercase(self, lowercase):
+        # First try to treat lowercase like an index array (boolean array,
+        # etc). If the coercion doesn't work, assume lowercase is a key into
+        # the current positional metadata.
+        index = self._munge_to_index_array(lowercase)
+        outbytes = self._bytes.copy()
+        outbytes[index] = _invert_case(outbytes[index])
+        return outbytes.tostring()
```
ebolyen: You will need to call `decode('ascii')` and wrap this in a `str()` call to make it pass py3 and py2

anderspitman: Makes sense

skbio/sequence/_iupac_sequence.py
```diff
@@ -19,6 +19,13 @@
 from ._sequence import Sequence
 
 
+def _invert_case(array):
```
ebolyen: I'm not 100% convinced this needs to be a function, but it should probably be named `_invert_ascii_case`

skbio/sequence/_iupac_sequence.py
```diff
@@ -19,6 +19,13 @@
...
+    # ASCII is built such that the difference between uppercase and
+    # lowercase is the 6th bit.
+    array ^= 32
```
ebolyen: It might be better to have a constant to represent the ascii case offset instead of the function. @jairideout thoughts?

anderspitman: Function gets my vote. The twiddling is opaque enough that I think it would require a comment to be duplicated everywhere it's used.

skbio/sequence/_iupac_sequence.py
anderspitman: Done

skbio/io/_base.py
```diff
@@ -195,7 +196,12 @@ def _format_fasta_like_records(generator, id_whitespace_replacement,
         qual = None
         if 'quality' in seq.positional_metadata:
             qual = seq.positional_metadata['quality'].values
-        yield header, str(seq), qual
+
+        if 'lowercase' in kwargs and hasattr(seq, 'lowercase'):
```
jairideout: I think this should error when the sequence can't be lowercased.

skbio/sequence/_iupac_sequence.py
```diff
@@ -164,23 +171,36 @@ def _motifs(self):
...
 
     def _convert_to_uppercase(self):
-        lowercase = self._bytes > self._ascii_lowercase_boundary
+        lowercase = self._lowercase_bytes()
         if np.any(lowercase):
             with self._byte_ownership():
-                # ASCII is built such that the difference between uppercase and
-                # lowercase is the 6th bit.
-                self._bytes[lowercase] ^= 32
+                self._bytes[lowercase] = _invert_case(self._bytes[lowercase])
```
jairideout: Since this mutates in-place it'd be clearer to have the function not return anything.

skbio/sequence/_iupac_sequence.py
jairideout: I'm fine with either a function or inline since this only occurs in two places. Inline might be clearer because it's obvious that it operates in place. It'll also be a little faster since there isn't the overhead of a function call (this could be called a lot of times, e.g., when processing a large FASTA file of lowercase sequences).

If we go inline, :+1: for using a constant to represent the case offset. If we go with a function, a name like `_convert_lower_to_upper` would be better since it isn't inverting the case.

skbio/io/_base.py
anderspitman: Oh yeah totally forgot that part

skbio/sequence/_iupac_sequence.py
anderspitman: Went with class constant

skbio/io/_base.py
anderspitman: Done

skbio/sequence/_iupac_sequence.py
```diff
@@ -164,23 +165,72 @@ def _motifs(self):
...
 
         if validate:
             self._validate()
 
     def _convert_to_uppercase(self):
-        lowercase = self._bytes > self._ascii_lowercase_boundary
+        lowercase = self._lowercase_mask
         if np.any(lowercase):
             with self._byte_ownership():
-                # ASCII is built such that the difference between uppercase and
```
jairideout: This comment was useful, can you add it back either here or on `_ascii_invert_case_bit_offset`?

skbio/sequence/_iupac_sequence.py
```diff
@@ -164,23 +165,72 @@ def _motifs(self):
...
-                 validate=True, case_insensitive=False):
+                 validate=True, lowercase=False):
         super(IUPACSequence, self).__init__(
             sequence, metadata, positional_metadata)
 
-        if case_insensitive:
+        # Note we are checking identity against the False object here and not
+        # just falsiness
+        if lowercase is not False:
+            self._lowercase_mask = self._bytes > self._ascii_lowercase_boundary
```
jairideout: Can you refactor this so that `_lowercase_mask` isn't being stored as an instance variable? We want users to be able to say `lowercase=True` and not have lowercase info tracked on a sequence object (e.g., for memory efficiency).

skbio/sequence/_iupac_sequence.py
anderspitman: Added it back in.

skbio/io/fasta.py
```diff
@@ -653,46 +653,49 @@ def _fasta_to_biological_sequence(fh, qual=FileSentinel, seq_num=1):
...
+        list(_fasta_to_generator(fh, qual=qual, constructor=constructor,
+                                 **kwargs)))
 
 
 @register_writer('fasta')
 def _generator_to_fasta(obj, fh, qual=FileSentinel,
                         id_whitespace_replacement='_',
-                        description_newline_replacement=' ', max_width=None):
+                        description_newline_replacement=' ', max_width=None,
+                        **kwargs):
```
ebolyen: I think this should explicitly be `length` that way it is documented. It doesn't look like kwargs actually get to go anywhere unlike in the `readers` where they get passed to the constructor (where they can fail). Using kwargs here will just result in swallowing any misspellings. `seq.write('filename', id_whitespace_replacment=' ')` won't raise an error and won't do what the user wants.

skbio/io/fastq.py
```diff
@@ -342,53 +342,58 @@ def _fastq_to_biological_sequence(fh, variant=None, phred_offset=None,
...
                                  phred_offset=phred_offset,
-                                 constructor=constructor)))
+                                 constructor=constructor, **kwargs)))
 
 
 @register_writer('fastq')
 def _generator_to_fastq(obj, fh, variant=None, phred_offset=None,
                         id_whitespace_replacement='_',
-                        description_newline_replacement=' '):
+                        description_newline_replacement=' ', **kwargs):
```
ebolyen: Same weakness as in `fasta`

skbio/io/fasta.py
jairideout: :+1:

skbio/io/tests/test_fasta.py
```diff
@@ -412,7 +428,7 @@ def test_fasta_to_generator_valid_files(self):
                 self.assertEqual(len(obs), len(exp))
                 for o, e in zip(obs, exp):
                     e = e.copy()
-                    del e.positional_metadata
+                    e.positional_metadata.drop('quality', 1, inplace=True)
```
ebolyen: Does `del e.positional_metadata['quality']` not work?

skbio/io/fasta.py
jairideout: Also we need to document this new parameter on every writer that supports it. I'm fine with that not happening in this PR, but please add a note to #949 so we can tackle before beta.

skbio/io/tests/test_fasta.py
```diff
@@ -435,12 +451,21 @@ def test_fasta_to_generator_invalid_files(self):
     def test_fasta_to_any_sequence(self):
         for constructor, reader_fn in ((Sequence,
                                         _fasta_to_biological_sequence),
-                                       (partial(DNA, validate=False),
-                                        _fasta_to_dna_sequence),
-                                       (partial(RNA, validate=False),
-                                        _fasta_to_rna_sequence),
-                                       (partial(Protein, validate=False),
-                                        _fasta_to_protein_sequence)):
+                                       (partial(DNA, validate=False,
```
ebolyen: This is likely not the case but now that `lowercase` exists, can you drop `validate=False` on some of these?

skbio/io/tests/test_fasta.py
```diff
@@ -560,7 +594,7 @@ def test_fasta_to_sequence_collection_and_alignment(self):
                     self.assertEqual(len(obs), len(exp))
                     for o, e in zip(obs, exp):
                         e = e.copy()
-                        del e.positional_metadata
+                        e.positional_metadata.drop('quality', 1, inplace=True)
```
ebolyen: Same question wrt using `del` on the column

skbio/io/tests/test_qseq.py
```diff
@@ -266,23 +265,29 @@ def test_invalid_files(self):
...
+        for constructor in [Sequence, DNA, RNA, Protein]:
             for valid, kwargs, components in self.valid_files:
-                for kwarg in kwargs:
-                    _drop_kwargs(kwarg, 'constructor', 'filter')
-
-                    seq_num = kwarg.get('seq_num', 1)
+                for observed_kwargs in kwargs:
+                    expected_kwargs = {}
+                    # Currently not validating the alphabet for fastq
+                    # files that are read in.
```
ebolyen: Mention that this is specific to the test, lest the reader extrapolate that statement to `fastq` generally.

skbio/sequence/_iupac_sequence.py
```diff
@@ -164,23 +167,71 @@ def _motifs(self):
...
             sequence, metadata, positional_metadata)
 
-        if case_insensitive:
-            self._convert_to_uppercase()
+        # Note we are checking identity against the False object here and not
+        # just falsiness
+        if lowercase is not False:
+            lowercase_mask = self._bytes > self._ascii_lowercase_boundary
+            self._convert_to_uppercase(lowercase_mask)
+            if not isinstance(lowercase, bool):
```
ebolyen: What happens when `lowercase=None`?

skbio/sequence/_iupac_sequence.py
```diff
@@ -164,23 +167,71 @@ def _motifs(self):
...
+
+        Constructor automatically populates a column in positional metadata
+        when the lowercase keyword argument is provided:
+
+        >>> s = DNA('ACgt', lowercase='introns')
+        >>> s.lowercase('introns')
+        'ACgt'
+        >>> s = DNA('ACGT', lowercase='introns')
+        >>> s.lowercase('introns')
+        'ACGT'
```
ebolyen: :+1:  I like this example!

ebolyen:
Finished with review @anderspitman, I think the logic in the constructor with `lowercase` has a hole in it. Also remove `**kwargs` from all of the writers. Other than that, just a couple little questions which don't matter that much.

skbio/io/fasta.py
anderspitman: Good point. Done.

skbio/io/fastq.py
anderspitman: Done.

skbio/io/tests/test_fasta.py
anderspitman: If there's lowercase data in `positional_metadata` (ie 'introns' column) we need it retained.

ebolyen: Right, we would be deleting only that `quality` column, assuming pandas implements this API, checking now.

anderspitman: Oh I see what you mean.

ebolyen: Yeah it works that way.

anderspitman: Tests still pass if I use `del`.

skbio/io/tests/test_fasta.py
anderspitman: Good catch, I was able to remove 1.

skbio/sequence/_iupac_sequence.py
anderspitman: I tried it and `None` is set as a valid column in `positional_metadata`. I believe @jairideout and I talked about a similar issue a while back having to do with situations like `lowercase=[]` and `lowercase=()`. We decided to go the route of letting pandas handle anything that's not a `bool`. Honestly I think we may want to constrain it to `bool` or `six.string_types`, but not sure.

skbio/io/_base.py
```diff
@@ -195,7 +196,17 @@ def _format_fasta_like_records(generator, id_whitespace_replacement,
         qual = None
         if 'quality' in seq.positional_metadata:
             qual = seq.positional_metadata['quality'].values
-        yield header, str(seq), qual
+
+        if lowercase is not None:
```
ebolyen: The semantics here should match the constructor, that means that either: `seq.lowercase` has to understand an inordinate amount of possible "keys" OR we restrict the constructor to be `str` or `bool`.  I would prefer the latter. @jairideout thoughts?

skbio/sequence/_iupac_sequence.py
ebolyen: I think it should be limited to `six.string_types` that way the same key can be used for the `lowercase` method. Without this `lowercase=` would behave differently between `io` and `IUPACSequence`

skbio/io/_base.py
jairideout: latter

ebolyen:
@anderspitman looks good, will merge when tests pass

PR #957, opened 2015-06-22
ENH: add Sequence.copy method

jairideout:
Adds `Sequence.copy(deep=False)` method for performing shallow and deep copies. Adds `Sequence.__copy__`/`__deepcopy__` for interoperability with `copy.copy`/`copy.deepcopy`.

Also updates codebase to use `Sequence.copy` instead of `Sequence._to`.

Addresses part of #949.

@ebolyen can you review?

### DIFF
--- a/skbio/sequence/_sequence.py
+++ b/skbio/sequence/_sequence.py
@@ -1594,47 +1594,15 @@ class Sequence(collections.Sequence, SkbioObject):
 
         Notes
         -----
-        This is a shallow copy, but since biological sequences are immutable,
-        it is conceptually the same as a deep copy.
+        By default, `metadata` and `positional_metadata` are shallow-copied and
+        the reference to `sequence` is used (without copying) for efficiency
+        since `sequence` is immutable. This differs from the behavior of
+        `Sequence.copy`, which will actually copy `sequence`.
 
         This method is the preferred way of creating new instances from an
         existing biological sequence, instead of calling
         ``self.__class__(...)``, as the latter can be error-prone (e.g.,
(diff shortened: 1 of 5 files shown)


ebolyen:
@jairideout Yup!

skbio/sequence/_sequence.py
```diff
@@ -987,6 +945,231 @@ def _format_str(self, s):
...
+
+    def _copy(self, deep, memo):
+        # strategy: copy the sequence without metadata first, then set metadata
+        # attributes with copies. we take this approach instead of simply
+        # passing the metadata through the Sequence constructor because we
+        # don't want to copy twice (this could happen when deep=True, where we
+        # deep copy here and then shallow copy in the Sequence constructor). we
+        # also directly set the private metadata attributes instead of using
+        # their public setters to avoid an unnecessary copy
+        if deep:
```
ebolyen: I don't think there is any difference between these branches for the type: `np.uint8`

jairideout: There shouldn't be any difference. I thought it'd be safer to use `deepcopy` in case the type changes in the future, but the performance hit probably isn't worth the defensive programming. I'll remove this branch and add a comment for why we're not using `deepcopy` here.

ebolyen: If the type changed to a non-integral type, so many things would need to change anyways...

ebolyen:
@jairideout Looks good! Ready for merge when the deep/shallow branching is removed for the numpy sequence.

ebolyen:
Thanks @jairideout will merge when tests pass

PR #963, opened 2015-06-25
Fix io

ebolyen:
~~Still need to fix/enhance the tests for `registry.py` and `util.py` but otherwise integration with the rest of skbio is complete. Also need lots of documentation.~~

This PR represents a really big change in what APIs are provided by the registry.
Formats are now objects! You can make one by using the `create_format` factory. This object possesses knowledge about the encoding of the format, and how newlines are handled. (binary formats will use `encoding='binary'`). These formats provide decorators as methods which makes the format code nicer to read.

Additionally, now the registry will provide an `io.BufferedReader/io.BufferedWriter` object for binary formats, and an `io.TextIOBase` implementation for text formats. It doesn't matter what the user gives the registry, the formats will always receive these objects. This also means that text formats will interact with `unicode` objects while binary formats will interact with `bytes`. This transition was a long time coming and needed to be dealt with. A big advantage is now skbio is now encoding agnostic rather than being implicitly ascii.

Compression is also handled by the registry automatically, but can be overridden by the user. This will be clearer when I add the rest of the documentation. @jrderuiter This should fix #909. 

Just for fun (and it helped me think through a lot of edge cases) you can also now do this:

`"".join(seq.write([]))` and you will get a fasta formatted string! :rabbit: :tophat:


ebolyen:
The motivation for most of these changes stemmed from the need to make `gzip` work in a Py2/Py3 compatible way while breaking off from our inability to deal with non-ascii encodings.

ebolyen:
Python 3.3 is likely to break, this isn't intentional, but there are probably some `u""` literals still hanging around.

anderspitman:
:+1:

ebolyen:
@jairideout ready for review

jairideout:
Great! I'll review first thing in the morning.

skbio/io/_iosources.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def get_reader(self):
+        return IterableStringReaderIO(self.repaired,
+                                      newline=self.options['newline'])
+
+    def get_writer(self):
+        return IterableStringWriterIO(self.file,
+                                      newline=self.options['newline'])
+
+    def _repair_iterable(self, head, tail):
```
jairideout: use `itertools.chain`

skbio/io/_iosources.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+class IOSource(object):
+    closeable = True
+
+    def __init__(self, file, options):
+        self.file = file
+        self.options = options
+
+    def can_read(self):
```
jairideout: `can_read` and `can_write` don't seem to be used anywhere, can they be removed?

jairideout: Sorry, missed that those are used during file resolution.

jairideout:
I'm seeing the following warning from the test suite, can you please fix?

```
skbio/io/registry.py:521: FormatIdentificationWarning: <skbio.io._fileobject.IterableStringReaderIO object at 0x2b34accb27d0> does not look like a clustal file
```

jairideout:
There are several of these types of warnings coming from the Python 3 tests, probably worth fixing here:

```
/home/travis/build/biocore/scikit-bio/skbio/io/tests/test_util.py:115: ResourceWarning: unclosed file <_io.BufferedReader name='/home/travis/build/biocore/scikit-bio/skbio/io/tests/data/example_file'>
  self.check_open_gc_behaviour(self.read_file)
```

skbio/io/format/clustal.py
```diff
@@ -215,7 +218,10 @@ def _clustal_sniffer(fh):
     #       * The sub-sequences have differing lengths
     #       * One of the sequence ids is not immediately
     #         followed by a subsequence
-    empty = True
+    # empty = True
```
jairideout: commented out code, can this be removed?

skbio/io/format/clustal.py
```diff
@@ -215,7 +218,10 @@ def _clustal_sniffer(fh):
...
+    if fh.read(7) != 'CLUSTAL':
```
jairideout: can this be a `peek` so you don't have to `seek(0)` below?

ebolyen: `peek` isn't necessarily defined on `io.TextIOBase` only on `io.BufferedReader`

skbio/io/format/newick.py
```diff
@@ -326,13 +328,14 @@ def _newick_to_tree_node(fh, convert_underscores=True):
 
         last_token = token
 
+    fh.seek(0)
```
jairideout: Can this be removed?

skbio/io/format/clustal.py
jairideout: makes sense, thanks

skbio/io/registry.py
```diff
@@ -0,0 +1,1118 @@
...
+        ----------
+        format_object : Format
+            The format to add to the registry.
+
+        """
+        # See comment in the constructor for an explanation for why this split
+        # occurs.
+        name = format_object.name
+        if name in self._binary_formats or name in self._text_formats:
+            raise DuplicateRegistrationError("A format with already exists"
```
jairideout: *A format with already* -> *A format already*

skbio/io/registry.py
```diff
@@ -0,0 +1,1118 @@
...
+            else:
+                # set to None to mask FileSentinel when user neglected argument
+                kwargs[param] = None
+
+        return file_keys, files
+
+    def _add_writer(self, cls, writer, monkey_patch, override):
+        if cls in self._writers and not override:
+            raise DuplicateRegistrationError("There is already a writer"
+                                             "registered to %s in format: %s"
```
jairideout: add space to error message here

skbio/io/registry.py
```diff
@@ -0,0 +1,1118 @@
...
+                                             % (cls, self._name))
+        self._writers[cls] = writer
+        if monkey_patch and cls is not None:
+            self._monkey_patch['write'].add(cls)
+
+    def _add_reader(self, cls, reader, monkey_patch, override):
+        if cls in self._readers and not override:
+            raise DuplicateRegistrationError("There is already a reader"
+                                             "registered to %s in format: %s"
```
jairideout: add space to error message here

jairideout:
@ebolyen this is great! I'm done reviewing.

jairideout:
Thanks @ebolyen!!!

PR #965, opened 2015-06-26
Ensure _munge_to_bytestring always returns bytes

anderspitman:


skbio/sequence/tests/test_sequence.py
```diff
@@ -2218,6 +2218,16 @@ def test_munge_to_index_array_invalid_string(self):
...
 
+    def test_munge_to_bytestring_return_bytes(self):
+        seq = Sequence('')
+        m = 'dummy_method'
+        possible_inputs = ('', 'a', 'acgt', u'', u'a', u'acgt', b'', b'a',
+                           b'acgt', Sequence(''), Sequence('a'),
+                           Sequence('acgt'))
+
+        for s in possible_inputs:
+            self.assertIs(type(seq._munge_to_bytestring(s, m)), bytes)
```
jairideout: Can you compare the observed value to the expected bytes value?

jairideout: Another test case with input that can't be decoded to ascii would be useful.

skbio/sequence/tests/test_sequence.py
```diff
@@ -2218,6 +2218,31 @@ def test_munge_to_index_array_invalid_string(self):
...
+
+    def test_munge_to_bytestring_unicode_out_of_ascii_range(self):
+        seq = Sequence('')
+        all_inputs = (u'\x80', u'abc\x80', u'\x80abc')
+        for input_ in all_inputs:
+            with self.assertRaisesRegexp(UnicodeEncodeError,
+                                         "'ascii' codec can't encode character"
+                                         ".*in position 0: ordinal not in"
+                                         " range\(128\)"):
+                seq._munge_to_bytestring(u'\x80', 'dummy_method')
```
jairideout: This should be `input_`

jairideout:
@anderspitman ready for merge once the test is fixed to use the correct input

PR #966, opened 2015-06-26
ENH: improved Sequence repr

jairideout:
Based on repr designed by @ebolyen, @anderspitman, @jairideout, and @gregcaporaso.

Addresses part of #949.


jairideout:
merge conflicts

jairideout:
note to self to verify number of tests against previous master because I added an option to the nosetests command used by `make test`

jairideout:
Verified tests and coverage. @ebolyen and/or @anderspitman can you please review?

anderspitman:
I can take a pass through now.

jairideout:
Thanks! @ebolyen and I just took a pass through and made some minor refactorings for readability.

skbio/io/tests/test_fasta.py
```diff
@@ -777,7 +777,6 @@ def blank_seq_gen():
     def test_generator_to_fasta_no_qual(self):
         # test writing standalone fasta (i.e., without a qual file)
         for obj, kwargs, fp, _ in self.objs_fps:
-            print(fp)
```
anderspitman: Oops. Probably mine, sorry

skbio/sequence/_sequence.py
```diff
@@ -887,7 +933,22 @@ def __str__(self):
         return str(self._string.decode("ascii"))
 
     def __repr__(self):
-        """Return a string representation of the biological sequence object.
+        r"""Return a string representation of the biological sequence object.
```
anderspitman: Seems like this docstring would be valuable information for users, but now that we're not putting `__` methods in our docs they'll never see it, right?

jairideout: Right, they'll never see it in the built HTML docs. They'll see it by calling `help` or `?` in an interactive python session or by reading the code though :P

skbio/sequence/_sequence.py
```diff
@@ -1843,3 +1945,272 @@ def _byte_ownership(self):
...
+        line_len = len(line)
+        if line_len > self._max_line_len:
+            self._max_line_len = line_len
+        self._lines.append(line)
+
+    def add_lines(self, lines):
+        for line in lines:
+            self.add_line(line)
+
+    def add_separator(self):
```
anderspitman: possibly rename to `add_horizontal_separator`

ebolyen: That would imply symmetry: where is the `add_vertical_seperator`? I think the name is fine.

skbio/sequence/_sequence.py
```diff
@@ -1843,3 +1945,272 @@ def _byte_ownership(self):
...
+        self._lines.append(None)
+        self._separator_idxs.append(len(self._lines) - 1)
+
+    def to_str(self):
+        separator = '-' * self._max_line_len
```
anderspitman: Seems like a strange place for the separator logic to live. I'm assuming there's a good reason you couldn't simply append separators as you go in `add_separator`?

skbio/sequence/_sequence.py
anderspitman: If a dev concludes symmetry and tries to use `add_veritcal_separator`, it will immediately fail. Meanwhile, `add_horizontal_separator` is perfectly clear from the name. In it's current state I had to look at the source to be sure it was a horizontal separator. It's a minor point though; I'm fine with the current name.

skbio/sequence/_sequence.py
jairideout: The separators need to expand to fit the maximum width (non-separator) line.

anderspitman:
@jairideout Took a quick pass through. Looks really good! Super excited for these reprs. Only 1 possible change, but I think I agree with @ebolyen on that.

skbio/sequence/_sequence.py
anderspitman: Makes sense.

skbio/sequence/_sequence.py
jairideout: I'm going to leave as-is since it's private functionality and pretty minor (real reason: Travis is almost done with its tests!!)

jairideout:
Thanks @anderspitman!

skbio/sequence/_sequence.py
anderspitman: :+1:

PR #968, opened 2015-06-27
Nucleotide mixin

anderspitman:
Not sure about ABC semantics but I think this is basically what we want.


skbio/sequence/_nucleotide_mixin.py
```diff
@@ -14,24 +14,16 @@
 import numpy as np
 
 from skbio.util import classproperty
-from ._iupac_sequence import IUPACSequence, _motifs as parent_motifs
+from ._iupac_sequence import _motifs as parent_motifs
 
 
-class NucleotideSequence(with_metaclass(ABCMeta, IUPACSequence)):
-    """Abstract base class for storing an IUPAC nucleotide sequence.
+class NucleotideMixin(with_metaclass(ABCMeta, object)):
```
anderspitman: Is using `with_metaclass` with `ABCMeta` and `object` the right way to make this abstract?

ebolyen: Yes, but I don't think it needs to be abstract, it's a mixin, I've never heard of anyone trying to instantiate a mixin and expecting anything to happen.

skbio/sequence/__init__.py
```diff
@@ -24,7 +24,7 @@
 
    Sequence
    IUPACSequence
-   NucleotideSequence
+   NucleotideMixin
```
jairideout: don't export

anderspitman: Is there a reason to export IUPACSequence and not NucleotideMixin?

jairideout: Good question -- we probably don't need to export IUPACSequence either. I was thinking it could be useful if devs wanted to subclass it, but it's probably less confusing to end users if we just hide this intermediate ABC. @ebolyen thoughts?

skbio/sequence/__init__.py
```diff
@@ -301,14 +301,14 @@
                          InvalidCodonError)
 from ._sequence import Sequence
 from ._iupac_sequence import IUPACSequence
-from ._nucleotide_sequence import NucleotideSequence
+from ._nucleotide_sequence import NucleotideMixin
```
jairideout: I think this import is breaking the tests currently

anderspitman: Huh not sure how I missed that one.

skbio/sequence/__init__.py
ebolyen: I'm ok with hiding it. Could we also just drop it from our docs? Then it is there, but our support only goes as far as what you are willing to read from the source code docs

jairideout: :+1:

skbio/sequence/_nucleotide_mixin.py
anderspitman: Went ahead and left it abstract

PR #971, opened 2015-06-29
CI: checklist.py verifies __future__ imports in .py files

jairideout:
@ebolyen and I pair-programmed this.

### DIFF
--- a/checklist.py
+++ b/checklist.py
@@ -458,7 +458,7 @@ class FluxCapacitorValidator(RepoValidator):
 
     def _validate(self, root, dirs, files):
         failures = []
-        expected = set(("absolute_import", "division", "print_function"))
+        expected = {"absolute_import", "division", "print_function"}
         for file in files:
             if file.endswith(".py"):
                 filename = os.path.join(root, file)
@@ -470,7 +470,8 @@ class FluxCapacitorValidator(RepoValidator):
                             continue
                         if isinstance(node, ast.ImportFrom):
                             if node.module == "__future__":
-                                if expected <= set(n.name for n in node.names):
+                                if expected.issubset(
+                                        {n.name for n in node.names}):
                                     failures.pop()
                             break
         return failures
--- a/checklist.py
+++ b/checklist.py
@@ -39,7 +39,7 @@ def main():
     root = 'skbio'
     validators = [InitValidator(), CopyrightHeadersValidator(),
                   ExecPermissionValidator(), GeneratedCythonValidator(),
-                  APIRegressionValidator()]
+                  APIRegressionValidator(), FluxCapacitorValidator()]
 
     return_code = 0
     for validator in validators:
@@ -93,7 +93,7 @@ class RepoValidator(object):
         msg = []
         if invalids:
             success = False
-            msg.append(self.reason + ':')
+            msg.append(self.reason)
 
             for invalid in invalids:
                 msg.append("    %s" % invalid)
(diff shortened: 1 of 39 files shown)


checklist.py
```diff
@@ -39,7 +39,7 @@ def main():
     root = 'skbio'
     validators = [InitValidator(), CopyrightHeadersValidator(),
                   ExecPermissionValidator(), GeneratedCythonValidator(),
-                  APIRegressionValidator()]
+                  APIRegressionValidator(), FluxCapacitorValidator()]
```
wasade: :watch::car::fire:

checklist.py
```diff
@@ -450,5 +450,31 @@ def _parse_file(self, fp, root):
...
 
+class FluxCapacitorValidator(RepoValidator):
+    """Ensure that the __future__ statements are fluxing correctly"""
+    reason = ("These files do not have the following import at the start:\n\n"
+              "from __future__ import absolute_import, division,"
+              " print_function\n")
+
+    def _validate(self, root, dirs, files):
+        failures = []
+        expected = set(("absolute_import", "division", "print_function"))
```
wasade: Could use the set notation `{"absolute_import", "division", "print_function"}`

checklist.py
```diff
@@ -450,5 +450,31 @@ def _parse_file(self, fp, root):
...
+        for file in files:
+            if file.endswith(".py"):
```
wasade: I believe this should also allow for `"*.pyx"`

checklist.py
```diff
@@ -450,5 +450,31 @@ def _parse_file(self, fp, root):
...
+                filename = os.path.join(root, file)
+                failures.append(filename)
+                with open(filename) as f:
+                    source = ast.parse(f.read())
+                    for node, _ in zip(ast.iter_child_nodes(source), range(2)):
+                        if isinstance(node, ast.Expr):
+                            continue
+                        if isinstance(node, ast.ImportFrom):
+                            if node.module == "__future__":
+                                if expected <= set(n.name for n in node.names):
```
wasade: Could also use set comprehension here: `if expected <= {n.name for n in node.names}`

The `<=` is a little confusing here though, what set operation is that? Shouldn't this be set difference, e.g.:

```python
16:21:43 (mcdonadt@Daniels-MacBook-Pro-3):/Users/mcdonadt/ResearchWork/software/scikit-bio
> {1,2,3} - {2,3,4,5,6}
Out[7]: {1}
```

It may be more explicit to use the method as well instead of the overloaded operator

checklist.py
jairideout: :scream: :goat:

checklist.py
jairideout: fixed, thanks!

checklist.py
jairideout: `ast` barfs on Cython code, no can do :(

I guess I could write a custom parser for Cython code... but I won't be able to add that to this PR. I can create an issue if you'd like.

wasade: Up to you, non-blocking, just depends on how strict you'd like this check
tobe

checklist.py
jairideout: `<=` is the same as `s.issubset(t)` (docs [here](https://docs.python.org/2/library/sets.html#set-objects)). I think it's the right check because we want to make sure all the expected future imports are present (there can be additional imports too). I'll update the code to use the more explicit method instead of the overloaded operator.

wasade: Thanks!

checklist.py
jairideout: Created an issue: #973

jairideout:
@wasade I addressed your comment, thanks for reviewing! This is ready for merge.

wasade:
:+1:

PR #972, opened 2015-06-30
Gc frequency content

anderspitman:
First pass. Applies to #949.

### DIFF
--- a/skbio/sequence/_nucleotide_mixin.py
+++ b/skbio/sequence/_nucleotide_mixin.py
@@ -33,6 +33,7 @@ class NucleotideMixin(with_metaclass(ABCMeta, object)):
 
     """
     __complement_lookup = None
+    __gc_codes = None
 
     @classproperty
     def _complement_lookup(cls):
@@ -44,6 +45,13 @@ class NucleotideMixin(with_metaclass(ABCMeta, object)):
             lookup[ord(key)] = ord(value)
         cls.__complement_lookup = lookup
         return lookup
+
+    @classproperty
+    def _gc_codes(cls):
+        if cls.__gc_codes is None:
+            gc_iupac_chars = 'GCS'
+            cls.__gc_codes = np.asarray([ord(g) for g in gc_iupac_chars])
+        return cls.__gc_codes
 
     @property
     def _motifs(self):
@@ -234,7 +242,9 @@ class NucleotideMixin(with_metaclass(ABCMeta, object)):
         """Calculate the relative frequency of G's and C's in the sequence.
 
         This includes G, C, and S characters. This is equivalent to calling
-        ``gc_frequency(relative=True)``.
+        ``gc_frequency(relative=True)``. Note that the sequence will be
+        degapped before the operation, so gap characters will not be included
+        when calculating the length of the sequence.
 
         Returns
         -------
@@ -258,6 +268,16 @@ class NucleotideMixin(with_metaclass(ABCMeta, object)):
         0.5
         >>> DNA('--..').gc_content()
         0
+
+        `S` means `G` or `C`, so it counts:
+
+        >>> DNA('ASST').gc_content()
+        0.5
+
+        Other degenerates don't count:
+
+        >>> DNA('RYKMBDHVN').gc_content()
+        0.0
 
         """
         return self.gc_frequency(relative=True)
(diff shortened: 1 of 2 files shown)


skbio/sequence/_nucleotide_mixin.py
```diff
@@ -230,6 +230,69 @@ def is_reverse_complement(self, other):
             # underlying sequence data
             return self.reverse_complement()._string == other._string
 
+    def gc_content(self):
+        """Calculate  the ratio of G's and C's in the sequence
```
jairideout: extra whitespace

skbio/sequence/_nucleotide_mixin.py
```diff
@@ -230,6 +230,69 @@ def is_reverse_complement(self, other):
...
+
+    def gc_frequency(self, relative=False):
+        """Calculate frequency of G's and C's in the sequence.
+
+        This calculates the minimum GC frequency, which corresponds to IUPAC
+        characters G, C, and S (which stands for G or C).
+
+        Parameters
+        ----------
+        relative : bool
```
jairideout: bool, optional

skbio/sequence/_nucleotide_mixin.py
```diff
@@ -230,6 +230,69 @@ def is_reverse_complement(self, other):
...
+            If False return GC count. If True return GC content, which is the
+            ratio of G's and C's in the sequence.
+
+        Returns
+        -------
+        int or float
+            Either count of GC occurrances or ratio, depending on input.
```
jairideout: occurrences

skbio/sequence/_nucleotide_mixin.py
```diff
@@ -230,6 +230,87 @@ def is_reverse_complement(self, other):
             # underlying sequence data
             return self.reverse_complement()._string == other._string
 
+    def gc_content(self):
+        """Calculate the relative frequency of G's and C's in the sequence.
+
+        This includes G, C, and S characters. This is equivalent to calling
+        ``gc_frequency(relative=True)``.
```
jairideout: Can you add a note here about the gap character behavior? In case the user doesn't read `gc_frequency` docs.

skbio/sequence/_nucleotide_mixin.py
```diff
@@ -230,6 +230,87 @@ def is_reverse_complement(self, other):
...
+        Returns
+        -------
+        float
+            Relative frequency of G's and C's in the sequence.
+
+        See Also
+        --------
+        gc_frequency
+
+        Examples
```
jairideout: Can you add an example with degenerates, including S (which should count towards GC content) and other degenerates (which shouldn't count towards GC content)?

jairideout: and add those examples to `gc_frequency` too

skbio/sequence/_nucleotide_mixin.py
```diff
@@ -230,6 +230,87 @@ def is_reverse_complement(self, other):
...
+        >>> DNA('ACGT--..').gc_frequency(relative=True)
+        0.5
+        >>> DNA('--..').gc_frequency(relative=True)
+        0
+
+        """
+
+        counts = np.bincount(self._bytes,
+                             minlength=self._number_of_extended_ascii_codes)
+        gc_ord = (ord(y) for y in 'CGS')
```
jairideout: Can you cache this value as a `classproperty`? This would be similar to how `IUPACSequence` is handling gap chars, degenerates, etc. (example [here](https://github.com/biocore/scikit-bio/blob/master/skbio/sequence/_iupac_sequence.py#L92-L97)).

skbio/sequence/_nucleotide_mixin.py
```diff
@@ -230,6 +230,87 @@ def is_reverse_complement(self, other):
...
+        >>> DNA('ACGT').gc_frequency(relative=True)
+        0.5
+        >>> DNA('ACGT--..').gc_frequency(relative=True)
+        0.5
+        >>> DNA('--..').gc_frequency(relative=True)
+        0
+
+        """
+
+        counts = np.bincount(self._bytes,
```
jairideout: Can you update your `bincount` strategy to match how we perform validation in `IUPACSequence` (example [here](https://github.com/biocore/scikit-bio/blob/master/skbio/sequence/_iupac_sequence.py#L203-L206))? Basically, cache a GC mask as a `classproperty` and multiply it against the sequence `bincount` here, then sum (it should be faster).

skbio/sequence/_nucleotide_mixin.py
anderspitman: Done

skbio/sequence/_nucleotide_mixin.py
anderspitman: Done

skbio/sequence/_nucleotide_mixin.py
ebolyen: It would actually be fast to run `count` 3 times and add. Once you hit 4 `count` calls it evens out with bincount

skbio/sequence/_nucleotide_mixin.py
anderspitman: Done

skbio/sequence/_nucleotide_mixin.py
anderspitman: @jairideout and I worked out an alternative implementation that should be just as fast.

PR #976, opened 2015-07-02
Optimize sequence init

anderspitman:
Early draft, not ready for review.

### DIFF
--- a/skbio/sequence/_sequence.py
+++ b/skbio/sequence/_sequence.py
@@ -492,12 +492,13 @@ class Sequence(collections.Sequence, SkbioObject):
     def __init__(self, sequence, metadata=None,
                  positional_metadata=None):
 
-        if type(sequence) == np.ndarray or isinstance(sequence, np.ndarray): 
+        if type(sequence) == np.ndarray or isinstance(sequence, np.ndarray):
             if sequence.dtype == np.uint8:
                 self._set_bytes_contiguous(sequence)
             elif sequence.dtype == '|S1':
                 sequence = sequence.view(np.uint8)
-                # TODO: This is effectively what was happening before, but is it necessary?
+                # Guarantee the sequence is an array (might be scalar before
+                # this).
                 if sequence.shape == ():
                     sequence = np.array([sequence], dtype=np.uint8)
                 self._set_bytes_contiguous(sequence)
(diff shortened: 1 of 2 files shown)


anderspitman:
@jairideout @ebolyen ready for review. Went with an uglier implementation that is faster and less likely to break things.

skbio/sequence/_sequence.py
```diff
@@ -491,7 +491,23 @@ def _string(self):
...
-        if isinstance(sequence, Sequence):
+
+        if type(sequence) == np.ndarray or isinstance(sequence, np.ndarray):
+            if sequence.dtype == np.uint8:
+                self._set_bytes_contiguous(sequence)
+            elif sequence.dtype == '|S1':
+                sequence = sequence.view(np.uint8)
+                # Guarantee the sequence is an array (might be scalar before
+                # this).
+                if sequence.shape == ():
```
ebolyen: Why does this branch occur in '|S1' and not in np.uint8?

skbio/sequence/_sequence.py
```diff
@@ -491,7 +491,23 @@ def _string(self):
 
     def __init__(self, sequence, metadata=None,
                  positional_metadata=None):
-        if isinstance(sequence, Sequence):
+
+        if type(sequence) == np.ndarray or isinstance(sequence, np.ndarray):
```
jairideout: This check seems redundant. Did it speed things up when you benchmarked? I'd be surprised because `isinstance` is slow in the negative case, but you're optimizing for the positive case, which should be relatively quick via `isinstance` already.

skbio/sequence/_sequence.py
```diff
@@ -503,43 +519,8 @@ def __init__(self, sequence, metadata=None,
                 positional_metadata = sequence.positional_metadata
             sequence = sequence._bytes
```
jairideout: I think this case needs `self._owns_bytes = False`

jairideout:
@anderspitman this looks really great, just a couple of comments.

skbio/sequence/_sequence.py
anderspitman: It's definitely redundant. I had thought I benchmarked it as faster but I just double checked and the difference seems negligible. Removed.

skbio/sequence/_sequence.py
anderspitman: I was gonna ask about this. Went ahead and added it but would like to discuss further at some point.

skbio/sequence/_sequence.py
anderspitman: Excellent question. I looked in to it, and interestingly enough `a = np.array([65], dtype=np.uint8)
` has shape `(1, )` but `a = np.array('A', dtype='c')` has shape `()`, so it's not necessary to check in the `np.uint8` case.

anderspitman:
Ready for another review round

skbio/sequence/_sequence.py
jairideout: Thanks! I think the previous factoring ended up setting `_owns_bytes` from within `_set_sequence` but now it's up to each branch in the constructor to determine it (unless passing through `_set_bytes_contiguous`). This flag lets us know whether a copy of the underlying sequence bytes needs to be made when using the `_byte_ownership` context manager to perform in-place modifications.

jairideout:
Thanks @anderspitman!

PR #989, opened 2015-07-08
PERF: use np.uint8 for QUAL scores

jairideout:


jairideout:
@ebolyen ready for review/merge

PR #999, opened 2015-07-09
Use requirements files for pip and conda

anderspitman:
Let's see if it works...


jairideout:
Should this be closed in favor of #1000?

anderspitman:
Technically they aren't related. I think this should be merged first.

anderspitman:
But I'm fine with combining them if you prefer, since they both touch travis

Issue #906, opened 2015-04-27
travis-ci improvements


jairideout:
The travis build could be improved by:

- [x] ~~building docs in a separate job~~
- [x] ~~testing coverage (without doctests) in a separate job to prevent doctests from inflating coverage~~

anderspitman:
Probably also useful:
- [ ] Many of the pip packages are also available in conda now.

- [ ] We should use a `requirements.txt` [file for conda](https://groups.google.com/a/continuum.io/forum/#!topic/conda/PiM9sjWyXFU)

anderspitman:
I've also noticed that travis seems to start 3 of our builds at the same time, but wait a while for the 4th. Might want to try and collapse it down to 3 builds.

PR #1000, opened 2015-07-09
Migrate to travis containers

anderspitman:


anderspitman:
Addresses part of #906

jairideout:
Do you think it makes sense to put the requirements files in a top-level `ci` directory? Otherwise users might think that these files are necessary to install scikit-bio.

PR #1011, opened 2015-07-12
Issue 1006 (Phylip parsing)

patena:
Implemented basic phylip sniffing/parsing - I don't know if you want to merge this version or wait for me to do more.

What this does:
 - strict sequential format only (I can look at ETE code to implement interleaved/relaxed)
 - reads data into Sequence objects, not DNA/RNA/Protein; no alphabet checking

I added tests with a bunch of new valid and invalid files.


skbio/io/format/phylip.py
```diff
@@ -208,12 +208,43 @@
...
+    line = next(_line_generator(fh))
+    try:
+        _validate_line(line, seq_len)
+    except PhylipFormatError:
+        return False, {}
+    return True, {}
+
+
+@phylip.reader(Alignment)
+def _phylip_to_alignment(fh):
```
ebolyen: Can you add a constructor argument which can be used?

`_phylip_to_alignment(fh, constructor=Sequence)`

skbio/io/format/phylip.py
```diff
@@ -208,12 +208,43 @@
...
+@phylip.sniffer()
+def _phylip_sniffer(fh):
+    # Strategy:
+    #   Read the header and a single sequence; verify that the sequence length
+    #   matches the header information.  Do not verify that the total number of
+    #   lines matches the header information, since that would require reading
+    #   the whole file.
+    try:
+        header = next(_line_generator(fh))
+    except StopIteration:
```
ebolyen: You can probably compress all of these try-excepts to a single one:

```
try:
    ...
except (StopIteration, PhylipFormatError):
    return False, {}
return True, {}
```

patena:
@ebolyen Thanks, done.

patena:
If I wanted to implement the interleaved format and the relaxed option, how should the sniffer and reader deal with that?  Make a separate format for each?  Just sniff for Phylip in general (pretty easy)? Make the reader take arguments specifying the type, or try to figure it out (or have both as an option)?

A few other format issues:
* the code from ETE implements skipping blank lines and comments (starting with #) - do we want that, or should we stick with the strict version?
* theoretically the sequential format can have multiple lines per sequence - currently not implemented. Do we want that? (Could be done by checking after each line whether the expected seq length has been reached... This format is really terrible.)
* According to the [original specification](http://evolution.genetics.washington.edu/phylip/doc/sequence.html) "a portion of a sequence like this: ``300 AAGCGTGAAC GTTGTACTAA TRCAG`` is perfectly legal" because "digits and blanks will be ignored, the sequence being taken as starting at the first base symbol". Do we want to implement ignoring all digits just like we ignore spaces?

Or we could just stick with the current strict sequential one-line format and see if anyone actually wants the other ones...

ebolyen:
Sorry for the delay on getting back to this @patena, doing a detailed review now!

ebolyen:
Looks great, thanks @patena!

anderspitman:
:+1:

patena:
@ebolyen Yay, thanks! I'll try to keep an eye on scikit-bio issues for other things I could do.

gregcaporaso:
@patena, thanks so much for the contribution! It was great meeting you at SciPy!

Issue #820, opened 2015-02-12
add Faith's Phylogenetic Diversity (PD) metric


jairideout:
Add implementation of Faith's Phylogenetic Diversity (PD) metric to `skbio.diversity.alpha`.

gregcaporaso:
This is implemented (see the studying biological diversity chapter). I give permission to port that functionality to scikit-bio.

jairideout:
Fixed in #1027.

Issue #46, opened 2014-01-15
Port FastUnifrac code to skbio


josenavas:
It can be assigned to me.

Is there any special improvement that we want to explore?

gregcaporaso:
I'm going to port my IAB implementations of PD and UniFrac to skbio next week. These will get an experimental interface in place, and support a benchmark of different implementations.

wasade:
:+1:

I can port fast unifrac sans changes except for removing the cogent dependency. There's some added stuff with the cogent implementation, like doing PCoA, which I wont' bring over -- the goal is minimal, adapt to the pending skbio interface, and remove cogent dep.

gregcaporaso:
I think it makes sense for only one of us to do this, since we don't want multiple implementations in skbio. It doesn't really matter what implementation is ported over since this would just be to get an experimental interface in place, as well as extensive tests that can be easily applied to different implementations. Mine would be really simple to port over so I don't mind taking this on (there are no cogent dependencies, etc, it's just moving code from IAB), but let me know if you'd rather do it. Writing the tests will be the slower part for this, but maybe some of that can come from cogent.

wasade:
Right,  I figured you'd have the interface in place quickly? Or do you only
want to do a slow educational one and delay a production one? Note, the
runtime between the classic unifrac algorithm and fast unifrac are
substantial

gregcaporaso:
Yep, I'll have the ([experimental](http://scikit-bio.org/docs/latest/user/api_stability.html#experimental)) interface and unit tests in next week. Once those are in, we can swap in whatever under the hood based on whichever implementation performs best. I was just going to take my IAB implementation since it already works with skbio so it'll be really quick to get in and we can use that as one of the functions in the benchmark. If it turns out to be slower but intuitively much easier to grasp than whatever implementation we land on, we could have this implementation raise an ``EfficencyWarning`` (i.e., be an "educational" implementation, as with the python aligners) .

wasade:
Yup, that was the plan

gregcaporaso:
Awesome, more soon...

jairideout:
Fixed in #1027.

wasade:
The title of this issue implies it should still be open re: #1040?

jairideout:
Good catch!

jairideout:
Fixed in #1040.

PR #1027, opened 2015-08-05
adds phylogenetic diversity metrics

gregcaporaso:
Adds PD, unweighted unifrac, and weighted unifrac (with and without normalization). 

To do before merge:
- [x] interface work (how will this be imported/used)
 - [x] add stability decorators
 - ~~should ``normalized=True`` be the default for ``weighted_unifrac``?~~ let's not for now, for consistency with QIIME
 - [x] modify ``pw_distances`` to take ``metric`` as the first argument and not have a default 
 - [x] update ``pw_distances`` to take skbio's distance functions as strings and this should be the preferred/documented way of passing them (similar to SciPy's ``pdist``)
 - [x] for PD, ``u`` becomes ``counts`` to match our other alpha diversity metrics
 - [x] for *unifrac, ``u`` and ``v`` become ``u_counts`` and ``v_counts``, to match the alpha diversity metrics and SciPy
 - [x] ``phylogenetic_diversity`` renamed to ``faith_pd`` and move to ``_base.py``; make it accessible from ``skbio.diversity.alpha``
 - [x] ``*unifrac`` should be accessible from ``skbio.diversity.beta``, and ``skbio/diversity/beta/_stats.py`` should be renamed ``skbio/diversity/beta/_unifrac.py``
 - [x] allow **kwargs (some optimized methods may need to pass in other data, such as a cache) 

- [x] testing
 - [x] expand unit tests to include direct computation of distance metrics with ``pw_distances``, including referencing the metric both by passing the function and the name (as a string)
 - [x] tree with no branch lengths
 - [x] tree with duplicated tip names
 - [x] tree with one node
 - [x] non-trivial case where tips in tree are superset of ``otu_ids``
 - [x] add test with input vector of ``length == 0``
 - [x] add test of vector with negative counts (see ``skbio.diversity.alpha._validate``)
 - [x] add direct tests of the new validate methods
 - [x] raises error on unrooted tree
 - [x] add boundary test cases for ``faith_pd``
 - [x] add another real-world test of unifracs

- [ ] documentation
 - [x] add numpy-doc (with [cite](http://aem.asm.org/content/73/5/1576.long))
 - [ ] add formulas for metrics (including normalization of weighted unifrac)
 - [x] update ``skbio.diversity.beta`` doctests to use a phylogenetic metric
 - [x] update ``skbio.diversity.alpha`` doctests to use ``faith_pd``
 - [x] describe branch length normalization (more detail will be added when I add the formulas)
 - [ ] *OTU* versus *observation* in docs?

- [x] misc
 - [x] inline ``_to_pdist_metric``
 - [x] add tree/otu ids ``_validate`` function
 - [x] add ``_validate`` for beta diversity metrics, similar (or the same as ``skbio.diversity.alpha._validate``)

- ~~squash commits~~ (we're not going to do this for this PR, we haven't done it much in the past, and this is a complex PR since @wasade is pulling the commits in, so we don't want to screw it up)
- [x] add ChangeLog note

- [ ] future
 - [ ] port original web docs
 - [ ] explore alternative weighting schemes
 - [ ] generalization to DAGs
 - [ ] sparse vectors as input

Addresses #46, #820.


wasade:
@gregcaporaso, would you like a review now, or the pending items get knocked out?

gregcaporaso:
Probably not worth it yet.

wasade:
kk

gregcaporaso:
This is now ready for a first review. Before reviewing, I recommend reading the list above - there is still a bunch of outstanding items, so it's probably not worth reviewing things related to those items (mostly tests that still need to be added) yet. Also, the purpose of this PR is primarily to define the API and extensive tests of ``faith_pd``, ``weighted unifrac``, and ``unweighted unifrac`` so we can then write and test other implementations that we'll benchmark against. So, it's not worth a lot of profiling or optimization of these functions now. We'll get to that after this initial PR is merged.

skbio/diversity/alpha/_base.py
```diff
@@ -295,6 +295,52 @@ def esty_ci(counts):
...
+@experimental(as_of="0.4.0-dev")
+def faith_pd(counts, otu_ids, tree):
+    """ Compute Faith's phylogenetic diversity metric (PD)
+
+    Parameters
+    ----------
+    counts : 1-D array_like, int
+        Vector of counts.
+    otu_ids: list, np.array
+        Vector of OTU ids corresponding to tip names in ``tree``. Must be the
```
wasade: Would it be more appropriate to use "observation" instead of OTU as it's more general?

skbio/diversity/alpha/_base.py
```diff
@@ -295,6 +295,52 @@ def esty_ci(counts):
...
+        same length as ``u``.
```
wasade: `u` is `counts`?

skbio/diversity/alpha/_base.py
```diff
@@ -295,6 +295,52 @@ def esty_ci(counts):
...
+    tree: skbio.TreeNode
+        Tree relating the OTUs in otu_ids. The set of tip names in the tree can
```
wasade: Same comment re: OTU, won't comment further on it as it's a general question

skbio/diversity/alpha/_base.py
```diff
@@ -295,6 +295,52 @@ def esty_ci(counts):
...
+
+    Returns
+    -------
+    float
+        The phylogenetic diversity (PD) of the samples.
+
+    Raises
+    ------
+    ValueError
+        If ``u`` and ``otu_ids`` are not equal in length.
```
wasade: Same re: `u` is `counts`?

skbio/diversity/beta/__init__.py
```diff
@@ -56,42 +56,51 @@
     [ 0.85714286  0.75        0.09392265  0.87777778  0.          0.68235294]
     [ 0.81521739  0.1627907   0.71597633  0.89285714  0.68235294  0.        ]]
 
-   Compute Jaccard distances between all pairs of samples and return a
-   ``DistanceMatrix`` object:
-
-   >>> j_dm = pw_distances(data, ids, "jaccard")
-   >>> print(j_dm)
+   Compute weighted unifrac distances between all pairs of samples and return a
```
wasade: unifrac -> UniFrac

skbio/diversity/beta/_base.py
```diff
@@ -44,15 +47,25 @@ def pw_distances(counts, ids=None, metric="braycurtis"):
...
     """
+    # this should probably go somewhere else. where?
+    _skbio_metrics = {'unweighted_unifrac': unweighted_unifrac,
+                      'weighted_unifrac': weighted_unifrac}
     num_samples = len(counts)
     if ids is not None and num_samples != len(ids):
         raise ValueError(
             "Number of rows in counts must be equal to number of provided "
             "ids.")
+    if metric in _skbio_metrics:
```
wasade: should this raise if it evals `False`?

skbio/diversity/beta/_base.py
```diff
@@ -44,15 +47,25 @@ def pw_distances(counts, ids=None, metric="braycurtis"):
...
     See Also
     --------
+    unweighted_unifrac
+    weighted_unifrac
     scipy.spatial.distance.pdist
     pw_distances_from_table
 
     """
+    # this should probably go somewhere else. where?
+    _skbio_metrics = {'unweighted_unifrac': unweighted_unifrac,
```
wasade: Probably at the global scope in this module or elsewhere, there isn't a benefit in constructing this lookup everytime the method is called

skbio/diversity/beta/_base.py
```diff
@@ -8,28 +8,31 @@
...
+def pw_distances(metric, counts, ids=None, **kwargs):
     """Compute distances between all pairs of columns in a counts matrix
 
     Parameters
     ----------
+    metric : str, callable
+        The pairwise distance function as a string or callable to use when
+        generating pairwise distances. See the scipy ``pdist`` docs and the
+        scikit-bio functions linked under *See Also* for available metrics.
     counts : 2D array_like of ints or floats
```
wasade: just a note, and likely out of scope for the PR, but it would be awesome if sparse vectors were allowed here

wasade: ...or a sparse matrix

skbio/diversity/beta/_base.py
```diff
@@ -44,15 +47,25 @@ def pw_distances(counts, ids=None, metric="braycurtis"):
...
     num_samples = len(counts)
     if ids is not None and num_samples != len(ids):
         raise ValueError(
             "Number of rows in counts must be equal to number of provided "
             "ids.")
+    if metric in _skbio_metrics:
+        metric = _skbio_metrics[metric]
+
+    if callable(metric):
+        metric = partial(metric, **kwargs)
```
wasade: Probably not relevant here, but if you don't know about `operator.methodcaller`, it's worth looking at. It's like `partial` except you can have it call a method of an object. Came across it the other day in [pyinstrument](https://github.com/joerick/pyinstrument/blob/master/pyinstrument/profiler.py#L293)

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,170 @@
...
+from skbio.util._decorator import experimental
+
+
+@experimental(as_of="0.4.0-dev")
+def unweighted_unifrac(u_counts, v_counts, otu_ids, tree):
+    """ Compute unweighted unifrac
+
+    Parameters
+    ----------
+    u_counts, v_counts: list, np.array
```
wasade: Minor, but what do you think about just naming them as `u` and `v`?

wasade: ...same here re: sparse support

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,170 @@
...
+        Vector of OTU ids corresponding to tip names in ``tree``. Must be the
+        same length as ``u_counts`` and ``v_counts``.
+    tree: skbio.TreeNode
+        Tree relating the OTUs in otu_ids. The set of tip names in the tree can
+        be a superset of ``otu_ids``, but not a subset.
+
+    Returns
+    -------
+    float
+        The unweighted unifrac distance between the two samples.
```
wasade: UniFrac

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,170 @@
...
+    unweighted (qualitative) versus weighted (quantitiative) diversity metrics
+    is presented in [2]_.
+
+    References
+    ----------
+    .. [1] Lozupone, C. & Knight, R. UniFrac: a new phylogenetic method for
+       comparing microbial communities. Appl. Environ. Microbiol. 71, 8228-8235
+       (2005).
+
+    .. [2] Lozupone, C. A., Hamady, M., Kelley, S. T. & Knight, R. Quantitative
```
wasade: I think it would be great to also add in [Lladser et al 2011](http://www.ncbi.nlm.nih.gov/pubmed/?term=lladser%5Bau%5D+AND+knight%5Bau%5D+AND+unifrac)

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,170 @@
...
+       and qualitative beta diversity measures lead to different insights into
+       factors that structure microbial communities. Appl. Environ. Microbiol.
+       73, 1576-1585 (2007).
+
+    """
+    if len(u_counts) != len(v_counts) != len(otu_ids):
+        raise ValueError("Input vectors u_counts, v_counts, and otu_ids must"
+                         " be of equal length.")
+    u_observed_otus = {o: c for o, c in zip(otu_ids, u_counts) if c >= 1}
```
wasade: This comprehension is done in numerous places, possible to decompose?

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,170 @@
...
+    v_observed_otus = {o: c for o, c in zip(otu_ids, v_counts) if c >= 1}
+    observed_nodes1 = set(tree.observed_node_counts(u_observed_otus))
+    observed_nodes2 = set(tree.observed_node_counts(v_observed_otus))
+    observed_branch_length = sum(
```
wasade: Could shorten this some: `sum(o.length or 0.0 for o in observed_nodes1 | observed_nodes2)`

wasade: I think it would be great if the variable names were shorter -- I think a balance can be found between descriptive names without needing to span multiple lines, but this is minor

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,170 @@
...
+    return unweighted_unifrac
+
+
+@experimental(as_of="0.4.0-dev")
+def weighted_unifrac(u_counts, v_counts, otu_ids, tree, normalized=False):
+    """ Compute weighted unifrac with or without branch length normalization
+
+    Parameters
+    ----------
+    u_counts, v_counts: list, np.array
```
wasade: Same general comments as above re: names and sparse

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,170 @@
...
+    References
+    ----------
+    .. [1] Lozupone, C. A., Hamady, M., Kelley, S. T. & Knight, R. Quantitative
+       and qualitative beta diversity measures lead to different insights into
+       factors that structure microbial communities. Appl. Environ. Microbiol.
+       73, 1576–1585 (2007).
+
+    """
+    u_observed_otus = {o: c for o, c in zip(otu_ids, u_counts) if c >= 1}
+    u_total_count = sum(u_counts)
```
wasade: BTW, it may be nice to ensure that the type coming in is `np.array`:

```python
12:41:57 (mcdonadt@Daniels-MacBook-Pro-2):/Users/mcdonadt
> data = range(100000)

14:26:23 (mcdonadt@Daniels-MacBook-Pro-2):/Users/mcdonadt
> npdata = array(data)

14:26:29 (mcdonadt@Daniels-MacBook-Pro-2):/Users/mcdonadt
> %timeit sum(data)
100 loops, best of 3: 4.77 ms per loop

14:26:34 (mcdonadt@Daniels-MacBook-Pro-2):/Users/mcdonadt
> %timeit sum(npdata)
10000 loops, best of 3: 49.1 µs per loop
```

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,170 @@
...
+    v_observed_otus = {o: c for o, c in zip(otu_ids, v_counts) if c >= 1}
+    v_total_count = sum(v_counts)
+    if u_total_count == 0 and v_total_count == 0:
```
wasade: since a negative count doesn't make sense, doesn't that imply that this test could be done first and with `if not len(u_counts) and not len(v_counts)`?

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,170 @@
...
+        # boundary case where both communities have no members
+        return 0.0
+    u_observed_nodes = tree.observed_node_counts(u_observed_otus)
+    v_observed_nodes = tree.observed_node_counts(v_observed_otus)
+    weighted_unifrac = 0
+    D = 0
+    for o in set(u_observed_nodes) | set(v_observed_nodes):
+        # handle the case of o.length is None
+        b = o.length or 0
+        u_branch_weight = _sample_branch_weight(
```
wasade: same comment about length of variable names

wasade: this method call isn't actually necessary if the counts produced on `v_observed_otus` and `u_observed_otus` were already all divided by the total count as could be done with a little reordering of the logic earlier in the method

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,170 @@
...
+            u_observed_nodes[o], u_total_count)
+        v_branch_weight = _sample_branch_weight(
+            v_observed_nodes[o], v_total_count)
+        branch_weight = abs(u_branch_weight - v_branch_weight)
+        weighted_unifrac += b * branch_weight
+        if o.is_tip() and normalized:
```
wasade: Minor, but I think reversing this would be better as in the unnormalized case, the method call is always incurred.

skbio/tree/_tree.py
```diff
@@ -651,6 +651,46 @@ def count(self, tips=False):
             return len(list(self.traverse(include_self=True)))
 
     @experimental(as_of="0.4.0")
+    def observed_node_counts(self, tip_counts):
+        """Returns counts of nodes observations from counts of tip observations
```
wasade: nodes -> node

skbio/tree/_tree.py
```diff
@@ -651,6 +651,46 @@ def count(self, tips=False):
...
+        ----------
+        tip_counts : dict of ints
+            Counts of observations of tips. Keys correspond to tip names in
+            ``self``, and counts are ints greater than zero.
+
+        Returns
+        -------
+        dict
+            Counts of observations of nodes. Keys correspond to node names
+            (internal nodes or tips), and counts are ints greater than zero.
```
wasade: unsigned ints?

skbio/tree/_tree.py
```diff
@@ -651,6 +651,46 @@ def count(self, tips=False):
...
+        ValueError
+            If a count less than one is observed.
+        MissingNodeError
+            If a count is provided for a tip not in the tree, or for an
+            internal node.
+
+        """
+        result = defaultdict(int)
+        for tip_name, count in tip_counts.items():
+            if count < 1:
```
wasade: this check is redundant given as it's been done in the callers, but not sure if it's worth doing both here and in the caller

skbio/diversity/alpha/tests/test_base.py
```diff
@@ -152,6 +162,43 @@ def _diversity(indices, f):
...
+                          self.t1)
+        expected = 0.0
+        self.assertAlmostEqual(actual, expected)
+        actual = faith_pd([0, 0, 0, 0, 0], self.oids1, self.t1)
+        expected = 0.0
+        self.assertAlmostEqual(actual, expected)
+
+    def test_faith_pd_all_observed(self):
+        actual = faith_pd([1, 1, 1, 1, 1], self.oids1, self.t1)
+        expected = sum([n.length for n in self.t1.traverse()
```
wasade: Minor, but the `[]` aren't needed with `sum`

skbio/diversity/beta/_base.py
wasade: oh, i see, the assumption is that if it isn't an skbio metric that scipy may know how to deal with it

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,170 @@
...
+        raise ValueError("Input vectors u_counts, v_counts, and otu_ids must"
+                         " be of equal length.")
+    u_observed_otus = {o: c for o, c in zip(otu_ids, u_counts) if c >= 1}
+    v_observed_otus = {o: c for o, c in zip(otu_ids, v_counts) if c >= 1}
+    observed_nodes1 = set(tree.observed_node_counts(u_observed_otus))
+    observed_nodes2 = set(tree.observed_node_counts(v_observed_otus))
+    observed_branch_length = sum(
+        o.length for o in observed_nodes1 | observed_nodes2
+        if o.length is not None)
+    if observed_branch_length == 0:
```
wasade: it may be worth an explicit check against the nodes (e.g., if the intersection is the empty set) as small negative branch lengths can result from FastTree. Perhaps something like:

```python
unioned_nodes = observed_nodes1 | observed_nodes2
if not unioned_nodes:
    return 0.0

observed_branch_length = sum(o.length or 0.0 for o in unioned_nodes)
```

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,170 @@
...
+    u_counts, v_counts: list, np.array
+        Vectors of counts of OTUs for two samples. Must be equal length.
+    otu_ids: list, np.array
+        Vector of OTU ids corresponding to tip names in ``tree``. Must be the
+        same length as ``u_counts`` and ``v_counts``.
+    tree: skbio.TreeNode
+        Tree relating the OTUs in otu_ids. The set of tip names in the tree can
+        be a superset of ``otu_ids``, but not a subset.
+    normalized: boolean, optional
+        If ``True``, apply branch length normalization. Resulting distances
```
wasade: Possible to include a mention in the notes about what this normalization is?

wasade:
Done with the review, this is awesome! Few general comments/questions:

- should OTU be changed to observation?
- should `u_counts` and `v_counts` just be `u` and `v`?
- unifrac -> UniFrac

Also note, build is failing against py27

gregcaporaso:
Thanks for the review @wasade, working through your comments this morning. 

> should ``u_counts`` and ``v_counts`` just be ``u`` and ``v``?

I started out with ``u`` and ``v``, but @jairideout, @ebolyen and I discussed yesterday. The motivation to name them ``u`` and ``v`` is consistency with the scipy beta diversity metrics, and the motivation to include the term ``counts`` is for clarity to users of other ecological software, and consistency with ``skbio.diversity.alpha``. @jairideout came up with these terminology distinctions when reviewing other ecology software, see the [note at the top of this page](http://scikit-bio.org/docs/latest/generated/skbio.diversity.alpha.html) for what we mean by counts specifically. I think we should leave this as is. 

> should OTU be changed to observation?

I'm torn on this... I think that in our documentation for ``TreeNode``, we should refer to tips as OTUs. This would be consistent with the literature on phylogeny reconstruction. If we go with that, then it makes sense to call these OTUs here since the ids are to map counts to tree tips (OTUs). I of course take your point that some/most/all of these metrics these could be applied to other data types, but I think it might make sense for the phylogenetic metrics to call these OTUs, since e.g. a generalized UniFrac for metagenomic data would probably operate on a DAG, and would ultimately probably be a different function. (For example, we specifically tell users to not use these metrics on shotgun data in the QIIME tutorials.)

> Also note, build is failing against py27

So hard to support this antiquated technology... :weary:

skbio/diversity/alpha/tests/test_base.py
gregcaporaso: fixed, thx.

skbio/diversity/alpha/_base.py
gregcaporaso: forgot to update when I updated the variable name.

skbio/diversity/alpha/_base.py
gregcaporaso: Responded on the *Conversation* tab.

ElDeveloper:
> > Also note, build is failing against py27
> 
> So hard to support this antiquated technology... :weary:

:movie_camera:

skbio/diversity/beta/_base.py
gregcaporaso: I completely agree. My thought is that that's something that can happen later without an interface change. When the BIOM table is stabilized, we'll likely revisit these interfaces, so I think that would be the natural point to stabilize support for this.

skbio/diversity/beta/_base.py
gregcaporaso: yep

skbio/diversity/beta/_unifrac.py
gregcaporaso: I think that one is less relevant here. These are the two that define the metrics themselves, while that one is a response to incorrect statements about the metrics.

skbio/diversity/beta/_unifrac.py
gregcaporaso: Good points, did some refactoring of variable names here.

skbio/diversity/beta/_unifrac.py
mortonjt: I agree with @wasade on this one.  I think this reference is a nice one to have because it includes some proofs on why Unifrac is an actual distance metric.

skbio/diversity/beta/_unifrac.py
mortonjt: A negative count wouldn't make sense, but it is still possible that a user erroneously inputs negative counts.  Perhaps it is worth while to raise an error if there are any negative counts?

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,170 @@
...
+            D += (d * normed_weight)
+    if normalized:
+        return weighted_unifrac / D
+    else:
+        return weighted_unifrac
+
+
+def _sample_branch_weight(observed_nodes, total_count):
+    if total_count == 0:
+        return 0
```
mortonjt: I'm just curious.  Does this scenario ever happen?
If the user inputs a list with no counts in it, shouldn't that raise an error?

skbio/diversity/beta/_unifrac.py
mortonjt: While we are on the topic, I implemented an identical function, `skbio.stats.composition.closure`.  It also does perform error checking on the inputs.

skbio/diversity/beta/_unifrac.py
gregcaporaso: Ok, good point. I'll add it.

skbio/diversity/beta/_unifrac.py
gregcaporaso: done

skbio/diversity/beta/_unifrac.py
gregcaporaso: UniFrac is defined for that case, so it should be supported. If  ``u_counts`` is all zeros and ``v_counts`` is all zeros, the distance between those samples is ``0.0``. If ``u_counts`` is all zeros and ``v_counts`` has at least one non-zero value, unweighted unifrac and weighted unifrac with branch length normalization are 1.0, and weighted unifrac without branch length normalization will be >= 0.0 (depending on the branch lengths).

skbio/diversity/beta/_unifrac.py
gregcaporaso: Good catch, fixed.

skbio/diversity/beta/_unifrac.py
gregcaporaso: I don't think it should be disallowed, though it might be a place to raise an ``EfficiencyWarning``. Let's revisit that (I'm going to add a ``_validate`` function to centralize some of these things, so we might add that there).

skbio/diversity/beta/_unifrac.py
gregcaporaso: Added to-do item for this.

wasade:
Okay, sounds good re: `u_counts` and `v_counts`. The PR isn't entirely
consistent in the use IIRC -- can flag if you'd like.

On OTUs, I disagree that the `TreeNode` should describe tips as OTUs given
that phylogenies can be used for orthologous or paralogous genes. Is OTU
really the terminology used in the phylogenetics literature?

skbio/diversity/beta/_base.py
wasade: Ah, so just in the future, overload the accepted datatypes? Okay. It wouldn't be out of reach to do that now since skbio already depends on scipy though, however, it may be better to have general support across the diversity metrics for sparse vectors and matrices which would imply a separate PR

skbio/diversity/beta/_unifrac.py
wasade: Thanks! The reason @mortonjt cited is why I thought it would be useful

skbio/diversity/beta/_unifrac.py
wasade: Thanks!

skbio/diversity/beta/_unifrac.py
wasade: Thanks!

skbio/diversity/beta/_unifrac.py
wasade: Awesome. On an aside, we should have the original UniFrac site documentation which contained a pretty awesome set of examples and descriptions. Can dig it up if that material would be of use

skbio/diversity/beta/_unifrac.py
wasade: Sounds good, thanks!

skbio/diversity/beta/_unifrac.py
wasade: ...and casting to `uint` would lead to some unexpected fun. I think it may be useful to put this into the `_validate` function that @gregcaporaso mentioned above

skbio/diversity/beta/_unifrac.py
wasade: These special cases can be caught earlier, the benefit is a reduction in the downstream complexity. I think the only annoyance then is the weignted w/o normalization:

```python
if not len(u_counts) or not len(v_counts):
    if len(u_counts) == len(v_counts):
        return 0.0
    elif not normalization:
        return 0.0
    else:
        counts = u_counts if len(u_counts) else v_counts
        # normalization counts -- would this break down to just PD?
```

wasade: ...note, just to clarify, suggesting that these checks be done `weighted_unifrac` near the top. Just wanted to add the note as the line this comment is on is not within that scope right now.

gregcaporaso:
@wasade, don't worry about it for now, I have a bunch of local changes that may have fixed it.

skbio/diversity/beta/_unifrac.py
gregcaporaso: Going to add that check in a ``validate`` function.

skbio/diversity/beta/_unifrac.py
gregcaporaso: I'm going to leave this as-is for now if you guys don't mind (it works, and we can work on optimizing later).

skbio/diversity/beta/_unifrac.py
gregcaporaso: fixed

skbio/tree/_tree.py
gregcaporaso: I think it's important to do here, since we don't know who the caller will be (this is a public method). May revisit doing this when I write ``_validate`` for *unifrac.

skbio/diversity/beta/_unifrac.py
gregcaporaso: It wouldn't just be PD (I think), b/c of weighting.

rob-knight: Correct. One note on weighted unifrac: we may want to investigate other ways of doing the difference in branch weight, e.g. squared differences. There are some other variants of unifrac out there e.g. VAW that we may also want to look into, if only to benchmark.

skbio/tree/_tree.py
gregcaporaso: missed initially, fixed in a local copy.

skbio/diversity/beta/_unifrac.py
wasade: oh, duh. Thanks. I do think catching the special cases early would still be worth considering if it could be done cleanly

skbio/diversity/beta/_base.py
gregcaporaso: That's what I'm thinking. Let's do this in a separate PR.

skbio/diversity/beta/_unifrac.py
gregcaporaso: I'm going to leave that till the optimization stage for now.

skbio/diversity/beta/tests/test_unifrac.py
```diff
@@ -0,0 +1,378 @@
...
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+from __future__ import absolute_import, division, print_function
+
+from unittest import TestCase, main
+
+import numpy as np
+
+from skbio.io._fileobject import StringIO
```
wasade: importing from a private module?

skbio/diversity/alpha/_base.py
```diff
@@ -311,6 +293,52 @@ def esty_ci(counts):
     return f1 / n - z * np.sqrt(W), f1 / n + z * np.sqrt(W)
 
 
+@experimental(as_of="0.4.0-dev")
+def faith_pd(counts, otu_ids, tree):
```
wasade: Should this also have a `suppress_validation` like the `unweighted_unifrac` and `weighted_unifrac`?

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,200 @@
...
+    return {o: c for o, c in zip(otu_ids, counts) if c >= 1}
+
+
+def _validate(u_counts, v_counts, otu_ids, tree):
+    _validate_counts_vectors(u_counts, v_counts, suppress_cast=True)
+    _validate_otu_ids_and_tree(u_counts, otu_ids, tree)
+
+
+@experimental(as_of="0.4.0-dev")
+def unweighted_unifrac(u_counts, v_counts, otu_ids, tree,
```
wasade: PyCogent's Fast UniFrac also supports [G](https://github.com/pycogent/pycogent/blob/master/cogent/maths/unifrac/fast_tree.py#L375) (normalized and unnormalized). Should the skbio API allow different metrics to be used?

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+def _validate(u_counts, v_counts, otu_ids, tree):
+    _validate_counts_vectors(u_counts=u_counts, v_counts=v_counts,
+                             suppress_cast=True)
+    _validate_otu_ids_and_tree(counts=u_counts, otu_ids=otu_ids, tree=tree)
+
+
+@experimental(as_of="0.4.0-dev")
+def unweighted_unifrac(u_counts, v_counts, otu_ids, tree,
```
wasade: The Fast UniFrac code gets a lot of benefit from having its data structure cached, but that comes at a large expense on each call to the implementation. The present API doesn't allow for passing in an existing cached data structure. One option is to pack a cached vectorized version of the tree into `tree`. Another option would be to expand the API to allow `**kwargs`. I think the latter is better as it is more explicit, and it is possible that other implementations have have similar needs for being able to pass in additional information. Thoughts?

gregcaporaso: I think expanding to allow kwargs would work.

wasade: Great, thanks!

skbio/diversity/beta/_unifrac.py
gregcaporaso: This is done.

skbio/diversity/beta/_unifrac.py
gregcaporaso: That could definitely be useful, e.g., as module level documentation or in a cookbook recipe (depending on what it looks like). Let's plan on that for when we get the optimized code in.

skbio/diversity/_base.py
```diff
@@ -0,0 +1,85 @@
...
+
+    if counts.ndim != 1:
+        raise ValueError("Only 1-D vectors are supported.")
+    elif (counts < 0).any():
+        raise ValueError("Counts vector cannot contain negative values.")
+
+    return counts
+
+
+def _validate_counts_vectors(u_counts, v_counts, suppress_cast=False):
```
gregcaporaso: Comment from @jairideout: maybe generalize this to take *args, so it can work for one or more counts vectors. This will allow consolidation with ``_validate_counts_vector``.

skbio/diversity/alpha/_base.py
```diff
@@ -311,6 +294,59 @@ def esty_ci(counts):
     return f1 / n - z * np.sqrt(W), f1 / n + z * np.sqrt(W)
 
 
+@experimental(as_of="0.4.0-dev")
+def faith_pd(counts, otu_ids, tree, suppress_validation=False):
```
gregcaporaso: Comment from @ebolyen: change ``suppress_validation=False`` to ``validate=True`` through-out, for consistency with the rest of the codebase.

skbio/diversity/beta/_base.py
```diff
@@ -8,28 +8,38 @@
...
 import numpy as np
 from scipy.spatial.distance import pdist, squareform
 
+from skbio.diversity.beta._unifrac import unweighted_unifrac, weighted_unifrac
 from skbio.stats.distance import DistanceMatrix
 from skbio.util._decorator import experimental, deprecated
 
 
+def _get_skbio_metrics():
+    result = {
```
gregcaporaso: return directly

skbio/diversity/beta/_base.py
```diff
@@ -66,18 +84,18 @@ def pw_distances(counts, ids=None, metric="braycurtis"):
 
 @deprecated(as_of="0.4.0", until="0.4.1",
             reason=pw_distances_from_table_deprecation_reason)
-def pw_distances_from_table(table, metric="braycurtis"):
+def pw_distances_from_table(metric, table):
```
gregcaporaso: Revert the API change to this function, since it's deprecated (no reason to modify the API, since it's going away).

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,203 @@
...
+                       suppress_validation=False, **kwargs):
+    """ Compute unweighted UniFrac
```
gregcaporaso: When we have multiple implementations of this (e.g., ``unweighted_unifrac`` and  ``unweighted_unifrac_slow``), the doc string should be inherited (e.g., with ``@overrides``).

gregcaporaso: This will be the subject of [an RFC](https://github.com/biocore/scikit-bio-rfcs/issues/2).

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,203 @@
...
+        Vector of OTU ids corresponding to tip names in ``tree``. Must be the
+        same length as ``u_counts`` and ``v_counts``.
+    tree: skbio.TreeNode
+        Tree relating the OTUs in otu_ids. The set of tip names in the tree can
+        be a superset of ``otu_ids``, but not a subset.
+    suppress_validation: bool, optional
+        If `True`, validation of the input won't be performed. This step can
+        be slow, so if validation is run elsewhere it can be disabled here.
+        However, invalid input data can lead to invalid results, so this step
+        should not be bypassed all together.
```
gregcaporaso: From @jairideout: Add ``kwargs``, and note that it's for internal use only.

skbio/diversity/beta/tests/test_unifrac.py
```diff
@@ -0,0 +1,624 @@
...
+from io import StringIO
+
+import numpy as np
+
+from skbio import TreeNode
+from skbio.tree import DuplicateNodeError, MissingNodeError
+from skbio.diversity.beta import unweighted_unifrac, weighted_unifrac
+
+
+class StatsTests(TestCase):
```
gregcaporaso: Change class name to ``UniFracTests``.

skbio/diversity/beta/tests/test_unifrac.py
```diff
@@ -0,0 +1,624 @@
...
+
+    def setUp(self):
+        self.b1 = np.array(
```
gregcaporaso: change variable name to ``self.table1``

scikitten-bio:
## Automated report from asv run
[http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html)

gregcaporaso:
Awesome to see the report @scikitten-bio! 

@jairideout, @ebolyen, and @wasade I'm going to wrap up my changes on this tomorrow so should be ready for merge then (and I'll create issues for the "future" items at that point as well).

skbio/diversity/beta/_unifrac.py
gregcaporaso: This is done now @wasade.

scikitten-bio:
## Automated report from asv run
[http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html)

scikitten-bio:
## Automated report from asv run
[http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html)

scikitten-bio:
## Automated report from asv run
[http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html)

scikitten-bio:
## Automated report from asv run
[http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html)

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,232 @@
...
+    otu_ids: list, np.array
+        Vector of OTU ids corresponding to tip names in ``tree``. Must be the
+        same length as ``u_counts`` and ``v_counts``.
+    tree: skbio.TreeNode
+        Tree relating the OTUs in otu_ids. The set of tip names in the tree can
+        be a superset of ``otu_ids``, but not a subset.
+    normalized: boolean, optional
+        If ``True``, apply branch length normalization, which is described in
+        [1]_. Resulting distances will then be in the range ``[0, 1]``.
+    suppress_validation: bool, optional
```
jairideout: should be `validate`

scikitten-bio:
## Automated report from asv run
[http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html)

skbio/diversity/alpha/_base.py
```diff
@@ -311,6 +294,70 @@ def esty_ci(counts):
...
+    ----------
+    counts : 1-D array_like, int
+        Vector of counts.
+    otu_ids: list, np.array
+        Vector of OTU ids corresponding to tip names in ``tree``. Must be the
+        same length as ``counts``.
+    tree: skbio.TreeNode
+        Tree relating the OTUs in otu_ids. The set of tip names in the tree can
+        be a superset of ``otu_ids``, but not a subset.
+    suppress_validation: bool, optional
```
jairideout: should be `validate`

skbio/diversity/beta/_unifrac.py
```diff
@@ -0,0 +1,232 @@
...
+    ----------
+    u_counts, v_counts: list, np.array
+        Vectors of counts of OTUs for two samples. Must be equal length.
+    otu_ids: list, np.array
+        Vector of OTU ids corresponding to tip names in ``tree``. Must be the
+        same length as ``u_counts`` and ``v_counts``.
+    tree: skbio.TreeNode
+        Tree relating the OTUs in otu_ids. The set of tip names in the tree can
+        be a superset of ``otu_ids``, but not a subset.
+    suppress_validation: bool, optional
```
jairideout: should be `validate`

skbio/tree/_tree.py
```diff
@@ -653,6 +653,46 @@ def count(self, tips=False):
             return len(list(self.traverse(include_self=True)))
 
     @experimental(as_of="0.4.0")
```
jairideout: as_of should be 0.4.0-dev

setup.py
```diff
@@ -128,6 +128,7 @@ def finalize_options(self):
           'skbio.io.format.tests': ['data/*'],
           'skbio.stats.tests': ['data/*'],
           'skbio.stats.distance.tests': ['data/*'],
+          'skbio.diversity.tests': ['data/*'],
```
jairideout: fix to include alpha and beta dirs

scikitten-bio:
## Automated report from asv run
[http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html)

scikitten-bio:
## Automated report from asv run
[http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html)

anderspitman:
@scikitten-bio is probably a little too active. I'll add some issues to autobencher to try and keep the number of automated comments down, delete outdated ones, etc.

scikitten-bio:
## Automated report from asv run
[http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/gregcaporaso/issue-46/html/index.html)

PR #1039, opened 2015-08-10
DistanceMatrix.from_iterable

ebolyen:
This serves as a replacement for the `SequenceCollection.distance` method which was the only method on `SequenceCollection` which could not be implemented as a simple list comprehension.

Hopefully we can remove `SequenceCollection` in favor of standard Python types such as `list` and `dict` in the near future.


jairideout:
Ready for merge when tests pass.

PR #1065, opened 2015-08-19
DOC: Only dunders we have created will appear in the documentation

ebolyen:
Some hacky things are happening, but it beats redistributing numpydoc!


jairideout:
:+1: this is *way* better than the `.. shownumpydoc` hack we used to have!

jairideout:
Ready for merge when tests pass

PR #1069, opened 2015-08-21
Add pielous evenness

### LINKED ISSUE #1068: Add Pielou's evenness as an alpha diversity metric
There is a cool alpha diversity metric of evenness which is based on Shannon entropy. I know this metric as ['Pielou's evenness'](https://en.wikipedia.org/wiki/Species_evenness).

Similar metrics are already included in scikit-bio these include [equitability](https://github.com/biocore/scikit-bio/blob/master/skbio/diversity/alpha/_base.py#L207), [Heip's evenness](https://github.com/biocore/scikit-bio/blob/master/skbio/diversity/alpha/_base.py#L454), and Shannon itself.

Colin

colinbrislawn:
Fixes #1068 

I'm new at this and I'm sure there are style issues.
Suggestions are welcome!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Version 0.4.0-dev (changes since 0.4.0 release go here)
 
 ### Features
-* Added ``skbio.diversity.alpha.pielou`` function as an evenness metric of alpha diversity. ([#1068](https://github.com/biocore/scikit-bio/issues/1068))
+* Added ``skbio.diversity.alpha.pielou_e`` function as an evenness metric of alpha diversity. ([#1068](https://github.com/biocore/scikit-bio/issues/1068))
 * Added `to_regex` method to `skbio.sequence._iupac_sequence` ABC - it returns a regex object that matches all non-degenerate versions of the sequence.
 * Added ``skbio.util.assert_ordination_results_equal`` function for comparing ``OrdinationResults`` objects in unit tests.
 * Added ``skbio.util.RepresentationWarning`` for warning about substitutions, assumptions, or particular alterations that were made for the successful completion of a process.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Version 0.4.0-dev (changes since 0.4.0 release go here)
 
 ### Features
+* Added ``skbio.diversity.alpha.pielou`` function as an evenness metric of alpha diversity. ([#1068](https://github.com/biocore/scikit-bio/issues/1068))
 * Added `to_regex` method to `skbio.sequence._iupac_sequence` ABC - it returns a regex object that matches all non-degenerate versions of the sequence.
 * Added ``skbio.util.assert_ordination_results_equal`` function for comparing ``OrdinationResults`` objects in unit tests.
 * Added ``skbio.util.RepresentationWarning`` for warning about substitutions, assumptions, or particular alterations that were made for the successful completion of a process.
--- a/skbio/diversity/alpha/_base.py
+++ b/skbio/diversity/alpha/_base.py
@@ -239,6 +239,7 @@ def equitability(counts, base=2):
     denominator = np.log(observed_otus(counts)) / np.log(base)
     return numerator / denominator
 
+
 @experimental(as_of="0.4.0")
 def esty_ci(counts):
     r"""Calculate Esty's CI.
@@ -848,6 +849,7 @@ def osd(counts):
     counts = _validate_counts_vector(counts)
     return observed_otus(counts), singles(counts), doubles(counts)
 
+
 @experimental(as_of="0.4.0")
 def pielou(counts):
     """Calculate Pielou's Evenness (Shannon index / ln(number of OTUs).
@@ -858,7 +860,7 @@ def pielou(counts):
 
        \frac{(H)}{\ln S}
 
-    where :math:`H` is the Shannon-Wiener entropy of counts and :math:`S` is 
+    where :math:`H` is the Shannon-Wiener entropy of counts and :math:`S` is
     the number of OTUs in the sample.
 
     Parameters
@@ -878,11 +880,11 @@ def pielou(counts):
 
     Notes
     -----
-    The implementation here is based on the description given by Carlo Heip [1]_
-    and is also described on Wikipedia [2]_.
-    
-	It is similar to Heip's evenness [1]_.
-	
+    The implementation here is based on the description given by Carlo Heip
+    [1]_ and is also described on Wikipedia [2]_.
+
+    It is similar to Heip's evenness [1]_.
+
     References
     ----------
     .. [1] Heip, C. 1974. A new index measuring evenness. J. Mar. Biol. Ass.
@@ -890,9 +892,10 @@ def pielou(counts):
     .. [2] https://en.wikipedia.org/wiki/Species_evenness
 
     """
-    
+
     counts = _validate_counts_vector(counts)
     return shannon(counts) / np.log(observed_otus(counts))
+
 
 @experimental(as_of="0.4.0")
 def robbins(counts):
--- a/skbio/diversity/alpha/_base.py
+++ b/skbio/diversity/alpha/_base.py
@@ -893,7 +893,7 @@ def pielou_e(counts):
 
     """
     counts = _validate_counts_vector(counts)
-    return shannon(counts) / np.log(observed_otus(counts))
+    return shannon(counts, base=np.e) / np.log(observed_otus(counts))
 
 
 @experimental(as_of="0.4.0")
--- a/skbio/diversity/alpha/_base.py
+++ b/skbio/diversity/alpha/_base.py
@@ -852,7 +852,7 @@ def osd(counts):
 
 @experimental(as_of="0.4.0-dev")
 def pielou_e(counts):
-    """Calculate Pielou's Evenness index J'.
+    r"""Calculate Pielou's Evenness index J'.
 
     Pielou's Evenness is defined as:
 
--- a/skbio/diversity/alpha/_base.py
+++ b/skbio/diversity/alpha/_base.py
@@ -880,13 +880,14 @@ def pielou_e(counts):
 
     Notes
     -----
-    The implementation here is based on the described in Wikipedia [1]_. It was
-    first proposed by E. C. Pielou [2]_ and is similar to Heip's evenness [3]_.
+    The implementation here is based on the description in Wikipedia [1]_.
+    It was first proposed by E. C. Pielou [2]_ and is similar to Heip's
+    evenness [3]_.
 
     References
     ----------
     .. [1] https://en.wikipedia.org/wiki/Species_evenness
-    .. [2] PIELOU, E. C., 1966. The measurement of diversity in different types
+    .. [2] Pielou, E. C., 1966. The measurement of diversity in different types
        of biological collections. Journal of Theoretical Biology, 13, 131-44.
     .. [3] Heip, C. 1974. A new index measuring evenness. J. Mar. Biol. Ass.
        UK., 54, 555-557.
--- a/skbio/diversity/alpha/_base.py
+++ b/skbio/diversity/alpha/_base.py
@@ -852,13 +852,13 @@ def osd(counts):
 
 @experimental(as_of="0.4.0-dev")
 def pielou_e(counts):
-    """Calculate Pielou's Evenness (Shannon index / ln(number of OTUs).
+    """Calculate Pielou's Evenness index J'.
 
     Pielou's Evenness is defined as:
 
     .. math::
 
-       \frac{(H)}{\ln S}
+       J' = \frac{(H)}{\ln S}
 
     where :math:`H` is the Shannon-Wiener entropy of counts and :math:`S` is
     the number of OTUs in the sample.
--- a/skbio/diversity/alpha/_base.py
+++ b/skbio/diversity/alpha/_base.py
@@ -222,7 +222,7 @@ def equitability(counts, base=2):
     See Also
     --------
     shannon
-    pielou
+    pielou_e
 
     Notes
     -----
@@ -477,7 +477,7 @@ def heip_e(counts):
     See Also
     --------
     shannon
-    pielou
+    pielou_e
 
     Notes
     -----
@@ -850,8 +850,8 @@ def osd(counts):
     return observed_otus(counts), singles(counts), doubles(counts)
 
 
-@experimental(as_of="0.4.0")
-def pielou(counts):
+@experimental(as_of="0.4.0-dev")
+def pielou_e(counts):
     """Calculate Pielou's Evenness (Shannon index / ln(number of OTUs).
 
     Pielou's Evenness is defined as:
--- a/skbio/diversity/alpha/_base.py
+++ b/skbio/diversity/alpha/_base.py
@@ -858,7 +858,7 @@ def pielou_e(counts):
 
     .. math::
 
-       J' = \frac{(H)}{\ln S}
+       J' = \frac{(H)}{\ln(S)}
 
     where :math:`H` is the Shannon-Wiener entropy of counts and :math:`S` is
     the number of OTUs in the sample.
@@ -881,9 +881,8 @@ def pielou_e(counts):
     Notes
     -----
     The implementation here is based on the description given by Carlo Heip
-    [1]_ and is also described on Wikipedia [2]_.
-
-    It is similar to Heip's evenness [1]_.
+    [1]_ and is also described on Wikipedia [2]_. It is similar to Heip's
+    evenness [1]_.
 
     References
     ----------
@@ -892,7 +891,6 @@ def pielou_e(counts):
     .. [2] https://en.wikipedia.org/wiki/Species_evenness
 
     """
-
     counts = _validate_counts_vector(counts)
     return shannon(counts) / np.log(observed_otus(counts))
 
--- a/skbio/diversity/alpha/_base.py
+++ b/skbio/diversity/alpha/_base.py
@@ -880,15 +880,16 @@ def pielou_e(counts):
 
     Notes
     -----
-    The implementation here is based on the description given by Carlo Heip
-    [1]_ and is also described on Wikipedia [2]_. It is similar to Heip's
-    evenness [1]_.
+    The implementation here is based on the described in Wikipedia [1]_. It was
+    first proposed by E. C. Pielou [2]_ and is similar to Heip's evenness [3]_.
 
     References
     ----------
-    .. [1] Heip, C. 1974. A new index measuring evenness. J. Mar. Biol. Ass.
+    .. [1] https://en.wikipedia.org/wiki/Species_evenness
+    .. [2] PIELOU, E. C., 1966. The measurement of diversity in different types
+       of biological collections. Journal of Theoretical Biology, 13, 131-44.
+    .. [3] Heip, C. 1974. A new index measuring evenness. J. Mar. Biol. Ass.
        UK., 54, 555-557.
-    .. [2] https://en.wikipedia.org/wiki/Species_evenness
 
     """
     counts = _validate_counts_vector(counts)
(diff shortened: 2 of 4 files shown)


scikitten-bio:
## Automated report from asv run
[http://powertrip.mggen.nau.edu:5000/runs/colinbrislawn/add_pielous_evenness/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/colinbrislawn/add_pielous_evenness/html/index.html)

skbio/diversity/alpha/tests/test_base.py
```diff
@@ -393,6 +394,20 @@ def test_observed_otus(self):
...
 
+    def test_pielou(self):
+        # Calculate "by hand".
+        arr = np.array([1, 2, 3, 1])
+        h = shannon(arr)
+        s = 4
+        expected = h / np.log(s)
+        self.assertAlmostEqual(pielou(arr), expected)
+
+        self.assertAlmostEqual(pielou(self.counts), 1.334283585856)
```
gregcaporaso: A couple of additional "boundary case" tests would also be good here. Can you test with passing in an empty vector (i.e., ``[]``), a vector where there are counts greater than 1 (e.g., ``[1,2,3,0]``), and a vector where ``observed_otus == 0``? 

Also, minor, but you only need to have the seven most significant digits in your expected values, b/c ``assertAlmostEqual`` only checks for equality to the sixth decimal place by default.

skbio/diversity/alpha/_base.py
```diff
@@ -847,6 +848,51 @@ def osd(counts):
...
+    --------
+    shannon
+    heip_e
+
+    Notes
+    -----
+    The implementation here is based on the description given by Carlo Heip [1]_
+    and is also described on Wikipedia [2]_.
+    
+	It is similar to Heip's evenness [1]_.
```
gregcaporaso: Looks like you have some extra indentation here.

gregcaporaso:
@colinbrislawn, you should run ``make test`` from your top-level scikit-bio directory. You're getting some failures from flake8 in travis. This will let you run those tests locally so you can correct them before your next push.

Thanks a lot for the contribution! Generally looks really good, just a few small things and it should be ready for merge.

skbio/diversity/alpha/_base.py
```diff
@@ -847,6 +848,51 @@ def osd(counts):
     counts = _validate_counts_vector(counts)
     return observed_otus(counts), singles(counts), doubles(counts)
 
+@experimental(as_of="0.4.0")
```
ebolyen: the experimental string should be `"0.4.0-dev"`

colinbrislawn: Thanks for the input @gregcaporaso, @ebolyen!

I did run `make test` and assumed it was fine when I received no errors from the code. I did receive an error with flake8, like you mentioned.
```
Ran 1313 tests in 66.137s

OK
flake8 skbio setup.py checklist.py
make: flake8: No such file or directory
make: *** [test] Error 1
```

How can I fix this?

ebolyen: `pip install flake8` there may be other packages that it will need in the same way. They should all be pip installable.

jairideout:
Thanks @colinbrislawn! Regarding the function name, I think it should be `pielou_e` to match the other "evenness" metrics (e.g., `heip_e`, `mcintosh_e`, `simpson_e`).

colinbrislawn:
@jairideout Good suggestion! I'll change the function name to `pielou_e`.

skbio/diversity/alpha/tests/test_base.py
colinbrislawn: Great idea @gregcaporaso! I was looking for more tests cases but was not sure what to add.

---

How should a metric respond when given a vector where `observed_otus == 0`? Currently, many metrics throws errors, like this:

``` python
>>> count = np.array([0, 0])
>>> shannon(count)
nan
>>> pielou(count)
nan
```
I think this is the 'correct' answer because calculating this metric with `observed_otus == 0` would involve dividing by zero. Should I test for this?

---

How should a metric respond when given an empty vector (`[]`)?  Currently, many metrics throw errors like this: 
``` python
>>> count = np.array([])
>>> shannon(count)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "skbio/diversity/alpha/_base.py", line 972, in shannon
    counts = _validate_counts_vector(counts)
  File "skbio/diversity/_base.py", line 25, in _validate_counts_vector
    counts = counts.astype(int, casting='safe', copy=False)
TypeError: Cannot cast array from dtype('float64') to dtype('int64') according to the rule 'safe'
```

skbio/diversity/alpha/_base.py
```diff
@@ -847,6 +848,51 @@ def osd(counts):
...
+def pielou(counts):
+    """Calculate Pielou's Evenness (Shannon index / ln(number of OTUs).
```
jairideout: Can you remove the equation from this summary line? numpydoc recommends putting this info in the body of the docstring (like you already have). A summary line such as the following should suffice:

```
Calculate Pielou's evenness index J'.
```

skbio/diversity/alpha/_base.py
```diff
@@ -847,6 +848,51 @@ def osd(counts):
...
+
+    Pielou's Evenness is defined as:
+
+    .. math::
+
+       \frac{(H)}{\ln S}
```
jairideout: Can you add `J' = ` here?

colinbrislawn: Fixed! :smile:

skbio/diversity/alpha/_base.py
colinbrislawn: I used conda to install the dependancies listed in the travis-ci folder. I am now getting this new error after running `make test`.

```
...
Ran 1313 tests in 64.961s

OK
flake8 skbio setup.py checklist.py
./checklist.py
check-manifest
make: check-manifest: No such file or directory
make: *** [test] Error 1
```

I don't think the issue is with flake8 because this runs fine: `flake8 skbio setup.py checklist.py`

scikitten-bio:
## Automated report from asv run
[http://powertrip.mggen.nau.edu:5000/runs/colinbrislawn/add_pielous_evenness/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/colinbrislawn/add_pielous_evenness/html/index.html)

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Version 0.4.0-dev (changes since 0.4.0 release go here)
 
 ### Features
+* Added ``skbio.diversity.alpha.pielou`` function as an evenness metric of alpha diversity. ([#1068](https://github.com/biocore/scikit-bio/issues/1068))
```
jairideout: `pielou` -> `pielou_e`

colinbrislawn:
There is some issue with building documentation.

```
Warning, treated as error:
/home/travis/miniconda3/envs/env_name/lib/python3.4/site-packages/skbio/diversity/alpha/_base.py:
docstring of skbio.diversity.alpha.pielou_e:14: 
WARNING: Block quote ends without a blank line; unexpected unindent.
```

Any advice?

gregcaporaso:
Did you see my comment about an indentation error in your docstring? That
might be it.

colinbrislawn:
Greg, I've cleaned up my code using `flake8`, so I think it's something different. 

What does the `14` mean, in this line?
`skbio.diversity.alpha.pielou_e:14:`

Colin

ebolyen:
@colinbrislawn That is typically going to be the line number.

filepath:line-number:character-of-line

ebolyen:
In this specfic case it is relative to the docstring, since that is all Sphinx really knows, so it thinks the 14th line of your doc for `pielou_e` has a problem. Because it is sphinx this should be interpreted as something being wrong in the *general vicinity* of that line

ebolyen:
Playing with it locally and this sphinx error is making no sense as usual. Going to try and figure out what is happening and I will report back.

colinbrislawn:
>general vicinity

That's what I was afraid of...

When I remove these lines, I don't get an error from this file. Do you see anything fishy about this code?
``` python
    .. math::

       J' = \frac{(H)}{\ln(S)}
```

ebolyen:
That all looks fine. Tried wrapping `(S)` in `{}` and that didn't help. I don't get it.

colinbrislawn:
So when I remove those lines from `alpha/_base.py`, I get a different error in a different file.

``` python
Warning, treated as error:
/Users/bris469/Documents/Projects/toy/scikit-bio/skbio/diversity/beta/__init__.py:docstring of skbio.diversity.beta:31: WARNING: Exception occurred in plotting skbio-diversity-beta-1
 from /Users/bris469/Documents/Projects/toy/scikit-bio/doc/source/generated/skbio.diversity.beta.rst:
Traceback (most recent call last):
  File "/Users/bris469/anaconda/envs/scikit-bio-dev/lib/python2.7/site-packages/matplotlib/sphinxext/plot_directive.py", line 506, in run_code
    six.exec_(code, ns)
  File "/Users/bris469/anaconda/envs/scikit-bio-dev/lib/python2.7/site-packages/six.py", line 672, in exec_
    exec("""exec _code_ in _globs_, _locs_""")
  File "<string>", line 1, in <module>
  File "<string>", line 36, in <module>
TypeError: initial_value must be unicode or None, not str

make: *** [html] Error 1
```

@ebolyen Thanks for helping me out with this! This is my first time doing 'social coding' and I really appreciate your willingness help me get started.

ebolyen:
@colinbrislawn No problem, that's what we're here for! RE: new error; I can't seem to reproduce that, the docs build completely once the math directive is removed...

colinbrislawn:
OK, I'll push again and see if it builds. I guess the Travis-CI build is the authoritative one, so we can pick up the error from there.

ebolyen:
@colinbrislawn I'm afraid we've been duped by the oldest sphinx related bug in the book!

The docstring doesn't being with an `r` which will tell python to treat it literally. So because it isn't present, python will expect each `\` to do something, which causes the docstring that sphinx actually gets to be wildly different from what you see. You can fix this by just starting it like this: `r""" rest of docstring...` or by double-escaping the `\` to `\\`. But using `r` is generally preferred.

scikitten-bio:
## Automated report from asv run
[http://powertrip.mggen.nau.edu:5000/runs/colinbrislawn/add_pielous_evenness/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/colinbrislawn/add_pielous_evenness/html/index.html)

colinbrislawn:
Ah ok! I've made this fix. `make test` now returns this error:
```
Ran 1313 tests in 66.403s

OK
flake8 skbio setup.py checklist.py
./checklist.py
check-manifest
make: check-manifest: No such file or directory
make: *** [test] Error 1
```

ElDeveloper:
I would try `pip install check-manifest`, which I think would work.

skbio/diversity/alpha/_base.py
```diff
@@ -848,6 +850,51 @@ def osd(counts):
...
+
+    References
+    ----------
+    .. [1] Heip, C. 1974. A new index measuring evenness. J. Mar. Biol. Ass.
+       UK., 54, 555-557.
+    .. [2] https://en.wikipedia.org/wiki/Species_evenness
+
+    """
+    counts = _validate_counts_vector(counts)
+    return shannon(counts) / np.log(observed_otus(counts))
```
jairideout: scikit-bio's `shannon` function oddly uses a default logarithm base of 2, so you'll need to pass `base=np.e` to specify the natural log.

skbio/diversity/alpha/_base.py
```diff
@@ -848,6 +850,51 @@ def osd(counts):
...
+        Pielou's Evenness.
+
+    See Also
+    --------
+    shannon
+    heip_e
+
+    Notes
+    -----
+    The implementation here is based on the description given by Carlo Heip
```
jairideout: Can you add a reference to the original paper where this metric is introduced?

```
PIELOU, E. C., 1966. The measurement of diversity in different types of biological collections. Journal of Theoretical Biology, 13, 131-44.
```

I think the current wording is still correct (i.e., we are basing the implementation on Heip's paper and Wikipedia) because the original paper doesn't specify a logarithm base.

scikitten-bio:
## Automated report from asv run
[http://powertrip.mggen.nau.edu:5000/runs/colinbrislawn/add_pielous_evenness/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/colinbrislawn/add_pielous_evenness/html/index.html)

skbio/diversity/alpha/tests/test_base.py
jairideout: @ebolyen @gregcaporaso what do you think about having `nan` returned for the divide-by-zero case? I don't know whether it's more useful to return that or have it raise an error saying the metric is undefined when there are no observed OTUs.

Regarding empty input `[]`, all metrics should currently be raising that error since it happens within `_validate_counts_vector`. I think `[]` should be explicitly disallowed because there are essentially no data for a metric to operate on. A better error message should be raised though. @ebolyen @gregcaporaso @colinbrislawn do you agree? I don't think this should get addressed in this PR since it has a broader scope.

skbio/diversity/alpha/_base.py
colinbrislawn: Wow! Good catch.

When building the unit test for this function, I realized that I was generating my 'correct' numbers using other parts of scikit-bio. Do you know of an independent implementation of shannon diversity I can compare it to?

jairideout: No problem, we should probably change the default logarithm base to `e` in `shannon` to avoid issues like this in the future (since there are several diversity metrics that use Shannon internally but assume natural log). Definitely doesn't need to be addressed in this PR.

Here's a few options to get unbiased expected values to use in your unit tests:

- [free online calculator](https://www.easycalculation.com/statistics/shannon-wiener-diversity.php)

- [this site](http://classroom.synonym.com/calculate-species-evenness-2851.html) runs through the calculations by hand with a small example, so you could add their example dataset to the unit tests and include a link (as a comment) to where the expected value came from

- perhaps the Heip paper you're linking to has an example dataset with expected values for Pielou's evenness?

- [SDR-IV](http://www.pisces-conservation.com/index.html?softdiversity.html$softwaremenu.html) has a free demo but probably only runs on Windows

wdwvt1: I think that the default of Shannon to base 2 is accurate because then the
Shannon score is in bits. This means its 'Shannon entropy', which I
understand is what the statistical theory is built on. Here is a post by
Dan Knights (
https://groups.google.com/forum/#!searchin/qiime-forum/shannon$20entropy/qiime-forum/QduSqx0K8Bg/O6FcpvAkYmcJ)
about this - though it takes no position on the correctness of this
practice. Obviously a change of base is no big deal from theoretical or
practical perspectives, but perhaps to be consistent we should stick with
base e.

Just my two cents.

colinbrislawn: I'll be using this site for external ref: http://ww2.mdsg.umd.edu/interactive_lessons/biofilm/diverse.htm#3

I'm now calling `shannon(base=np.e)`, which works fine. The choice of a default is beyond my pay grade and both will work fine in the code.

Open question: when only one species is in the community, what should we return for Pielou's evenness?
- Option one: return `0` because Shannon entropy is zero.
- Option two: return `nan` because we are dividing by log(1), which is zero.

scikitten-bio:
## Automated report from asv run
[http://powertrip.mggen.nau.edu:5000/runs/colinbrislawn/add_pielous_evenness/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/colinbrislawn/add_pielous_evenness/html/index.html)

skbio/diversity/alpha/_base.py
jairideout: Thanks for the details @wdwvt1! I agree that changing the default base to `e` is probably safest for consistency, though outside this PR's scope.

@ebolyen and I discussed these border cases, let's go with the following for now:

- for empty input `[]`, let it continue to raise an error (don't worry about adding unit tests for this case). The current error message isn't that helpful but we will revisit how to handle `[]` universally when we bring the alpha diversity package from experimental to stable (timeline TBD).

- for counts of all zeros (e.g., `[0, 0, 0]`) add a unit test ensuring it returns `nan`

- for a single species occurrence (e.g., `[0, 42, 0]` and `[42]`) add a unit test ensuring it returns `nan`. It is unclear whether an evenness of 1.0 should be returned, so for now we'll match the equation and return `nan`. We'll revisit this decision in the future when working on the alpha diversity package.

colinbrislawn: Sounds like a plan @jairideout!

How do I test for `nan`. This does not work:
``` python
FAIL: test_pielou_e (skbio.diversity.alpha.tests.test_base.BaseTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/bris469/Documents/Projects/toy/scikit-bio/skbio/diversity/alpha/tests/test_base.py", line 414, in test_pielou_e
    self.assertEqual(pielou_e([0, 0, 200, 0, 0]), np.nan)
AssertionError: nan != nan
```

Both correctly return the `numpy.nan` object which is a `numpy.float64`. However, `assertEqual` gets scared when it see this.

ebolyen: @colinbrislawn You can use [`isnan`](http://docs.scipy.org/doc/numpy/reference/generated/numpy.isnan.html). Generally, any operation done to `nan` will result in False or another `nan` by the floating point specification.

skbio/diversity/alpha/_base.py
```diff
@@ -848,6 +850,52 @@ def osd(counts):
...
+        Pielou's Evenness.
+
+    See Also
+    --------
+    shannon
+    heip_e
+
+    Notes
+    -----
+    The implementation here is based on the described in Wikipedia [1]_. It was
```
jairideout: *the described* -> *the description*

skbio/diversity/alpha/_base.py
```diff
@@ -848,6 +850,52 @@ def osd(counts):
...
+    first proposed by E. C. Pielou [2]_ and is similar to Heip's evenness [3]_.
+
+    References
+    ----------
+    .. [1] https://en.wikipedia.org/wiki/Species_evenness
+    .. [2] PIELOU, E. C., 1966. The measurement of diversity in different types
```
jairideout: *PIELOU* -> *Pielou*

skbio/diversity/alpha/tests/test_base.py
```diff
@@ -393,6 +394,26 @@ def test_observed_otus(self):
...
+        self.assertAlmostEqual(pielou_e(self.counts), 0.92485490560)
+
+        self.assertEqual(pielou_e([1, 1]), 1.0)
+        self.assertEqual(pielou_e([1, 1, 1, 1]), 1.0)
+        self.assertEqual(pielou_e([1, 1, 1, 1, 0, 0]), 1.0)
+
+        # Examples from
+        # http://ww2.mdsg.umd.edu/interactive_lessons/biofilm/diverse.htm#3
+        self.assertAlmostEqual(pielou_e([1, 1, 196, 1, 1]), 0.078, 3)
+        print(pielou_e([0, 0, 200, 0, 0]))
```
jairideout: extra print statement

skbio/diversity/alpha/_base.py
colinbrislawn: @ebolyen So, like this?
``` python
        self.assertEqual(np.isnan(pielou_e([0, 0, 0, 0, 0])), True)
```
I feel like there is a more elegant way to code that.

ebolyen: @colinbrislawn use `assertTrue` :smile:

colinbrislawn: Perfect!

scikitten-bio:
## Automated report from asv run
[http://powertrip.mggen.nau.edu:5000/runs/colinbrislawn/add_pielous_evenness/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/colinbrislawn/add_pielous_evenness/html/index.html)

jairideout:
Looks great, thanks for your contribution @colinbrislawn!

colinbrislawn:
Thanks :smile:

gregcaporaso:
Thanks @colinbrislawn!

PR #1080, opened 2015-08-28
ENH: add Sequence.frequencies and Sequence.observed_chars

### LINKED ISSUE #1074: add Sequence method to compute character frequencies
It would be useful to have a way of computing the frequencies of characters (or a subset of characters) in a sequence. Currently the way to do this is `seq.kmer_frequencies(k=1)`, which handles the frequencies of characters present in the sequence but does not support a subset of characters.

It could also be useful to have this functionality on `Sequence` because iterating over positions in an alignment will yield `Sequence` objects. See the [`TabularMSA` RFC](https://github.com/biocore/scikit-bio-rfcs/pull/1) for potential `Alignment` API changes, which will *not* include the current `Alignment.position_frequencies` functionality.

@ebolyen and I propose adding the method `Sequence.frequencies(chars=None, relative=False)` where:

- `chars=None` produces the frequencies of characters in the sequence. Equivalent to `seq.kmer_frequencies(k=1)`.
- `chars` will accept either a single-character string or a `set` of single-character strings indicating characters to include in the frequency calculations. We only support single-character strings to avoid confusion with substring matching (use `Sequence.count` or `'foo' in seq` for this). We don't want users to run something like `seq.frequencies(chars='ACGT')` and expect to get a frequency distribution back for each character.
- `relative` matches `Sequence.kmer_frequencies`, returning relative frequencies when `True`
- the method returns the same type of output as `Sequence.kmer_frequencies`

### LINKED ISSUE #1075: add Sequence method to return set of observed characters
@ebolyen and I propose adding the method `Sequence.observed_chars()`, which returns the `set` of characters observed in the sequence.

This seems like a useful piece of the sequence API as it mirrors class methods describing the alphabet such as `gap_chars`, `degenerate_chars`, etc. The set of observed characters can then be referenced or composed with other pieces of the sequence API, such as `Sequence.frequencies` (proposed in #1074).

jairideout:
Added `Sequence.frequencies` method for computing character frequencies in a sequence.

Added `Sequence.observed_chars` property for obtaining the set of observed characters in a sequence. 

`Sequence.kmer_frequencies` now returns a `dict` instead of `Counter` or `defaultdict`. See changelog for detailed explanation of this backward-incompatible stable API change.

Fixes #1074. Fixes #1075.

### DIFF
--- a/skbio/sequence/_sequence.py
+++ b/skbio/sequence/_sequence.py
@@ -1900,11 +1900,9 @@ class Sequence(collections.Sequence, SkbioObject):
 
         Returns
         -------
-        collections.Counter or collections.defaultdict
+        dict
             Frequencies of words of length `k` contained in the biological
-            sequence. This will be a ``collections.Counter`` if `relative` is
-            ``False`` and a ``collections.defaultdict`` if `relative` is
-            ``True``.
+            sequence.
 
         Raises
         ------
@@ -1913,19 +1911,19 @@ class Sequence(collections.Sequence, SkbioObject):
 
         Examples
         --------
-        >>> from collections import defaultdict, Counter
+        >>> from pprint import pprint
         >>> from skbio import Sequence
         >>> s = Sequence('ACACATTTATTA')
         >>> freqs = s.kmer_frequencies(3, overlap=False)
-        >>> freqs == Counter({'TTA': 2, 'ACA': 1, 'CAT': 1})
-        True
+        >>> pprint(freqs) # using pprint to display dict in sorted order
+        {'ACA': 1, 'CAT': 1, 'TTA': 2}
         >>> freqs = s.kmer_frequencies(3, relative=True, overlap=False)
-        >>> freqs == defaultdict(float, {'ACA': 0.25, 'TTA': 0.5, 'CAT': 0.25})
-        True
+        >>> pprint(freqs)
+        {'ACA': 0.25, 'CAT': 0.25, 'TTA': 0.5}
 
         """
         kmers = self.iter_kmers(k, overlap=overlap)
-        freqs = collections.Counter((str(seq) for seq in kmers))
+        freqs = dict(collections.Counter((str(seq) for seq in kmers)))
 
         if relative:
             if overlap:
@@ -1933,7 +1931,7 @@ class Sequence(collections.Sequence, SkbioObject):
             else:
                 num_kmers = len(self) // k
 
-            relative_freqs = collections.defaultdict(float)
+            relative_freqs = {}
             for kmer, count in viewitems(freqs):
                 relative_freqs[kmer] = count / num_kmers
             freqs = relative_freqs
--- a/skbio/sequence/_sequence.py
+++ b/skbio/sequence/_sequence.py
@@ -539,11 +539,11 @@ class Sequence(collections.Sequence, SkbioObject):
         --------
         >>> from skbio import Sequence
         >>> s = Sequence('AACGAC')
-        >>> s.observed_chars == {b'G', b'A', b'C'}
+        >>> s.observed_chars == {'G', 'A', 'C'}
         True
 
         """
-        return set(self.values)
+        return set(str(self))
 
     @property
     def _string(self):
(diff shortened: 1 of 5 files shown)


anderspitman:
@scikitten-bio seems to be on holiday. Can someone resend the last webhooks 'opened' event? Thanks

anderspitman:
PS logging is enable now so I'll be able to keep resending it myself once I capture it.

jairideout:
Just resent it, let me know if that's not the right one.

anderspitman:
That's perfect, thanks!

scikitten-bio:
## Automated report
[http://powertrip.mggen.nau.edu:5000/runs/jairideout/char-freqs-and-obs-chars/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/jairideout/char-freqs-and-obs-chars/html/index.html)

scikitten-bio:
Back in the saddle

skbio/sequence/_sequence.py
```diff
@@ -1764,6 +1784,144 @@ def mismatch_frequency(self, other, relative=False):
...
+        # inclusive.
+        obs_chars = obs_bytes[0].astype(np.uint8).tostring().decode('ascii')
+        obs_counts = freqs[obs_bytes]
+        if relative:
+            obs_counts = obs_counts / len(self)
+        freqs = dict(zip(obs_chars, obs_counts))
+
+        if chars is not None:
+            # Get a copy of the keys because we're deleting in the loop.
+            for obs_char in list(freqs.keys()):
```
jairideout: This can probably be vectorized and pushed into the numpy code above. I'll look into this on Monday. Performance-wise there won't be a difference but it'll be cleaner than this.

jairideout:
> Back in the saddle

You can ride a horse?!?! :horse_racing: :cat:

scikitten-bio:
## Automated report
[http://powertrip.mggen.nau.edu:5000/runs/jairideout/char-freqs-and-obs-chars/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/jairideout/char-freqs-and-obs-chars/html/index.html)

scikitten-bio:
unicat

jairideout:
:open_mouth:

skbio/sequence/_sequence.py
jairideout: Looking into this further, I don't think it'll be much cleaner or shorter than the current code. I'll keep as-is for review and will update if necessary.

ebolyen: I think you could skip a lot of this code entirely if you did something like this:
```python
freqs = np.bincount(self._bytes,
                    minlength=self._number_of_extended_ascii_codes)

if chars is not None:
    chars = list(self._munge_to_char_set(chars))
    indices = [ord(c) for c in chars]
else:
    indices = np.nonzero(freqs)[0]
    chars = indices.astype(np.uint8).tostring().decode('ascii')

obs_counts = freqs[indices]
if relative:
    obs_count /= len(self)

return dict(zip(chars, obs_count))
```

This also has the advantage of making the keys of the dict the same type as what is in the set (when provided).

skbio/sequence/_sequence.py
```diff
@@ -1764,6 +1784,144 @@ def mismatch_frequency(self, other, relative=False):
...
+                isinstance(chars, six.binary_type):
+            chars = set([chars])
+        elif not isinstance(chars, set):
+            raise TypeError(
+                "`chars` must be of type `set`, not %r" % type(chars).__name__)
+
+        munged_chars = set()
+        for char in chars:
+            if isinstance(char, six.binary_type):
+                char = char.decode('ascii')
```
ebolyen: This wouldn't be necessary in the above alternative.

skbio/sequence/tests/test_sequence.py
```diff
@@ -1640,6 +1656,148 @@ def test_index_on_subclass(self):
...
+        self.assertEqual(seq.frequencies(chars=None, relative=True), exp)
+        self.assertEqual(
+            seq.frequencies(chars=seq.observed_chars, relative=True),
+            exp)
+
+    def test_frequencies_invalid_chars(self):
+        seq = Sequence('abcabc')
+
+        with six.assertRaisesRegex(self, ValueError, '0 characters'):
+            seq.frequencies(chars='')
```
ebolyen: :+1: :+1: :+1: 

(Although np.inf could have be a response...)

jairideout: heh

skbio/sequence/_sequence.py
jairideout: Thanks! This is much cleaner and shorter. I also like that the keys you get back match those provided by the user.

skbio/sequence/_sequence.py
jairideout: :shipit:

jairideout:

skbio/sequence/_sequence.py
```diff
@@ -1764,6 +1784,138 @@ def mismatch_frequency(self, other, relative=False):
...
+        else:
+            indices, = np.nonzero(freqs)
+            # Downcast from int64 to uint8 then convert to str. This is safe
+            # because we are guaranteed to have indices in the range 0 to 255
+            # inclusive.
+            chars = indices.astype(np.uint8).tostring().decode('ascii')
+
+        obs_counts = freqs[indices]
+        if relative:
+            obs_counts = obs_counts / len(self)
```
ebolyen: Is it possible to change the dtype of `obs_counts` so that division works inplace?

jairideout: I don't think there is an advantage to doing this since we'll have to call `astype` to convert to float, which will always return a copy (so in either case another vector gets created).

scikitten-bio:
## Automated report
[http://powertrip.mggen.nau.edu:5000/runs/jairideout/char-freqs-and-obs-chars/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/jairideout/char-freqs-and-obs-chars/html/index.html)

skbio/sequence/_sequence.py
ebolyen: Oh right, that makes sense, there's no way to create a floating point view of ints in place.

scikitten-bio:
## Automated report
[http://powertrip.mggen.nau.edu:5000/runs/jairideout/char-freqs-and-obs-chars/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/jairideout/char-freqs-and-obs-chars/html/index.html)

Issue #897, opened 2015-04-17
Multivariate regression of distance matrices


JWDebelius:
It would be really useful to have a multivariate regression for distance matrices in Python. It seems like the only one currently implemented is Adonis in R, and that isn't a regression, it's a MANOVA.

Jorge-C:
:+1: Do you have pointers to existing implementations/papers?

mortonjt:
Maybe something like [ecodist](http://cran.r-project.org/web/packages/ecodist/ecodist.pdf)?

JWDebelius:
The `corgen` in ecodist is one example.  The lmdist package (which is proprietary) is another. 

[Lichstein 2007](http://people.biology.ufl.edu/jlichstein/downloads/publications/Lichstein_2007_Plant_Ecol.pdf) (although I think he wrote ecodist) and [Smouse et al 1986](ftp://zhoulab.rccc.ou.edu/Wu/Reference/Multiple%20regression%20and%20correlation%20extensions%20of%20the%20Mantel%20test%20of%20matrix%20correspondence.pdf) also both talk about the problem.

jairideout:
:+1:, this sounds like a useful addition to the distance-based stats in scikit-bio. Related to #407 since Lichstein 2007 says MRM is an extension of partial Mantel. Maybe it makes sense to get partial Mantel added first?

If anyone is interested/available to work on #407 and/or this issue, let me know and I'll assign.

JWDebelius:
@jairideout: I'd like to start on #407, and maybe move into this. I may need some help with the math, though.

Thanks!

Jorge-C:
@JWDebelius I don't know much about #407, but I'm always happy to help with math!

jairideout:
Thanks @JWDebelius for offering to work on #407! That one is now assigned to you. I'd also be happy to help with the math, API design, or any other questions.

PR #1104, opened 2015-09-01
Multiple regression on distance matrices

mortonjt:
This addresses #897

Tried to base the formatting off of the mantel test.

A couple of things
1. Added in an option to add `np.RandomState`.  We'll need to add this eventually, so might as well add it in this PR
2. Thinking about adding in a function in `_base.py` for creating distance matrices from metadata / abundances.  This function will have the following format

```
def make_distance_matrix(x, f):
"""
Parameters
----------------
x : array_like or pd.DataFrame or pd.Series
f : function
...
"""
```

Basically, it'll be the equivalent of `dist()` in R.  I think it'll make it a tad bit easier for users to create distance matrices.  Any thoughts?

The following will need to be completed before this can get merged
- [x] Write tests for handling the `strict` and `lookup` parameters
- [x] Add in some tests against MRM from ecodist
- [x] Add some examples
- [x] Add in publicly visible helper function for discretizing categorical distances
- [ ] squash commits

Am I forgetting anything else @JWDebelius?


skbio/stats/distance/_regression.py
```diff
@@ -0,0 +1,170 @@
...
+    This module performs a multiple linear regression on distance
+    matrices.  This is done by extracting the upper (or lower)
+    triangular portions of the symmetric distance matrices,
+    and collapsing them to compute a multiple linear regression
+    on the distances.  The signficances are computed by
+    permuting the residuals.
+
+    Parameters
+    ----------
+    y : DistanceMatrix, array_like, or filepath
```
JWDebelius: Is there a reason not to simply import a distance matrix? The conversion doesn't make sense to me?

skbio/stats/distance/_regression.py
```diff
@@ -0,0 +1,170 @@
...
+    # a single matrix
+    n = len(y.data)
+    X = np.vstack([np.ones((1, n*(n-1)/2))] + \
+                  [k.data[np.triu_indices(n, 1)] for k in xargs]).T
+    Y = np.atleast_2d(y[np.triu_indices(n, 1)]).T
+    n, p = X.shape
+    J = np.ones((n, n))
+    I = np.identity(n)
+
+    # Define regression function
```
JWDebelius: Should this be its own function, so it can be applied to other data types?

JWDebelius:
This is feature incomplete. It would be really helpful if there was also API in place to 
1. Import a pandas data frame  or numpy array of categorical data and convert it to a series of distance matrices (a list?) that can be used with this function.
2. Import a pandas dataframe or numpy array of continuous data and covert it to a distance matrix. I think the scipy API around this exists.

I'd ideally also like to see the steps (metadata conversion and linear regression) combined as a wrapper function.

skbio/stats/distance/_regression.py
mortonjt: I did this to be consistent with [mantel](https://github.com/biocore/scikit-bio/blob/master/skbio/stats/distance/_mantel.py#L303).  

@jairideout , could you clarify why filepaths are an option in the mantel test?

skbio/stats/distance/_regression.py
mortonjt: :+1: That'll actually clean things up. And also open the doors for permutative OLS.

The only issue that I can see is that having a permutative OLS would no longer fit within the `skbio.stats.distance` package.  It'll have to become its own separate thing...

skbio/stats/distance/_regression.py
ebolyen: @mortonjt That is only for `pwmantel` where loading each (of potentially hundreds) of distance matrices into memory is unproductive, when instead `pwmantel` could just open them from disk one at a time as needed.

jairideout:
@mortonjt @JWDebelius take a look at `DistanceMatrix.from_iterable` and `skbio.diversity.beta.pw_distances`. Do these methods get you the functionality you're currently implementing in `makedm`?

skbio/stats/distance/_regression.py
jairideout: @mortonjt I think you should remove filepath support and just accept `DistanceMatrix` and `array_like` since they'll all have to be loaded into memory at once.

mortonjt: Yup.  I just removed it

skbio/stats/distance/_regression.py
```diff
@@ -0,0 +1,170 @@
...
+    lookup : dict, optional
+        Maps each ID in the distance matrices to a new ID. Used to match up IDs
+        across distance matrices prior to running the Mantel test. If the IDs
+        already match between the distance matrices, this parameter is not
+        necessary. This parameter is disallowed if `x` and `y` are
+        ``array_like``.
+
+    Returns
+    -------
+    B : np.array
```
jairideout: Do you think these data could fit in a `pandas` structure? We're taking this approach for many of the distance-based stats (e.g., `pwmantel`, `anosim`, `permanova`). Or if `pandas` isn't appropriate perhaps this can be a `namedtuple`?

skbio/stats/distance/_regression.py
```diff
@@ -0,0 +1,170 @@
...
+    ----------
+    .. [1] Legendre, P. and Legendre, L. (2012) Numerical Ecology. 3rd English
+       Edition. Elsevier.
+    .. [2] Legendre, P. Lapointe, F., Casgrain P. (1994) Modeling Brain
+       Evolution from Behavior: A Permutational Regression Approach
+    .. [3] https://cran.r-project.org/web/packages/ecodist/index.html
+    """
+
+    # Unpack kwargs
+    params = {'permutations': 10000,
```
jairideout: Is there a reason this default isn't 999? That's the default used in other permutation-based stats in scikit-bio and often in R.

mortonjt:
Hmm.  That functionality is suspiciously similar ...

I am getting the same outputs for 1D arrays and pd.Series.  But, I get
different outputs when I increase the dimensions.

```
In [134]: x = DistanceMatrix.from_iterable([[1, 2, 3], [1, 3, 5]],
euclidean)

In [135]: x.data
Out[135]:
array([[ 0.  ,  2.24],
          [ 2.24,  0.  ]])
```
This distance matrix should be (assuming rows=samples and columns=features)
```
exp = DistanceMatrix([[0, 5],
                                    [5, 0]])
```
However, for the sake of this PR, `from_iterable` does everything we need.
I'm going to remove makedm in this PR.

skbio/stats/distance/_regression.py
mortonjt: There isn't a good reason.  Just used 10000 since that's what the scipy folks wanted on my permutation ttest

skbio/stats/distance/_regression.py
```diff
@@ -0,0 +1,170 @@
...
+import six
+import numpy as np
+from skbio.util._decorator import experimental
+from skbio.util._misc import check_random_state
+from skbio.stats.distance._mantel import _order_dms
+
+from skbio.stats.distance import DistanceMatrix
+
+@experimental(as_of="0.4.0")
+def linregress(y, *args, **kwargs):
```
jairideout: Is this method implementing [MRM](http://mb3is.megx.net/gustame/constrained-analyses/multiple-regression-on-dissimilarity-matrices) or [MLR](https://sites.google.com/site/mb3gustame/constrained-analyses/multiple-regression)? It sounds like the latter but you mentioned checking the output against ecodist's MRM function. Just want to make sure I'm on the same page :)

skbio/stats/distance/_regression.py
jairideout: Thanks!

skbio/stats/distance/_regression.py
mortonjt: That would be actually really awesome.

However, I'm not sure what the best way to do this is.  Each of the coefficients corresponds to a covariate distance matrix.  One option is to pass in a bunch of ids, one for each distance matrix.  But that would bulk up the number of input parameters and could make it confusing how to handle input parameters.

The other option is to pass in an optional ID into the DistanceMatrix constructor.  Any thoughts?

jairideout:
> I am getting the same outputs for 1D arrays and pd.Series.  But, I get
different outputs when I increase the dimensions.

`DistanceMatrix.from_iterable` should work for arbitrary inputs, as long as the metric function can handle comparing those inputs. It passes each pair of items in the iterable to `metric`. Can you post a full code sample reproducing the unexpected behavior when passing something > 1D?

> However, for the sake of this PR, `from_iterable` does everything we need.
I'm going to remove makedm in this PR.

Awesome, thanks!

skbio/stats/distance/_regression.py
jairideout: What does ecodist use as the default?

mortonjt: 1000 permutations.  Here's the [link](https://cran.r-project.org/web/packages/ecodist/ecodist.pdf)

I don't think this is a major issue.  For the sake of consistency with the other methods in skbio, I'll make the default 999.  What do you think?

mortonjt:
Here's an example

```
>>> from skbio.stats.distance import DistanceMatrix
>>> from scipy.spatial.distance import cityblock # manhattan distance
>>> x = DistanceMatrix.from_iterable([[1, 2, 3], [1, 3, 5]], cityblock)
>>> x.data

array([[ 0.,  3.],
       [ 3.,  0.]])
```

That actually works on 2D lists.  Also works on 2D arrays.
Things break when you try to pass in pandas dataframes
```
>>> import pandas as pd
>>> df = pd.DataFrame([[1, 2, 3], [1, 3, 5]])
>>> x = DistanceMatrix.from_iterable(df, cityblock)
>>> x.data

array([[ 0.,  1.,  2.],
       [ 1.,  0.,  1.],
       [ 2.,  1.,  0.]])
```
Doesn't matter if I transpose
```
>>> x = DistanceMatrix.from_iterable(df.T, cityblock)
>>> x.data

array([[ 0.,  1.],
       [ 1.,  0.]])
```

skbio/stats/distance/_regression.py
mortonjt: This method is _supposed_ to implement MRM.  

MRM should just be a variant of MLR, except it vectorizes the upper triangular distance matrix before performing a linear regression.

skbio/stats/distance/_regression.py
jairideout: :+1: to 999 for consistency. Also, 999 makes sense as a default because you're computing the p-value as `(count_extreme + 1) / (permutations + 1)` (i.e., counting the *original* test statistic as one of the permutations).

skbio/stats/distance/_regression.py
mortonjt: But perhaps that name is misleading. Let me change it to `mrm`.

In the meantime, @JWDebelius talked about stripping out the MLR portion and making it its own functionality.  While I think this is a good idea, I'm not sure where this MLR should belong.
I could put it in a new folder `skbio.stats.regression.py`.  Any thoughts?

skbio/stats/distance/_regression.py
jairideout: That's a great question! We had this same issue with `pwmantel` and worked around it with an optional `labels` parameter. Do you think that same strategy would work here?

mortonjt: :+1: I think that'll work just fine.

skbio/stats/distance/_regression.py
jairideout: > But perhaps that name is misleading. Let me change it to mrm.

:+1: for `mrm`. The MRM page I linked to above indicates the regression doesn't necessarily have to be linear. Is that the case for this function? The docstring indicates linear regression.

> In the meantime, @JWDebelius talked about stripping out the MLR portion and making it its own functionality. While I think this is a good idea, I'm not sure where this MLR should belong.
I could put it in a new folder skbio.stats.regression.py. Any thoughts?

MLR appears to be implemented in [statsmodels](http://www.datarobot.com/blog/multiple-regression-using-statsmodels/), [scikit-learn](http://stackoverflow.com/a/11479279/3776794), and [scipy](http://stackoverflow.com/questions/17679140/multiple-linear-regression-with-python). What about using one of these implementations instead?

skbio/stats/distance/_regression.py
jairideout: Awesome!

jairideout:
> Things break when you try to pass in pandas dataframes

Thanks for the code sample. This method doesn't work as expected with DataFrames because iterating over a `DataFrame` gives you column names instead of row vectors:

```python
In [1]: import pandas as pd

In [2]: df = pd.DataFrame([[1, 2, 3], [1, 3, 5]], columns=['a','b','c'])

In [3]: df
Out[3]:
   a  b  c
0  1  2  3
1  1  3  5

In [4]: for e in df:
   ...:     print(e)
   ...:
a
b
c
```

So technically the method is working correctly. As a workaround you should be able to pass `df.values` into the method.

It could be misleading to not have this method work with `DataFrames`, though type-checking to account for this case smells and would (unsafely) assume a specific orientation of the `DataFrame`. @ebolyen do you have any ideas for how to handle this? Maybe the method's docs just need to be expanded to describe what each element in the iterable should typically be.

skbio/stats/distance/_regression.py
mortonjt: 1. Statsmodels assumes normality, that that won't work out of the box.  However, if we _could_ perform permutations on that - but this will only work if statsmodels become a dependency.  This module is also pretty nice since it includes quite a few result statistics
2. Scikit-learn has the same problem with statsmodels, but worse since this will mean more dependencies
3. The scipy solution can be integrated into the solution that I have right now.  It could speed up the calculations of the coefficients.  But this will probably need profiling

mortonjt: tl;dr The first two options require dependencies, while the 3rd option doesn't output all of the information that I need.  Do you think we are willing to import either of those two libraries?

Also considering that it is a short function, would it hurt to leave it where it is?

mortonjt: Alright, here's a good reason to keep it where it is.

When doing the permutations, I noticed that there is a bit of redundant work (`H` and `XX1` being computed multiple times).  If we were to opt for options 1 or 2, we would be doing _alot_ of repetitive work.  I may do option 3, but that'll only speed up some of the bottlenecks.

So I vote to keep that `regress` function inside of the `mrm`, so that we can optimize more appropriately.  Any objections?

ElDeveloper:
I haven't quite looked at the code, etc, but I would like to suggest that the function be named something else but `linregress`? I would advocate for a fully descriptive name. The quasi-abbreviations are (I think) a vestige of MATLAB users moving over to Python and are not all that helpful.

jairideout:
@ElDeveloper see discussion [here](https://github.com/biocore/scikit-bio/pull/1104#discussion_r38446404) about function naming

mortonjt:
I've been making quite a few changes.  Sorry that not all of these changes have been pushed.

skbio/stats/distance/_regression.py
jairideout: Thanks for looking into these implementations. I'm not opposed to making scikit-learn or statsmodels a dependency at some point but it doesn't appear to be worth it just for this one function IMO.

What about making `regress` private (renaming it `_regress`)? That way the code is easier to read, avoiding a long function containing an inner function. Then we won't have to support this as part of scikit-bio's public API and can swap it out for another implementation at any point in the future.

Also, :+1: to trying out scipy's functionality in `_regress` for improved performance and/or code reduction.

mortonjt: Just timed it, my calculation of the coefficients is 20x faster than the scipy version, because I can split up the calculation

```
# scipy 
>>> x = np.arange(1, 1000)
>>> y = np.arange(1, 1000)
>>> A = np.vstack([x, np.ones(len(x))]).T
>>> lstsq(A, y)[0]
array([  1.00e+00,   2.98e-14])

>>>  %timeit lstsq(A, y)
10000 loops, best of 3: 63 us per loop

# mine
>>> Y = np.atleast_2d(y)
>>> X = np.vstack([np.ones(len(x)), x]).T
>>> XX1 = np.linalg.pinv(X.T.dot(X))
>>> XX1.dot(X.T.dot(Y))
array([[ -5.93e-11],
       [  1.00e+00]])

>>> %timeit XX1.dot(X.T.dot(Y))
The slowest run took 8.61 times longer than the fastest. This could mean that an intermediate result is being cached 
100000 loops, best of 3: 3.71 us per loop
```

ebolyen:
Based on offline discussion with @jairideout RE: `from_iterable`. 

I think the current behavior is fine as it is matching the python protocol for iterating (it's name even says as much). There also isn't a clear axis on which to iterate in the case of a `DataFrame` and I think knowledge of `DataFrames` is out of scope for that method. If this becomes a very clear problem (where multiple people are running into this) then it might be worth addressing with an explicit exception, but I don't think that is the case yet. 

Just as I wouldn't hand a `dict` to `from_iterable` without saying `.values()` I also wouldn't had a `DataFrame` to it.

Perhaps in the future we could add something like a `from_arraylike` which understands the `__array__` protocol and does something efficient in that case (at which point `DataFrame` would clearly belong in that method rather than `from_iterable`

skbio/stats/distance/_regression.py
jairideout: :+1: that's awesome, thanks for testing it!

mortonjt:
Just raised #1105

At the moment, it is impossible to incorporate missing value imputation into the regression module.
Fixing #1105 should be relatively simple - it'll just require tweaking `DistanceMatrix._validate`.

However, I think this PR should move ahead.  Missing value imputation is an entirely different beast, so that should be its own separate PR anyways.

mortonjt:
That should fix most of the errors.  Not sure what is going on with the errors with the p-values.
Its possible something is wonky with the random seed ...

jairideout:
@mortonjt sounds good re: handling missing values in a future PR.

Re: the failing tests, the issue only appears to happen with Python 3.3. Can you confirm this locally? Also, is the failure consistent or does it only happen sometimes?

mortonjt:
Looks like that problem is consistent across the travis checks for Python3.3.  Will test on my local machine in a bit

mortonjt:
Ecodist tests has been added.  Other than that remaining python3.3 error, this PR is now ready for review.

mortonjt:
I raised #1108 since dealing with categorical variables in this PR is really finicky.

I don't think it is a blocking issue for this PR.  However as it stands, it is rather un-useable for the average user.  So I think this will require further discussion.

Added the [DO NOT MERGE] back to the title since I'm still making changes to `make_categorical_dms`.

mortonjt:
@ebolyen @jairideout quick question about the `DistanceMatrix.from_iterable()` function

I have the following code
```
>>> import pandas as pd
>>> from skbio.stats.distance import DistanceMatrix
>>> from scipy.spatial.distance import cityblock
>>> x = pd.Series([1, 2, 3], index=['a','b','c'])
>>> y = DistanceMatrix.from_iterable(x, metric=cityblock)
>>> print(y)
3x3 distance matrix
IDs:
'0', '1', '2'
Data:
[[ 0.  1.  2.]
 [ 1.  0.  1.]
 [ 2.  1.  0.]]
```o

**Comment by mortonjt** [comment]
Had a git glitch ...

Anyways, back to my original question
```
In [2]: import pandas as pd
In [3]: from skbio.stats.distance import DistanceMatrix
In [4]: x = pd.Series([1, 2, 3], index=['a','b','c'])
In [5]: y = DistanceMatrix.from_iterable(x, metric=cityblock)
In [6]: print(y)
3x3 distance matrix
IDs:
'0', '1', '2'
Data:
[[ 0.  1.  2.]
 [ 1.  0.  1.]
 [ 2.  1.  0.]]
```
Is there a way to directly pass the index from `x` into `y` in `DistanceMatrix.from_iterable`? Or do I need to do `y.ids = x.index`?

**Comment by jairideout** [comment]
> Added the [DO NOT MERGE] back to the title since I'm still making changes to make_categorical_dms.

Thanks! FYI we have a "Do Not Merge" label for this.

> Is there a way to directly pass the index from x into y in DistanceMatrix.from_iterable?

Not really, one way you could work around this is by passing a generator as `key` that yields successive elements from your `index`. `skbio.diversity.beta.pw_distances` accepts an `ids` parameter, so that function may be an easier way to achieve this.

@ebolyen created #1111 to add `iterable` support to `key`, so when that's merged you'll be able to pass the index directly via `key`.

> Or do I need to do y.ids = x.index?

This would also work.

> I don't think I completely understand what's going on with that key in DistanceMatrix.from_iterable.

`key` tells `from_iterable` what each sample ID should be in the resulting distance matrix. It can accept a callable that takes a single argument (each sample) or a key that is passed through to each sample's `metadata` attribute (if it exists).

For example, to create a `DistanceMatrix` from an iterable of `Sequence` objects, using each sequence's ID (stored as `metadata`) as the resulting ID in the distance matrix:

```python
In [1]: from scipy.spatial.distance import hamming

In [2]: from skbio import Sequence, DistanceMatrix

In [3]: seqs = [Sequence('ACGT', metadata={'id': 'a'}), Sequence('AAAA', metadata={'id': 'b'}), Sequence('ACGA', metadata={'id': 'c'})]

In [4]: dm = DistanceMatrix.from_iterable(seqs, lambda seq1, seq2: hamming(seq1.values, seq2.values), key='id')

In [5]: print(dm)
3x3 distance matrix
IDs:
'a', 'b', 'c'
Data:
[[ 0.    0.75  0.25]
 [ 0.75  0.    0.5 ]
 [ 0.25  0.5   0.  ]]
```

jairideout:
FYI we created a new "in progress" label to mark active PRs that aren't ready for merge yet. I updated your PR to reflect this, hopefully it'll look less hostile than a red "Do Not Merge" label (we'll continue to use that for inactive PRs).

mortonjt:
Thanks!

jairideout:
@mortonjt, we now have better support for keys that have been precomputed (#1112, thanks @ebolyen!). You can now pass an iterable of keys via the `keys` parameter in `from_iterable`. Let us know how it goes!

mortonjt:
Thanks for the quick fix @ebolyen and @jairideout !  I have already made the changes in my local version.

mortonjt:
Made a couple of changes

1. Testing this method out with the AG dataset made me realize that there were some exploding memory problems.  Those have been fixed.
2. After looking at the original mrm paper more closely, I realized that the original method uses a permutation of labels instead of a permutation of residuals.  And it seems to work better too.

Since this method is using permutation of labels, I may try to optimize if with my own permutation algorithm (this'll be after this PR is finished of course).

scikitten-bio:
## Automated report
[http://powertrip.mggen.nau.edu:5000/runs/mortonjt/regression/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/mortonjt/regression/html/index.html)

scikitten-bio:
## Automated report
[http://powertrip.mggen.nau.edu:5000/runs/mortonjt/regression/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/mortonjt/regression/html/index.html)

mortonjt:
I'm going to close this PR - since I don't have time to continue this.  If anyone else wants to complete this, I'll be happy to review.

PR #1107, opened 2015-09-02
Improve error for sequence with lowercase characters

### LINKED ISSUE #1041: Better error in the situation when a sequence has valid, but lowercase characters
Currently it just complains that the letter isn't in the alphabet, but in practice there is a good chance that it is just lowercase. We should augment the error when this is the case so that it informs the user about the `lowercase` keyword. That way, they have an actionable path forward.

colinbrislawn:
Fix #1041


ebolyen:
Thanks @colinbrislawn!

PR #1109, opened 2015-09-03
BUG: Changed ``__nonzero__`` to ``__bool__`` and aliased.

### LINKED ISSUE #1106: Switch `__nonzero__` to `__bool__` in Sequence and then alias `__nonzero__` to `__bool__`
We should be preferring the Python 3 way of doing truthiness. The missing alias isn't a "problem" for Python 3 because it will check `__len__` which happens to work for `Sequence`, but it would be better to not be accidentally successful. 

```python
def __nonzero__(self):
   ...
```
becomes:
```python
def __bool__(self):
   ...

__nonzero__ = __bool__
```

Kleptobismol:
Fixes #1106

### DIFF
--- a/skbio/sequence/_sequence.py
+++ b/skbio/sequence/_sequence.py
@@ -959,7 +959,7 @@ class Sequence(collections.Sequence, SkbioObject):
         return self._bytes.size
 
     @stable(as_of="0.4.0")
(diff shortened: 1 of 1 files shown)


jairideout:
Thanks @Kleptobismol!

PR #1116, opened 2015-09-08
The Alignment constructor does not take a validate parameter.

dansondergaard:
This corrects a small documentation error.


jairideout:
Thanks @dansondergaard!

PR #1120, opened 2015-09-10
ENH: add TabularMSA.sort, from_dict, and to_dict

### LINKED ISSUE #1082: `TabularMSA`: from_dict/to_dict

- `from_dict(dict)`

   Return a new TabularMSA from the keys and sequences of a dict. The values
   must all be an appropriate `IUPACSequence` class of the same length.

- `to_dict() -> dict`

   Returns a dictionary that maps the keys of the MSA to the aligned sequences.

### LINKED ISSUE #1095: `TabularMSA`: sort

- `sort(key=None, reverse=False) -> None`

   Sort sequences in-place. If `key=None` the existing keys will be used to
   sort. Otherwise `key` matches the constructor. If there are no keys for the
   MSA and `key=None`, an exception should be raised.

jairideout:
Also made ``TabularMSA.keys`` use ``object`` dtype to prevent lossy conversions.

Fixes #1082, fixes #1095.

### DIFF
--- a/skbio/alignment/_tabular_msa.py
+++ b/skbio/alignment/_tabular_msa.py
@@ -690,14 +690,14 @@ class TabularMSA(SkbioObject):
 
         """
         if key is None:
-            sort_keys = self.keys
+            sort_keys = self.keys.tolist()
         else:
             sort_keys = [resolve_key(seq, key) for seq in self._seqs]
 
         if len(self) > 0:
             if self.has_keys():
                 _, sorted_seqs, sorted_keys = self._sort_by_first_element(
-                    [sort_keys, self._seqs, self.keys], reverse)
+                    [sort_keys, self._seqs, self.keys.tolist()], reverse)
                 self.keys = sorted_keys
             else:
                 _, sorted_seqs = self._sort_by_first_element(
--- a/skbio/alignment/_tabular_msa.py
+++ b/skbio/alignment/_tabular_msa.py
@@ -622,8 +622,7 @@ class TabularMSA(SkbioObject):
     def sort(self, key=None, reverse=False):
         """Sort sequences in-place.
 
-        Performs a stable sort of the sequences in-place using Python's
-        ``sorted`` function.
+        Performs a stable sort of the sequences in-place.
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -493,6 +525,350 @@ def test_reindex_non_hashable_keys(self):
...
+        # original state is maintained
+        self.assertEqual(msa, TabularMSA([]))
+
+        msa = TabularMSA([DNA('TC'), DNA('AA')])
+        with self.assertRaises(OperationError):
+            msa.sort()
+        self.assertEqual(msa, TabularMSA([DNA('TC'), DNA('AA')]))
+
+    def test_sort_on_unorderable_msa_keys(self):
+        # TODO is there a better way to handle this?
```
jairideout: Couldn't figure out a better way to get this test to pass on Python 2 and 3. @ebolyen any ideas?

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -493,6 +525,350 @@ def test_reindex_non_hashable_keys(self):
...
+        if future.utils.PY3:
+            msa = TabularMSA([DNA('AAA'), DNA('ACG')], keys=[42, None])
+            with self.assertRaises(TypeError):
+                msa.sort()
+            self.assertEqual(
+                msa,
+                TabularMSA([DNA('AAA'), DNA('ACG')], keys=[42, None]))
+
+    def test_sort_on_unorderable_key(self):
+        # TODO is there a better way to handle this?
```
jairideout: Same goes for this one.

skbio/alignment/_tabular_msa.py
```diff
@@ -564,9 +608,140 @@ def reindex(self, key=None, keys=None):
...
+        Sort the sequences using the MSA's existing keys:
+
+        >>> msa.sort()
+        >>> msa == TabularMSA([DNA('AC-'), DNA('AC-'), DNA('ACG')],
+        ...                   keys=['a', 'b', 'c'])
+        True
+
+        """
+        if key is None:
+            sort_keys = self.keys
```
ebolyen: Can you call `tolist()` on `self.keys`?

skbio/alignment/tests/test_tabular_msa.py
ebolyen: Try using `object()` it shouldn't be orderable

skbio/alignment/_tabular_msa.py
jairideout: Yep, fixed here and below

skbio/alignment/tests/test_tabular_msa.py
jairideout: `object` is orderable in py2. The following dummy class seems to work though:

```python
class Unorderable(object):
    def __lt__(self, other):
        raise TypeError()
    __cmp__ = __lt__
```

jairideout:

ebolyen:
Thanks @jairideout!

PR #1121, opened 2015-09-11
Implement TabularMSA.append

### LINKED ISSUE #1098: `TabularMSA`: append

- `append(IUPACSequence) -> None`

   Add a sequence to the end of the alignment. Must be the same `dtype` and
   length must match second dimension. It should be documented that this
   operation, if successful, is not necessarily meaningful.

anderspitman:
Closes #1098


anderspitman:
Still need to do documentation. Just want to make sure implementation looks ok.

scikitten-bio:
## Automated report
[http://powertrip.mggen.nau.edu:5000/runs/anderspitman/tabular_msa_append/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/anderspitman/tabular_msa_append/html/index.html)

anderspitman:
Ready for review once tests pass.

scikitten-bio:
## Automated report
[http://powertrip.mggen.nau.edu:5000/runs/anderspitman/tabular_msa_append/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/anderspitman/tabular_msa_append/html/index.html)

jairideout:
Can you handle the case where there are no sequences in the MSA? The constructor can probably be modified to just call `append` since its logic/validation needs to happen in `append` too.

This wasn't discussed in the RFC, but how should we handle adding sequences to an MSA that has keys? Perhaps this method could accept an optional `key` parameter that is required if the MSA has keys already and will raise an error if provided and the MSA doesn't have keys.

cc: @ebolyen

skbio/alignment/_tabular_msa.py
```diff
@@ -570,3 +570,48 @@ def reindex(self, key=None, keys=None):
...
+        ------
+        TypeError
+            If the type of the sequence does not match the dtype of the MSA.
+        ValueError
+            If the length of the sequence does not match the length of the
+            second dimension of the MSA.
+
+        Notes
+        -----
+        This operation, even if successful, is not necessarily meaningful.
```
jairideout: Can you explain why it is not meaningful? We were wanting to make it clear to users that the sequences are not re-aligned to account for the new sequence. This is simply adding a sequence to the MSA that matches its # of positions and sequence type.

anderspitman:
I believe we're already handling the case where the MSA doesn't have an sequences, since `_seqs` is initialized to `[]`. Is that what you meant? However, I like the idea of changing the constructor to use append. So maybe the dtype for the MSA is set to the first sequence in the iterable, then append will raise an exception if any of the subsequent sequences don't match the first one?

skbio/alignment/_tabular_msa.py
anderspitman: Will do. I figured that was the reason but wasn't sure.

jairideout:
The current code will always raise an error when trying to add a sequence to an MSA without sequences. I think all of the logic in the constructor can be moved into `append` because we'll need to handle this case in both situations.

anderspitman:
Ah I see because there's no dtype. Got it

anderspitman:
@jairideout how's that look?

anderspitman:
Obviously still need to update the docstring but I want to nail down the implementation first. May also be able to remove some of the constructor tests that are covered by append now.

scikitten-bio:
## Automated report
[http://powertrip.mggen.nau.edu:5000/runs/anderspitman/tabular_msa_append/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/anderspitman/tabular_msa_append/html/index.html)

jairideout:
This looks great! I'll do a more detailed review shortly. I think we should leave the constructor tests as-is in case `append`/`extend` aren't used internally in the future. There will be some duplication of tests for the constructor and `append` but I don't think this is a big deal.

anderspitman:
Cool. I'll update the docs. Also, I like your idea for handling appending of keys. It seems consistent with the way keys are used currently. I'll draft an implementation real quick as well.

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -493,6 +493,43 @@ def test_reindex_non_hashable_keys(self):
...
+        # msa[-1]
+        seq = None
+        for seq in msa:
+            pass
+        self.assertEqual(seq, DNA('AAAA'))
+
+    def test_append_to_empty_msa(self):
+        msa = TabularMSA([])
+        msa.append(DNA('ACGT'))
+        self.assertEqual(len(msa), 1)
```
ebolyen: Can you check the `shape` as well?

jairideout: These tests should use `self.assertEqual(msa1, msa2)` to compare `TabularMSA` objects (in order to check shape and other properties).

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -493,6 +493,43 @@ def test_reindex_non_hashable_keys(self):
 
         npt.assert_array_equal(msa.keys, keys)
 
+    def test_append(self):
+        msa = TabularMSA([DNA('ACGT'), DNA('TGCA')])
+        msa.append(DNA('AAAA'))
+        self.assertEqual(msa.shape, (3, 4))
+        # TODO: Hack to get last seq. once __getitem__ is implemented use
```
jairideout: You can use `self.assertEqual` to compare `msa` to an expected `TabularMSA` (and the shape assertion above can be removed too!).

skbio/alignment/_tabular_msa.py
```diff
@@ -570,3 +549,61 @@ def reindex(self, key=None, keys=None):
...
+        if msa_is_empty:
+            dtype = type(sequence)
+            if not issubclass(dtype, IUPACSequence):
+                raise TypeError(
+                    "`sequences` must contain scikit-bio sequence objects "
+                    "that have an alphabet, not type %r" % dtype.__name__)
+            self._dtype = dtype
+            self._shape = _Shape(sequence=1, position=len(sequence))
+            self._seqs = [sequence]
+        elif type(sequence) != self.dtype:
```
jairideout: Can you change this back to:

```python
if type(sequence) is not dtype:
```

This way of checking is faster.

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -493,6 +493,43 @@ def test_reindex_non_hashable_keys(self):
...
+            pass
+        self.assertEqual(seq, DNA('AAAA'))
+
+    def test_append_to_empty_msa(self):
+        msa = TabularMSA([])
+        msa.append(DNA('ACGT'))
+        self.assertEqual(len(msa), 1)
+
+
+    def test_append_to_empty_msa_invalid_dtype(self):
```
jairideout: For each of these error-checking tests, can you assert that the object's state isn't altered? This will catch the case where `self._seqs` is appended to before an error is raised, which would put the object into an inconsistent state. This is important to avoid, for example, if the user catches the error and continues to interact with the MSA. There are some examples in the other `TabularMSA` unit tests for how we're checking this.

skbio/alignment/_tabular_msa.py
```diff
@@ -570,3 +549,61 @@ def reindex(self, key=None, keys=None):
...
+        >>> msa = TabularMSA([DNA('ACG'), DNA('AC-')])
+        >>> len(msa)
+        2
+        >>> msa.append(DNA('AAA'))
+        >>> len(msa)
+        3
+
+        """
+        msa_is_empty = len(self) == 0
+        if msa_is_empty:
```
jairideout: Minor, but any reason to not just do this?

```python
if not len(self):
```

skbio/alignment/_tabular_msa.py
```diff
@@ -570,3 +549,61 @@ def reindex(self, key=None, keys=None):
...
 
         self._keys = keys_
+
+    @experimental(as_of='0.4.0-dev')
+    def append(self, sequence):
+        """Append a sequence to the MSA.
+
+        Parameters
+        ----------
+        sequence : IUPACSequence
```
jairideout: Can you change the type to be `alphabet-aware scikit-bio sequence object`? We're about to change the inheritance hierarchy a bit to accommodate custom sequence classes that aren't IUPAC.

skbio/alignment/_tabular_msa.py
```diff
@@ -570,3 +549,61 @@ def reindex(self, key=None, keys=None):
...
+            Sequence to be appended. Must match the dtype of the MSA and the
+            length of the second dimension of the MSA.
+
+        Raises
+        ------
+        TypeError
+            If the type of the sequence does not match the dtype of the MSA.
+        ValueError
+            If the length of the sequence does not match the length of the
```
jairideout: Something like this may be easier for users to understand:

"If the length of the sequence does not match the number of positions in the MSA."

skbio/alignment/_tabular_msa.py
```diff
@@ -570,3 +549,61 @@ def reindex(self, key=None, keys=None):
...
+        >>> msa.append(DNA('AAA'))
+        >>> len(msa)
+        3
+
+        """
+        msa_is_empty = len(self) == 0
+        if msa_is_empty:
+            dtype = type(sequence)
+            if not issubclass(dtype, IUPACSequence):
+                raise TypeError(
```
jairideout: Can you document this error in `Raises`?

skbio/alignment/_tabular_msa.py
```diff
@@ -570,3 +549,61 @@ def reindex(self, key=None, keys=None):
...
+            self._shape = _Shape(sequence=1, position=len(sequence))
+            self._seqs = [sequence]
+        elif type(sequence) != self.dtype:
+            raise TypeError(
+                "`sequences` cannot contain mixed types. Type %r does "
+                "not match type %r" %
+                (type(sequence).__name__, self.dtype.__name__))
+        elif len(sequence) != self.shape.position:
+            raise ValueError(
+                "`sequences` must contain sequences of the same "
```
jairideout: Can you please update the error messages in this method so that they make sense for `append`? They're still phrased as they were in the constructor, where we were passed an iterable of `sequences`.

ebolyen:
@jairideout RE: keys.

Here is an issue #1125  that describes our solution for handling keys in methods that modify the collection.

Basically we can keep track of the `key` used in the constructor, and reuse it when the user fails to supply a key. This prevents them from repeating themselves all of the time and has some other useful properties in the context of alignment functions.

@anderspitman the suggested changes in #1125 should only require some very trivial modifications to `reindex`, a small method, and a couple of new test cases. Would you be up to including them so that `append` has convenient behaviour in the presence of keys?

anderspitman:
@ebolyen yep sounds good.

skbio/alignment/tests/test_tabular_msa.py
anderspitman: @jairideout yeah much better way of doing it!

anderspitman:
@jairideout I believe I've addressed everything discussed up to this point. Ready for review once tests pass.

scikitten-bio:
## Automated report
[http://powertrip.mggen.nau.edu:5000/runs/anderspitman/tabular_msa_append/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/anderspitman/tabular_msa_append/html/index.html)

jairideout:
Thanks @anderspitman! I just followed up on #1125 which will affect the implementation here so I'll hold off on further review for now.

skbio/alignment/_tabular_msa.py
Jorge-C: Is `type(obj)` guaranteed to return a singleton? (ie, are types
singletons?) Otherwise, I don't know if `type(sequence) is not dtype:` is
valid, even if it seems to work (`1 is 1` also works, but is invalid). The
only objects I can find on Python's docs that are guaranteed to be
singletons are `None`, `NotImplemented`, `Ellipsis`, `True` and `False`.

2015-09-11 23:21 GMT+02:00 Jai Ram Rideout <notifications@github.com>:

anderspitman: Good question.

ebolyen: While classes certainly aren't value types (like ints, to a point), Python does use nominal typing. I don't see how nominal typing can work if the same classobject can exist at more than one memory address.

Worth looking into further...

ebolyen: Here is an SO post [indicating](http://stackoverflow.com/a/152596/579416) that this behavior will work. (Followed by the requisite obnoxious comments to never check types in Python, of course)

jairideout: [This post](http://bugs.python.org/issue13387) on the Python issue tracker also indicates this is the correct way to check for an exact type. See also the docs for `assertIsInstance` [here](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertNotIsInstance), which give an example of exact type checking (`assertIs(type(obj), cls)`).

scikitten-bio:
## Automated report
[http://powertrip.mggen.nau.edu:5000/runs/anderspitman/tabular_msa_append/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/anderspitman/tabular_msa_append/html/index.html)

skbio/alignment/_tabular_msa.py
wasade: Would `isinstance` work here? e.g., `elif not isinstance(sequence,
self.dtype)`

skbio/alignment/_tabular_msa.py
```diff
@@ -717,6 +707,92 @@ def reindex(self, key=None, keys=None):
...
+        if msa_is_empty:
+            dtype = type(sequence)
+            if not issubclass(dtype, IUPACSequence):
+                raise TypeError(
+                    "`sequence` must be a scikit-bio sequence object "
+                    "that has an alphabet, not type %r" % dtype.__name__)
+            self._dtype = dtype
+            self._shape = _Shape(sequence=1, position=len(sequence))
+            self._seqs = [sequence]
+        elif type(sequence) is not self.dtype:
```
wasade: `type(sequence)` is issued in two of the four branches, with 3 total calls. This is dropped into a variable on like 773 but that is only scoped for one of the branches. Possible to just do this call once prior to the `if`?

scikitten-bio:
## Automated report
[http://powertrip.mggen.nau.edu:5000/runs/anderspitman/tabular_msa_append/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/anderspitman/tabular_msa_append/html/index.html)

skbio/alignment/_tabular_msa.py
anderspitman: Not sure. The else branch is the normal case, and it doesn't require the call. I didn't write this. This was modified from code originally in `__init__` so @jairideout or @ebolyen might have more insight.

anderspitman: Modified my previous comment. I changed the original code more than I had remembered, so there's a good chance it's not as optimal as before. Probably would be a good idea to get a second opinion.

anderspitman:
@jairideout I believe this is ready for another review.

scikitten-bio:
## Automated report
[http://powertrip.mggen.nau.edu:5000/runs/anderspitman/tabular_msa_append/html/index.html](http://powertrip.mggen.nau.edu:5000/runs/anderspitman/tabular_msa_append/html/index.html)

anderspitman:
@jairideout @ebolyen any word on reviewing this?

jairideout:
Sorry for the delay, one of us should be available to review tomorrow

skbio/alignment/_tabular_msa.py
jairideout: That would work but will be slower and we want to do exact type-matching here (for example, to avoid the case where the `dtype` is `Sequence` and the user tries adding `DNA`, `RNA` or `Protein` to the collection -- we don't want to allow mixed types in `TabularMSA`).

skbio/alignment/_tabular_msa.py
```diff
@@ -37,8 +37,8 @@ class TabularMSA(SkbioObject):
     metadata : dict, optional
         Arbitrary metadata which applies to the entire MSA. A shallow copy of
         the ``dict`` will be made.
-    key : callable or metadata key, optional
-        If provided, defines a unique, hashable key for each sequence in
+    minter : callable or metadata key, optional
+        If provided, defines a unique, hashable key minter for each sequence in
```
jairideout: This wording here sounds like `minter` needs to be unique and hashable. Can you reword to indicate the key minter provides a unique, hashable key for each sequence?

skbio/alignment/_tabular_msa.py
```diff
@@ -201,6 +201,16 @@ def keys(self):
 
     @property
     @experimental(as_of='0.4.0-dev')
+    def minter(self):
```
jairideout: docstring plz

skbio/alignment/_tabular_msa.py
```diff
@@ -201,6 +201,16 @@ def keys(self):
...
+        if self._minter is not None:
+            return self._minter
+        else:
+            raise OperationError(
+                "Attempted use of key minter on MSA, but none exists. Use "
```
jairideout: Minor: the user isn't really *using* the minter on the MSA, just requesting it. Maybe reword to something like "Minter does not exist. Use `reindex` with the `minter` parameter to set one."

skbio/alignment/_tabular_msa.py
```diff
@@ -308,44 +318,23 @@ def from_dict(cls, dictionary):
...
-        self._seqs = seqs
-        self._dtype = dtype
-        self._shape = _Shape(sequence=len(seqs), position=length)
+
+        self._seqs = []
+        self._dtype = None
+        self._shape = _Shape(sequence=0, position=0)
+        self._minter = None
+
+        for seq in sequences:
```
jairideout: Can you move the `iter` call here? It'll make the code clearer now that it's been refactored.

```python
for seq in iter(sequences):
    ...
```

skbio/alignment/_tabular_msa.py
```diff
@@ -619,15 +607,16 @@ def has_metadata(self):
...
         """Reassign keys to sequences in the MSA.
 
         Parameters
         ----------
-        key : callable or metadata key, optional
-            If provided, defines a unique, hashable key for each sequence in
-            the MSA. Can either be a callable accepting a single argument (each
-            sequence) or a key into each sequence's ``metadata`` attribute.
+        minter : callable or metadata key, optional
+            If provided, defines a unique, hashable key minter for each
```
jairideout: same comment about rewording as in constructor docstring

skbio/alignment/_tabular_msa.py
```diff
@@ -685,13 +674,14 @@ def reindex(self, key=None, keys=None):
...
-        if key is not None and keys is not None:
+        if minter is not None and keys is not None:
             raise ValueError(
-                "Cannot use both `key` and `keys` at the same time.")
+                "Cannot use both `minter` and `keys` at the same time.")
 
         keys_ = None
-        if key is not None:
-            keys_ = [resolve_key(seq, key) for seq in self._seqs]
+        if minter is not None:
```
jairideout: There are a couple of border cases to consider:

- a `TabularMSA` with a minter, calling `reindex` with `keys` should set new keys and remove the minter
- a `TabularMSA` with a minter, calling `reindex` without passing `keys` or `minter` should remove the keys *and* the minter

Can you modify the logic here and add unit tests for these cases?

skbio/alignment/_tabular_msa.py
```diff
@@ -685,13 +674,14 @@ def reindex(self, key=None, keys=None):
...
+            self._minter = minter
```
jairideout: Can you refactor to avoid modifying object state until after we're sure no errors will be thrown? The line below could fail if the user provides an invalid minter.

skbio/alignment/_tabular_msa.py
jairideout: Also can you add a `has_minter` method? Sorry this wasn't super clear in #1125.

skbio/alignment/_tabular_msa.py
```diff
@@ -717,6 +707,92 @@ def reindex(self, key=None, keys=None):
...
+        if minter is None:
+            if self.has_keys():
+                minter = self.minter
+        else:
+            if not self.has_keys():
+                raise OperationError(
+                    "key was provided but MSA does not have keys.")
+
+        self._add_sequence(sequence)
+        self.reindex(minter=minter)
```
jairideout: This works but is inefficient because you're running the minter on all sequences each time a sequence is added. Is it safe to only run the minter on the sequence to be appended?

skbio/alignment/_tabular_msa.py
jairideout: :+1: for adding `dtype = type(sequence)` and using that throughout

skbio/alignment/_tabular_msa.py
```diff
@@ -717,6 +707,92 @@ def reindex(self, key=None, keys=None):
...
+                "`sequence` Must match the type of any other sequences "
+                "already in the MSA. Type %r does not match type %r" %
+                (type(sequence).__name__, self.dtype.__name__))
+        elif len(sequence) != self.shape.position:
+            raise ValueError(
+                "`sequence` length must match the length of any other "
+                "sequences already in the MSA: %r != %r"
+                % (len(sequence), self.shape.position))
+        else:
+            self._shape = _Shape(sequence=self._shape.sequence + 1,
```
jairideout: Setting the shape is happening in three places. Can you modify the `shape` property to construct a `Shape` object based on the number of rows and positions in the first sequence? That way this book-keeping won't be necessary.

skbio/alignment/_tabular_msa.py
```diff
@@ -717,6 +707,92 @@ def reindex(self, key=None, keys=None):
...
+            self._shape = _Shape(sequence=1, position=len(sequence))
+            self._seqs = [sequence]
+        elif type(sequence) is not self.dtype:
+            raise TypeError(
+                "`sequence` Must match the type of any other sequences "
+                "already in the MSA. Type %r does not match type %r" %
+                (type(sequence).__name__, self.dtype.__name__))
+        elif len(sequence) != self.shape.position:
+            raise ValueError(
+                "`sequence` length must match the length of any other "
```
jairideout: *`sequence` length must match the number of positions in the MSA: %d != %d*

skbio/alignment/_tabular_msa.py
```diff
@@ -717,6 +707,92 @@ def reindex(self, key=None, keys=None):
...
 
     @experimental(as_of='0.4.0-dev')
+    def append(self, sequence, minter=None):
+        """Append a sequence to the MSA.
+
+        Parameters
+        ----------
+        sequence : alphabet-aware scikit-bio sequence object
+            Sequence to be appended. Must match the dtype of the MSA and the
+            length of the second dimension of the MSA.
```
jairideout: *and the number of positions in the MSA*

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -342,11 +345,34 @@ def test_keys_deleter_no_keys(self):
         self.assertFalse(msa.has_keys())
 
     def test_keys_deleter_with_keys(self):
-        msa = TabularMSA([RNA('UUU'), RNA('AAA')], key=str)
+        msa = TabularMSA([RNA('UUU'), RNA('AAA')], minter=str)
         self.assertTrue(msa.has_keys())
         del msa.keys
         self.assertFalse(msa.has_keys())
 
+    def test_get_cached_minter_str(self):
```
jairideout: Can you name these three methods `test_minter_*` or `test_minter_getter_*` to match the naming scheme of the other property tests?

jairideout: Another interesting test to add would be testing that an `msa` initialized with `keys` raises an error when `minter` is accessed.

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -1015,5 +1042,82 @@ def test_from_dict_to_dict_roundtrip(self):
         self.assertIs(d1[42], d2[42])
 
 
+class TestAppend(unittest.TestCase):
```
jairideout: I like this style of testing: a test class for each method. It doesn't match the style used in the rest of the tests though -- @ebolyen what do you think about mixing styles of testing here? I'm on the fence.

jairideout:
@anderspitman I'm done reviewing, in general this looks great! In addition to the inline comments, can you address the comment in bullet point 4 [here](https://github.com/biocore/scikit-bio/issues/1125#issuecomment-140161467)?

anderspitman:
@jairideout starting on this now. I'll provide updates as there's progress.

skbio/alignment/_tabular_msa.py
anderspitman: @jairideout is this going to supercede `has_keys`? They seem to pretty much mean the same thing.

skbio/alignment/_tabular_msa.py
anderspitman: Done, but what is the purpose of using `iter` here? We're not storing the iterator or passing it anywhere, won't `sequences` always be iterable? The tests pass with or without it.

skbio/alignment/_tabular_msa.py
jairideout: `has_keys` tells whether or not there are keys on the MSA, regardless of whether there is a minter. Keys can be set either via `keys` or a minter in the constructor and `reindex`. `has_minter` tells whether or not there is a minter on the MSA.

skbio/alignment/_tabular_msa.py
jairideout: Good catch, `iter` was previously necessary because we were calling `next` but now we're only looping. This can be removed.

skbio/alignment/tests/test_tabular_msa.py
ebolyen: I'm fine with mixing. We'll have more test mixins eventually, so we are going to end up with a mixed style regardless.

skbio/alignment/_tabular_msa.py
anderspitman: I've implemented this. One side affect of the second bullet is that deleting the keys also deletes the minter. This makes sense, right? Also see [my comment here](https://github.com/biocore/scikit-bio/pull/1121#discussion_r41322924)

skbio/alignment/_tabular_msa.py
anderspitman: :+1:

skbio/alignment/_tabular_msa.py
anderspitman: We'd still need to check for duplicates. I could refactor the duplication check code in `reindex` into a separate method and call it in both places. That would save us calling `resolve_key` for all the sequences. This feels like premature optimization to me, though. Since this would increase complexity and the chance for introducing errors I would recommend not doing it until we have a measured use case telling us it's too slow.

skbio/alignment/_tabular_msa.py
anderspitman: I like this, but I think it requires changing `__len__` from:

`return self.shape.position`

to:

`return len(self._seqs)`

Is that ok?

skbio/alignment/tests/test_tabular_msa.py
jairideout: Okay, thanks

skbio/alignment/_tabular_msa.py
jairideout: Yeah, reindexing without keys or a minter should remove the keys and the minter. Do you have a question about the comment you linked to? I previously followed up on that one.

skbio/alignment/_tabular_msa.py
jairideout: In addition to performance gains, I think refactoring makes sense because we need to handle the case where the sequence has been successfully appended but the key/minting process fails. In this case, we'd want the MSA's state to remain unmodified so that it isn't left in an inconsistent state (i.e., having a sequence appended without adding a new key).

skbio/alignment/_tabular_msa.py
jairideout: That's fine

jairideout:
@anderspitman see my comment [here](https://github.com/biocore/scikit-bio/issues/1125#issuecomment-146356126) regarding storing the minter. Are you up for making those changes in this PR? They should mostly be done, I think it's just some code deletion.

skbio/alignment/_tabular_msa.py
```diff
@@ -715,8 +792,92 @@ def reindex(self, key=None, keys=None):
             keys_.flags.writeable = False
 
         self._keys = keys_
+        self._minter = minter
 
     @experimental(as_of='0.4.0-dev')
+    def append(self, sequence, minter=None):
```
jairideout: Given my latest comment [here](https://github.com/biocore/scikit-bio/issues/1125#issuecomment-146356126), this method should now accept a minter to apply to `sequence` and a `key` parameter specifying a key for `sequence`. These options should be mutually exclusive. If keys don't exist on the MSA, an error should be thrown if `minter` or `key` is provided.

Much simpler behavior indeed! :shipit:

anderspitman:
@jairideout @ebolyen I think this makes sense. Just to make sure I understand the scope of the changes in the context of the current state of this PR: I'm going to remove the `minter` property and any code having to do with it. The minter functionality will remain but will be stateless. ie if you call `append` and you want the sequence to have a key you either provide a key or minter with the call. Sound correct?

ebolyen:
@anderspitman correct!

anderspitman:
@ebolyen is it necessary to handle a `minter` kwarg in `append`? Since it would now only apply to the sequence being appended, wouldn't it make more sense to have the caller compute the key and pass it in?

skbio/alignment/_tabular_msa.py
anderspitman: And I'm assuming if keys do exist, and no minter or key is provided to append, then that would also be an error? ie we can never have an MSA where some sequences have keys and others don't?

anderspitman: Please see my comment [here](https://github.com/biocore/scikit-bio/issues/1125#issuecomment-146705316).

jairideout: Yep that's right: a minter/key is required if there are keys on the MSA. I just followed up on #1125.

jairideout:
@anderspitman we want to support a `minter` and `key` to provide a consistent interface with the constructor and `reindex`.

skbio/alignment/_tabular_msa.py
```diff
@@ -716,8 +730,119 @@ def reindex(self, key=None, keys=None):
...
+        True
+
+        >>> msa = TabularMSA([DNA('', metadata={'id': 'a'})], minter='id')
+        >>> msa.append(DNA('', metadata={'id': 'b'}), minter='id')
+        >>> msa == TabularMSA([DNA('', metadata={'id': 'a'}),
+        ...                    DNA('', metadata={'id': 'b'})], minter='id')
+        True
+        """
+
+        if key is not None and minter is not None:
```
anderspitman: If anyone can think of a way to make this block simpler that would be great. If python supported anonymous functions I would be tempted to implement this as a decision table. The current form is already reduced by one level and I still don't like it.

anderspitman:
@ebolyen if you're able to review this that would be great. I have other things to work on so don't worry if you don't have the bandwidth.

jairideout:
@anderspitman I'll take a pass through now

skbio/alignment/_tabular_msa.py
```diff
@@ -309,44 +315,20 @@ def from_dict(cls, dictionary):
...
-                        "length: %r != %r" % (len(seq), length))
-                seqs.append(seq)
-
-        self._seqs = seqs
-        self._dtype = dtype
-        self._shape = _Shape(sequence=len(seqs), position=length)
+    def __init__(self, sequences, metadata=None, minter=None, keys=None):
+        self._seqs = []
+        self._dtype = None
+        self._minter = None
```
jairideout: remove

skbio/alignment/_tabular_msa.py
```diff
@@ -590,6 +579,34 @@ def has_keys(self):
         return self._keys is not None
 
     @experimental(as_of='0.4.0-dev')
+    def has_minter(self):
```
jairideout: this method can be removed, we're no longer storing the minter

skbio/alignment/_tabular_msa.py
```diff
@@ -716,8 +730,119 @@ def reindex(self, key=None, keys=None):
             keys_.flags.writeable = False
 
         self._keys = keys_
+        self._minter = minter
```
jairideout: remove

skbio/alignment/_tabular_msa.py
```diff
@@ -123,7 +124,12 @@ def shape(self):
         3
 
         """
-        return self._shape
+        len_sequence = len(self)
```
jairideout: Minor comment: I was confused by the names `len_sequence` as it sounds like it's the length of the sequence (it's actually the number of sequences in the MSA). Possible to rename these local vars to something like `num_sequences` and `num_positions`? or `sequence_count` and `position_count`

skbio/alignment/_tabular_msa.py
```diff
@@ -477,6 +458,14 @@ def __eq__(self, other):
         bool
             Indicates whether this MSA is equal to `other`.
 
+        Notes
```
jairideout: This section of text can be removed now that we're not storing the minter

skbio/alignment/_tabular_msa.py
```diff
@@ -716,8 +730,119 @@ def reindex(self, key=None, keys=None):
             keys_.flags.writeable = False
 
         self._keys = keys_
+        self._minter = minter
+
+    def _fail_if_duplicate_keys(self, keys):
```
jairideout: Can you rename this to `_assert_valid_keys`? This will match the style we've used elsewhere (e.g., `Sequence._assert_can_cast_to`). The name is also more general because we're validating uniqueness *and* hashability (and in the future this method could place additional constraints on keys).

skbio/alignment/_tabular_msa.py
```diff
@@ -716,8 +730,119 @@ def reindex(self, key=None, keys=None):
...
+        # Hashability of keys is implicitly checked here.
+        duplicates = find_duplicates(keys)
+        if duplicates:
+            raise UniqueError(
+                "Keys must be unique. Duplicate keys: %r" % duplicates)
 
     @experimental(as_of='0.4.0-dev')
+    def append(self, sequence, key=None, minter=None):
```
jairideout: Can you swap the order of `key` and `minter` in the signature? (for consistency with `reindex` and constructor)

Also `key` is missing its parameter docstring

skbio/alignment/_tabular_msa.py
```diff
@@ -716,8 +730,119 @@ def reindex(self, key=None, keys=None):
...
+        """Append a sequence to the MSA.
+
+        Parameters
+        ----------
+        sequence : alphabet-aware scikit-bio sequence object
+            Sequence to be appended. Must match the dtype of the MSA and the
+            number of positions in the MSA.
+        minter : callable or metadata key, optional
```
jairideout: `minter` description is out of date (still written for when we were storing the minter)

skbio/alignment/_tabular_msa.py
```diff
@@ -716,8 +730,119 @@ def reindex(self, key=None, keys=None):
...
+        ------
+        TypeError
+            If the sequence object is a type that doesn't have an alphabet
+        TypeError
+            If the type of the sequence does not match the dtype of the MSA.
+        ValueError
+            If the length of the sequence does not match the number of
+            positions in the MSA.
+        OperationError
+            If no key is provided, but the MSA has keys, and no cached key
```
jairideout: remove reference to cached keys

skbio/alignment/_tabular_msa.py
```diff
@@ -716,8 +730,119 @@ def reindex(self, key=None, keys=None):
...
+
+        >>> msa = TabularMSA([DNA('', metadata={'id': 'a'})], minter='id')
+        >>> msa.append(DNA('', metadata={'id': 'b'}), minter='id')
+        >>> msa == TabularMSA([DNA('', metadata={'id': 'a'}),
+        ...                    DNA('', metadata={'id': 'b'})], minter='id')
+        True
+        """
+
+        if key is not None and minter is not None:
+            raise OperationError("Cannot provide key and minter.")
```
jairideout: rewording for clarity and consistency with other error messages in TabularMSA:

```
"Cannot use both `minter` and `key` at the same time."
```

skbio/alignment/_tabular_msa.py
```diff
@@ -716,8 +730,119 @@ def reindex(self, key=None, keys=None):
...
+
+        if self.has_keys():
+            if key is None and minter is None:
+                raise OperationError(
+                    "MSA has keys but no key or minter was privided.")
```
jairideout: privided -> provided

skbio/alignment/_tabular_msa.py
```diff
@@ -716,8 +730,119 @@ def reindex(self, key=None, keys=None):
...
+            elif key is not None:
+                new_key = key
+            elif minter is not None:
+                new_key = resolve_key(sequence, minter)
+            else:
+                raise Exception("Explicitly handle impossible branch")
```
jairideout: Can you refactor this to remove the impossible branch? This will drop coverage.

skbio/alignment/_tabular_msa.py
jairideout: Not sure if this will help simplify the code, but can you handle the case where the key has been successfully added but appending the sequence fails? In this case we'd want the MSA's state to remain consistent -- both the key *and* the sequence must succeed otherwise no changes are made to MSA state. Can you also add a unit test for this behavior?

skbio/alignment/_tabular_msa.py
```diff
@@ -716,8 +730,119 @@ def reindex(self, key=None, keys=None):
...
+            self._dtype = dtype
+            self._seqs = [sequence]
+        elif dtype is not self.dtype:
+            raise TypeError(
+                "`sequence` Must match the type of any other sequences "
+                "already in the MSA. Type %r does not match type %r" %
+                (dtype.__name__, self.dtype.__name__))
+        elif len(sequence) != self.shape.position:
+            raise ValueError(
+                "`sequence` length must match the length of any other "
```
jairideout: Can you change the error message to what's described [here](https://github.com/biocore/scikit-bio/pull/1121#discussion-diff-41078068)?

skbio/alignment/_tabular_msa.py
```diff
@@ -716,8 +730,119 @@ def reindex(self, key=None, keys=None):
...
+        if msa_is_empty:
+            if not issubclass(dtype, IUPACSequence):
+                raise TypeError(
+                    "`sequence` must be a scikit-bio sequence object "
+                    "that has an alphabet, not type %r" % dtype.__name__)
+            self._dtype = dtype
+            self._seqs = [sequence]
+        elif dtype is not self.dtype:
+            raise TypeError(
+                "`sequence` Must match the type of any other sequences "
```
jairideout: Must -> must

skbio/alignment/_tabular_msa.py
```diff
@@ -716,8 +730,119 @@ def reindex(self, key=None, keys=None):
...
+        elif len(sequence) != self.shape.position:
+            raise ValueError(
+                "`sequence` length must match the length of any other "
+                "sequences already in the MSA: %r != %r"
+                % (len(sequence), self.shape.position))
+        else:
+            self._seqs.append(sequence)
+
+    def _add_key(self, key):
+        keys = list(self._keys)
```
jairideout: Can you use `self.keys`? In general we try to avoid using private state even within the class itself.

skbio/alignment/_tabular_msa.py
```diff
@@ -716,8 +730,119 @@ def reindex(self, key=None, keys=None):
...
+        keys.append(key)
+        self._fail_if_duplicate_keys(keys)
```
jairideout: These two lines can become:

```python
self.keys = keys
```

This will perform additional logic missing in the current implementation, e.g. keys need to be cast to an immutable numpy array of object dtype.

skbio/alignment/_tabular_msa.py
jairideout: There are some additional error cases in this method, can you please document them?

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -342,11 +345,65 @@ def test_keys_deleter_no_keys(self):
         self.assertFalse(msa.has_keys())
 
     def test_keys_deleter_with_keys(self):
-        msa = TabularMSA([RNA('UUU'), RNA('AAA')], key=str)
+        msa = TabularMSA([RNA('UUU'), RNA('AAA')], minter=str)
         self.assertTrue(msa.has_keys())
         del msa.keys
         self.assertFalse(msa.has_keys())
 
+    def test_keys_removed_after_reindex(self):
```
jairideout: This is already tested in `test_reindex_empty` and `test_reindex_non_empty`

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -342,11 +345,65 @@ def test_keys_deleter_no_keys(self):
...
+        msa = TabularMSA([], keys=[])
+        msa.reindex()
+        self.assertFalse(msa.has_keys())
+
+    def test_minter_str(self):
```
jairideout: This is already tested in `test_constructor_non_empty_with_keys`

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -342,11 +345,65 @@ def test_keys_deleter_no_keys(self):
...
+        msa = TabularMSA([DNA('', metadata={'id': 42}),
+                          DNA('', metadata={'id': 43})], minter='id')
+        npt.assert_array_equal(msa.keys, np.array([42, 43]))
+
+    def test_minter_callable(self):
```
jairideout: This is already tested in `test_constructor_non_empty_with_keys`

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -342,11 +345,65 @@ def test_keys_deleter_no_keys(self):
...
+        def minter_func(x):
+            return x.metadata['id']
+        msa = TabularMSA([DNA('', metadata={'id': 42}),
+                          DNA('', metadata={'id': 43})], minter=minter_func)
+        npt.assert_array_equal(msa.keys, np.array([42, 43]))
+
+    def test_minter_removed_after_adding_keys(self):
```
jairideout: remove, this is left over from storing the minter

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -342,11 +345,65 @@ def test_keys_deleter_no_keys(self):
...
+        msa = TabularMSA([], minter=str)
+        msa.reindex(keys=[])
+        self.assertFalse(msa.has_minter())
+        self.assertTrue(msa.has_keys())
+
+    def test_minter_removed_after_reindex(self):
```
jairideout: remove, this is left over from storing the minter

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -342,11 +345,65 @@ def test_keys_deleter_no_keys(self):
...
+        msa = TabularMSA([], minter=str)
+        msa.reindex()
+        self.assertFalse(msa.has_minter())
+
+    def test_minter_invalid_does_not_change_msa(self):
```
jairideout: This is already tested in `reindex` tests

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -342,11 +345,65 @@ def test_keys_deleter_no_keys(self):
...
+        msa = TabularMSA([DNA('', metadata={'id': 'a'}),
+                          DNA('', metadata={'id': 'b'})],
+                         minter='id')
+        with self.assertRaises(KeyError):
+            msa.reindex(minter='invalid')
+        npt.assert_array_equal(msa.keys, np.array(['a', 'b']))
+
+    def test_has_minter_constructor_no_minter(self):
```
jairideout: remove please

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -342,11 +345,65 @@ def test_keys_deleter_no_keys(self):
...
+        msa = TabularMSA([])
+        self.assertFalse(msa.has_minter())
+
+    def test_has_minter_constructor_has_minter(self):
```
jairideout: remove please

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -342,11 +345,65 @@ def test_keys_deleter_no_keys(self):
...
+        msa = TabularMSA([], minter=str)
+        self.assertTrue(msa.has_minter())
+
+    def test_has_minter_reindex_no_minter(self):
```
jairideout: remove please

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -342,11 +345,65 @@ def test_keys_deleter_no_keys(self):
...
+        msa = TabularMSA([])
+        msa.reindex()
+        self.assertFalse(msa.has_minter())
+
+    def test_has_minter_reindex_has_minter(self):
```
jairideout: remove please

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -672,7 +745,7 @@ def test_reindex_non_hashable_keys(self):
 
         npt.assert_array_equal(msa.keys, keys)
 
-    def test_sort_no_msa_keys_and_key_not_specified(self):
+    def test_sort_no_msa_keys_and_minter_not_specified(self):
```
jairideout: revert to previous name, `key` is still a valid parameter for `sort`

skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -1015,5 +1088,101 @@ def test_from_dict_to_dict_roundtrip(self):
...
+        expected = TabularMSA([DNA('ACGT')])
+        msa = TabularMSA([])
+        msa.append(DNA('ACGT'))
+        self.assertEqual(msa, expected)
+
+    def test_to_empty_msa_invalid_dtype(self):
+        msa = TabularMSA([])
+        with six.assertRaisesRegex(self, TypeError,
+                                   'sequence.*alphabet.*Sequence'):
+            msa.append(Sequence(''))
```
jairideout: In these error-case tests can you check that the original MSA state isn't mangled?

jairideout:
@anderspitman I'm done with my review, let me know when you're ready for another pass.

anderspitman:
Thanks @jairideout. I'm surprised how much I missed that time through. I'll try to make it less work for you on the next one.

jairideout:
No problem, thanks @anderspitman!

skbio/alignment/_tabular_msa.py
anderspitman: :+1:

skbio/alignment/_tabular_msa.py
anderspitman: Excellent. I also moved the called to `_add_key` into `_add_sequence`, since there's really a direct dependency there and I want it to be explicit.

skbio/alignment/tests/test_tabular_msa.py
anderspitman: That test doesn't appear to exercise the callable?

anderspitman:
Alrightly @jairideout I think we're ready for another go. I also did a read through the entire file to see if any of the minter- or key-related stuff seemed out of place. Fixed a couple cases where `key` needed to be `minter` but that's all I found.

skbio/alignment/_tabular_msa.py
jairideout: Ping on this one -- the current implementation allows for a sequence to be appended but appending the key can still error (e.g., if the new key isn't unique or hashable), leaving the MSA in an inconsistent state.

jairideout:
Thanks @anderspitman! [This comment](https://github.com/biocore/scikit-bio/pull/1121/files#r42269298) is the only thing remaining then this is ready for merge.

skbio/alignment/_tabular_msa.py
anderspitman: Oooo very interesting chicken and egg problem. We don't want to mutate `self._seqs` until we know the key is valid, but the way we're validating the key is by attempting to set `self.keys`, which won't work unless the number of sequences is correct, which it won't be until we've appended the new sequence...

See how you like the new approach. Basically I'm doing the important part of the key validation (hashability), then appending the sequence, then setting `self.keys`. I added a big comment and unit test. Personally I hate it but I haven't come up with a better solution yet. Transactional updates to 2 separate sections is difficult apparently...

anderspitman:
@jairideout interesting problem. Please see [my comment](https://github.com/biocore/scikit-bio/pull/1121/files#r42292361) and the latest attempt.

skbio/alignment/_tabular_msa.py
```diff
@@ -717,7 +694,136 @@ def reindex(self, key=None, keys=None):
...
+
+        >>> msa = TabularMSA([DNA('', metadata={'id': 'a'})], minter='id')
+        >>> msa.append(DNA('', metadata={'id': 'b'}), minter='id')
+        >>> msa == TabularMSA([DNA('', metadata={'id': 'a'}),
+        ...                    DNA('', metadata={'id': 'b'})], minter='id')
+        True
+        """
+
+        if key is not None and minter is not None:
+            raise OperationError(
```
jairideout: Sorry I didn't catch this before, this should be a `ValueError`

anderspitman: @jairideout How could you miss that? I mean really. I think you might be losing your edge...

jairideout: :scream_cat:

skbio/alignment/_tabular_msa.py
jairideout: Hmmm... this is tricky. The current approach isn't too bad but there's an extra round of key validation happening that would be nice to get rid of. Here are a couple of ideas:

1. Modify `_assert_valid_keys` to validate keys, cast them to the correct type, and return them (this is currently happening in `reindex`, which is what the `keys` setter calls). In `append`, if the sequence and keys are valid, update `_keys` and `_seqs` directly (bypassing `reindex`/`keys` setter). Modify `reindex` to use `_assert_valid_keys` to avoid duplicate logic. `_assert_valid_keys` should be renamed since it is validating and casting.

2. In a try/except, append the sequence and use `keys` setter to set the new keys (which will also validate them). If any exception is raised, remove the appended sequence.

(1) is harder to implement but seems less of a hack than the current approach and (2).

(2) is easy to implement but feels very much like a hack, similar to the current implementation. (2) will be more efficient than the current approach.

What do you think? @ebolyen can you weigh in here?

skbio/alignment/_tabular_msa.py
```diff
@@ -717,7 +694,136 @@ def reindex(self, key=None, keys=None):
...
+            elif minter is not None:
+                raise OperationError(
+                    "minter was provided but MSA does not have keys.")
+
+        self._add_sequence(sequence, new_key)
+
+    def _add_sequence(self, sequence, key=None):
+        msa_is_empty = not len(self)
+        dtype = type(sequence)
+        if msa_is_empty:
```
jairideout: This block doesn't allow users to append a sequence *and* key to an empty MSA. Can you add a unit test to catch this case too?

anderspitman: One of these days you're going to have to train me to catch stuff like this.

anderspitman:
@jairideout good to go.

jairideout:
Thanks @anderspitman! Let me ping @ebolyen on the transactional update thing, I think that was the last remaining item.

skbio/alignment/_tabular_msa.py
ebolyen: I like `(1)` better.

ebolyen: You could call it `munge_keys`

jairideout: Thanks @ebolyen! :+1: for (1). @anderspitman are you up for implementing that here or should we make that refactor in another PR? I'm fine with either since the unit tests are in place to ensure this behavior.

anderspitman: @jairideout @ebolyen sorry I didn't see this before. Is there an easy way to see line comments which you're participating in that have been updated? I feel like I keep missing these.

Anyway, option 1 sounds great to me. I actually tried option 2 at one point and it felt really wrong. I don't see any reason I can't add it to this PR.

anderspitman:
@jairideout feeling pretty good about this one ;)

skbio/alignment/_tabular_msa.py
jairideout: > @jairideout @ebolyen sorry I didn't see this before. Is there an easy way to see line comments which you're participating in that have been updated? I feel like I keep missing these.

I get an email whenever someone posts a comment, are you not getting those?

Thanks for working on this, I'll take a final pass now!

skbio/alignment/_tabular_msa.py
```diff
@@ -690,12 +674,124 @@ def reindex(self, key=None, keys=None):
...
+            if not issubclass(dtype, IUPACSequence):
+                raise TypeError(
+                    "`sequence` must be a scikit-bio sequence object "
+                    "that has an alphabet, not type %r" % dtype.__name__)
+            self._dtype = dtype
```
jairideout: This block has the transactional update problem (the block below looks great though!). Can you fix this and add a unit test? Then this PR is ready to go!

anderspitman: Nice catch. It was surprisingly difficult to exercise an error conditional at that location, since it's already done so many checks up to that point. An unhashable key did the trick.

anderspitman:
@jairideout you can take another look. I've got my laptop now and I'll keep an eye on my email so I can make additional changes on demand. Back to the roof!

jairideout:
Looks great, thanks @anderspitman!

PR #1146, opened 2015-10-12
BUG: scale needs to convert input to float64 to work in-place

### LINKED ISSUE #1144: Build failing in 2.7 and 3.4
```
======================================================================

ERROR: test_corr (skbio.stats.ordination.tests.test_util.TestUtils)

----------------------------------------------------------------------

Traceback (most recent call last):

  File "/home/travis/miniconda3/envs/env_name/lib/python3.4/site-packages/skbio/stats/ordination/tests/test_util.py", line 45, in test_corr

    obs = corr(self.small_mat)

  File "/home/travis/miniconda3/envs/env_name/lib/python3.4/site-packages/skbio/stats/ordination/_utils.py", line 173, in corr

    x = scale(x)

  File "/home/travis/miniconda3/envs/env_name/lib/python3.4/site-packages/skbio/stats/ordination/_utils.py", line 123, in scale

    a -= avg

TypeError: Cannot cast ufunc subtract output from dtype('float64') to dtype('int64') with casting rule 'same_kind'

----------------------------------------------------------------------
```

Looks like a package (numpy probably) has just been updated.

Jorge-C:
Test result was wrong.

Fixes #1144

### DIFF
--- a/skbio/stats/ordination/_utils.py
+++ b/skbio/stats/ordination/_utils.py
@@ -117,6 +117,7 @@ def scale(a, weights=None, with_mean=True, with_std=True, ddof=0, copy=True):
     """
     if copy:
         a = a.copy()
+    a = np.asarray(a, dtype=np.float64)
     avg, std = mean_and_std(a, axis=0, weights=weights, with_mean=with_mean,
                             with_std=with_std, ddof=ddof)
     if with_mean:
--- a/skbio/stats/ordination/tests/test_util.py
+++ b/skbio/stats/ordination/tests/test_util.py
@@ -43,9 +43,9 @@ class TestUtils(TestCase):
 
     def test_corr(self):
         obs = corr(self.small_mat)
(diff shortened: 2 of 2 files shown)


Jorge-C:
Our implementation of `corr` is equivalent to the following:
```
def corr(x, y=None):
    if y is None:
        return np.corrcoef(x.T)
    else:
        return np.corrcoef(x.T, y.T)[:x.shape[1], -y.shape[1]:]
```

but when `y is not None` using numpy wastes resources creating a rather large matrix (we only need a little rectangle out of it, which is the only thing our implementation computes).

I don't have any strong feelings either way.

ebolyen:
Thanks for fixing @Jorge-C! 

I think we should probably run a benchmark or two before deciding one way or the other. This functionality probably doesn't need to be public facing either way as numpy already has an API for it.

PR #1148, opened 2015-10-15
ENH: add classonlymethod decorator

ebolyen:
The [descriptor protocol](http://users.rcn.com/python/download/Descriptor.htm) is kewl.

This is very useful for things like `Sequence.concat` where the class which it was called on defines the return type. It used to be possible to say `seq1.concat([seq2, seq3])` but it is ambiguous as to the behaviour (it would only concat seq2 and 3 and cast as the type of seq1, but a future maintainer of such code may not know this, indeed the original author probably didn't either).


jairideout:
There's a couple unit tests failures in the monkey patch tests

ebolyen:
@jairideout fixed the tests

PR #1150, opened 2015-10-15
MAINT: use mixin for classes storing metadata

jairideout:
We were duplicating code on ``Sequence`` and ``TabularMSA`` to support arbitrary metadata. Most, if not all, classes in scikit-bio will eventually support metadata so this refactoring will avoid further code duplication as the project grows.

This commit adds a private mixin ``skbio._base.MetadataMixin`` which centralizes the following metadata logic:

- `metadata` property (getter, setter, deleter)
- `has_metadata` method
- metadata handling in `__init__`
- metadata handling in `__eq__`/`__ne__`

Additionally, a private unit test mixin `skbio.util._testing.MetadataMixinTests` is used to test `MetadataMixin`, `Sequence`, and `TabularMSA`, avoiding redundant unit tests and ensuring that metadata API and behavior is consistent across classes.

### DIFF
--- a/skbio/sequence/_sequence.py
+++ b/skbio/sequence/_sequence.py
@@ -671,7 +671,7 @@ class Sequence(MetadataMixin, collections.Sequence, SkbioObject):
 
             self._set_bytes(sequence)
 
-        super(Sequence, self).__init__(metadata=metadata)
+        MetadataMixin.__init__(self, metadata=metadata)
 
         if positional_metadata is None:
             self._positional_metadata = None
@@ -808,7 +808,7 @@ class Sequence(MetadataMixin, collections.Sequence, SkbioObject):
         if self.__class__ != other.__class__:
             return False
 
-        if not super(Sequence, self).__eq__(other):
+        if not MetadataMixin.__eq__(self, other):
             return False
 
         if self._string != other._string:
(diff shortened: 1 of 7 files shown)


anderspitman:
:+1: I was just thinking about this today. Thanks for picking up the brainwaves and implementing it!

skbio/sequence/_sequence.py
```diff
@@ -887,18 +808,7 @@ def __eq__(self, other):
...
-        # positional metadata
-        if self.has_metadata() and other.has_metadata():
-            if self.metadata != other.metadata:
-                return False
-        elif not (self.has_metadata() or other.has_metadata()):
-            # both don't have metadata
-            pass
-        else:
-            # one has metadata while the other does not
+        if not super(Sequence, self).__eq__(other):
```
anderspitman: Nice! I was wondering what a good way to do this without explicitly depending on the implementation of the mixin would be.

jairideout:
No problem, I was starting to work on adding positional metadata to TabularMSA and couldn't bring myself to copy and paste again ;)

ebolyen:
@jairideout looks like we have some merge conflicts

jairideout:
Noooooooooooooo

jairideout:
@ebolyen merge conflicts are resolved

ebolyen:
:+1: :x: :100: 

This is awesome!

Issue #1149, opened 2015-10-15
Arrange Act Assert in unit tests?


anderspitman:
A lot of our unit tests have multiple assertions, reuse local variables, etc. This isn't really in alignment with the general practice of ["Arrange, Act, Assert"](http://c2.com/cgi/wiki?ArrangeActAssert) style tests. Ideally we would have one assertion per test. This doesn't necessarily mean a single assert call, but it means you only test one thing. Moving in this direction does require more tests (though they're smaller and simpler), which necessitates breaking down into test suite classes and more fully utilizing test fixtures.

I think this would be a desirable direction to move, at least for new tests. Thought?

ebolyen:
I like this pattern, especially as it basically negates using `setUp` which I feel makes most tests pretty hard to maintain. Way too much of our code still operates as a single test for a given method. I'm not arguing for or against, but frankly I wouldn't be against creating a test-class for each method. Or at least a test-class for each protocol/interface. 

Do you think AAA precludes using a loop to deal with multiple arrange steps with identical act/assert? There are some tests which would expand to 5 or 6 individual tests that would be identical if not for the arrange step.

jairideout:
:+1: for starting to use this pattern (I really like having a class for testing each method). I'm curious too about @ebolyen's question, these sorts of tests happen a lot in practice.

wasade:
:+1:

anderspitman:
@ebolyen I'm not sure what you mean when you say this negates using `setUp`. I try to use `setUp` extensively even when using AAA to reduce duplication. Maybe we're not talking about quite the same thing here.

As for loops, I think we should try to keep them if possible, at least for some cases. I think the key principle is that you want to minimize the risk that a test depends on the results of a previous test. So loops do introduce a higher risk of that, but I think it's reasonable. On the other hand loop tests are very difficult to maintain. I remember back when we were changin `Sequence` that was a huge pain to try and figure out how to shoehorn new (or removed) functionality into the test loops.

I guess overall I'm not too religious about what approach we take, so much as I feel we don't really have many conventions currently so there's a lot of variation in testing methodology.

RNAer:
:+1: on AAA

gregcaporaso:
I think this is a good idea for new tests, but not worth the time to port
existing tests (too much other stuff that's higher priority).

ebolyen:
[...]

PR #1151, opened 2015-10-21
ENH: add TabularMSA.gap_frequencies

### LINKED ISSUE #1089: `TabularMSA`: gap_frequencies

- `gap_frequencies(axis='position', relative=False) -> np.ndarray`

   Return vector of gap frequencies along an axis.

jairideout:
Fixes #1089. Tried out Arrange Act Assert (#1149) for these unit tests. I like it!

This is the first method with an `axis` parameter. The RFC specifies that `'sequence'` and `0` are equivalent, likewise with `'position'` and `1`. However, our notion of `axis` differs from numpy:

```python
In [1]: import numpy as np

In [2]: a = np.array([[1,2,3], [4,5,6]])

In [3]: a
Out[3]:
array([[1, 2, 3],
       [4, 5, 6]])

In [4]: a.sum(axis=0)
Out[4]: array([5, 7, 9])

In [5]: a.sum(axis=1)
Out[5]: array([ 6, 15])

In [6]: from skbio import DNA, TabularMSA

In [7]: msa = TabularMSA([DNA('ACG'), DNA('A--')])

In [8]: msa.gap_frequencies(axis=0)
Out[8]: array([0, 2])

In [9]: msa.gap_frequencies(axis=1)
Out[9]: array([0, 1, 1])
```

Is scikit-bio's notion of `axis` correct or should we be matching numpy? Intuitively our API makes more sense to me but not matching numpy may be a mistake...

### DIFF
--- a/skbio/alignment/_tabular_msa.py
+++ b/skbio/alignment/_tabular_msa.py
@@ -11,7 +11,7 @@ from __future__ import absolute_import, division, print_function
 import collections
 import operator
 
-from future.builtins import zip
+from future.builtins import zip, range
 from future.utils import viewkeys, viewvalues
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


ebolyen:
It looks like our semantics are backwards. `axis` means *across* not *along*. 
Here is an example in `pandas`
```
In [9]: df
Out[9]: 
   1  2
0  1  5
1  2  6
2  3  7

In [10]: df.sum(axis='columns')
Out[10]: 
0     6
1     8
2    10
dtype: int64

In [11]: df.sum(axis='rows')
Out[11]: 
1     6
2    18
dtype: int64

```

So positional gap frequencies would be `gap_frequencies(axis='sequence')` because we are finding the gap frequencies across the sequence axis.

Since we intend to interoperate closely with `pandas` and `numpy` (and maybe `xray`!) we need to make sure we match the semantics of `axis` which means *across*.

It also makes sense why `numpy` had to use this meaning in the first place. Consider a 5 dimensional array, if we wanted to sum across a single axis we can either specify that single axis, or state that we want to sum *along* all of the other dimensions. It just looks a little backwards in `2d`.

jairideout:
Thanks @ebolyen! That makes sense, will update shortly.

jairideout:
How about removing the integer axis names (`0` and `1`) in the spirit of named axes? From DS4DS it sounded like numpy might support named axes in the future, and devs frowned on code that used numbers instead of descriptive names. We could always add integer support in the future if users ask for it without breaking backward-compatibility, and in the meantime, scikit-bio *user* code will be more readable.

jairideout:
@wasade what do you think re: having only named axis arguments? (See my previous comment for context)

wasade:
I really like descriptive axis labels

jairideout:
Cool! The specific question is: do we want to support both named axes and integer axes, or just named axes?

wasade:
Oh, I see. I guess the benefit of integer axes is it maps onto numpy right now. I don't really see any other benefit except performance within a conditional as an integer test is faster than a string comparison. But, numpy is (likely) shifting away from unnamed axes, and I don't know if the performance difference warrants the semantic confusion. I think supporting multiple ways of describing the axes can lead to annoying bugs, although I admit there were times I wished it were possible to use the numbered axes in BIOM to reduce line length.

jairideout:
Awesome, thanks for the input! Sounds like @ebolyen, you, and I are on the fence on this one. I'm leaning towards keeping integer and string axis support so that there's a convenient, short way to reference axes via integers. In all documentation examples we'll use the named (string) version of the axes to encourage users to do the same.

Anyone else have an opinion?

fence

jairideout:
@ebolyen ready for review

gregcaporaso:
> I'm leaning towards keeping integer and string axis support

I think this is a good way to go too (and we can always revisit before we leave ``experimental``).

jairideout:
:+1:

ebolyen:
@jairideout merge conflicts

jairideout:
merge conflicts resolved

skbio/alignment/_tabular_msa.py
```diff
@@ -484,6 +484,97 @@ def __ne__(self, other):
...
+        array([ 0.  ,  0.25,  0.75])
+
+        Compute frequency of gap characters for each sequence (i.e., *across*
+        the position axis):
+
+        >>> msa.gap_frequencies(axis='position')
+        array([0, 2, 1, 1])
+
+        """
+        if axis == 'sequence' or axis == 0:
```
ebolyen: Is there any value to encoding this logic (and the validation) into a helper method like `_is_sequence_axis(axis)`?

The the code could be:
```python
if self._is_sequence_axis(axis):
    #stuff
else:
    #other
#The error was already raised by `_is_sequence_axis`
```

jairideout: done

PR #1157, opened 2015-10-28
use hashable object as positional_metadata column name to store its feature metadata

RNAer:
@jairideout , @ebolyen , or @gregcaporaso , mind taking a quick look at this?


ebolyen:
Also needs to define a `__hash__` such that `a == b` implies `hash(a) == hash(b)`. We will also probably want to make the attributes read-only, that can be done a couple ways, but using Python's descriptor protocol is probably the most "straight-forward"

skbio/sequence/_feature.py
```diff
@@ -0,0 +1,46 @@
...
+    def __eq__(self, other):
+        if self.__class__ != other.__class__:
+            return False
+        for attr, value in other:
+            if getattr(self, attr) != value:
+                return False
+        return True
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
```
ebolyen: This can be `return not (self == other)`

skbio/sequence/_feature.py
```diff
@@ -0,0 +1,46 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2013--, scikit-bio development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
```
ebolyen: missing some `__future__` imports which is why travis is failing

RNAer:
oh yeah. I thought the builtin `hash` already did the job. Not realize it
uses `id()`. Will add `__hash__`` function. Isn't the current
implementation already read-only? cause there is no `__setitem__`.

RNAer:
oh sorry, forgot one can just do `instance.attr = "foo"` to overwrite.

RNAer:
@ebolyen , instead of use `descriptor`, I took @jairideout 's original suggestion of frozen dict, which seems simpler. Is this update ok?

another question is whether we want to define our own `__eq__`. The inherited `__eq__` is more general:

     $ x = Feature(x=3,b='a')
     $ x == {'x':3, 'b':'a'}
     : True

ebolyen:
@RNAer sounds good! As for `__eq__` if `__hash__` is also implemented by `Mappable` then I say we take it. If anything this makes it easier for users to find things, and `dicts` aren't hashable anyway so the comparison there won't really make a difference.

RNAer:
This is ready for review.

RNAer:
@ebolyen , can I add the #1167 and `SparseDataFrame` (they are related) into this pull request? I know you guys focus on other things, would you have time to review and merge it? Before we have a final solution, I'd like to have skbio better posed for it.

jairideout:
@RNAer what changes are you planning to add for `SparseDataFrame`? AFAIK it's kinda broken in pandas which is why we didn't use it originally, and there isn't a clear path forward in #1159.

RNAer:
It was broken but the latest version is not. It is better but not good
enough for memory as discussed in #1159. I am basically talking about the
code here RNAer@c00bb7c
<https://github.com/RNAer/scikit-bio/commit/c00bb7cb4b93619405c78d7482f7f4d2671a99ec>
<https://github.com/RNAer/scikit-bio/commit/c00bb7cb4b93619405c78d7482f7f4d2671a99ec>

ebolyen:
@RNAer given that the sparse doesn't necessarily solve our problems, I think we should probably hold off on this. I would recommend creating an intentional "fork" of scikit-bio and making basically any API changes you feel are needed for micronota (this probably involves creating a "feature" property on sequence)

The advantage of this is you can quickly iterate without needing to stop for our input. Once you guys have a solution you are happy with we could create an RFC indicating what changes have to be made to support large annotations and get those changes merged in.

We should be able to move forward without the hashable features for now.

RNAer:
@ebolyen , I am playing different solutions and creating RFC for the ultimate solution as Greg suggested. The motivation for what I was asking is that, even though it doesn't solve the problem, it mitigates it. Better than none. And more importantly, it does not change API and the code is already written. So it is trivial to put them in place. Is there any other concern I don't see?

ebolyen:
@jairideout and I have been discussing this PR. Based on the discussion in #1159 it's not clear how we ultimately want to handle sequence features in scikit-bio, so the hashable features may not even be how we actually want to solve this problem. Introducing hashable features is an API change that may not make sense to make right now. If you wanted to include this as part of your fork while you develop a more robust solution that is totally fine. But given the subpar performance of `SparseDataFrame` it probably isn't worth making this change now, as we need a different API solution for these features anyways. 

I'm also not convinced that using `SparseDataFrame` will work either, as I expect that slicing the sequence by a `SparseSeries` in the `SparseDataFrame` will produce an incorrect sequence.

As for #1167, I would open a new PR for that. That functionality is totally fine, and we should be able to review it quickly.

RNAer:
fair enough. thanks.

jairideout:
@RNAer can this be closed in favor of the newer interval metadata PRs?

RNAer:
yep, it is safe to close.

PR #1162, opened 2015-10-28
Fixes #757

### LINKED ISSUE #757: Add method to distance matrix class that returns a pandas dataframe
It would be really nice to have the ability to create a `pandas.DataFrame` from a `DistanceMatrix` object with a simple method for instance `DistanceMatrix.to_DataFrame()`

Kleptobismol:
Adds ``to_dataframe`` method and test.


ebolyen:
@Kleptobismol looks good!

PR #1163, opened 2015-10-30
MAINT: add positional metadata mixin

jairideout:


skbio/_base.py
```diff
@@ -168,6 +169,197 @@ def has_metadata(self):
         return self._metadata is not None and bool(self.metadata)
 
 
+class PositionalMetadataMixin(with_metaclass(ABCMeta, object)):
+    @abstractmethod
+    def _positional_metadata_axis_len(self):
```
ebolyen: Did we want to use the "sunder" idea: `_positional_metadata_axis_len_`?

jairideout: Yep, good catch!

jairideout:
@ebolyen comment addressed!

PR #1173, opened 2015-11-04
ENH: add TabularMSA.__copy__/__deepcopy__

### LINKED ISSUE #1158: `TabularMSA`: copy / __copy__/ __deepcopy__
Forgot these methods on the RFC (likely other dunder methods too). I'll take this one after `positional_metadata` (#1092). API will match `skbio.Sequence`.

jairideout:
Refactored common code into metadata mixins. Fixes #1158.

@ebolyen shallow copying creates a new list with references to the original sequences. Is that right or should I have it create a new list *and* shallow copy each sequence? See the `TabularMSA` `copy` unit tests for current behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,9 @@ constructor. ([#6240](https://github.com/biocore/scikit-bio/issues/624))
 * ``DissimilarityMatrix.plot()`` no longer leaves a white border around the
   heatmap it plots (PR #1070).
 
+### Deprecated functionality [stable]
+* `skbio.Sequence.copy` has been deprecated in favor of `copy.copy(seq)` and `copy.deepcopy(seq)`.
+
 ### Deprecated functionality [experimental]
 * ``SequenceCollection.distances`` has been deprecated in favor of ``DistanceMatrix.from_iterable``. Use `key="id"` to exactly match original behavior.
 
--- a/skbio/alignment/_tabular_msa.py
+++ b/skbio/alignment/_tabular_msa.py
@@ -9,6 +9,7 @@
 from __future__ import absolute_import, division, print_function
 
 import collections
+import copy
 import operator
 
 from future.builtins import zip, range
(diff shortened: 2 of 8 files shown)


ebolyen:
@jairideout A new list and shallow copies would be a deepcopy., well doing that recursively would be.

I would expect a shallow copy to create fresh collections and not much else. A deep copy would create fresh collections and recursively deepcopy anything in them.

skbio/alignment/_tabular_msa.py
```diff
@@ -474,6 +475,68 @@ def __ne__(self, other):
...
+
+        keys = None
+        if self.has_keys():
+            keys = self.keys
+
+        # Constructor makes shallow copy of keys but this should be sufficient
+        # even for a deep copy since keys must be hashable (read: immutable).
+        # Technically it is possible to create a hashable, mutable key but the
+        # user has bigger problems at that point... the concept of keys is
+        # going to change soon so may want to revisit then.
```
ebolyen: :+1:

ebolyen:
Thinking further. I disagree with myself now.

Part of our user story for metadata is delegating some of the metadata into the sequences. If I created a copy of an MSA would probably intuitively expect the sequence data to be copied while the  metadata itself might be shallow having the same references.

More specifically I could see:
```
msa1 = <magically align some seqs them>
msa2 = copy.copy(msa1)
msa2[0].metadata['first'] = "First!"

msa1[0].metadata['first']
KeyError
```

I guess it depends on whether the user will see this as a "grid" of data, or a list of sequences. I think I lean towards the "grid" concept.

And a deep copy is still unambiguous as to it's behaviour.

CHANGELOG.md
```diff
@@ -73,6 +73,9 @@ constructor. ([#6240](https://github.com/biocore/scikit-bio/issues/624))
 * ``DissimilarityMatrix.plot()`` no longer leaves a white border around the
   heatmap it plots (PR #1070).
 
+### Deprecated functionality [stable]
+* `skbio.Sequence.copy` has been deprecated in favor of `copy.copy(seq)` and `copy.deepcopy(seq)`.
```
ebolyen: :+1:, Especially with our current docs with explicitly point out the existence of `copy.copy` and `copy.deepcopy`

jairideout:
:+1: for shallow copying each sequence as that most closely matches our notion of shallow/deep copies of other "grid" data (e.g., positional metadata). I think it's pretty arbitrary at this point but we might as well be self-consistent.

Thanks for the input, will update the code, unit tests, and fix merge conflicts.

jairideout:
@ebolyen ready for review

PR #1129, opened 2015-09-14
Issue 1091 tab msa consensus

anderspitman:
Closes #1091

Should probably merge #1128 first

PR #1183, opened 2015-11-13
ENH: TabularMSA.consensus

### LINKED ISSUE #1091: `TabularMSA`: consensus

- `consensus() -> IUPACSequence`

   Return majority consensus sequence matching alignment's `dtype`. Other
   types of consensus may be added later.

jairideout:
Contains commits in #1129. Fixes #1091.

To-do:

- [x] ~~should each gap character be treated distinctly when computing consensus for each position? This is the current behavior here and in `Alignment.majority_consensus`, not sure if this is right or not.~~ @gregcaporaso, @ebolyen, and I discussed this. For now we will treat different gap characters the same and we can add a parameter to treat them distinctly if requested by users in the future. This change will require an `IUPACSequence.default_gap_char`, which will be used to represent a consensus gap.

### DIFF
--- a/skbio/alignment/tests/test_tabular_msa.py
+++ b/skbio/alignment/tests/test_tabular_msa.py
@@ -1141,6 +1141,106 @@ class TestIterPositions(unittest.TestCase):
         self.assertIsNone(msa._positional_metadata)
 
 
+class TestConsensus(unittest.TestCase):
+    def test_no_sequences(self):
+        msa = TabularMSA([])
+
+        cons = msa.consensus()
+
+        self.assertEqual(cons, Sequence(''))
+
+    def test_no_positions(self):
+        msa = TabularMSA([DNA(''),
+                          DNA('')])
+
+        cons = msa.consensus()
+
+        self.assertEqual(cons, DNA(''))
+
+    def test_single_sequence(self):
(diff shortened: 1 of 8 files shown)


skbio/alignment/tests/test_tabular_msa.py
```diff
@@ -1141,6 +1141,125 @@ def test_handles_missing_positional_metadata_efficiently(self):
...
+        self.assertIsNone(msa._positional_metadata)
+        self.assertIsNone(cons._positional_metadata)
+
+    def test_mixed_gap_characters_as_majority(self):
+        seqs = [
+            DNA('A'),
+            DNA('A'),
+            DNA('A'),
+            DNA('.'),
+            DNA('.'),
```
ebolyen: Could you add one more `'.'` and two more `'A'`s to show that `-` is used even though `.` would win?

ebolyen:
@jairideout only a minor comment, then this looks ready to go.

skbio/alignment/tests/test_tabular_msa.py
jairideout: done

ebolyen:
Thanks @jairideout!

Issue #1117, opened 2015-09-08
Adding metadata in skbio.TreeNode


mortonjt:
This is related to the discussion in #1108 

Noticed that the same problem of dealing with metadata also shows up in `skbio.TreeNode`
It would super awesome if there is a way to assign metadata in the same fashion as the `skbio.SequenceCollection`

ebolyen:
:+1: This would also allow us to support the [`NHX`](http://home.cc.umanitoba.ca/~psgendb/doc/atv/NHX.pdf) format for interoperability with `ete`.

jhcepas:
regarding the interoperability layer between the upcoming ete3 version and scikit-bio tree objects, any feedback here would be great:  https://github.com/jhcepas/ete/issues/137

mortonjt:
I'm going to close this.  Metadata can be added to the tree objects pretty easily.
For example, the following can be run
```python
from skbio import TreeNode
>>> t = TreeNode.read(['(a, b)c;'])
>>> t.foo_am_not = False
>>> print(t.foo_am_not)
False
```

Issue #809, opened 2015-01-20
Refactor TreeNode to use NetworkX internally


ebolyen:
This will avoid replicating common functionality

anderspitman:
What would the scope of this be? I'm not too familiar with NetworkX. It looks like it's mostly used to represent graph objects, right? Is there a way to only use NetworkX's node implementation as the backend for TreeNode and leave the interface the same, or am I completely missing the point?

mortonjt:
I've got a use-case that may need to convert TreeNode objects to networkx. Can submit pull request shortly

wasade:
`to_networkx` and `from_networkx` would be very nice to have, thank you!!

Issue #465, opened 2014-06-19
add TreeNode.__eq__?


jairideout:
I'm adding some tests for `TreeNode` and it'd come in handy to have an `__eq__` method defined for this class. @wasade thoughts on this?

PR #1193, opened 2015-11-20
Allow subclasses of TreeNode to be used

terrycojones:
I've just started using scikit-bio (thanks!) and ran into something a little inflexible that I think could be improved quite easily.

I'd like to have my own `TreeNode` class so I can store attributes and methods on it. This issue has come up in various other guises, see e.g., #1117, #809, and #465.

If I can't use my own subclass, I need to keep per-node metadata somewhere else, and keep it up to date, etc. E.g., I could make a class, `MyClass`, that keeps a `TreeNode` instance (e.g., called `tree`) plus some other metadata. This is quite inconvenient and un-Pythonic. It means I need to call the extra methods of my class using things like

```python
myClass = MyClass()
for node in myClass.tree.children:
    if myClass.nodeSupport(node) > 0:
        # pass
```

rather than the simpler approach in which I subclass `TreeNode` and add the methods and attributes I want.  The simpler approach also makes testing easier.

Anyway, the usual way to do this is to pass around a class that is used to make new tree nodes, and have it default to the original simple base class (`TreeNode` in this case).

I'm not proposing that this pull request gets merged at the moment, but would like to see what people think of the idea and the simple diff.  If this seems like a good direction, I can add some tests that subclass `TreeNode` and which pass a `tree_node_class` argument to `majority_rule`.

Thanks again for scikit-bio!

### DIFF
--- a/skbio/tree/_majority_rule.py
+++ b/skbio/tree/_majority_rule.py
@@ -116,7 +116,8 @@ def _filter_clades(clade_counts, cutoff_threshold):
     return accepted_clades
 
 
-def _build_trees(clade_counts, edge_lengths, support_attr):
+def _build_trees(clade_counts, edge_lengths, support_attr,
+                 tree_node_class=TreeNode):
     """Construct the trees with support
 
     Parameters
@@ -127,6 +128,9 @@ def _build_trees(clade_counts, edge_lengths, support_attr):
         Keyed by the frozenset of the clade and valued by the weighted length
     support_attr : str
         The name of the attribute to hold the support value
+    tree_node_class: Python class
+        Either TreeNode (the default) or a class that implements the same
+        interface (most usefully, a subclass of TreeNode).
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


wasade:
:+1:, this used to be handled through a `constructor` `kwarg` in pycogent. It was dropped as, in practice, it wasn't used and led to confusion in the code base. We do need some mechanism that cleanly allows for subclassing support in ancillary methods. 

With the current object, there shouldn't be an issue with decorating on your own attributes, and they'll automatically be represented during copy operations, for instance. But, I also completely agree that the metadata representation does need to be rethought on this object.

terrycojones:
Hi @wasade 

Thanks. I can add attributes myself to `TreeNode` instances, but it's quite cumbersome in some cases. E.g., if I call a `TreeNode` method that alters the tree, I may need to walk the tree myself to see if I need to add/update my attributes on all nodes (some nodes may be new, some nodes may have different numbers of siblings or whatever, etc). Or if I want to change the behavior of any of the existing methods (even something simple, like `__str__`), it'd be a real pain. Adding something like `__eq__`, as several people seem to have wanted to do (perhaps in quite different ways, which can be done trivially via subclasses), is also hard. Of course these things can be done with additional (external) functions, but that's awkward too.

wasade:
Ya, agree it is not ideal. Thanks for opening the issue!

jairideout:
Closing in favor of #1364, which contains the commits in this PR and some updates. Thanks for your contribution @terrycojones!

Issue #1185, opened 2015-11-18
investigate beta diversity parallelization using sklearn.metrics.pairwise.pairwise_distances


gregcaporaso:
We might be able to [call this](http://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.pairwise_distances.html) instead of ``pdist`` internally and use the scikit-learn parallel framework. This should be separate from #1040 though, as that one is already very big and this wouldn't involve an API change.

wasade:
It doesn't look like this is doing the block parallelization, but is instead parallelizing over rows similar to how the partitioning was done in QIIME. The code is being passed through to joblib but is partitioning only over a single axis (see [here](https://github.com/scikit-learn/scikit-learn/blob/c957249/sklearn/metrics/pairwise.py#L1060) and the preceding line).  

The downsides of this strategy for fast unifrac is that all samples need to be resident. The impact is that the counts array requires a substantial memory footprint (estimated at 3.2TB for the EMP open reference). Additionally, it is not possible to prune the input tree down leading to performance overhead.

gregcaporaso:
related to #1181

gregcaporaso:
@wasade, true that there are likely better approaches. with this hooked up (see my [PR](https://github.com/biocore/scikit-bio/pull/1194)) we'll have working parallel beta diversity now which we can use as our baseline during development of other approaches.

jairideout:
Fixed in #1194.

PR #1194, opened 2015-11-20
parameterizes function for computing pairwise distances, supports parallel beta diversity calcs

gregcaporaso:
This PR parameterizes the function used for computing pairwise distances in ``skbio.diversity.beta_diversity``. The default is still ``scipy.spatial.distance.pdist``, but this can additionally now used with ``sklearn.metrics.pairwise_distances`` which  takes an ``n_jobs`` parameter supporting sample-level parallelization. This will be really useful as it gives us a way to parallelize beta diversity computation now that is better than what QIIME 1 does (it's not file-based), but also creates a framework that could supporting other parallelization approaches, such as the [block-based approach that @wasade is interested in](https://github.com/biocore/scikit-bio/issues/1185).


gregcaporaso:
This one failure is expected - noted in my comment above.

jairideout:
The pickling error is likely because `_setup_multiple_weighted_unifrac` returns a closure, which gets passed to the pairwise distances function (`pickle` doesn't support closures).

gregcaporaso:
Discussed with @jairideout, we decided the following:
* we'll drop the sklearn dependency, but leave the parameterized ``pdist``
* ``n_jobs`` will be only passed as a kwarg, this will involve modifying how our kwargs are validated

mortonjt:
I'm curious.  How would dropping the sklearn dependency help?  Wouldn't the pickle problem still remain?

Let me know how I can help!  This functionality would also be really useful for the network distance metric that I'm developing.

gregcaporaso:
> How would dropping the sklearn dependency help? Wouldn't the pickle problem still remain?

It wouldn't, it's a separate issue really. The sklearn dependency isn't really need, it's only actually used in my test code. This PR will be updated to not include the sklearn dependency. Metrics that can't be pickled won't be able to be used with sklearn's ``pairwise_distances``, but parameterizing the function it opens the door for us to explore other options for parallelization (or we refactor our methods to be pickle-able).

mortonjt:
Gotcha.  That makes sense.

skbio/diversity/tests/test_driver.py
```diff
@@ -298,20 +298,20 @@ def test_invalid_input(self):
 
         # additional kwargs
         self.assertRaises(TypeError, beta_diversity, 'euclidean',
-                          [[0, 1, 3, 4], [0, 3, 12, 42]],
+                          [[0, 1, 3], [0, 3, 12]],
```
gregcaporaso: Noticed that these values were incorrect before, but the ``TypeError`` was getting thrown due to the extra kwarg before the ``MissingNodeError``. (The test was testing the correct thing, it just would fail for a different reason if the ``TypeError`` wasn't thrown.)

jairideout: Can you update these tests to use the following form:

```python
with six.assertRaisesRegex(self, TypeError, "regex that must match error message"):
    beta_diversity('euclidean', ...)
```

Using `six.assertRaisesRegex` ensures the correct exception type is raised and that the error message is the one you're expecting. Using the context manager is the preferred way of testing error cases because it is easier to read and to run the error-raising code (e.g., to view the actual error that's raised) b/c you can just comment out the `with` line and dedent.

Another note (not necessary to change in this PR unless you want to; we should have caught this during review): try to avoid test cases that test many different things. Most of the tests in this file are long and test several cases in a single test method. We've started using Arrange, Act, Assert (#1149) and it has really improved the quality of the unit tests.

gregcaporaso: I can update to use ``six.assertRaisesRegex``, but note that that wouldn't have caught this issue. The test was passing for the right reason.

jairideout: Oh sorry, I misunderstood what the issue was

skbio/diversity/_driver.py
```diff
@@ -207,6 +209,11 @@ def beta_diversity(metric, counts, ids=None, validate=True, **kwargs):
         bypassed if you're not certain that your input data are valid. See
         :mod:`skbio.diversity` for the description of what validation entails
         so you can determine if you can safely disable validation.
+    pairwise_func : callable, optional
```
jairideout: Can you note the default behavior? `pairwise_func=None`

jairideout: Also note what it should return (so that you've completely described the function signature)

gregcaporaso:
Thanks!

PR #1199, opened 2015-11-25
ENH: TabularMSA.join

### LINKED ISSUE #1093: `TabularMSA`: join

- `join(TabularMSA, how='strict') -> TabularMSA`

   Joins two `TabularMSA` by sequence (horizontally). `how` may be
   `'strict'`, in which case both MSAs must have the same keys (or be the same
   length if there are no keys). `how` may also be one of the common SQL joins:
   `'inner'`, `'outer'`, `'left'` (outer), or `'right'` (outer); these require
   keys or they will raise an exception. `how` is a superset of the interface
   provided by pandas. Outer based modes will insert gaps for missing keys.
   This should be documented as *not* preforming an alignment.

jairideout:
Fixes #1093.

### DIFF
--- a/skbio/alignment/_tabular_msa.py
+++ b/skbio/alignment/_tabular_msa.py
@@ -1095,6 +1095,254 @@ class TabularMSA(MetadataMixin, PositionalMetadataMixin, SkbioObject):
                     "positions in the MSA: %d != %d"
                     % (length, expected_length))
 
+    def join(self, other, how='strict'):
+        """Join this MSA with another by sequence (horizontally).
+
+        Sequences will be joined by index labels. MSA ``positional_metadata``
+        will be joined by columns. Use `how` to control join behavior.
+
+        Alignment is **not** recomputed during join operation (see *Notes*
+        section for details).
+
+        Parameters
+        ----------
+        other : TabularMSA
+            MSA to join with. Must have same ``dtype`` as this MSA.
+        how : {'strict', 'inner', 'outer', 'left', 'right'}, optional
+            How to join the sequences and MSA `positional_metadata`:
+
+            * ``'strict'``: MSA indexes and `positional_metadata` columns must
+              match
+
+            * ``'inner'``: an inner-join of the MSA indexes and
+              ``positional_metadata`` columns (only the shared set of index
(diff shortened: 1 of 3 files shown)


skbio/alignment/_tabular_msa.py
```diff
@@ -1095,6 +1095,252 @@ def _assert_valid_sequences(self, sequences):
...
+        sort
+        skbio.sequence.Sequence.concat
+
+        Notes
+        -----
+        The join operation does not automatically perform re-alignment;
+        sequences are simply joined together. Therefore, this operation is not
+        necessarily meaningful on its own.
+
+        The index of this MSA and the index of `other` must each be unique;
```
ebolyen: Add "respectively"? The wording as is could be construed as no matching labels between either index

skbio/alignment/_tabular_msa.py
```diff
@@ -1095,6 +1095,252 @@ def _assert_valid_sequences(self, sequences):
...
+        if self.dtype is not other.dtype:
+            raise TypeError(
+                "`other` dtype %r does not match this MSA's dtype %r" %
+                (other.dtype if other.dtype is None else other.dtype.__name__,
+                 self.dtype if self.dtype is None else self.dtype.__name__))
+
+        if not self.index.is_unique or not other.index.is_unique:
+            raise ValueError(
+                "Cannot join `TabularMSA` objects. Index labels must be "
+                "unique.")
```
ebolyen: Same here, users might get the wrong impression as to how unique things have to be.

ebolyen:
@jairideout aside from my nitpicky comment about uniqueness this looks great!

skbio/alignment/_tabular_msa.py
jairideout: I reworded and split the sentence in two, I think it reads clearer

skbio/alignment/_tabular_msa.py
jairideout: Split this conditional into two if statements to give more precise and clear error messages

jairideout:

ebolyen:
Thanks @jairideout!

PR #1207, opened 2015-11-30
reorganization of skbio.diversity documentation

gregcaporaso:
One issue that needs to be sorted out is that the plots are now showing up much bigger than in the 0.4.0 docs and as a result are being cut off in some cases. For example:

screenshot 2015-11-30 11 53 11

Has anyone else run into this?


gregcaporaso:
The issue with the docs noted above is now #1210.

PR #1208, opened 2015-11-30
fixes #1077

### LINKED ISSUE #1077: `root_at_midpoint` methods fails on certain `TreeNode` objects
```python 
from skbio import TreeNode
tree = TreeNode.read('rep_set.tre', format='newick')
tree.root_at_midpoint()
```

Fails with:

```TypeError: unorderable types: TreeNode() > TreeNode()```

I apologize, I am not able to reproduce this error with a smaller tree object, as I don't entirely understand the issue. @wasade I think you may understand what is specifically happening here?

### PR DESCRIPTION
**Pull request by wasade**
Resolves #1077. The issue was not in `root_at_midpoint` but actually in `_set_max_distance`. This solution fixes the bug albeit at a computational expense as the inner loop logic has been expanded. 

A unit test was added to with the tree described in #1077 that produced the bug.

### CODE REVIEW

File: skbio/tree/tests/test_tree.py [hunk 1]
```diff
@@ -633,6 +633,18 @@ def test_set_max_distance(self):
...
+        """Corresponds to #1077"""
+        s = StringIO("((a:1,b:1)c:2,(d:3,e:4)f:5)root;")
+        t = TreeNode.read(s)
+
+        # the above tree would trigger an exception in max. The central issue
+        # was that the data being passed to max were a tuple of tuple:
+        # ((left_d, left_n), (right_d, right_n))
+        # the call to max would break in this scenario as it would fall onto
+        # idx 1 of each tuple to assess the "max".
+        t._set_max_distance()
```
jairideout: Can you test the expected value?

jairideout:
Thanks for the fix @wasade! Had one inline comment then this is ready for merge.

wasade:
Thanks! Will address this morning

jairideout:
@wasade any update on this? Would be good to get in for 0.4.1. Also note there's merge conflicts, likely just the changelog.

CHANGELOG.md
```diff
@@ -81,6 +81,7 @@ constructor. ([#6240](https://github.com/biocore/scikit-bio/issues/624))
 * ``Sequence`` objects now handle slicing of empty positional metadata correctly. Any metadata that is empty will no longer be propagated by the internal ``_to`` constructor. ([#1133](https://github.com/biocore/scikit-bio/issues/1133))
 * ``DissimilarityMatrix.plot()`` no longer leaves a white border around the
   heatmap it plots (PR #1070).
+* ``TreeNode._set_max_distance`` would fail if ties were encountered. ([#1077](https://github.com/biocore/scikit-bio/issues/1077))
```
gregcaporaso: Since this is intended to be a user-facing doc, I recommend the following instead of (or in addition to) this:

* ``TreeNode.root_at_midpoint`` no longer fails when a node with two equal length child branches exists in the tree. ([#1077](https://github.com/biocore/scikit-bio/issues/1077))

jairideout: :+1:

skbio/tree/_tree.py
```diff
@@ -2327,7 +2327,11 @@ def _set_max_distance(self):
                 if len(n.children) == 1:
                     raise TreeError("No support for single descedent nodes")
                 else:
-                    tip_info = [(max(c.MaxDistTips), c) for c in n.children]
+                    tip_info = []
+                    for c in n.children:
+                        (left_d, _), (right_d, _) = c.MaxDistTips
```
gregcaporaso: I got part way through fixing this before I saw this PR. This lets you do it without creating a new list. Doesn't matter to me which way we solve this, just wanted to share this since I started on it. 

```
                    tip_info = []
                    for c in n.children:
                        tip_info.append((max(c.MaxDistTips, key=lambda x: x[0]), c))
```

skbio/tree/tests/test_tree.py
jairideout: It'd also be great to add a separate test for `root_at_midpoint` with a tree that has ties, in case `root_at_midpoint` no longer uses `_set_max_distance` internally in the future.

skbio/tree/tests/test_tree.py
```diff
@@ -633,6 +633,18 @@ def test_set_max_distance(self):
         self.assertEqual(tip_a[0] + tip_b[0], 1.6)
         self.assertEqual(sorted([tip_a[1].name, tip_b[1].name]), ['b', 'e'])
 
+    def test_set_max_distance_tie_bug(self):
```
gregcaporaso: I think we should also have a test that midpoint rooting now works when there are equal branch lengths. I started a test of this in a PR that I had:

```
    def test_root_at_midpoint_equal_lengths(self):
        nwk = u"(((a:1,b:1)c:2),((d:3,e:4)f:5))root;"
        t = TreeNode.read(StringIO(nwk))
        obs = t.root_at_midpoint()
        exp_nwk = u""
        self.assertEqual(str(obs), exp_nwk)
```

gregcaporaso:
A few additional comments on this one.

skbio/tree/_tree.py
jairideout: :+1: , the existing list comprehension can just be modified to pass `key`, seems the simplest and most Pythonic IMO

gregcaporaso:
I'm getting another error with this branch when I try to midpoint root a tree that I generated with ssu-align:

screenshot 2015-12-04 08 04 38

@wasade, any idea what's going on here? I can't see the end of the traceback b/c my notebook server is being killed.

wasade:
Possible to reexecute on the command line?

gregcaporaso:
Yes, I will today and post the result. I'm trying the midpoint rooting with ete3 first though b/c I need the tree for an analysis that I need to finish today.

gregcaporaso:
@wasade, will you be able to address these comments for the 0.4.1 release? We'd probably need the PR merged by the end of the day today.

wasade:
I think so, sorry for the delay

wasade:
Test in. I noticed an unexpected aspect of this method though, `MaxDistTips` was only valid for the node on which the method was called. An example of the unexpected issue is below. Commit coming momentarily to resolve it with an additional test to verify.

```python
10:11:57 (daniel@sandbar):scikit-bio> import skbio

10:12:02 (daniel@sandbar):scikit-bio> from io import StringIO

10:12:10 (daniel@sandbar):scikit-bio> s = StringIO("((a:1,b:1)c:2,(d:3,e:4)f:5)root;")

10:12:12 (daniel@sandbar):scikit-bio> t = skbio.TreeNode.read(s)

10:12:19 (daniel@sandbar):scikit-bio> t
Out[6]: <TreeNode, name: root, internal node count: 2, tips count: 4>

10:12:21 (daniel@sandbar):scikit-bio> for node in t.postorder():
   ...:     print(node.name, node.length)
   ...:
a 1.0
b 1.0
c 2.0
d 3.0
e 4.0
f 5.0
root None

10:12:43 (daniel@sandbar):scikit-bio> t._set_max_distance()

10:12:52 (daniel@sandbar):scikit-bio> for node in t.postorder():
    print(node.name, node.length, [(l, n.name) for l, n in node.MaxDistTips])
   ...:
a 1.0 [(3.0, 'a'), (0.0, 'a')]
b 1.0 [(1.0, 'b'), (0.0, 'b')]
c 2.0 [(3.0, 'a'), (1.0, 'b')]
d 3.0 [(3.0, 'd'), (0.0, 'd')]
e 4.0 [(9.0, 'e'), (0.0, 'e')]
f 5.0 [(3.0, 'd'), (9.0, 'e')]
root None [(3.0, 'a'), (9.0, 'e')]
```

wasade:
@gregcaporaso @jairideout, good for review once travis completes. Sorry for the delaoy

CHANGELOG.md
gregcaporaso: This comment wasn't addressed. Similarly, could you adapt your other ChangeLog addition to be more user centric (i.e., what would this have impacted from the user's perspective)? I can't make a specific suggestion for that one as I don't really know what it would impact.

skbio/tree/tests/test_tree.py
gregcaporaso: This comment wasn't addressed.

wasade: Not sure I agree, the bug wasn't in this method

wasade: ...we'd be duplicating regression tests effectively

jairideout: [One reason](https://github.com/biocore/scikit-bio/pull/1208/files#r46687739) to add this test. Also, it's a good test to have regardless of the bug (i.e., testing midpoint rooting on trees with and without ties).

skbio/tree/_tree.py
```diff
@@ -2320,22 +2320,26 @@ def _set_max_distance(self):
...
         for n in self.postorder():
             if n.is_tip():
-                n.MaxDistTips = [[0.0, n], [0.0, n]]
+                n.MaxDistTips = ((0.0, n), (0.0, n))
             else:
                 if len(n.children) == 1:
                     raise TreeError("No support for single descedent nodes")
                 else:
-                    tip_info = [(max(c.MaxDistTips), c) for c in n.children]
+                    tip_info = []
```
jairideout: remove this line

skbio/tree/tests/test_tree.py
gregcaporaso: Yes, we need to have a test of the public function that was previously failing when provided with valid data, to protect against a similar bug coming back in the future if we change how this method works internally.

skbio/tree/tests/test_tree.py
wasade: but it doesn't use it internally, it uses `get_max_distance`

skbio/tree/tests/test_tree.py
wasade: putting in a test regardless, but i still maintain that it is extraneous.

gregcaporaso: I disagree, but thanks for putting in the test.

jairideout: I disagree as well -- there are at least two valid reasons to add this test.

wasade: good point, your comment wasn't visible when i had responded

skbio/tree/tests/test_tree.py
jairideout: Okay, but we still need a test for each public method that was failing or producing incorrect results, in case internal implementation changes. This will prevent regression.

CHANGELOG.md
```diff
@@ -97,6 +97,8 @@ constructor. ([#6240](https://github.com/biocore/scikit-bio/issues/624))
 * ``Sequence`` objects now handle slicing of empty positional metadata correctly. Any metadata that is empty will no longer be propagated by the internal ``_to`` constructor. ([#1133](https://github.com/biocore/scikit-bio/issues/1133))
 * ``DissimilarityMatrix.plot()`` no longer leaves a white border around the
   heatmap it plots (PR #1070).
+* TreeNode.root_at_midpoint`` no longer fails when a node with two equal length child branches exists in the tree. ([#1077](https://github.com/biocore/scikit-bio/issues/1077))
+* ``TreeNode._set_max_distance`` would store distance information as ``list``s in the attribute ``MaxDistTips`` on each node, however, these distances were only valid for the node in which the call to ``_set_max_distance`` was called on. ([#1223](https://github.com/biocore/scikit-bio/issues/1223))
```
gregcaporaso: My previous comment on this note may have been lost, but can you update to note which public methods were effected by this so users know what results they'd previously generated may have been incorrect? Users won't know how this bug impacts them since we don't expect them to know what private functionality they've accessed (e.g., I don't know how it's impacted me). 

For the same reasons we are talking about below, public methods which were impacted by this should also have new tests added (the fact that this wasn't caught means that there was a condition that those methods weren't testing).

wasade: No public methods were impacted by this, as in, I'm not aware of a method in scikit-bio that produced incorrect results due to #1223; no method that I'm aware of traversed the tree and interrogated `MaxDistTips`. A public attribute was impacted, and its impact is described.

gregcaporaso: Not ``get_max_distance``? I'm sorry, I don't really understand what this bug impacted from the description (hence my request).

wasade: No, `get_max_distance` does not traverse the tree directly. There was no impact of this bug that I'm aware of, however it left misleading information on the tree. I'm revising the entry to be more specific.

gregcaporaso:
Thanks for the fixes @wasade, merging.

PR #1219, opened 2015-12-04
ENH: Rename IUPACSequence to GrammaredSequence

### LINKED ISSUE #1175: support custom grammared sequence types
Changes:

- [x] rename  `skbio.sequence._iupac_sequence.IUPACSequence` class to `GrammaredSequence`: it is a sequence that conforms to the scikit-bio sequence grammar (based on IUPAC's notion of degenerates, gaps, nondegenerates, etc. but not necessarily constrained to the IUPAC alphabet).

- [x] make `GrammaredSequence` publicly importable from `skbio.sequence`

- [x] document `GrammaredSequence` as a place for users to inherit from to create grammared sequences with custom alphabets (add a doctest example of this too).

- [x] for `GrammaredSequence` make a metaclass that performs validation of its alphabet (e.g., `default_gap_char` should be present in `gap_chars`; degenerates must expand into valid nondegenerates; etc.)

- [x] make it impossible to inherit from DNA, RNA, and Protein and document why

anderspitman:
Closes #1175


anderspitman:
Changes:
- [x] rename skbio.sequence._iupac_sequence.IUPACSequence class to GrammaredSequence: it is a sequence that conforms to the scikit-bio sequence grammar (based on IUPAC's notion of degenerates, gaps, nondegenerates, etc. but not necessarily constrained to the IUPAC alphabet).

- [ ] make GrammaredSequence publicly importable from skbio.sequence

- [ ] document GrammaredSequence as a place for users to inherit from to create grammared sequences with custom alphabets (add a doctest example of this too).

- [ ] for GrammaredSequence make a metaclass that performs validation of its alphabet (e.g., default_gap_char should be present in gap_chars; degenerates must expand into valid nondegenerates; etc.)

- [ ] make it impossible to inherit from DNA, RNA, and Protein and document why

anderspitman:
@jairideout ready for review

jairideout:
Thanks @anderspitman! I should have time on Monday to review.

skbio/alignment/_tabular_msa.py
```diff
@@ -1730,7 +1730,7 @@ def append(self, sequence, minter=None, index=None):
             If neither `minter` nor `index` are provided and the MSA has a
             non-default index.
         TypeError
-            If the sequence object isn't an ``IUPACSequence``.
+            If the sequence object isn't an ``GrammaredSequence``.
```
jairideout: *an* -> *a*

skbio/alignment/_tabular_msa.py
```diff
@@ -1816,7 +1816,8 @@ def extend(self, sequences, minter=None, index=None):
         ValueError
             If `index` is not the same length as `sequences`.
         TypeError
-            If `sequences` contains an object that isn't an ``IUPACSequence``.
+            If `sequences` contains an object that isn't an
```
jairideout: *an* -> *a*

skbio/io/format/clustal.py
```diff
@@ -49,7 +49,8 @@
 -----------------
 The only supported format parameter is ``constructor``, which specifies the
 type of in-memory sequence object to read each aligned sequence into. This must
-be a subclass of ``IUPACSequence`` (e.g., ``DNA``, ``RNA``, ``Protein``) and is
+be a subclass of ``GrammaredSequence`` (e.g., ``DNA``, ``RNA``, ``Protein``)
+and is
```
jairideout: fix formatting (premature line break)

skbio/io/format/tests/test_clustal.py
```diff
@@ -25,19 +25,19 @@
 from skbio.io import ClustalFormatError
 
 
-class CustomSequence(IUPACSequence):
+class CustomSequence(GrammaredSequence):
     @classproperty
-    @overrides(IUPACSequence)
+    @overrides(GrammaredSequence)
```
jairideout: You added `default_gap_char` to previous `CustomSequence` classes in the unit tests. Is `default_gap_char` not abstract?

skbio/sequence/_dna.py
```diff
@@ -64,6 +66,14 @@ class DNA(IUPACSequence, NucleotideMixin):
     See Also
     --------
     RNA
+    GrammaredSequence
+
+    Notes
+    -----
+    Subclassing is disabled for DNA, because subclassing makes
+    it possible to change the alphabet, and certain methods rely on the
+    IUPAC alphabet. If a custom sequence is needed, inherit directly from
```
jairideout: *If a custom sequence alphabet is needed, ...*

skbio/sequence/_grammared_sequence.py
```diff
@@ -7,9 +7,10 @@
 # ----------------------------------------------------------------------------
 
 from __future__ import absolute_import, division, print_function
-from future.utils import with_metaclass
 
 from abc import ABCMeta, abstractproperty
+from six import add_metaclass
```
jairideout: put this 3rd party import with numpy imports, group itertools and abc imports together b/c they're stdlib

skbio/sequence/_grammared_sequence.py
```diff
@@ -22,11 +23,71 @@
...
+                raise TypeError("Subclassing disabled for class %s. To create"
+                                " a custom sequence class, inherit directly"
+                                " from skbio.sequence.%s" %
+                                (b.__name__, GrammaredSequence.__name__))
+        return type.__new__(cls, name, bases, dict(classdict))
+
+
+# Note: apparently ABCMeta needs to be applied before GrammaredSequenceMeta
+@add_metaclass(GrammaredSequenceMeta)
+@add_metaclass(ABCMeta)
```
jairideout: can this be removed since `GrammaredSequenceMeta` inherits from `ABCMeta`?

skbio/sequence/_grammared_sequence.py
```diff
@@ -22,11 +23,71 @@
...
+
+            if len(cls.degenerate_chars & cls.nondegenerate_chars) > 0:
+                raise GrammaredSequenceException(
+                    "degenerate_chars and nondegenerate_chars must not share "
+                    "any characters for class %s" % name)
+
+        return cls
+
+
+class DisableSubclassingMeta(GrammaredSequenceMeta):
```
jairideout: Can this inherit from `type`? I don't understand why it's inheriting from `GrammaredSequenceMeta`.

This metaclass may make more sense in `skbio.util._meta` since it is very general and could be used elsewhere. I think we'd want to keep it package-private though so imports would come from `skbio.util._meta`.

skbio/sequence/_grammared_sequence.py
```diff
@@ -22,11 +23,71 @@
...
+    def __new__(cls, name, bases, classdict):
```
jairideout: The parameter names in this `__new__` are different from `__new__` in `GrammaredSequenceMeta`. Recommend using the most common Python convention for `__new__` (not sure what that is offhand).

skbio/sequence/_grammared_sequence.py
jairideout: This class is based on [this SO answer](http://stackoverflow.com/a/16056691/3776794). Can you add a comment or docstring saying where this came from? That's what we generally try to do even though it isn't strictly required by SO.

jairideout: I see that this class has a handy error message specific to the `sequence` subpackage. Ignore my comment about moving it to `skbio.utit._meta`, the nice error message is much more useful. We can always move the class if there are other parts of the codebase that need it.

skbio/sequence/_grammared_sequence.py
```diff
@@ -22,11 +23,71 @@
...
+        for b in bases:
+            if isinstance(b, DisableSubclassingMeta):
+                raise TypeError("Subclassing disabled for class %s. To create"
+                                " a custom sequence class, inherit directly"
+                                " from skbio.sequence.%s" %
+                                (b.__name__, GrammaredSequence.__name__))
+        return type.__new__(cls, name, bases, dict(classdict))
```
jairideout: Should this be `super` so that inheritance works correctly? I'm not sure what the MRO is here, is it short-circuiting the `ABCMeta` and `GrammaredSequenceMeta` validations?

skbio/sequence/_grammared_sequence.py
```diff
@@ -210,10 +312,11 @@ def _validate(self):
                 invalid_characters > 0)[0].astype(np.uint8).view('|S1'))
             raise ValueError(
                 "Invalid character%s in sequence: %r. \n"
-                "Lowercase letters are not used in IUPAC notation. You can "
+                "Note that lowercase letters are not valid by default. "
```
jairideout: There isn't really a default, how about the following?

*Use `lowercase` if your sequence contains lowercase characters not in the sequence's alphabet.*

skbio/sequence/_grammared_sequence.py
```diff
@@ -22,11 +23,71 @@
...
+                                " a custom sequence class, inherit directly"
+                                " from skbio.sequence.%s" %
+                                (b.__name__, GrammaredSequence.__name__))
+        return type.__new__(cls, name, bases, dict(classdict))
+
+
+# Note: apparently ABCMeta needs to be applied before GrammaredSequenceMeta
+@add_metaclass(GrammaredSequenceMeta)
+@add_metaclass(ABCMeta)
+class GrammaredSequence(Sequence):
```
jairideout: Can you make `default_gap_char` and `gap_chars` abstract? These are defaulting to IUPAC gap characters.

skbio/sequence/_grammared_sequence.py
```diff
@@ -22,11 +23,71 @@
 from ._sequence import Sequence
 
 
-class IUPACSequence(with_metaclass(ABCMeta, Sequence)):
-    """Store biological sequence data conforming to the IUPAC character set.
+class GrammaredSequenceException(Exception):
```
jairideout: What about using `TypeError` for these exceptions? That'd match `ABCMeta` and `DisableSubclassingMeta` and I think makes sense (an error defining a new type). We've been preferring using stdlib error types over custom exceptions if there's one that fits.

skbio/sequence/_protein.py
```diff
@@ -61,6 +64,17 @@ class Protein(IUPACSequence):
...
 
+    See Also
+    --------
+    GrammaredSequence
+
+    Notes
+    -----
+    Subclassing is disabled for Protein, because subclassing makes
+    it possible to change the alphabet, and certain methods rely on the
+    IUPAC alphabet. If a custom sequence is needed, inherit directly from
```
jairideout: *If a custom sequence alphabet is needed, ...*

skbio/sequence/_rna.py
```diff
@@ -64,6 +66,14 @@ class RNA(IUPACSequence, NucleotideMixin):
     See Also
     --------
     DNA
+    GrammaredSequence
+
+    Notes
+    -----
+    Subclassing is disabled for RNA, because subclassing makes
+    it possible to change the alphabet, and certain methods rely on the
+    IUPAC alphabet. If a custom sequence is needed, inherit directly from
```
jairideout: *If a custom sequence alphabet is needed, ...*

skbio/util/__init__.py
```diff
@@ -63,11 +63,12 @@
 from ._testing import (get_data_path, TestRunner,
                        assert_ordination_results_equal,
                        assert_data_frame_almost_equal)
+from ._decorator import classproperty
```
jairideout: Can you add `classproperty` to this file's docstring?

skbio/sequence/__init__.py
```diff
@@ -304,7 +304,9 @@
 from ._dna import DNA
 from ._rna import RNA
 from ._genetic_code import GeneticCode
+from ._grammared_sequence import GrammaredSequence
 
-__all__ = ['Sequence', 'Protein', 'DNA', 'RNA', 'GeneticCode']
+__all__ = ['Sequence', 'Protein', 'DNA', 'RNA', 'GeneticCode',
+           'GrammaredSequence']
```
jairideout: add `GrammaredSequence` to this file's list of classes in the docstring

skbio/sequence/tests/test_grammared_sequence.py
```diff
@@ -0,0 +1,604 @@
...
+            "name1": lambda x, _, __: str(x),
+            "name2": lambda x, _, __: len(x)
+        }
+
+
+class TestGrammaredSequence(TestCase):
+    def test_default_gap_must_be_in_gap_chars(self):
+        with six.assertRaisesRegex(
+            self, GrammaredSequenceException,
+            ("default_gap_char must be in gap_chars for class "
```
jairideout: these parens aren't needed here or in the tests below

skbio/sequence/tests/test_grammared_sequence.py
```diff
@@ -0,0 +1,604 @@
...
+import six
+
+from unittest import TestCase, main
+
+import numpy as np
+import numpy.testing as npt
+
+from skbio.sequence import GrammaredSequence
+from skbio.sequence._grammared_sequence import GrammaredSequenceException
+from skbio.util._decorator import classproperty
```
jairideout: Interesting that checklist.py didn't catch this... can you change to `from skbio.util import classproperty`?

jairideout:
Done reviewing, this looks great @anderspitman! Mostly minor comments/questions. Can you also update the changelog to indicate the addition of `skbio.sequence.GrammaredSequence` and `skbio.util.classproperty` and that `DNA`, `RNA`, and `Protein` are no longer inheritable?

skbio/sequence/_grammared_sequence.py
anderspitman: Refer to the discussion @ebolyen and I had in #1175 about metaclasses. Basically there are lots of issues with inheriting metaclasses. This is the only way I know of to make it work, even though it seems like the metaclass should be general.

jairideout: Ok, thanks. Can you add a comment here so it's clearer why the inheritance is set up this way?

anderspitman: Yeah can do. I'll also try to cut out some time to finish this out ASAP.

skbio/sequence/_grammared_sequence.py
anderspitman: [This](https://docs.python.org/2/reference/datamodel.html#customizing-class-creation) is the closest thing I found. Apparently other than `cls` the other parameters to `new` are optional, unless it's a metaclass in which case it's like in the link.

anderspitman: The only thing I don't like about that is using `dict`. I changed it to `dct`, hopefully that's ok.

skbio/sequence/_grammared_sequence.py
anderspitman: Honestly not sure. `super` seems to work though so we'll do that...

skbio/sequence/tests/test_grammared_sequence.py
anderspitman: Yeah that makes me nervous

skbio/sequence/_grammared_sequence.py
anderspitman: @jairideout this will introduce a bit of duplication between `DNA`, `RNA`, and `Protein`. That ok?

anderspitman: What do you think about making an intermediate class `IUPACGapSequence` which only defines `gap_chars` and `default_gap_char`? That gets rid of the duplication but I'm not sure it's worth making our inheritance tree deeper.

anderspitman:
@jairideout I've pretty much got this ready to go. Before I push, I hoped you and @ebolyen could weigh in on something. I was able to make `gap_chars` and `default_gap_char` abstract, but ran into an issue in the process. In order to prevent "`abstractproperty` not iterable" errors (@ebolyen knows all about that), I have to wrap our import-time metaclass logic in guard if statements, which basically makes it so we don't perform the checks unless we're dealing with a subclass of `GrammaredSequence`. Unfortunately, having those guard statements makes it possible to declare a class with invalid class properties, because the checks don't run if some of the abstract properties having been defined, but we can't guarantee all the abstract properties are defined until someone tries to instantiate an object.

Hope that makes sense. Let me know if you need clarification.

skbio/sequence/_grammared_sequence.py
jairideout: That's fine, can you just make sure the parameter names are consistent for GrammaredSequenceMeta and DisableSubclassingMeta?

skbio/sequence/_grammared_sequence.py
jairideout: The duplication is pretty minimal so it's probably simplest to copy the gap alphabet parts to those classes. Also fine to add an intermediate class if you want

jairideout:
That's a tricky problem... I guess for now we could force the class extending `GrammaredSequence` to be fully concrete at import time, otherwise an error is thrown if there are any abstract attributes. This is more restrictive than I'd like though... @ebolyen any ideas?

ebolyen:
@anderspitman I think I am confused. Wouldn't a subclass with `abstract` properties fail to instantiate by virtue of the `ABCMeta`? So having invalid properties isn't an issue. Subclassing this incomplete class again where we fill in the abstract properties would cause the metaclass to run the validation as nothing is abstract anymore. The only issue I can see is where someone modifies an abstract class after class-construction to remove the abstract properties resulting in a working class as far as the `ABCMeta` is concerned and where validation did not occur.

This is likely the reason `ABCMeta` does its validation at instantiation as that is the only "guaranteed" point at which you know everything about the class that would be instantiated. 

So I think it would work as is, and if we want to support every edge case, then import-time checks are a paradox anyways as I can always mutate the class before instantiation.

anderspitman:
@ebolyen I agree it's an edge case and I think we should be fine. By the time an object is instantiated, everything will have to be in order for it to work, which is the most important thing.

CHANGELOG.md
```diff
@@ -46,6 +46,8 @@ constructor. ([#6240](https://github.com/biocore/scikit-bio/issues/624))
 * ``TreeNode.to_array`` now supports replacing ``nan`` branch lengths in the resulting branch length vector with the value provided as ``nan_length_value``.
 * ``skbio.io.format.phylip`` now supports sniffing and reading strict, sequential PHYLIP-formatted files into ``skbio.Alignment`` objects. ([#1006](https://github.com/biocore/scikit-bio/issues/1006))
 * Added `default_gap_char` class property to ``DNA``, ``RNA``, and ``Protein`` for representing gap characters in a new sequence.
+* Replaced `skbio.sequence.IUPACSequence` with `skbio.sequence.GrammaredSequence` and added `skbio.util.classproperty`.
```
jairideout: `skbio.sequence.IUPACSequence` doesn't exist because `IUPACSequence` is private. Can you change the text to:

* Added `skbio.sequence.GrammaredSequence`, which can be inherited from to create grammared sequences with custom alphabets (e.g., for use with `TabularMSA`) ([#1175](https://github.com/biocore/scikit-bio/issues/1175))
* Added `skbio.util.classproperty` decorator

CHANGELOG.md
```diff
@@ -46,6 +46,8 @@ constructor. ([#6240](https://github.com/biocore/scikit-bio/issues/624))
...
+* DNA, RNA, and Protein are no longer inheritable.
```
jairideout: *DNA, RNA, and Protein are no longer inheritable because they assume an IUPAC alphabet.*

jairideout:
Commits got messed up in this PR, closing in favor of #1273. Will finish final review there.

Issue #967, opened 2015-06-26
add Stockholm format to I/O registry


jairideout:

jairideout:
Fixed in #1243 and #1270.

PR #1243, opened 2015-12-16
Adds Stockholm format to IO

Kleptobismol:


Kleptobismol:
All done guys, ready for review! @jairideout @anderspitman

skbio/io/_exception.py
```diff
@@ -84,6 +84,11 @@ class QSeqFormatError(FileFormatError):
     pass
 
 
+class StockholmFormatError(FileFormatError):
+    """Raised when a ``stockholm`` formatted file cannot be parsed."""
+    pass
```
anderspitman: `pass` shouldn't be necessary as long as there's a docstring. If it's not affecting coverage though I'd leave it in since it's more explicit.

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,415 @@
+"""
+Stockholm format (:mod:`skbio.io.format.stockholm`)
+===================================================
+
+.. currentmodule:: skbio.io.format.stockholm
+
+The Stockholm format is a multiple sequence alignment format written in Markup
+as opposed to a simple text-based format. Data is stored on seperate lines,
```
anderspitman: seperate -> separate

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,415 @@
...
+    J04373.1/6212-6234             UAAGUUCUCGAUCUUUAAAAUCG
+    M24803.1/1-23                  UAAGUUCUCGAUCUCUAAAAUCG
+    #=GC SS_cons                   .AAA....<<<<aaa....>>>>
+    //
+
+Format Support
+==============
+**Has Sniffer: Yes**
+
+**State: Experimental as of 0.4.0-dev.**
```
anderspitman: @jairideout shouldn't this be auto-generated (and also say 0.5.0)? I'm not as familiar with our decorator system; maybe it doesn't work for classes yet.

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,415 @@
...
+
++------+------+---------------------------------------------------------------+
+|Reader|Writer|                          Object Class                         |
++======+======+===============================================================+
+|Yes   |No    |:mod:`skbio.alignment.TabularMSA`                              |
++------+------+---------------------------------------------------------------+
+
+Format Specification
+====================
+Stockholm format contains two types of data: raw DNA, RNA, or Protein data and
```
anderspitman: This is a little confusing the way it's wording, saying there are two types of data then listing 3. Maybe reword it to something like: "Stockholm format contains two types of data: nucleotides (DNA or RNA) or Protein...". Or some other rearrangement.

skbio/io/format/stockholm.py
jairideout: We don't have the API stability decorators hooked up to I/O yet so this is the convention we're using for now.

skbio/io/_exception.py
jairideout: Also recommend leaving as-is since the other exceptions in this module are defined that way.

Kleptobismol: Done, followed @jairideout 's comment.

skbio/io/format/stockholm.py
Kleptobismol: Done.

skbio/io/format/stockholm.py
Kleptobismol: Done.

skbio/io/_exception.py
anderspitman: :+1:

skbio/io/format/stockholm.py
anderspitman: :+1:

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+
++------+------+---------------------------------------------------------------+
+|Reader|Writer|                          Object Class                         |
++======+======+===============================================================+
+|Yes   |No    |:mod:`skbio.alignment.TabularMSA`                              |
++------+------+---------------------------------------------------------------+
+
+Format Specification
+====================
+Stockholm format contains two types of data. The first can contain raw DNA,
```
anderspitman: "Stockholm format" -> "The Stockholm format"

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+two-letter marker which describes the metadata type, for example '#=GF' (see
+Metadata Types). Each metadata line also contains an extra two-letter tag,
+such as 'AS' or 'CC' which tells what type of data it precedes. All metadata
+is optional.
+
+Metadata Types
+++++++++++++++
+GF
+--
+Data relating to multiple sequence alignment as a whole, such as authors or
```
anderspitman: "to multiple sequence alignment" -> "to the multiple sequence alignment"

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+to.
+For example:
+
+.. code-block:: none
+
+    #=GR O31698/18-71 SS    CCCHHHHHHHHHHHHHHH..EEEEEEEE....EEEEEEEEHHH
+
+Example taken from [2]_.
+
+GC
```
anderspitman: The ordering of GF, GS, GR, and GC seems out of order. I would either group by things that affect the whole alignment and things that affect individual sequences, or group by things that relate to columns and things that don't. So for example I would rearrange this documentation to explain them in the order GF, GC, GS, GR; or GF, GS, GC, GR. Hopefully that makes sense. Just a flow thing.

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+
+stockholm = create_format('stockholm')
+
+
+@stockholm.sniffer()
+def _stockholm_sniffer(fh):
+    # Smells a Stockholm file if the first line contains 'Stockholm' and the
+    # version number.
+    line = next(fh)
+    if line[:15] == "# STOCKHOLM 1.0":
```
anderspitman: Is this guaranteed to always be exactly the first 15 characters of the line? For example is it valid for there to be 2 spaces after the "#" or something similar?

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+
+Example taken from [2]_.
+
+Examples
+========
+Suppose we have a Stockholm file:
+
+>>> import skbio.io
+>>> from io import StringIO
+>>> from skbio.sequence import RNA
```
anderspitman: You can import `RNA` and `TabularMSA` directly from skbio, like this:
`from skbio import RNA, TabularMSA`

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+    # Uses OrderedDict() to make sure dna_data isn't arranged randomly
+    dna_data = OrderedDict()
+    metadata = {}
+    positional_metadata = {}
+    seqs = []
+
+    # Retrieves data from file
+    for line in fh:
+        if not (line.startswith("#") or line.startswith("//") or
+                line.isspace() == True):
```
anderspitman: `line.isspace() == True` can be replaced with simply `line.isspace()`

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+def _stockholm_to_tabular_msa(fh, constructor=Protein):
+    # Uses OrderedDict() to make sure dna_data isn't arranged randomly
+    dna_data = OrderedDict()
+    metadata = {}
+    positional_metadata = {}
+    seqs = []
+
+    # Retrieves data from file
+    for line in fh:
+        if not (line.startswith("#") or line.startswith("//") or
```
anderspitman: I recommend moving this if into a function with a name that describes what it's checking. For example:

```python
if is_data_line(line):
    dna_data = _parse_stockholm_line_data(line, dna_data)
````

then you would have the function:

```python
def is_data_line(line):
    return not (line.startswith("#") or line.startswith("//") or line.isspace())
```
This makes the code more explicit and saves anyone reading it the few seconds it takes to figure out what it's supposed to do.

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+                line.isspace() == True):
+            dna_data = _parse_stockholm_line_data(line, dna_data)
+    fh.seek(0)
+
+    # Retrieves metadata from file
+    for line in fh:
+        if line.startswith("#=GF"):
+            metadata = _parse_stockholm_line_gf(line, metadata)
+        if line.startswith("#=GS"):
```
anderspitman: Pretty sure these should be `elif`s

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+@stockholm.reader(TabularMSA)
+def _stockholm_to_tabular_msa(fh, constructor=Protein):
+    # Uses OrderedDict() to make sure dna_data isn't arranged randomly
+    dna_data = OrderedDict()
+    metadata = {}
+    positional_metadata = {}
+    seqs = []
+
+    # Retrieves data from file
+    for line in fh:
```
anderspitman: It looks like this can be combined with the loop below. That way you don't have to do `fh.seek(0)` and loop twice.

anderspitman: Ah is this to maintain ordering because the DNA data always goes first? If so maybe add a comment.

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+            dna_data = _parse_stockholm_line_gs(line, dna_data)
+        if line.startswith("#=GR"):
+            dna_data = _parse_stockholm_line_gr(line, dna_data)
+        if line.startswith('#=GC'):
+            positional_metadata = _parse_stockholm_line_gc(line,
+                                                           positional_metadata)
+
+    for key in dna_data.keys():
+        # Sets blank dictionaries and lists to None instead
+        if not dna_data[key][1]:
```
anderspitman: I recommend using temp variables to clarify what these keys and indices represent. For example:

```python
metadata = dna_data[key][1]
positional_metadata = dna_data[key][2]

if not metadata:
    ...
```

This is more efficient since it doesn't have to perform the index lookup multiple times, but more importantly makes the code easier to read.

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+    # Constructs TabularMSA
+    return TabularMSA(seqs, metadata=metadata,
+                      positional_metadata=positional_metadata,
+                      index=list(dna_data.keys()))
+
+
+def _parse_stockholm_line_gf(line, metadata):
+    """Takes ``#=GF`` line and returns parsed data."""
+    line = _remove_newline(line.split(' ', 2))
+    if line[1] in metadata.keys():
```
anderspitman: Again recommend using a temp here:
```python
msa_id = line[1]
```

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+    else:
+        metadata[line[1]] = line[2]
+    return metadata
+
+
+def _parse_stockholm_line_gs(line, dna_data):
+    """Takes ``#=GS`` line and returns parsed data."""
+    line = _remove_newline(line.split(' ', 3))
+    if line[1] in dna_data.keys():
+        dna_data[line[1]][1][line[2]] = line[3]
```
anderspitman: Temps

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+    else:
+        raise StockholmFormatError("Markup line references nonexistent "
+                                   "data %r." % line[1])
+    return dna_data
+
+
+def _parse_stockholm_line_gr(line, dna_data):
+    """Takes ``#=GR`` line and returns parsed data."""
+    line = _remove_newline(line.split(' '))
+    if len(line) != 4:
```
anderspitman: Can you document why this is done and maybe put it in a function with a name that describes it?

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+        del(line[3:len(line)-1])
+    if line[1] in dna_data.keys():
+        if line[2] in dna_data[line[1]][2].keys():
+            raise StockholmFormatError("Found duplicate GR label %r associated"
+                                       " with data label %r" % (line[2],
+                                                                line[1]))
+        dna_data[line[1]][2][line[2]] = list(line[3])
```
anderspitman: Temps

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+                                   "data %r." % line[1])
+    return dna_data
+
+
+def _parse_stockholm_line_gc(line, positional_metadata):
+    """Takes ``#=GC`` line and returns parsed data."""
+    line = _remove_newline(line.split(' '))
+    if len(line) > 3:
+        del(line[2:len(line)-1])
+    if line[1] in positional_metadata.keys():
```
anderspitman: Same comment as above for temps

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+        raise StockholmFormatError("Found duplicate GC label %r." % (line[1]))
+    positional_metadata[line[1]] = list(line[2])
+    return positional_metadata
+
+
+def _parse_stockholm_line_data(line, dna_data):
+    """Takes data line and returns parsed data."""
+    line = line.split()
+    if line[0] not in dna_data.keys():
```
anderspitman: Same comment as above for temps

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,301 @@
...
+        dna_data[line[0]] = [line[1], {}, {}]
+    elif line[0] in dna_data.keys():
+        raise StockholmFormatError("Found multiple data lines under same "
+                                   "name: %r" % line[0])
+    return dna_data
+
+
+def _remove_newline(line):
+    """Removes '\n' from line and returns line."""
+    if '\n' in line[len(line)-1]:
```
anderspitman: Same comment as above for temps.

anderspitman:
@Kleptobismol @jairideout done with my review. Everything looks really good for the most part. Most of my comments revolve around a couple coding style recommendations to make things more explicit and readable. Great work!

skbio/io/format/stockholm.py
Kleptobismol: I ordered the data types in the order they usually appear in a Stockholm file, let me know if you'd rather I swap the order.

jairideout:
Thanks for reviewing @anderspitman! @Kleptobismol let @anderspitman know when you're ready for him to take another pass through.

Kleptobismol:
Should we try to fix that weird Travis error before I address his comments @jairideout ?

jairideout:
I'm not sure why the Py3 builds are timing out. I'd recommend making the changes and pushing again in case it's a transient issue with Travis.

Kleptobismol:
No problem.

skbio/io/format/stockholm.py
Kleptobismol: Done.

skbio/io/format/stockholm.py
Kleptobismol: Done.

skbio/io/format/stockholm.py
Kleptobismol: Done.

skbio/io/format/stockholm.py
Kleptobismol: In all examples I've found, yes.

skbio/io/format/stockholm.py
Kleptobismol: Done

skbio/io/format/stockholm.py
Kleptobismol: Done

skbio/io/format/stockholm.py
Kleptobismol: Done

skbio/io/format/stockholm.py
Kleptobismol: Done

skbio/io/format/stockholm.py
Kleptobismol: Done

skbio/io/format/stockholm.py
Kleptobismol: The index lookup would still need to happen twice.

skbio/io/format/stockholm.py
Kleptobismol: Done

skbio/io/format/stockholm.py
Kleptobismol: Done

skbio/io/format/stockholm.py
Kleptobismol: Oops, classic overthinking. The line, after getting parsed kept returning something like ``['#=GC', 'SS_cons', '', '', '', 'CAAGGGAAATACTACGGGAC']`` and for some reason it slipped my mind that I was using ``line.split(' ')`` when I could've been using ``line.split()``. Fixed both here and in GR.

skbio/io/format/stockholm.py
Kleptobismol: Done.

skbio/io/format/stockholm.py
Kleptobismol: Done.

skbio/io/format/stockholm.py
Kleptobismol: Done.

skbio/io/format/stockholm.py
Kleptobismol: Done.

Kleptobismol:
The build is still failing, but passes locally. If anyone has any ideas, I'm all ears.

antgonza:
IMOO is because of matplotlib: `warnings.warn('Matplotlib is building the font cache using fc-list. This may take a moment.')`. This can take a lot of time and might cause time outs. Not sure if the solution is to force a version or try to fix whatever is generating that warning.

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,308 @@
...
+Format Specification
+====================
+The Stockholm format contains two types of data. The first can contain raw DNA,
+RNA, or Protein data and the second is comprised of associated metadata.
+Raw data lines begin with an associated 'name', which
+often times is an id comprised of letters and numbers, such as
+'M24804.1/82-104' ([1]_). After the id and a few tab characters,
+the data is displayed. Metadata lines, however, begin with a '#' and a
+two-letter marker which describes the metadata type, for example '#=GF' (see
+Metadata Types). Each metadata line also contains an extra two-letter tag,
```
Kleptobismol: So you think it would be more clear to call them "features" instead of "tags"?

RNAer: yeah, that's what I would do.

jairideout:
I don't think it's matplotlib because this same warning appears in the Python 2.7 tests and they are passing. Also the warning appears at the top of the output when matplotlib is first imported. The timeout is happening somewhere in the middle of the tests.

Kleptobismol:
@anderspitman done addressing your comments.

jairideout:
The Travis timeout is unrelated to this PR, master is timing out also. I'm debugging in #1255.

anderspitman:
Looking better @Kleptobismol. I should probably clarify my comments about the temp variables. I shouldn't have tried to point out every instance I wanted you to change. It's more the general principle that's important, namely you want the code to explain itself as much as possible. One big way to do this is to avoid using "magic numbers", and instead have one central place that knows which indices match which parts of your data structure. There are several ways to do this. The most common is to have a "constant" that you use in place of the number, for example:

```python
SEQ_NAME_INDEX = 0
data_seq_name = line[SEQ_NAME_INDEX]
```

Unfortunately for what you're doing, this would end up pretty messy because you would need to declare a global variable to use as your index constant.

If we take a step back, the trickiness is coming from needing to keep track of which index is storing what for each element of your `OrderedDict`. If you look at this line:

```python
dna_data[data_seq_name] = [line[1], {}, {}]
```

Basically what we're doing is inventing a new object type. Each object of this type is a list where the first element is a string and the second and third are dicts. Each of these has a unique purpose which can be named. My recommendation would be to create a class to represent these objects. That would allow you to have named properties for each object of the class, rather than relying on indices. It could look something like this (untested):

```python
class DNAData(object):
    @property
    def seq_data(self):
        return self._seq_data
    @seq_data.setter
    def seq_data(self, data):
        self._seq_data = data

    @property
    def gc_features(self):
        return self._gc_features
    @gc_features.setter
    def gc_features(self, features):
        self._gc_features = features

    @property
    def gr_features(self):
        return self._gr_features
    @gr_features.setter
    def gr_features(self, features):
        self._gr_features = features

    def __init__(self, seq_data, gc_feature, gr_feature):
        self.seq_data = seq_data
        self.gc_feature = gc_feature
        self.gr_feature = gr_feature
```

I think you'll find if you implement something like this, the rest of the code will become much cleaner to work with.

For example, the line above would become:
```python
dna_data[data_seq_name] = DNAData(line[1], {}, {})
```

And here's an example (untested) of what rewriting one of your functions would look like using these principles.

```python
def _parse_stockholm_line_gr(line, dna_data):
    """Takes ``#=GR`` line and returns parsed data."""
    line = _remove_newline(line.split())
    data_seq_name = line[1]
    gr_feature = line[2]
    whatever_line3_is = line[3]
    if data_seq_name in dna_data.keys():
        seq_data = dna_data[data_seq_name]
        if gr_feature in seq_data.gr_features.keys():
            raise StockholmFormatError(
               "Found duplicate GR label %r associated"
               " with data label %r" % (gr_feature, data_seq_name))
        seq_data.gr_features[gr_feature] = list(whatever_line3_is)
    else:
        raise StockholmFormatError(
            "Markup line references nonexistent "
            "data %r." % data_seq_name)
    return dna_data
```


That ended up being a lot of words. Let me know if this makes sense.

jairideout:
To add to @anderspitman suggestions, a `collections.namedtuple` might be helpful here instead of creating a custom class. `namedtuple` provides index and named property access.

anderspitman:
Great suggestion

Kleptobismol:
@anderspitman  if you'd like to take a final pass, it's ready.

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,322 @@
...
+            dna_data = _parse_stockholm_line_gr(line, dna_data)
+        elif line.startswith('#=GC'):
+            positional_metadata = _parse_stockholm_line_gc(line,
+                                                           positional_metadata)
+
+    for key in dna_data.keys():
+        # Sets blank dictionaries and lists to None instead
+        # Note: _replace is not a private function, see
+        # https://docs.python.org/2/library/collections.html#namedtuple-factory
+        # -function-for-tuples-with-named-fields
```
anderspitman: Useful link! Maybe link directly to the subsection that mentions the underscores though:
https://docs.python.org/2/library/collections.html#namedtuple-factory-function-for-tuples-with-named-fields

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,322 @@
...
+
+
+def _parse_stockholm_line_gf(line, metadata):
+    """Takes ``#=GF`` line and returns parsed data."""
+    line = _remove_newline(line.split(' ', 2))
+    gf_feature = line[1]
+    if gf_feature in metadata.keys():
+        metadata[gf_feature] = metadata[gf_feature] + ' ' + line[2]
+    else:
+        metadata[line[1]] = line[2]
```
anderspitman: `line[1]` -> `gf_feature`

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,322 @@
...
+                      positional_metadata=positional_metadata,
+                      index=list(dna_data.keys()))
+
+
+def _parse_stockholm_line_gf(line, metadata):
+    """Takes ``#=GF`` line and returns parsed data."""
+    line = _remove_newline(line.split(' ', 2))
+    gf_feature = line[1]
+    if gf_feature in metadata.keys():
+        metadata[gf_feature] = metadata[gf_feature] + ' ' + line[2]
```
anderspitman: What is `line[2]` here?

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,322 @@
...
+    return metadata
+
+
+def _parse_stockholm_line_gs(line, dna_data):
+    """Takes ``#=GS`` line and returns parsed data."""
+    line = _remove_newline(line.split(' ', 3))
+    data_seq_name = line[1]
+    if data_seq_name in dna_data.keys():
+        if not dna_data[data_seq_name].metadata:
+            dna_data[data_seq_name].metadata[line[2]] = line[3]
```
anderspitman: What are `line[2]` and `line[3]`?

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,322 @@
...
+    """Takes ``#=GR`` line and returns parsed data."""
+    line = _remove_newline(line.split())
+    data_seq_name = line[1]
+    gr_feature = line[2]
+    if data_seq_name in dna_data.keys():
+        if gr_feature in dna_data[data_seq_name].pos_metadata.keys():
+            raise StockholmFormatError("Found duplicate GR label %r associated"
+                                       " with data label %r" % (gr_feature,
+                                                                data_seq_name))
+        dna_data[data_seq_name].pos_metadata[gr_feature] = list(line[3])
```
anderspitman: `line[3]` is pretty obvious here since everything on the left of the equals has a name.

anderspitman:
Done with another pass @Kleptobismol. It looks great! Your use of `_SeqData` is perfect. Just a couple minor comments then I'd say this is good to go.

skbio/io/format/stockholm.py
jairideout: That's the same link, it's just split over two lines for pep8

Kleptobismol: Done.

skbio/io/format/stockholm.py
Kleptobismol: Done.

skbio/io/format/stockholm.py
Kleptobismol: ``line[2]`` is the string associated with a GF data tag.
For example:
``#=GF DE CBS domain``
`line[1]` would be `DE` and `line[2]` would be `CBS domain`

Kleptobismol: I'll just assign it to a value to increase readability

Kleptobismol: Done.

skbio/io/format/stockholm.py
Kleptobismol: Done. Added name for `line[2]`

skbio/io/format/stockholm.py
Kleptobismol: :+1:

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,324 @@
...
+
+def _parse_stockholm_line_gf(line, metadata):
+    """Takes ``#=GF`` line and returns parsed data."""
+    line = _remove_newline(line.split(' ', 2))
+    gf_feature = line[1]
+    gf_feature_data = line[2]
+    if gf_feature in metadata.keys():
+        metadata[gf_feature] = metadata[gf_feature] + ' ' + gf_feature_data
+    else:
+        metadata[gf_feature] = line[2]
```
anderspitman: I missed this one:
`line[2]` -> `gf_feature_data`

Kleptobismol: Done.

Kleptobismol:
Thank you for reviewing, @anderspitman , @jairideout I think it's ready for the final pass!

jairideout:
Thanks for working on this @Kleptobismol and thanks for reviewing @anderspitman! I'll take a final pass through today or tomorrow.

skbio/io/format/stockholm.py
jairideout: Change to `0.4.1-dev`

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,324 @@
...
+                        unicode_literals)
+
+from collections import OrderedDict, namedtuple
+
+from skbio.alignment import TabularMSA
+from skbio.sequence import Protein
+from skbio.io import create_format, StockholmFormatError
+
+stockholm = create_format('stockholm')
+_SeqData = namedtuple("SeqData", ["seq", "metadata", "pos_metadata"])
```
jairideout: Can you add a comment describing this data structure? Also, `positional_metadata` might be a better field name to match the `Sequence` constructor.

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,324 @@
...
+
+@stockholm.sniffer()
+def _stockholm_sniffer(fh):
+    # Smells a Stockholm file if the first line contains 'Stockholm' and the
+    # version number.
+    try:
+        line = next(fh)
+    except StopIteration:
+        return False, {}
+    if line[:15] == "# STOCKHOLM 1.0":
```
jairideout: Minor change but avoids magic numbers:

```python
if line == "# STOCKHOLM 1.0\n":
```

skbio/io/format/stockholm.py
```diff
@@ -0,0 +1,324 @@
...
+        return True, {}
+
+    return False, {}
+
+
+@stockholm.reader(TabularMSA)
+def _stockholm_to_tabular_msa(fh, constructor=Protein):
```
jairideout: Can you make the reader not have a default `constructor`? The format doesn't specify a default and we can't reliably infer the sequence type. Defaulting to `constructor=None` and adding this check should do the trick (the I/O registry doesn't allow additional positional arguments):

```python
if constructor is None:
    raise ValueError("Must provide `constructor` parameter.")
elif not issubclass(constructor, IUPACSequence):
    raise TypeError("`constructor` must be a subclass of `IUPACSequence`.")
```

skbio/io/format/tests/test_stockholm.py
```diff
@@ -0,0 +1,148 @@
...
+# Copyright (c) 2013--, scikit-bio development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+from six import assertRaisesRegex
```
jairideout: Use `import six` and `six.assertRaisesRegex`

skbio/io/format/tests/test_stockholm.py
```diff
@@ -0,0 +1,148 @@
...
+
+import unittest
+
+from skbio import TabularMSA
+from skbio.io import StockholmFormatError
+from skbio.io.format.stockholm import (_stockholm_to_tabular_msa,
+                                       _stockholm_sniffer)
+from skbio import Protein, DNA, RNA
```
jairideout: Merge these imports with `from skbio import TabularMSA` line above

skbio/io/format/tests/test_stockholm.py
```diff
@@ -0,0 +1,148 @@
...
+from skbio.util import get_data_path
+
+
+class TestStockholmSniffer(unittest.TestCase):
+    def setUp(self):
+        self.positives = [get_data_path(e) for e in [
+            'stockholm_extensive',
```
jairideout: Can you add the other valid stockholm files here? The sniffer is only being tested against one "positive" file.

skbio/io/format/tests/test_stockholm.py
```diff
@@ -0,0 +1,148 @@
...
+            ]]
+
+        self.negatives = [get_data_path(e) for e in [
+            'stockholm_data_only',
```
jairideout: Can you rename this file to something like `stockholm_no_header` or `stockholm_missing_header` to better reflect this test file's purpose? `stockholm_data_only` could be interpreted as meaning "a stockholm file without metadata" (which is valid).

skbio/io/format/tests/test_stockholm.py
```diff
@@ -0,0 +1,148 @@
...
+            'empty',
```
jairideout: Can you add `whitespace_only` to this list?

skbio/io/format/tests/test_stockholm.py
```diff
@@ -0,0 +1,148 @@
...
+        fp = get_data_path('stockholm_invalid_nonexistent_gr')
+        with assertRaisesRegex(self, StockholmFormatError,
+                               'Markup line references.*RL1355.'):
+            _stockholm_to_tabular_msa(fp, constructor=RNA)
+        fp = get_data_path('stockholm_invalid_nonexistent_gs')
+        with assertRaisesRegex(self, StockholmFormatError,
+                               'Markup line references.*AC14.'):
+            _stockholm_to_tabular_msa(fp, constructor=RNA)
+
+    def test_duplicate_data_error(self):
```
jairideout: Can you rename this test case and test file to something like `test_duplicate_sequence_names` and `stockholm_duplicate_sequence_names`, respectively?

skbio/io/format/stockholm.py
anderspitman: Are Stockholm files guaranteed to use Unix line endings?

skbio/io/format/stockholm.py
Kleptobismol: Done.

skbio/io/format/stockholm.py
Kleptobismol: Done.

skbio/io/format/stockholm.py
Kleptobismol: Done. Will also add tests for these errors.

skbio/io/format/tests/test_stockholm.py
Kleptobismol: Done.

skbio/io/format/tests/test_stockholm.py
Kleptobismol: Done.

skbio/io/format/tests/test_stockholm.py
Kleptobismol: Done.

skbio/io/format/tests/test_stockholm.py
Kleptobismol: Done.

skbio/io/format/tests/test_stockholm.py
Kleptobismol: Done.

skbio/io/format/tests/test_stockholm.py
Kleptobismol: Done.

skbio/io/format/stockholm.py
jairideout: Good catch! We should definitely support this, though I see that other readers (stockholm, phylip, qseq, blast7) have the same problem. Will chat with @ebolyen about this and follow up. @Kleptobismol in the meantime hold off on this change, I'm not sure of the safest/correct way to handle it.

Jorge-C: If the file has been open in Py2 with the `'U'` mode, the program will see unix line endings no matter what the file uses, so the sniffers work well in that case.

We can guarantee that when we get a file name and  skbio opens the file, but given a handle we probably need to add something like `line.rstrip()` for safety. It's a minimal change and deals with any whitespace inconsistencies at the line end. Any other solutions you can think of?

Any way, every reader should be updated with the same solution :)

jairideout: @ebolyen and I chatted and decided to remove the `newline` parameter from `read` so that universal newline support is always enabled when reading, such that all newline characters are transformed into `\n`. This is currently the default behavior but can be overridden by the user, which will cause problems here and in the readers I mentioned above if something other than `\n` is used.

Writers will still have the `newline` parameter to control how `\n` is transformed on write. By default, `\n` will be transformed into the system's native newline character, and we can still safely give users control over the newline character that gets written.

This will allow devs to implement sniffers, readers, and writers in a newline-agnostic way by always using `\n` in code.

I'll work on implementing these changes in a separate PR. @Kleptobismol you can make the suggested change here because it will be a safe operation once this is fixed upstream.

Thanks for catching this @anderspitman!

skbio/io/format/tests/test_stockholm.py
```diff
@@ -0,0 +1,148 @@
...
+    def test_positives(self):
+        for fp in self.positives:
+            self.assertEqual(_stockholm_sniffer(fp), (True, {}))
+
+    def test_negatives(self):
+        for fp in self.negatives:
+            self.assertEqual(_stockholm_sniffer(fp), (False, {}))
+
+
+class TestStockholmReader(unittest.TestCase):
```
jairideout: Can you add the following reader test cases?

- interleaved stockholm file (not currently supported): this should raise the same error as reading a file with duplicate sequence names. Example interleaved file [here](http://www.bioperl.org/wiki/Stockholm_multiple_alignment_format), but note that we can't use that data due to license incompatibilities, so just make up a small example in interleaved format.
- `stockholm_data_only`, `empty`, and `whitespace_only` files raise error stating that required header is missing
- file missing `//` raises an error
- file with blank and/or whitespace-only lines works correctly (those lines are ignored)
- file with unrecognized markup line raises an error (e.g., `#=GZ foo bar`)
- file with malformed metadata markup line raises an error (e.g., `#=GF foo \n`)
- file with malformed sequence data line raises an error (e.g., `my-seq \n`)
- file with sequences of differing length raises error
- file with multiple stockholm alignments only reads the first alignment
- file with GF/GC/GS/GR metadata in recommended placement vs. different placement results in the same `TabularMSA`
- file with GC or GR metadata of wrong length raises an error

skbio/io/format/tests/test_stockholm.py
```diff
@@ -0,0 +1,148 @@
...
+        fp = get_data_path('stockholm_duplicate_data')
+        with assertRaisesRegex(self, StockholmFormatError,
+                               'Found multiple data.*ASR132'):
+            _stockholm_to_tabular_msa(fp, constructor=RNA)
+
+    def test_no_data_error(self):
+        fp = get_data_path('empty')
+        with assertRaisesRegex(self, StockholmFormatError, 'No data present'):
+            _stockholm_to_tabular_msa(fp)
+        fp = get_data_path('stockholm_no_data')
```
jairideout: I think the reader should support an empty stockholm file since it can be loaded into an empty `TabularMSA` object. Can you add tests for loading a file with no sequences or metadata:

```
# STOCKHOLM 1.0
//
```

and a file without sequences but with GF metadata?

skbio/io/format/tests/test_stockholm.py
```diff
@@ -0,0 +1,148 @@
...
+                          RNA('UGCUAAGAGUGGGGAUGAUU'),
+                          RNA('GCCACAACCGAUUAGAUAGA'),
+                          RNA('UUAGAAACCGAUGGACCGAA')],
+                         metadata={'AC': 'G2134T23', 'ID': 'ARD'},
+                         positional_metadata=(
+                         {'AC_cons': list('GGGACUGGACAUCUAUUCAG')}),
+                         index=['RTC2231', 'RTF2124', 'RTH3322', 'RTB1512'])
+        self.assertEqual(msa, exp)
+
+    def test_stockholm_runon_gf(self):
```
jairideout: Can you also add a test for run-on GS metadata?

jairideout:
@Kleptobismol this looks great! I had some inline comments, mostly focused on adding more unit tests to exercise some border cases. I haven't reviewed the code or docs in detail, will do that on my next pass.

One other thing I'd like to resolve is handling of trees. The format specifies:

```
For embedding trees:

    #=GF NH <tree in New Hampshire eXtended format>

    #=GF TN <Unique identifier for the next tree>
Notes:
* A tree may be stored on multiple #=GF NH lines.
* If multiple trees are stored in the same file, each tree must be preceded by a #=GF TN line with a unique tree identifier. If only one tree is included, the #=GF TN line may be omitted.
```

I think NH and TN are the only features that need to be handled differently than simple key/value pairs in GF metadata.

Here's one way to store the tree(s) in the `TabularMSA`'s `metadata` dictionary (all other GF metadata are loaded into this too):

Single tree, no ID:

```python
metadata = {
  'NH': 'tree in NHX format'
}
```

Single tree, has ID:

```python
metadata = {
  'NH': {
    'tree-id': 'tree in NHX format'
  }
}
```

Multiple trees (must have IDs):

```python
metadata = {
  'NH': {
    'tree-id-1': 'tree in NHX format',
    'tree-id-2': 'tree in NHX format'
  }
}
```

@ebolyen @anderspitman @Kleptobismol does this scheme/convention make sense? We could also store trees under `'TN'` but the schemes I could think of weren't as clean.

One other change in behavior is repeated NH lines. Other repeated GF features are treated as text and concatenated with spaces. For trees stored in NHX format, it's probably safer to concatenate without adding a space to avoid introducing errors into the NHX string.

Does anyone have an example Stockholm file with a tree split over multiple NH lines, or a file with multiple trees? That'd be useful for testing.

skbio/io/format/stockholm.py
anderspitman: Coincidentally just a few hours after writing that I ended up with almost this exact issue for a school assignment. Teacher provided an input file with `\r`'s. Still took a bit to track down the problem.

ebolyen: @Jorge-C Just as a point of information, the registry doesn't allow the user to specify the open mode, additionally we use `io.open` instead (which doesn't have the defunct `'U'` mode and uses `newline`). This let's us compose the file objects for sniffing in binary and textual contexts.

ebolyen:
@jairideout I agree with the conventions listed.

jairideout:
Thanks for taking a look @ebolyen! @Kleptobismol can you update the reader to handle trees as described above? We'll want some test cases that exercise:

- single tree, no ID
- single tree, has ID
- multiple trees
- tree split across multiple NH lines
- duplicate tree IDs raises error

In my previous comment I mentioned:

> One other change in behavior is repeated NH lines. Other repeated GF features are treated as text and concatenated with spaces. For trees stored in NHX format, it's probably safer to concatenate without adding a space to avoid introducing errors into the NHX string.

This is incorrect -- when joining features that are split across multiple GF lines, we should just concatenate the strings without adding a space. The example Stockholm files I've looked at all have a trailing space for text-like features, I missed this somehow.

skbio/io/format/stockholm.py
Jorge-C: Thanks @ebolyen!

skbio/io/format/tests/test_stockholm.py
Kleptobismol: Done.

skbio/io/format/tests/test_stockholm.py
Kleptobismol: Done.

skbio/io/format/tests/test_stockholm.py
Kleptobismol: Interleaved item done

Kleptobismol: Header item done

Kleptobismol: Footer item done

Kleptobismol: Blank/whitespace lines item done

Kleptobismol: Unrecognized markup line item done

Kleptobismol: Malformed metadata item done

Kleptobismol: Malformed sequence data item done

Kleptobismol: Differing sequence length item done

Kleptobismol: Multiple alignment item done

Kleptobismol: Reccomended placement item done

Kleptobismol: GR/GC metadata length item done

Kleptobismol:
Thank you all for your interest in reviewing my code! @jairideout I addressed all your comments, so it's ready for a final check.

jairideout:
Thanks for making these changes @Kleptobismol! I'll take another pass through soon.

The RDA and composition test failures are unrelated to your changes (see #1267). Note that there is a legitimate Stockholm test failure in the build.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Version 0.4.1-dev (changes since 0.4.1 go here)
 
 ### Features
+* Added `skbio.io.format.stockholm` for reading Stockholm files into a `TabularMSA`. ([#937](https://github.com/biocore/scikit-bio/issues/967))
```
jairideout: change to #967

jairideout:
Thanks for all your work on this @Kleptobismol!

PR #1242, opened 2015-12-13
BUG: Addressing #1241

mortonjt:
Quick fix addressing #1241 

Thanks for pointing out this issue @johnchase .  Let me know if there is anything more that should be improved on the documentation.

PR #1245, opened 2015-12-17
Warning fixes

ebolyen:
This PR does 3 things:

- Drops the version of HTTPretty because their latest release has a syntax error in Py3
- Subclass all scikit-bio warnings from `SkbioWarning` this allows us to filter only those warnings in the test. Ideally this will make real warnings more noticeable.
- Fix a bunch of warnings throughout the codebase so things don't fall apart come `3.6` or new `pandas`/`scipy` release. 

@gregcaporaso could you review? The `HTTPretty` part is blocking #1242

### DIFF
--- a/skbio/util/_decorator.py
+++ b/skbio/util/_decorator.py
@@ -13,6 +13,7 @@ import textwrap
 import decorator
 
 from ._exception import OverrideError
+from ._warning import DeprecationWarning as SkbioDeprecationWarning
 
 
 class _state_decorator(object):
@@ -254,7 +255,7 @@ class deprecated(_state_decorator):
             warnings.warn('%s is deprecated as of scikit-bio version %s, and '
                           'will be removed in version %s. %s' %
                           (func.__name__, self.as_of, self.until, self.reason),
-                          DeprecationWarning)
+                          SkbioDeprecationWarning)
             # args[0] is the function being wrapped when this is called
             # after wrapping with decorator.decorator, but why???
             return func(*args[1:], **kwargs)
--- a/ci/pip_requirements.txt
+++ b/ci/pip_requirements.txt
@@ -1,4 +1,4 @@
-HTTPretty
+HTTPretty <= 0.8.10
 bz2file
 contextlib2
 coveralls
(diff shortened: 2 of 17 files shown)


skbio/util/_decorator.py
```diff
@@ -13,6 +13,7 @@
 import decorator
 
 from ._exception import OverrideError
+from ._warning import DeprecationWarning as SkbioDeprecationWarning
```
gregcaporaso: Would this make it harder for someone to ignore this warning, since ``SkbioDeprecationWarning`` isn't importable?

ebolyen: Good point, I'll bring it up to `skbio.util` like `SkbioWarning`.

gregcaporaso: :+1: Looks good other than that.

ebolyen: Wait never mind, I subclass `DeprecationWarning` for that very reason. So it wouldn't change anything for the user.

gregcaporaso: @ebolyen and I chatted offline and agree this is fine as-is.

jairideout:
:+1: this is great!

Can you add a changelog mention about `SkbioWarning`? I think that's the only public-facing change. Also I'm seeing a couple `RuntimeWarning` from PCoA tests, is that something we want to silence too?

ebolyen:
Just added the changelog to #1246.
As for the `RuntimeWarning` I left those out on purpose... they really should get fixed/updated. We surely can come up with a different example that doesn't flag our warning out of the box.

jairideout:
Thanks! Agree about changing tests to not raise the warning

On Friday, December 18, 2015, Evan Bolyen <notifications@github.com> wrote:

> Just added the changelog to #1246
> <https://github.com/biocore/scikit-bio/pull/1246>.
> As for the RuntimeWarning I left those out on purpose... they really
> should get fixed/updated. We surely can come up with a different example
> that doesn't flag our warning out of the box.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/biocore/scikit-bio/pull/1245#issuecomment-165645778>.
>


-- 
Sent from my phone.

Issue #1256, opened 2016-01-13
Python 3 Travis build timeout


jairideout:
Master branch Python 3 builds are timing out on Travis. Able to reproduce locally after updating packages in `ci/pip_requirements.txt`.

jairideout:
Still not sure what's going on. I was only able to reproduce this locally after updating packages in `ci/pip_requirements.txt` and it only happens in Python 3.4/3.5. The issue didn't surface when I had only updated the conda packages in `ci/conda_requirements.txt`.

The test that's hanging indefinitely is `python skbio/io/format/tests/test_fastq.py TestReaders.test_fastq_to_generator_solexa`. Commenting out this test "fixes" the issue. Running the code in this test from an IPython session freezes so it isn't the testing framework.

I think this has something to do with the changes to generators, `StopIteration`, and `RuntimeError` detailed in [this PEP](https://www.python.org/dev/peps/pep-0479/). The code freezes when raising a `NotImplementedError`, which is a subclass of `RuntimeError`. Changing the error type to something other than `RuntimeError` fixes the issue (e.g., `ValueError` works fine).

If anyone has time to dig in that'd be great, otherwise I'll keep working on it on Friday.

jairideout:
Accidentally closed from commit.

ebolyen:
This bug is truly awful. :-1: :x: :100:

ebolyen:
For anyone who wants to play along, this example will spin-lock if run as a unittest, or in IPython. Normal Python interpreter does not produce this bug.
```python
import unittest
import contextlib

@contextlib.contextmanager
def single_context():
    yield

class Example(unittest.TestCase):
    def test(self):
        with contextlib.ExitStack() as stack:
            stack.enter_context(single_context())
            raise NotImplementedError("!")

if __name__ == '__main__':
    unittest.main()
```

This appears to be related to the following:
[`contextlib.ExitStack introduces a cycle in exception __context__` ](https://bugs.python.org/issue25786)


... :-(

jairideout:
This is a bug in Python and there isn't much we can do to fix it. Bumping this issue to 0.5.0 milestone since we have a workaround (contextlib2 version cap).

kdmurray91:
I've just hit this bug running the tests as a part of the Debian packaging. Aside from disabling this test, is there any workaround independent of capping contextlib2's version (which is verboten in Debian)

ebolyen:
[...]

PR #1260, opened 2016-02-01
Addresses #1259

mortonjt:
Swapping out the mann-whitney U-test with a t-test.

Note, this only swaps out this statistical test in the broken test cases.

### DIFF
--- a/skbio/stats/tests/test_composition.py
+++ b/skbio/stats/tests/test_composition.py
@@ -12,7 +12,6 @@ from unittest import TestCase, main
 import numpy as np
 import numpy.testing as npt
 import pandas.util.testing as pdt
-
 from numpy.random import normal
 import pandas as pd
 import scipy
@@ -336,7 +335,6 @@ class AncomTests(TestCase):
             [10, 11, 10, 10, 10, 11],
             [10, 13, 10, 10, 10, 12]]).T
         self.cats1 = pd.Series([0, 0, 0, 1, 1, 1])
-
         # Real valued data with 2 groupings
         D, L = 40, 80
         np.random.seed(0)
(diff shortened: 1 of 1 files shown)


Jorge-C:
I was about to complain that swapping a robust Mann–Whitney U test with a t-test doesn't seem valid, then I read your note, and it's just a small test case edit :)

Before changing this I feel we should really understand what changed between versions, as Jai mentioned in #1259. Were the values wrong previously?

mortonjt:
I'm not exactly sure what is going on.  Looks like these changes were made in this [commit](https://github.com/scipy/scipy/commit/424128b563f291c71f1a123071f37e1cb22312bd)

Jorge-C:
Thanks! I'll have a look too.

jairideout:
Thanks @mortonjt and @Jorge-C for looking into this. I agree it would be nice to figure out why we're seeing differences between versions. However, `mannwhitneyu` has had (and still has) a slew of bugs related to tie correction and/or p-values (take a look at issues/PRs related to this function [here](https://github.com/scipy/scipy/search?q=mann&type=Issues&utf8=%E2%9C%93)). IMO `mannwhitneyu`'s buggy behavior is outside the scope of scikit-bio because we aren't advertising usage of `mannwhitneyu` in the docs and the function was only used in a couple of unit tests to make sure that passing an alternative significance test works as expected. In other words, the user-supplied significance test is buggy but ANCOM is not.

I think this PR is good to go but will hold off on merging in case you guys want more time in tracking down the differences, or if there are any other objections to merging. Thanks again!

mortonjt:
I have no objections.  @JorgeC, what do you think?

Jorge-C:
Thanks a lot for the summary Jai! :+1: from me too (I've restarded the build to see if the timeout disappears).

jairideout:
Sounds good, will merge! Note that master is timing out for Python 3 builds (#1256).

PR #1263, opened 2016-02-03
Fixes #1153

### LINKED ISSUE #1153: test_genbank fails on non-US locale
I was unable to pass all test for `scikit-bio` due to an assumed US locale and `datetime.strftime` in `skbio/io/format/genbank.py`. The error can be reproduced using:

    LC_TIME=de_DE.UTF-8 make test

will fail on '18-DEZ-2001' != '18-DEC-2001', while

    LC_TIME=en_US.UTF-8 make test

runs without error. I'm not sure if this is a bug i.e. if genbank requires US date formatting. If US formatting is required there does not seem to be a 100% waterproof solution. You could probably

    import locale
    locale.setlocale(locale.LC_TIME, "en_US")

but this will fail if the `en_US` locale is not installed on the system. Using a third party library such as `babel` would add another dependency but would also provide a cleaner solution to handling dates

    >>> from datetime import datetime
    >>> dt = datetime(2001, 12, 18, 0, 0)
    >>> from babel.dates import format_date
    >>> import locale
    >>> locale.setlocale(locale.LC_TIME, 'de_DE') # set 'wrong' locale
    >>> format_date(dt, 'yyyy-MMM-dd').upper() # will produce unintended result
    u'2001-DEZ.-18'
    >>> format_date(dt, 'yyyy-MMM-dd', locale='en_US').upper() # will produce expected result
    u'2001-DEC-18'

jairideout:
When reading GenBank files, the date field of the LOCUS line is no longer
parsed into a `datetime.datetime` object and is left as a string. When writing
GenBank files, the locus date metadata is expected to be a string instead of a
`datetime.datetime` object.

Fixes #1153.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,10 @@
 ### Backward-incompatible changes [stable]
 
 ### Backward-incompatible changes [experimental]
+* `skbio.io.format.genbank`: When reading GenBank files, the date field of the LOCUS line is no longer parsed into a `datetime.datetime` object and is left as a string. When writing GenBank files, the locus date metadata is expected to be a string instead of a `datetime.datetime` object ([#1153](https://github.com/biocore/scikit-bio/issues/1153))
 
 ### Bug fixes
-* Changed `skbio.stats.composition.multiplicative_replacement` to raise an error whenever a large value of `delta` is chosen.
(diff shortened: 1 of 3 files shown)


RNAer:
:+1:

ebolyen:
Thanks @jairideout!

PR #1273, opened 2016-02-15
Issue 1175 new

anderspitman:


CHANGELOG.md
```diff
@@ -55,6 +55,8 @@ constructor. ([#6240](https://github.com/biocore/scikit-bio/issues/624))
 * ``TreeNode.to_array`` now supports replacing ``nan`` branch lengths in the resulting branch length vector with the value provided as ``nan_length_value``.
 * ``skbio.io.format.phylip`` now supports sniffing and reading strict, sequential PHYLIP-formatted files into ``skbio.Alignment`` objects. ([#1006](https://github.com/biocore/scikit-bio/issues/1006))
 * Added `default_gap_char` class property to ``DNA``, ``RNA``, and ``Protein`` for representing gap characters in a new sequence.
+* Replaced `skbio.sequence.IUPACSequence` with `skbio.sequence.GrammaredSequence` and added `skbio.util.classproperty`.
```
jairideout: `skbio.sequence.IUPACSequence` doesn't exist because `IUPACSequence` is private. Can you change the text to:

* Added `skbio.sequence.GrammaredSequence`, which can be inherited from to create grammared sequences with custom alphabets (e.g., for use with `TabularMSA`) ([#1175](https://github.com/biocore/scikit-bio/issues/1175))
* Added `skbio.util.classproperty` decorator

CHANGELOG.md
```diff
@@ -55,6 +55,8 @@ constructor. ([#6240](https://github.com/biocore/scikit-bio/issues/624))
...
+* DNA, RNA, and Protein are no longer inheritable.
```
jairideout: *DNA, RNA, and Protein are no longer inheritable because they assume an IUPAC alphabet.*

anderspitman:
@jairideout any idea? I've never seen that build error before.

anderspitman:
Docs build fine locally.

skbio/sequence/_grammared_sequence.py
```diff
@@ -122,6 +240,7 @@ def gap_chars(cls):
         """
         return set('-.')
```
jairideout: Can all abstract properties be defined with `pass` instead of sets? Since these are abstract we really shouldn't have defaults.

ebolyen:
@anderspitman Looks like it is related to a new release: [#1565](http://www.sphinx-doc.org/en/stable/changes.html#release-1-3-5-released-jan-24-2016). Not related to this PR.

anderspitman:
@ebolyen what's our standard procedure in this situation? Do we tell travis to use a specific version of sphinx? Or try to work around the issue?

ebolyen:
I'll see if I can fix it in another PR, there could be a couple docs that are broken in the same (relatively trivial) way.

skbio/sequence/_grammared_sequence.py
```diff
@@ -22,11 +22,79 @@
 from ._sequence import Sequence
 
 
-class IUPACSequence(with_metaclass(ABCMeta, Sequence)):
-    """Store biological sequence data conforming to the IUPAC character set.
+class GrammaredSequenceException(TypeError):
```
jairideout: Sorry I wasn't clearer before, I'd just remove this custom exception type and raise `TypeError` directly in the validations.

skbio/sequence/__init__.py
```diff
@@ -32,6 +32,7 @@
    RNA
    Protein
    GeneticCode
+   GrammaredSequence
```
jairideout: This class makes more sense to appear after `Sequence` and before `DNA`

skbio/alignment/_tabular_msa.py
```diff
@@ -19,7 +19,7 @@
 
 from skbio._base import SkbioObject, MetadataMixin, PositionalMetadataMixin
 from skbio.sequence import Sequence
-from skbio.sequence._iupac_sequence import IUPACSequence
```
jairideout: There are still 4 references to `IUPACSequence` in this file

skbio/io/format/stockholm.py
```diff
@@ -351,8 +351,9 @@ def _stockholm_to_tabular_msa(fh, constructor=None):
     if constructor is None:
         raise ValueError("Must provide `constructor` parameter.")
     # Checks that contructor parameter is supported
-    elif not issubclass(constructor, IUPACSequence):
-        raise TypeError("`constructor` must be a subclass of `IUPACSequence`.")
+    elif not issubclass(constructor, GrammaredSequence):
+        raise TypeError("`constructor` must be a subclass of"
```
jairideout: Add a space between `of` and `GrammaredSequence`

jairideout:
@anderspitman, the Travis build failures are fixed upstream (#1274, thanks @ebolyen!).

skbio/sequence/_dna.py
```diff
@@ -132,6 +142,35 @@ def degenerate_map(cls):
             "H": set("ACT"), "V": set("ACG"), "N": set("ACGT")
         }
 
+    @classproperty
+    def default_gap_char(cls):
+        """Gap character to use when constructing a new gapped sequence.
```
jairideout: The docstring will be inherited if you use `overrides`:

```python
@classproperty
@overrides(GrammaredSequence)
def default_gap_char(cls):
    return '-'
```

Same comment below for `gap_chars`.

skbio/sequence/_dna.py
```diff
@@ -132,6 +142,35 @@ def degenerate_map(cls):
...
+        Returns
+        -------
+        str
+            Default gap character.
+
+        """
+        return '-'
+
+    @classproperty
+    @stable(as_of='0.4.0')
```
jairideout: Can you remove this API stability decorator, as well as the API decorators for `complement_map`, `nondegenerate_chars`, and `degenerate_map`? I just noticed that the API stability description is being duplicated in the final docstring.

skbio/sequence/_protein.py
```diff
@@ -109,19 +123,19 @@ def _stop_codes(cls):
 
     @classproperty
     @stable(as_of="0.4.0")
```
jairideout: remove this line

skbio/sequence/_protein.py
```diff
@@ -109,19 +123,19 @@ def _stop_codes(cls):
...
-    @overrides(IUPACSequence)
+    @overrides(GrammaredSequence)
     def alphabet(cls):
         return super(Protein, cls).alphabet | cls.stop_chars
 
     @classproperty
     @stable(as_of="0.4.0")
```
jairideout: remove this line

skbio/sequence/_protein.py
```diff
@@ -109,19 +123,19 @@ def _stop_codes(cls):
...
-    @overrides(IUPACSequence)
+    @overrides(GrammaredSequence)
     def nondegenerate_chars(cls):
         return set("ACDEFGHIKLMNPQRSTVWY")
 
     @classproperty
     @stable(as_of="0.4.0")
```
jairideout: remove this line

skbio/sequence/_protein.py
```diff
@@ -141,6 +155,34 @@ def stop_chars(cls):
         """
         return set('*')
 
+    @classproperty
+    def gap_chars(cls):
+        """Return characters defined as gaps.
```
jairideout: remove docstring, use `overrides`

skbio/sequence/_protein.py
```diff
@@ -141,6 +155,34 @@ def stop_chars(cls):
...
+        -------
+        set
+            Characters defined as gaps.
+
+        """
+        return set('-.')
+
+    @classproperty
+    def default_gap_char(cls):
+        """Gap character to use when constructing a new gapped sequence.
```
jairideout: remove docstring, use `overrides`

skbio/sequence/_rna.py
```diff
@@ -118,13 +128,13 @@ def complement_map(cls):
 
     @classproperty
     @stable(as_of="0.4.0")
```
jairideout: remove this line (`stable`)

also remove `stable` line from `complement_map` above (I can't comment on the specific line b/c it's not included in the diff)

skbio/sequence/_rna.py
```diff
@@ -118,13 +128,13 @@ def complement_map(cls):
...
-    @overrides(IUPACSequence)
+    @overrides(GrammaredSequence)
     def nondegenerate_chars(cls):
         return set("ACGU")
 
     @classproperty
     @stable(as_of="0.4.0")
```
jairideout: remove this line (`stable`)

skbio/sequence/_rna.py
```diff
@@ -132,6 +142,35 @@ def degenerate_map(cls):
             "H": set("ACU"), "V": set("ACG"), "N": set("ACGU")
         }
 
+    @classproperty
+    def default_gap_char(cls):
+        """Gap character to use when constructing a new gapped sequence.
```
jairideout: remove docstring, use `overrides`

skbio/sequence/_rna.py
```diff
@@ -132,6 +142,35 @@ def degenerate_map(cls):
...
+        str
+            Default gap character.
+
+        """
+        return '-'
+
+    @classproperty
+    @stable(as_of='0.4.0')
+    def gap_chars(cls):
+        """Return characters defined as gaps.
```
jairideout: remove docstring, use `overrides`

skbio/sequence/_rna.py
```diff
@@ -132,6 +142,35 @@ def degenerate_map(cls):
...
+        Returns
+        -------
+        str
+            Default gap character.
+
+        """
+        return '-'
+
+    @classproperty
+    @stable(as_of='0.4.0')
```
jairideout: remove this line (`stable`)

skbio/sequence/_grammared_sequence.py
```diff
@@ -22,11 +22,79 @@
...
+    pass
+
+
+class GrammaredSequenceMeta(ABCMeta, type):
+    def __new__(mcs, name, bases, dct):
+        cls = super(GrammaredSequenceMeta, mcs).__new__(mcs, name, bases, dct)
+
+        # Only perform validation on classes that aren't abstract.
+        if (type(cls.default_gap_char) is not abstractproperty and
```
jairideout: I think this should check that *all* abstract properties are defined, not just `default_gap_char` and `gap_chars`?

Another way to do this would be to validate what you can and skip the validations that can't be performed due to abstractness. This is probably a better way to go because then you can validate what's available, hopefully notifying the user sooner than when the class definition is fully concrete. I'd define some bools at the top and use those as guards to the validations:

```python
default_gap_char_defined = type(cls.default_gap_char) is not abstractproperty
gap_chars_defined = type(cls.gap_chars) is not abstractproperty
degenerate_chars_defined = type(cls.degenerate_chars) is not abstractproperty # this duplicated logic should probably be refactored into a helper method...
...
```

jairideout:
@anderspitman done reviewing

skbio/sequence/_grammared_sequence.py
anderspitman: I actually already have it set up in a hierarchy. There's a second set of guards nested inside the first. Also I don't think defining the bools really makes it more readable (and believe me it doesn't take much of any excuse for me to throw in a temp for clarity). The raw Python already says pretty much exactly why it's doing: only process it's not abstract. "x_defined" seems ambiguous in this context.

anderspitman:
Back at ya @jairideout

skbio/sequence/_grammared_sequence.py
jairideout: I'm confused about the current checks: there's a top-level guard for `gap_chars` and `default_gap_char`, and the validations only have the chance to run if those are present. However, there are guarded validations that could be performed even if `gap_chars` and `default_gap_char` are abstract (e.g., checks dealing with degenerate/nondegenerate characters). Can you remove the hierarchy and instead perform independent validations where possible? Defining a few bools at the top may be helpful because you're going to have that code duplicated quite a bit (maybe a better name for the bools is `concrete_default_gap_char`, `concrete_gap_chars`, etc....?)

skbio/sequence/_grammared_sequence.py
```diff
@@ -22,11 +22,75 @@
...
+                type(cls.gap_chars) is not abstractproperty):
+            if cls.default_gap_char not in cls.gap_chars:
+                raise TypeError(
+                    "default_gap_char must be in gap_chars for class %s" %
+                    name)
+
+            if len(cls.gap_chars & cls.degenerate_chars) > 0:
```
jairideout: `cls.degenerate_chars` isn't guarded against abstractness here

skbio/sequence/_grammared_sequence.py
anderspitman: @jairideout see if this is like what you had in mind. I drew up a dependency graph to try and keep the duplication down and scope properly. Honestly I'm not sure this is worth the extra complexity. If any of these properties are still abstract at instantiation (which for most users will come soon after class declaration time), it's going to fail anyway. I think it would be more straightforward for abstractness to be a class-level distinction. If any member is static, the class should be considered static and the checks shouldn't run. Then once the user gets all the ABC ducks in a row, the static checks will start complaining and they can fix any of those necessary. As it currently stands the user could have ABC errors and static check errors interleaved.

jairideout: > If any member is static

@anderspitman I'm assuming when you say "static" you're meaning "abstract", please correct me if I'm wrong.

I don't see how the ABC errors and GrammaredSequenceMeta errors could be interleaved because the ABC errors will happen at instantiation time and the GrammaredSequenceMeta errors will happen at class definition time.

I'm fine with having a single "abstract" guard that only performs validations if all properties are concrete. The more granular approach to validations is nice because the user will know immediately (at class definition time) if there's something wrong with their alphabet, even if some pieces are still abstract. I agree that it may not be worth the complexity and we can always refactor to provide these validations if there's a need in the future.

Are you good with updating the code to only perform validations when all properties are concrete?

CHANGELOG.md
```diff
@@ -55,6 +55,9 @@ constructor. ([#6240](https://github.com/biocore/scikit-bio/issues/624))
 * ``TreeNode.to_array`` now supports replacing ``nan`` branch lengths in the resulting branch length vector with the value provided as ``nan_length_value``.
 * ``skbio.io.format.phylip`` now supports sniffing and reading strict, sequential PHYLIP-formatted files into ``skbio.Alignment`` objects. ([#1006](https://github.com/biocore/scikit-bio/issues/1006))
 * Added `default_gap_char` class property to ``DNA``, ``RNA``, and ``Protein`` for representing gap characters in a new sequence.
+* Added ``skbio.sequence.GrammaredSequence``, which can be inherited from to create grammared sequences with custom alphabets (e.g., for use with TabularMSA) (#1175)
```
jairideout: Can you make the issue number a link? See other changelog entries for examples.

jairideout: These changelog entries are in the wrong location (0.4.1). Can you please move them under 0.4.1-dev? I think the first two bullet points should go under "Features" and the third bullet point (DNA/RNA/Protein subclassing) should go under "Bug fixes".

skbio/sequence/_grammared_sequence.py
anderspitman: Yeah I meant "abstract", thanks. Sorry, "interleaved" was a bad way to describe that. What I meant was interleaved in time. So if user of skbio was working on writing a subclass of `GrammaredSequence` in a TDD fashion, the first error they would get would be from ABC. Then they would add the required property, try again, but this time the error might be from one of the static checks, then once they fix that the next error might be ABC again. I think this might be confusing, whereas if we force everything to not be abstract before any of the static checks run, all the related errors will come together.

I can implement the one check. Just a thought, what do you think about looping through all the attributes of the class and making sure nothing is abstract, rather than trying to maintain a list?

jairideout: Ah that makes sense, thanks for clarifying. :+1: for looping through attributes and checking nothing is abstract.

jairideout:
@anderspitman I took a final pass through, looks great! There are a couple of issues with the changelog and the metaclass validation guards need updating, once those are addressed this is ready for merge.

jairideout:
Thanks for all of your work on this @anderspitman!

PR #1279, opened 2016-02-17
HACK: raise ValueError instead of NotImplementedError when variant='solexa'

jairideout:
Temporary hack for #1256 to avoid spin-locked process. Reverted commit that capped contextlib2 version.


ebolyen:
:+1:

wasade:
Is HACK an official tag? :)

ebolyen:
totes

PR #1285, opened 2016-02-17
WEB: update home page to use conda channel

ebolyen:


index.html
```diff
@@ -56,7 +56,7 @@
 <pre>pip install scikit-bio
 </pre>
 <p>Equivalently, you can use the <code>conda</code> package manager available in <a href="http://continuum.io/downloads">Anaconda</a> or <a href="http://conda.pydata.org/miniconda.html">miniconda</a> to install scikit-bio and its dependencies without having to compile them:</p>
-<pre>conda install scikit-bio
+<pre>conda install -c https://conda.anaconda.org/biocore scikit-bio
```
Jorge-C: FYI, in a default conda you can type `conda install -c biocore scikit-bio`, which looks cleaner. However, it's possible for a user to modify the prefix by editing `channel_alias` in `.condarc`, and then the short and sweet instructions could fail.

ebolyen: Didn't know that, thanks for the info!

PR #1298, opened 2016-03-01
WIP: Update genbank io

mortonjt:
Depends on #1299 and #1324

This is the first prototype of the interval metadata object.

I realize that there are still quite a few unittests that needed to be added in.
Right now, I'd really appreciate any comments about the file organization, since there are some pretty large changes.  For instance, this is adding in a new metadata package


jairideout:
Thanks @mortonjt! Would it be possible to have the package organization changes made in a separate PR that we rapidly review/merge, then put the interval metadata functionality in a new PR? I think that'll make things easier to review.

The first PR would create the `skbio.metadata` subpackage, which initially will only contain package-private functionality (no public user imports). `skbio._base.MetadataMixin` and `skbio._base.PositionalMetadataMixin` would be moved to `skbio.metadata._mixin` and `skbio.util._metadata_repr` would be moved to `skbio.metadata._repr`.

mortonjt:
This is still a rather huge PR.  @RNAer can you do another pass over this?

skbio/io/format/genbank.py
```diff
@@ -438,17 +445,24 @@ def _serialize_single_genbank(obj, fh):
...
         if header in md:
-            serializer = _SERIALIZER_TABLE.get(
-                header, _serialize_section_default)
             out = serializer(header, md[header])
+        elif header == 'FEATURES':
+            # This will need to change
+            features = obj.interval_metadata.features.keys()
+            out = serializer(header, sorted(features, key=lambda x: x['location']))
             # test if 'out' is a iterator.
             # cf. Effective Python Item 17
```
RNAer: the comment is out of place.

skbio/io/format/genbank.py
```diff
@@ -629,18 +642,24 @@ def _serialize_features(header, obj, indent=21):
...
+        list of lines
+    length : int
+        sequence length
+
     Returns
     -------
-    tuple
-        Tuple of a dict of `metadata` and a pandas.Series of
-        `positional_metadata` for the feature.
+    a pandas.Series of `positional_metadata` for the feature with its name
```
RNAer: `positional_metadata`? this doc seems outdated.

skbio/metadata/_feature.py
```diff
@@ -0,0 +1,88 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2013--, scikit-bio development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+from __future__ import absolute_import, division, print_function
```
RNAer: this can be removed now

skbio/metadata/mixin.py
```diff
@@ -0,0 +1,519 @@
+# ----------------------------------------------------------------------------
```
RNAer: you didn't pull the latest from master branch? This is renamed to skbio/metadata/_mixin.py

mortonjt: Good catch.  Its now deleted.

skbio/metadata/_feature.py
mortonjt: Done

mortonjt: Done

skbio/metadata/mixin.py
mortonjt: Good catch.  done.

skbio/metadata/_interval.py
```diff
@@ -0,0 +1,205 @@
...
+        Returns
+        -------
+        list, Feature
+            A list of features satisfying the search criteria.
+        """
+        feats = []
+
+        # Find queries by interval
+        for value in args:
+            feats += self._query_interval(value)
```
wasade: why not have `feats = set()`, and add to it with `feats.update(...)`? 

the return is currently a list, but that suggests order matters. if order does matter, then the current approach will disregard order due to the cast to `set` on return

mortonjt: 👍

skbio/metadata/intersection.pyx
```diff
@@ -0,0 +1,513 @@
+"""
```
wasade: Please verify that this code is BSD compatible. From a google search, it looks like this code [originates](https://github.com/dnanexus/rseqc/blob/master/rseqc/lib/bx/intervals/intersection.pyx) from the rseqc project, which [appears](https://github.com/dnanexus/rseqc/blob/master/rseqc/lib/bx/intervals/intersection.pyx) to be GPLv3.

cc @gregcaporaso 
https://github.com/dnanexus/rseqc/blob/master/COPYING

mortonjt: I grabbed this from bx-python, which has an [MIT license](https://bitbucket.org/james_taylor/bx-python/src/da37e3aa45dc610ab4fc2f60949de90f93f84401/LICENSE?at=default&fileviewer=file-view-default).

I was under the impression that the MIT license would be compatible with BSD.  @jxtx, are there any issues merging your IntervalTree into scikit-bio?

gregcaporaso: Is this just a copy of their file? Generally you'd want to check with the developers of the file first (not sure if you've done that here), and then possibly put a note in our copy of the file. You also would need to add a licence to [``scikit-bio/licenses``](https://github.com/biocore/scikit-bio/tree/master/licenses). We should always be discussing these kind of additions beforehand (probably at the RFC stage, when we're discussing any new dependencies) to make sure that all our bases are covered.

wasade: It looks like rseqc may have been using that code without a more explicit attribution. 

@mortonjt, would it be possible to include the relevant license in the licenses [directory](https://github.com/biocore/scikit-bio/tree/master/licenses)?

mortonjt: Done

jxtx: I don't know anything about `rseqc` but it appears to have copied all of `bx-python` without attribution. 

This code is indeed from `bx-python` which is MIT licensed and I have no objection to including it here. In fact, I'd be happy to see anything people find useful from `bx-python` moved into `scikit-bio`. 

ping @brentp @ianschenck

brentp: fine with me. might be nice to package the interval tree stuff separately (still under MIT) for those who don't want the full, bx-python install.

mortonjt: @gregcaporaso @jairideout @ebolyen , what do you guys think about making the IntervalTree data structure publically accessible?

skbio/metadata/intersection.pyx
```diff
@@ -0,0 +1,513 @@
...
+
+    def __repr__(self):
+        return "IntervalNode(%i, %i)" % (self.start, self.end)
+
+    def __cinit__(IntervalNode self, int start, int end, object interval):
+        # Python lacks the binomial distribution, so we convert a
+        # uniform into a binomial because it naturally scales with
+        # tree size.  Also, python's uniform is perfect since the
+        # upper limit is not inclusive, which gives us undefined here.
+        self.priority = ceil(nlog * log(-1.0/(1.0 * rand()/RAND_MAX - 1)))
```
wasade: leverage numpy?

mortonjt:
Created another branch removing the genbank io modifications.  Let's try to get #1324 merged first before merging this ok?

skbio/metadata/intersection.pyx
wasade: actually, how important is the quality of the random numbers? this looks like C `rand` given the `extern` from `stdlib.h`. might be useful to replace numpy's binomial which derives from Mersenne Twister?

skbio/metadata/intersection.pyx
ebolyen: I'm not opposed, but something that @jairideout and I have been talking a lot about recently is creating our own index objects for tabular-ish datastructures (in light of the issues we are experiencing in having our codebase so closely coupled with pandas). If we were to go that route, representing this IntervalTree as sane index-like object could make a lot of sense for the codebase generally. I'm imagining indexes would have meaningful dispatch and provide a dict-like interface (unlike pandas indexes). So it could be used as "just" an IntervalTree, but it would also plug into things like TabularMSA and FeatureTable.

That being said, scikit-bio indexes might not happen for a while (or ever).

jairideout:
@mortonjt @RNAer it's great to see all the work being done on interval metadata! Opening these PRs is totally fine but we'll need an RFC to discuss the design and API before adding this functionality to scikit-bio. Feel free to reference these PRs in the RFC if it helps to include some code examples or implementation. In the RFC we can also discuss including bx-python's `IntervalTree` in scikit-bio and @ebolyen's idea to have our own index objects. Does that sound good to you guys?

skbio/metadata/intersection.pyx
jairideout: @mortonjt @ebolyen see my comment [here](https://github.com/biocore/scikit-bio/pull/1298#issuecomment-207142069) re: including `IntervalTree` and custom index objects

jairideout:
@mortonjt @RNAer can this be closed in favor of #1356?

skbio/metadata/intersection.pyx
ebolyen: I echo @wasade's concern over `rand` and would question why the priority is random. Also, using casts instead of multiplying by one would be preferable. (I'm not sure how smart the Cython compiler is, but explicit casting is clearer anyways).

Issue #1305, opened 2016-03-15
skbio cannot read newick trees with partitions


kdm9:
Hi all,

I'm working a lot recently with genome simulations using tools like `scrm` and `ms`. These simulators output newick trees with a partition length (representing recombinations). The format of this is:

```
[partition_length](newick tree);
```

An example of this [comes with seq-gen](https://github.com/rambaut/Seq-Gen/blob/master/examples/recombinant.tree).

Currently `skbio.read('filename.nwk', into=TreeNode)` fails with:

``` python
>>> samp = skbio.read(treefile, into=TreeNode)
  File "/usr/lib/python3/dist-packages/skbio/io/registry.py", line 1135, in read
    **kwargs)
  File "/usr/lib/python3/dist-packages/skbio/io/registry.py", line 497, in read
    return self._read_ret(file, format, into, verify, kwargs)
  File "/usr/lib/python3/dist-packages/skbio/io/registry.py", line 503, in _read_ret
    io_kwargs)
  File "/usr/lib/python3/dist-packages/skbio/io/registry.py", line 525, in _init_reader
    fmt, skwargs = self.sniff(file, **io_kwargs)
  File "/usr/lib/python3/dist-packages/skbio/io/registry.py", line 428, in sniff
    % file)
skbio.io._exception.UnrecognizedFormatError: Could not detect the format of <_io.BufferedReader name='125467289/population.nwk'>
```

A minimal failing example is:

```
import skbio
from skbio import TreeNode
t = skbio.read('https://github.com/rambaut/Seq-Gen/raw/master/examples/recombinant.tree', into=TreeNode)
```

I'm not sure how best to store or represent this information, for now I'm parsing out the components so that I can do:

``` python
from skbio import TreeNode

def iter_part_nwk(fname):
    with open(fname) as fh:
        for line in fh:
            m = re.match(r'\[(.*)\](\(.*;)', line)
            if m:
                l, t = m.groups()
                yield int(float(l)), TreeNode.read([t])

for partlen, tree in iter_part_nwk('recombinant.tree'):
    print(partlen, tree)
```

I'm happy to make a PR implementing a function that does a more rigorous version of this if that is the best way you can think of to handle this data type.

Cheers,
K

jairideout:
@kdmurray91 this sounds like a useful feature to have in scikit-bio, thanks for offering to work on it! Do you have a reference to where this format extension is defined?

[...]

PR #1312, opened 2016-03-22
Parser function for Partitioned Newick files

kdm9:
Initial work in progress addressing #1305

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,8 @@ develop-eggs
 lib
 lib64
 __pycache__
+.eggs
+.cache
 
 # Installer logs
 pip-log.txt
--- a/ci/pip_requirements.txt
+++ b/ci/pip_requirements.txt
@@ -9,3 +9,4 @@ git+git://github.com/sphinx-doc/sphinx.git
 sphinx-bootstrap-theme
 git+git://github.com/numpy/numpydoc.git@1a848331c2cf53d4fe356f4607799524bcc577ed
 check-manifest
+tempdir
--- a/skbio/tree/__init__.py
+++ b/skbio/tree/__init__.py
@@ -33,6 +33,7 @@ Utility Functions
    :toctree: generated/
 
     majority_rule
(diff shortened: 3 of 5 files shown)


ebolyen:
This should ideally be a module in `skbio.io.format`. Either an augmentation of the existing newick parser, or a new format. I don't have strong feelings on that. Then this format could be multiplexed into `read` and `write` and can be used in a variety of contexts. It also looks like having a generator for newick trees is needed. The current newick parser doesn't have a binding for that, but it would be nice to have.

jairideout:
I think making this its own format makes sense since it isn't part of "standard" newick. What we'd want minimally is a generator reader that yields partition lengths and `TreeNode` objects, and have this generator hooked up to the I/O registry. Having a sniffer would be nice to have but not essential.

Adding a generator reader for `newick` format shouldn't be necessary in this PR, agree it would be nice to have at some point.

@kdmurray91 @ebolyen what are your thoughts for a format name? `format='partitioned-newick'`?

kdmurray91:
@jairideout that sounds reasonable. Do you know of another "sub-format" like this that delegates some of the parsing to another parser? I'm not that familiar with the IO submodule.

The above was purely illustrative of the output of the code, I just ported the dodgy function I was using. I'll have a go at a formal format.

I also agree that a generator parser for newick trees would be very useful, there are many cases where you would need one. I will have a go at making one if it turns out to be required for this PR.

jairideout:
@kdmurray91 that all sounds good, thanks!

We don't have another "sub-format" yet (besides Stockholm, which currently doesn't parse the NHX). You should be able to import whatever package-private functions you need from `skbio.io.format.newick` and use those to parse the newick portions of the format. @ebolyen does that sound right to you?

ebolyen:
:+1:

kdmurray91:
Cool, I'll have a go.

wasade:
Closing as this appears to have gone stale, please reopen if this should be finished and merged

PR #1321, opened 2016-03-30
MAINT: remove empty metadata optimizations

### LINKED ISSUE #1308: master is broken using latest pandas release
Using the latest pandas release breaks many many unit tests and doctests. The default pandas index is now `RangeIndex` instead of `Int64Index`, and data frame/series `repr` adds trailing `.0` to whole numbers.

jairideout:
Removed memory optimizations for storing empty `metadata` and `positional_metadata`, currently on `Sequence` and `TabularMSA`.

Previously, empty metadata was stored internally as `None`, and an empty metadata representation (e.g., `{}` or `pd.DataFrame(index=range(axis_len))` was created and stored on the object when `.metadata` or `.positional_metadata` properties were accessed for the first time.

While this optimization resulted in some memory savings, its implementation is not sustainable because scikit-bio devs must remember to use `has_metadata` or `has_positional_metadata` to handle empty metadata efficiently, otherwise any memory savings are lost. Every routine interacting with metadata needed a unit test ensuring empty metadata is handled efficiently, and worse, these tests had to interact with private state. As metadata will be applied to more, if not all, scikit-bio objects in the future, the maintenance and cognitive burden placed on developers is not worth the memory savings at this time. Further, this memory optimization caused the memory footprints of scikit-bio objects to be unpredictable. For example, a user may not understand why they are running out of memory when accessing `.metadata`, `.positional_metadata`, or a routine that doesn't handle empty metadata efficiently.

Now, empty `metadata` will be stored as `{}` and empty `positional_metadata` as a `DataFrame` with default index matching the object's axis length. The empty dict shouldn't be too heavy but the empty `DataFrame` is problematic. To fix #1308 we will use the new default `RangeIndex` in pandas 0.18.0, which is much more memory-efficient (essentially storing 3 integers: start, stop, and end). Thus, empty `positional_metadata` will have a much smaller memory footprint soon. If more memory savings are needed in the future, we could implement proxy objects.

Deprecated `has_metadata` and `has_positional_metadata`, as they mainly existed for scikit-bio devs to check whether metadata exists without creating empty metadata objects.

Related to #1308.

### DIFF
--- a/skbio/metadata/_mixin.py
+++ b/skbio/metadata/_mixin.py
@@ -12,7 +12,7 @@ import copy
 import numpy as np
 import pandas as pd
 
-from skbio.util._decorator import stable
+from skbio.util._decorator import stable, deprecated
 
 
 class MetadataMixin(metaclass=abc.ABCMeta):
@@ -61,18 +61,11 @@ class MetadataMixin(metaclass=abc.ABCMeta):
 
         Delete metadata:
 
-        >>> seq.has_metadata()
-        True
         >>> del seq.metadata
         >>> seq.metadata
         {}
-        >>> seq.has_metadata()
-        False
 
         """
-        if self._metadata is None:
-            # Not using setter to avoid copy.
-            self._metadata = {}
         return self._metadata
 
     @metadata.setter
@@ -84,7 +77,8 @@ class MetadataMixin(metaclass=abc.ABCMeta):
 
     @metadata.deleter
     def metadata(self):
-        self._metadata = None
+        # Not using setter to avoid copy.
+        self._metadata = {}
 
     @abc.abstractmethod
     def __init__(self, metadata=None):
@@ -92,7 +86,7 @@ class MetadataMixin(metaclass=abc.ABCMeta):
 
     def _init_(self, metadata=None):
         if metadata is None:
-            self._metadata = None
+            del self.metadata
         else:
             self.metadata = metadata
 
@@ -101,20 +95,7 @@ class MetadataMixin(metaclass=abc.ABCMeta):
         pass
 
     def _eq_(self, other):
-        # We're not simply comparing self.metadata to other.metadata in order
-        # to avoid creating "empty" metadata representations on the objects if
-        # they don't have metadata.
-        if self.has_metadata() and other.has_metadata():
-            if self.metadata != other.metadata:
-                return False
-        elif not (self.has_metadata() or other.has_metadata()):
-            # Both don't have metadata.
-            pass
-        else:
-            # One has metadata while the other does not.
-            return False
-
-        return True
+        return self.metadata == other.metadata
 
     @abc.abstractmethod
     def __ne__(self, other):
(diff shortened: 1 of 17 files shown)


skbio/metadata/_mixin.py
```diff
@@ -311,7 +283,7 @@ def __init__(self, positional_metadata=None):
 
     def _init_(self, positional_metadata=None):
         if positional_metadata is None:
-            self._positional_metadata = None
+            del self.positional_metadata
```
ebolyen: This is kind of neat

skbio/sequence/_sequence.py
```diff
@@ -505,13 +505,10 @@ def concat(cls, sequences, how='strict'):
...
-            cols = []
+            cols = set()
             for s in seqs:
-                if s.has_positional_metadata():
-                    cols.append(frozenset(s.positional_metadata))
-                else:
-                    cols.append(frozenset())
-            if len(set(cols)) > 1:
+                cols.add(frozenset(s.positional_metadata))
+            if len(cols) > 1:
```

skbio/metadata/_mixin.py
jairideout: Thanks! It looks a little odd but allows the "empty" metadata representation to be defined in a single place (the property deleter).

skbio/sequence/_sequence.py
```diff
@@ -564,10 +559,9 @@ def __init__(self, sequence, metadata=None, positional_metadata=None,
             # creating "empty" metadata representations on both sequence
```
ebolyen: Comment is out of date

jairideout: Noice catch!

skbio/sequence/_sequence.py
```diff
@@ -886,17 +880,17 @@ def __getitem__(self, indexable):
         seq = self._bytes[indexable]
         positional_metadata = self._slice_positional_metadata(indexable)
 
-        return self._to(sequence=seq, positional_metadata=positional_metadata)
+        return self._constructor(
+            sequence=seq,
+            metadata=self.metadata,
```
ebolyen: Does `Sequence._constructor` not propagate attributes automatically? This was one of the reasons for `_to`. I think `TabularMSA` propagates.

skbio/sequence/_sequence.py
```diff
@@ -1467,11 +1461,8 @@ def replace(self, where, character):
         index = self._munge_to_index_array(where)
         seq_bytes = self._bytes.copy()
         seq_bytes[index] = character
-        metadata = self.metadata if self.has_metadata() else None
-        positional_metadata = self.positional_metadata if \
-            self.has_positional_metadata() else None
-        return self.__class__(seq_bytes, metadata=metadata,
-                              positional_metadata=positional_metadata)
+        return self.__class__(seq_bytes, metadata=self.metadata,
```
ebolyen: Should this have been `_constructor`?

skbio/sequence/_sequence.py
jairideout: `Sequence._constructor` doesn't propagate, it mainly exists over `__class__` to allow subclasses to suppress validation. `TabularMSA._constructor_` (note it is a *sunder*) propagates but doesn't exist on `Sequence`.

skbio/sequence/_sequence.py
jairideout: It's `__class__` to run validations, since the user could introduce invalid characters. I'll add a comment

ebolyen:
Thanks for the clarifications. This looks pretty much ready to go when (**if?**) tests get around to passing...

PR #1323, opened 2016-04-02
Rename nondegenerate

### LINKED ISSUE #1071: Deprecate `nondegenerate_chars` in favor of a better name.
Using `nondegenerate` was probably a mistake. It is hard to talk about (I know I have mixed myself up a couple of times) and it almost implies the inverse of `degenerate` which isn't really the intention. Fortunately, it shouldn't be too painful to deprecate and gives us a chance to exercise that process.

`standard` might be a better description. So to kick off the conversation, I propose we deprecate the following:

`nondegenerate_chars` -> `standard_chars`
`has_nondegenerates()` -> `has_standards()`
`nondegenerates()` -> `standards()`

The only thing I'm not too jazzed about is how awkward `standard` is when plural. `has_standards()` is a particularly amusing demonstration of this. 

Does anyone have a better name to describe the "normal" bases of a sequence?

anderspitman:
Fixes #1071


skbio/sequence/_grammared_sequence.py
```diff
@@ -277,9 +279,9 @@ def degenerate_chars(cls):
         """
         return set(cls.degenerate_map)
 
-    @abstractproperty
     @classproperty
-    @stable(as_of='0.4.0')
+    @deprecated(as_of='0.5.0', until='0.5.2',
```
jairideout: Can you change `as_of` to `0.4.2-dev`? That way if 0.5.0 isn't the next release we'll be able to catch and update the `as_of` version.

skbio/sequence/_grammared_sequence.py
```diff
@@ -289,19 +291,33 @@ def nondegenerate_chars(cls):
             Non-degenerate characters.
 
         """
+        return cls.definite_chars
+
+    @abstractproperty
+    @classproperty
+    @stable(as_of='0.4.0')
```
jairideout: This should be `0.4.2-dev` since the new name will be stable in the next release.

skbio/sequence/_grammared_sequence.py
```diff
@@ -450,6 +466,32 @@ def has_degenerates(self):
         return bool(self.degenerates().any())
 
     @stable(as_of='0.4.0')
```
jairideout: This should be `0.4.2-dev` since the new name will be stable in the next release.

skbio/sequence/_grammared_sequence.py
```diff
@@ -450,6 +466,32 @@ def has_degenerates(self):
...
+        --------
+        >>> from skbio import DNA
+        >>> s = DNA('ACWGN')
+        >>> s.definites()
+        array([ True,  True, False,  True, False], dtype=bool)
+
+        """
+        return np.in1d(self._bytes, self._definite_char_codes)
+
+    @deprecated(as_of='0.5.0', until='0.5.2',
```
jairideout: `as_of=0.4.2-dev`

skbio/sequence/_grammared_sequence.py
```diff
@@ -473,9 +514,40 @@ def nondegenerates(self):
         array([ True,  True, False,  True, False], dtype=bool)
 
         """
-        return np.in1d(self._bytes, self._nondegenerate_codes)
+        return self.definites()
 
     @stable(as_of='0.4.0')
```
jairideout: `as_of=0.4.2-dev`

skbio/sequence/_grammared_sequence.py
```diff
@@ -473,9 +514,40 @@ def nondegenerates(self):
...
+        False
+        >>> t = DNA('ANCACWWGACGTT')
+        >>> t.has_definites()
+        True
+
+        """
+        # TODO: cache results
+        return bool(self.definites().any())
+
+    @deprecated(as_of='0.5.0', until='0.5.2',
```
jairideout: `as_of=0.4.2-dev`

skbio/sequence/_grammared_sequence.py
```diff
@@ -625,8 +697,8 @@ def to_regex(self):
         -------
         regex
             Pre-compiled regular expression object (as from ``re.compile``)
-            that matches all non-degenerate versions of this sequence, and
-            nothing else.
+            that matches all definites versions of this sequence, and nothing
```
jairideout: *definites* -> *definite*

skbio/sequence/_grammared_sequence.py
```diff
@@ -562,8 +634,8 @@ def expand_degenerates(self):
```
jairideout: I can't comment on the relevant line because it's outside the diff, but can you update the *Yields* section above to use *definite*?

skbio/sequence/_grammared_sequence.py
```diff
@@ -562,8 +634,8 @@ def expand_degenerates(self):
 
         Notes
         -----
-        There is no guaranteed ordering to the non-degenerate sequences that
-        are yielded.
+        There is no guaranteed ordering to the definite sequences that are
+        yielded.
 
         Each non-degenerate sequence will have the same type, metadata,
```
jairideout: *non-degenerate* -> *definite*

jairideout:
Thanks @anderspitman, this looks great! Can you also update the changelog's `Deprecated functionality [stable]` section to note that `nondegenerate_chars`, `nondegenerates`, and `has_nondegenerates` are deprecated in favor of the new names? Relevant public classes that are affected: `GrammaredSequence`, `DNA`, `RNA`, `Protein`

anderspitman:
Thanks for reviewing @jairideout. Should be able to address review comments tomorrow!

anderspitman:
@jairideout all review comments up to this point should be addressed. I was thinking it might be worthwhile to add a bit more documentation on what exactly `definite` means. It makes sense when it's explained, but it's not perfectly clear just from the name what it's used for. Any thoughts on whether more explanation is necessary and what form it might take?

jairideout:
Thanks @anderspitman! I think some documentation would be nice but can be added later since this is just a rename. Honestly I'm not sure where this would make the most sense -- `GrammaredSequence` would be the logical place but users are unlikely to look there. We really need some user docs (similar to pandas) that are separate from the API and explain some of these overarching ideas, naming conventions, etc. Basically an intro to scikit-bio. I think an explanation of the `GrammaredSequence` naming conventions/alphabet would make sense there.

anderspitman:
:+1:  sounds good

Issue #1221, opened 2015-12-07
Support partial beta diversity calculations


wasade:
Currently, the beta diversity of all pairwise combinations of sample IDs are computed. However, it would be nice to compute only a partial set in the case of combining studies (i.e., only compute the necessary distances to merge a set of samples) or if performing a block-oriented computation of the subsequent full distance matrix to avoid duplicated compute. 

A possible modification to the `beta_diversity` API would be to allow `ids` to either be a iterable of `str` or an iterable of `tuple`. In the former case, all pairwise combinations would be computed. In the later case, only the distances between the described pairs would be computed. It is not clear though if and how this should impact the return.

josenavas:
Just want to ping on this issue as I think the potential use cases for this are increasing.
@gregcaporaso @jairideout @wasade @ebolyen which is the best way of starting a formal discussion to get this API changes into scikit-bio?

wasade:
Thanks, @josenavas, for following up on this. 

From looking at the code, I think I may have a solution that does not necessitate an API change:

```python
def partial_pw(id_pairs, counts, metric, **kwargs):
    n_ids = len({a for a, b in id_pairs}.union({b for a, b in id_pairs}))
    dm = np.zeros((n_ids, n_ids), dtype=float)
    # do some fancy indexing that isn't shown here
    for u, v in id_pairs_indexed:
        dm[u, v] = metric(counts[u], counts[v], **kwargs)
    return dm

# ...then as a user/developer
pw = functools.partial(partial_pw, my_id_pairs)
ids = {a for a, b in my_id_pairs}.union({b for a, b in my_id_pairs})
partial_dm = beta_diversity('unweighted_unifrac', counts, ids=ids, validate=True, pairwise_func=pw, tree=tree, otu_ids=otu_ids)
```

I think this would accomplish what is necessary. So then the only addition to the public API would be a method `partial_pw` which is metric agnostic. There would not need to be any modification to existing APIs. 

If this seems reasonable, I can issue a PR on this. I'm not too jazzed (yet) on the ID ordering, as it would be important to minimize any juggling when merging DMs in the case of a block decomposition.

gregcaporaso:
@wasade, I think that would work. I guess the main question is would we want to support this through the ``pairwise_func``, or though the ``beta_diversity`` interface directly. Note that we can add new parameters to ``beta_diversity``, so it seems like we might be able to build this into the existing API and take some burden off the user.

wasade:
[...]

PR #1337, opened 2016-04-19
Partial pairwise calculation for beta diversity.

wasade:
Addresses #1221.

Specific questions:

* ~~Validation was pushed into `_partial_pw` to not clog `beta_diversity`, is this preferred?~~
* ~~The mechanism for resolving string forms of metrics (e.g., `"euclidean"`) is not [exposed](https://github.com/scipy/scipy/blob/v0.17.0/scipy/spatial/distance.py#L1249) in scipy's `pdist` making it difficult to support scipy-native metrics. Is this a blocker?~~ The method resolution in scipy makes a large number of internal assumptions, and likely requires a fairly substantial PR to scipy in order to even expose `euclidean`. As such, I'd prefer to merge this with support only for skbio metrics and passed in callables, as there is a near-term need for this functionality for UniFrac.  
* ~~The condensed form of the matrix is returned, but it isn't clear if this is absolutely necessary or not. Avoiding it would be nice.~~
* ~~Do we need to assert that duplicate ID pairs are not specified? As written, if both `('a', 'b')` and `('b', 'a')` are provided, it will result in a "doubling" of the distance due to the `dm + dm.T`. One option would be to carry forward `validate`.~~

~~I have not added in tests specifically for `_partial_pw` as resolution of the questions above may result in shuffling of the logic and tests. The tests added so far should exercise all branches, but this PR should not be considered complete until unit tests are added for `_partial_pw`.~~

To dos:

- [x] `_partial_pw` specific unit tests
- [ ] expand example documentation. Would a reviewer please advise on if and where an example should go? I looked at adding to `skbio.diversity.__init__`, but that is already comprehensive and I was concerned about breaking up the flow of the example. 
- [x] add mention in the ChangeLog

:neckbeard:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Expanded error message in `skbio.io.format.stockholm` reader when `constructor` is not passed, in order to provide better explanation to user. ([#1327](https://github.com/biocore/scikit-bio/issues/1327))
 * Added `skbio.sequence.distance.kmer_distance` for computing the kmer distance between two sequences. ([#913](https://github.com/biocore/scikit-bio/issues/913))
 * Added `skbio.sequence.Sequence.replace` for assigning a character to positions in a `Sequence`. ([#1222](https://github.com/biocore/scikit-bio/issues/1222))
+* Added support for partial pairwise calculations in `skbio.diversity.beta_diversity`. ([#1221](https://github.com/biocore/scikit-bio/issues/1221), [#1337](https://github.com/biocore/scikit-bio/pull/1337))
 
 ### Backward-incompatible changes [stable]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 * Added `skbio.sequence.Sequence.replace` for assigning a character to positions in a `Sequence`. ([#1222](https://github.com/biocore/scikit-bio/issues/1222))
 * Added support for `pandas.RangeIndex`, lowering the memory footprint of default integer index objects. `Sequence.positional_metadata` and `TabularMSA.positional_metadata` now use `pd.RangeIndex` as the positional metadata index. `TabularMSA` now uses `pd.RangeIndex` as the default index. Usage of `pd.RangeIndex` over the previous `pd.Int64Index` [should be transparent](http://pandas.pydata.org/pandas-docs/version/0.18.0/whatsnew.html#range-index), so these changes should be non-breaking to users. scikit-bio now depends on pandas >= 0.18.0 ([#1308](https://github.com/biocore/scikit-bio/issues/1308))
 * Added `reset_index=False` parameter to `TabularMSA.append` and `TabularMSA.extend` for resetting the MSA's index to the default index after appending/extending.
-* Added support for partial pairwise calculations in `skbio.diversity.beta_diversity`. ([#1221](https://github.com/biocore/scikit-bio/issues/1221), [#1337](https://github.com/biocore/scikit-bio/pull/1337))
+* Added support for partial pairwise calculations via `skbio.diversity.partial_beta_diversity`. ([#1221](https://github.com/biocore/scikit-bio/issues/1221), [#1337](https://github.com/biocore/scikit-bio/pull/1337))
 
 ### Backward-incompatible changes [stable]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 * Added `skbio.sequence.Sequence.replace` for assigning a character to positions in a `Sequence`. ([#1222](https://github.com/biocore/scikit-bio/issues/1222))
 * Added support for `pandas.RangeIndex`, lowering the memory footprint of default integer index objects. `Sequence.positional_metadata` and `TabularMSA.positional_metadata` now use `pd.RangeIndex` as the positional metadata index. `TabularMSA` now uses `pd.RangeIndex` as the default index. Usage of `pd.RangeIndex` over the previous `pd.Int64Index` [should be transparent](http://pandas.pydata.org/pandas-docs/version/0.18.0/whatsnew.html#range-index), so these changes should be non-breaking to users. scikit-bio now depends on pandas >= 0.18.0 ([#1308](https://github.com/biocore/scikit-bio/issues/1308))
 * Added `reset_index=False` parameter to `TabularMSA.append` and `TabularMSA.extend` for resetting the MSA's index to the default index after appending/extending.
-* Added support for partial pairwise calculations via `skbio.diversity.partial_beta_diversity`. ([#1221](https://github.com/biocore/scikit-bio/issues/1221), [#1337](https://github.com/biocore/scikit-bio/pull/1337))
+* Added support for partial pairwise calculations via `skbio.diversity.partial_beta_diversity`. ([#1221](https://github.com/biocore/scikit-bio/issues/1221), [#1337](https://github.com/biocore/scikit-bio/pull/1337)). This function is immediately deprecated as its return type will change in the future and should be used with caution in its present form (see the function's documentation for details).
 
 ### Backward-incompatible changes [stable]
 
--- a/skbio/diversity/__init__.py
+++ b/skbio/diversity/__init__.py
@@ -399,10 +399,10 @@ Create a matrix containing 6 samples (rows) and 7 OTUs (columns):
 
 from skbio.util import TestRunner
 
-from ._driver import (alpha_diversity, beta_diversity,
+from ._driver import (alpha_diversity, beta_diversity, partial_beta_diversity,
                       get_alpha_diversity_metrics, get_beta_diversity_metrics)
 
 __all__ = ["alpha_diversity", "beta_diversity", "get_alpha_diversity_metrics",
-           "get_beta_diversity_metrics"]
+           "get_beta_diversity_metrics", "partial_beta_diversity"]
 
 test = TestRunner(__file__).test
--- a/skbio/diversity/__init__.py
+++ b/skbio/diversity/__init__.py
@@ -377,15 +377,15 @@ Create a matrix containing 6 samples (rows) and 7 OTUs (columns):
    >>> plt.close('all') # not necessary for normal use
    >>> fig = sample_md.boxplot(column='Faith PD', by='body_site')
 
-We can also compute Spearman correlations between all pairs of columns in this
-``DataFrame``. Since our alpha diversity metrics are the only two numeric
-columns (and thus the only columns for which Spearman correlation is relevant),
-this will give us a symmetric 2x2 correlation matrix.
+    We can also compute Spearman correlations between all pairs of columns in
+    this ``DataFrame``. Since our alpha diversity metrics are the only two
+    numeric columns (and thus the only columns for which Spearman correlation
+    is relevant), this will give us a symmetric 2x2 correlation matrix.
 
->>> sample_md.corr(method="spearman")
-               Observed OTUs  Faith PD
-Observed OTUs       1.000000  0.939336
-Faith PD            0.939336  1.000000
+    >>> sample_md.corr(method="spearman")
+                   Observed OTUs  Faith PD
+    Observed OTUs       1.000000  0.939336
+    Faith PD            0.939336  1.000000
 
 """
 
--- a/skbio/diversity/__init__.py
+++ b/skbio/diversity/__init__.py
@@ -162,6 +162,7 @@ Functions
 
     alpha_diversity
     beta_diversity
+    partial_beta_diversity
     get_alpha_diversity_metrics
     get_beta_diversity_metrics
 
@@ -377,15 +378,15 @@ Create a matrix containing 6 samples (rows) and 7 OTUs (columns):
    >>> plt.close('all') # not necessary for normal use
    >>> fig = sample_md.boxplot(column='Faith PD', by='body_site')
 
-    We can also compute Spearman correlations between all pairs of columns in
-    this ``DataFrame``. Since our alpha diversity metrics are the only two
-    numeric columns (and thus the only columns for which Spearman correlation
-    is relevant), this will give us a symmetric 2x2 correlation matrix.
+We can also compute Spearman correlations between all pairs of columns in
+this ``DataFrame``. Since our alpha diversity metrics are the only two
+numeric columns (and thus the only columns for which Spearman correlation
+is relevant), this will give us a symmetric 2x2 correlation matrix.
 
-    >>> sample_md.corr(method="spearman")
-                   Observed OTUs  Faith PD
-    Observed OTUs       1.000000  0.939336
-    Faith PD            0.939336  1.000000
+>>> sample_md.corr(method="spearman")
+               Observed OTUs  Faith PD
+Observed OTUs       1.000000  0.939336
+Faith PD            0.939336  1.000000
 
 """
 
--- a/skbio/diversity/_driver.py
+++ b/skbio/diversity/_driver.py
@@ -194,11 +194,12 @@ def _partial_pw(ids, id_pairs, counts, metric, **kwargs):
         of OTUs in a given sample.
     ids : iterable of strs
         Identifiers for each sample in ``counts``.
-    id_pairs : iterable of tuple, optional An iterable of tuples of IDs to
-        compare (e.g., ``[('a', 'b'), ('a', 'c'), ...])``. If specified, the
-        set of IDs described must be a subset of ``ids``. ``id_pairs`` is a
-        mutually exclusive argument with ``pairwise_func``. ``metric`` must be
-        resolvable by scikit-bio (e.g., UniFrac methods), or must be callable.
+    id_pairs : iterable of tuple, optional
+        An iterable of tuples of IDs to compare (e.g., ``[('a', 'b'), ('a',
+        'c'), ...])``. If specified, the set of IDs described must be a subset
+        of ``ids``. ``id_pairs`` is a mutually exclusive argument with
+        ``pairwise_func``. ``metric`` must be resolvable by scikit-bio (e.g.,
+        UniFrac methods), or must be callable.
     kwargs : kwargs, optional
         Metric-specific parameters.
 
@@ -214,6 +215,11 @@ def _partial_pw(ids, id_pairs, counts, metric, **kwargs):
         If ``id_pairs`` are not a subset of ``ids``.
         If ``metric`` is not a callable.
         If duplicates are observed in ``id_pairs``.
+
+    See Also
+    --------
+    skbio.diversity.get_beta_diversity_metrics
+
     """
     if ids is None:
         raise ValueError("`ids` must be specified if `id_pairs` is specified")
@@ -275,11 +281,12 @@ def beta_diversity(metric, counts, ids=None, validate=True, pairwise_func=None,
         that can be provided are ``scipy.spatial.distance.pdist`` and
         ``sklearn.metrics.pairwise_distances``. By default,
         ``scipy.spatial.distance.pdist`` will be used.
-    id_pairs : iterable of tuple, optional An iterable of tuples of IDs to
-        compare (e.g., ``[('a', 'b'), ('a', 'c'), ...])``. If specified, the
-        set of IDs described must be a subset of ``ids``. ``id_pairs`` is a
-        mutually exclusive argument with ``pairwise_func``. ``metric`` must be
-        resolvable by scikit-bio (e.g., UniFrac methods), or must be callable.
+    id_pairs : iterable of tuple, optional
+        An iterable of tuples of IDs to compare (e.g., ``[('a', 'b'), ('a',
+        'c'), ...])``. If specified, the set of IDs described must be a subset
+        of ``ids``. ``id_pairs`` is a mutually exclusive argument with
+        ``pairwise_func``. ``metric`` must be resolvable by scikit-bio (e.g.,
+        UniFrac methods), or must be callable.
     kwargs : kwargs, optional
         Metric-specific parameters.
 
--- a/skbio/diversity/_driver.py
+++ b/skbio/diversity/_driver.py
@@ -238,7 +238,7 @@ def _partial_pw(ids, id_pairs, counts, metric, **kwargs):
     for u, v in id_pairs_indexed:
         dm[u, v] = metric(counts[u], counts[v], **kwargs)
 
-    return scipy.spatial.distance.squareform(dm + dm.T)
+    return dm + dm.T
 
 
 @experimental(as_of="0.4.0")
--- a/skbio/diversity/_driver.py
+++ b/skbio/diversity/_driver.py
@@ -182,20 +182,21 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
     return pd.Series(results, index=ids)
 
 
-@deprecated(as_of='0.4.2-dev', until='0.5.0',
+@deprecated(as_of='0.4.2-dev', until='0.5.1',
             reason=('The return type is unstable. Developer caution is '
-                    'advised. The resulting DistanceMatrix object may include '
-                    'zeros when distance has not been calculated, and '
+                    'advised. The resulting DistanceMatrix object will '
+                    'include zeros when distance has not been calculated, and '
                     'therefore can be misleading.'))
-def partial_beta_diversity(metric, counts, ids=None, validate=None,
-                           id_pairs=None, **kwargs):
+def partial_beta_diversity(metric, counts, ids, id_pairs, validate=True,
+                           **kwargs):
     """Compute distances only between specified ID pairs
 
     Parameters
     ----------
-    metric : callable
-        The pairwise distance function to apply. ``metric`` must be resolvable
-        by scikit-bio (e.g., UniFrac methods), or must be callable.
+    metric : str or callable
+        The pairwise distance function to apply. If ``metric`` is a string, it
+        must be resolvable by scikit-bio (e.g., UniFrac methods), or must be
+        callable.
     counts : 2D array_like of ints or floats
         Matrix containing count/abundance data where each row contains counts
         of OTUs in a given sample.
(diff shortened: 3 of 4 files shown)


josenavas:
Awesome @wasade - you're on fire!

Just my thoughts on the topics:
> - Validation was pushed into _partial_pw to not clog beta_diversity, is this preferred?

From what I can see in the code the validation only applies in the _partial_pw code, so I think it is fine to put it there.

> - The mechanism for resolving string forms of metrics (e.g., "euclidean") is not exposed in scipy's pdist making it difficult to support scipy-native metrics. Is this a blocker?

I think it is ok in the meantime and with good docs. Would it also be possible to open an issue in the scipy repo to see if  they can expose such transaltion?

> - The condensed form of the matrix is returned, but it isn't clear if this is absolutely necessary or not. Avoiding it would be nice.

I think this is ok to return in order to keep consistency in the output types of the `beta_diversity` function.

> - Do we need to assert that duplicate ID pairs are not specified? As written, if both ('a', 'b') and ('b', 'a') are provided, it will result in a "doubling" of the distance due to the dm + dm.T. One option would be to carry forward validate.

I would suggest to validate that list and simple raise a Warning saying that duplicate ID pairs where passed and they've been ignored.

Again, thanks @wasade !

skbio/diversity/_driver.py
```diff
@@ -180,9 +181,64 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+        If ``ids`` are not specified.
+        If ``id_pairs`` are not a subset of ``ids``.
+        If ``metric`` is not a callable.
+    """
+    if ids is None:
+        raise ValueError("`ids` must be specified if `id_pairs` is specified")
+
+    # flattening list benchmark here
+    # http://stackoverflow.com/a/952952
+    all_ids_in_pairs = {id_ for pair in id_pairs for id_ in pair}
```
josenavas: Minor, but `list(itertools.chain.from_iterable(l))` is actually faster from one of the comments in such post (and at least we are using it in Qiita).

wasade: Great, thanks!

wasade:
@josenavas, btw, given the structure of the `pdist` method for resolving `str` -> callable, I don't think it would be an easy fix. I also don't think it would help us as it redirects to private methods in many cases, so for instance, I don't think we could just reuse their `euclidean` method.

josenavas:
Thanks @wasade!

I was just looking over the code and tests again and I think I misunderstood your comment about the 'condensed' form of the matrix. Let me see if I understand correctly what is going on:

If I have ids A, B and C in the table but I only wanted to compute the distance for the pair (A, B), I will still receive back a 3x3 distance matrix? Shouldn't it be a 2x2? I'm just thinking in really tables in which we just want to compute a handful of comparisons, it will be a lot of 0's...

wasade:
@josenavas, we need to operate within the acceptable inputs for the `DistanceMatrix` constructor. Keep in mind though that you can filter the table to `{A, B}` before computing. It would be a lot of zeros if, for instance, you wanted to compute the distance between a set of new American Gut samples against all the rest of the American Gut samples. In my opinion, I'd prefer to get this merged in in a reasonable state, and we can refine the returns from `_partial_pw` at a later time if necessary relatively easily since it is a private method.

josenavas:
Thanks @wasade !

After your comment I went back to the code and I see now the complexity of reducing the distance matrix to the subset of ids. I agree with you that it is a small price to pay now compared with adding this functionality.

skbio/diversity/_driver.py
```diff
@@ -180,9 +182,68 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+    Parameters
+    ----------
+    metric : callable
+        The pairwise distance function to apply.
+    counts : 2D array_like of ints or floats
+        Matrix containing count/abundance data where each row contains counts
+        of OTUs in a given sample.
+    ids : iterable of strs
+        Identifiers for each sample in ``counts``.
+    id_pairs : iterable of tuple, optional An iterable of tuples of IDs to
```
josenavas: Can you add `An iterable...` in a new line?

wasade: Good catch, my "gq" got hosed.

skbio/diversity/_driver.py
```diff
@@ -180,9 +182,68 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+        compare (e.g., ``[('a', 'b'), ('a', 'c'), ...])``. If specified, the
+        set of IDs described must be a subset of ``ids``. ``id_pairs`` is a
+        mutually exclusive argument with ``pairwise_func``. ``metric`` must be
+        resolvable by scikit-bio (e.g., UniFrac methods), or must be callable.
```
josenavas: Probably useful to add a link to [this](http://scikit-bio.org/docs/latest/generated/skbio.diversity.get_beta_diversity_metrics.html#skbio.diversity.get_beta_diversity_metrics) function?

skbio/diversity/_driver.py
```diff
@@ -214,6 +275,11 @@ def beta_diversity(metric, counts, ids=None, validate=True, pairwise_func=None,
         that can be provided are ``scipy.spatial.distance.pdist`` and
         ``sklearn.metrics.pairwise_distances``. By default,
         ``scipy.spatial.distance.pdist`` will be used.
+    id_pairs : iterable of tuple, optional An iterable of tuples of IDs to
```
josenavas: Same comments as above

josenavas:
Just a couple of comments about docs - otherwise this looks good! 👍 thanks @wasade

skbio/diversity/_driver.py
wasade: done

skbio/diversity/_driver.py
wasade: done

skbio/diversity/_driver.py
wasade: done

ElDeveloper:
👍

wasade:
Is this good to merge?

josenavas:
@gregcaporaso @jairideout @ebolyen 

@ElDeveloper and I already went over this but we would like to have you guys to take a last look and this and merge if you don't see any issues. We are going to start running bdiv on the EMP pretty soon and having this in will be handy.

Thanks!

wasade:
@jairideout is working on a PartialDistanceMatrix for this PR and should be
in soon

jairideout:
Closing in favor of #1354.

wasade:
@jairideout, I do not believe this PR should be closed yet. Given the scaling issues in #1354, I'd prefer to get this current PR merged so that we can get #1352 finalized, and resolve a partial distance matrix in a separate PR.

wasade:
Just to update, @jairideout @ebolyen and I spoke on the phone this afternoon. The plan is to:

- defer #1354 until we can more precisely resolve a `PairwiseDistanceMatrix`
- add in a `DangerWarning` when using `id_pairs` as the `DistanceMatrix` objects returned will contain zeros at locations where there are uncomputed distances.

This approach will allow us to get this PR merged relatively soon so that we can get #1352 done as well for processing EMP data. `beta_diversity` is marked `experimental` so having a non-ideal return is okay for a short while. Since the immediate use-case is block decomposition, having the warnings gives some protection against unanticipated use of partial distance matrices.

jairideout:
@wasade, @ebolyen, and I talked offline and decided to temporarily use the solution here so EMP analysis isn't blocked when the data arrives. This solution works today (from a practical standpoint) but is dangerous to expose to users who aren't aware of its caveats. Solving this correctly/safely (from an API perspective) while maintaining scalability will likely require an RFC, as there are many ways to solve this and it isn't obvious which will scale (I learned that the hard way in #1354!). Analyzing the EMP dataset will help identify a solution that scales, at which point it can be integrated into scikit-bio and this temporary solution can be removed.

@wasade, we had initially discussed having `beta_diversity` raise a warning when run in partial mode describing its caveats and eminent removal from the codebase. @gregcaporaso, @ebolyen, and I talked offline and thought it would be better to have a separate function for partial beta diversity that is deprecated outright. We'd prefer this approach because it'll make it clearer to the user what functionality is slated for removal, it "packages" the deprecated functionality in a single function, keeps temporary code out of  `beta_diversity`, and avoids further complicating the function (it already has a lot going on). It also avoids mutually exclusive parameters and other inconsistencies in `beta_diversity`: partial beta diversity doesn't work with `pairwise_func`, requires `ids`, and only supports resolvable metrics and callables. It'll be a lot cleaner to have a separate function with its own signature. Are you okay with making this change?

wasade:
*jinx*

jairideout:
😶

wasade:
Totally fine with the proposed change, thanks!

jairideout:
Awesome, thanks @wasade!

gregcaporaso:
Thanks for sorting this out guys!

wasade:
@jairideout @ebolyen, I think we're good. I shuffled the logic into `partial_beta_diversity` maintaining consistency with the signature of `beta_diversity`. `partial_beta_diversity` is denoted as `deprecated` the reason being the return type can be misleading. I set the version of removal at `0.5.0` but I'm not sure if that's best.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 * Added `skbio.sequence.Sequence.replace` for assigning a character to positions in a `Sequence`. ([#1222](https://github.com/biocore/scikit-bio/issues/1222))
 * Added support for `pandas.RangeIndex`, lowering the memory footprint of default integer index objects. `Sequence.positional_metadata` and `TabularMSA.positional_metadata` now use `pd.RangeIndex` as the positional metadata index. `TabularMSA` now uses `pd.RangeIndex` as the default index. Usage of `pd.RangeIndex` over the previous `pd.Int64Index` [should be transparent](http://pandas.pydata.org/pandas-docs/version/0.18.0/whatsnew.html#range-index), so these changes should be non-breaking to users. scikit-bio now depends on pandas >= 0.18.0 ([#1308](https://github.com/biocore/scikit-bio/issues/1308))
 * Added `reset_index=False` parameter to `TabularMSA.append` and `TabularMSA.extend` for resetting the MSA's index to the default index after appending/extending.
+* Added support for partial pairwise calculations via `skbio.diversity.partial_beta_diversity`. ([#1221](https://github.com/biocore/scikit-bio/issues/1221), [#1337](https://github.com/biocore/scikit-bio/pull/1337))
```
josenavas: Is it worth mentioning this here given the functionality is directly deprecated?

wasade: Good question. This is a bit of a grey area as it's possible the method is
"experimental" but the return is "deprecated"

josenavas: Thanks! I don't have a strong preference and I don't see any other issues on the PR, so if @ebolyen and @jairideout agree with this, it should be ready to go!

Thanks @wasade !

skbio/diversity/_driver.py
```diff
@@ -180,9 +182,114 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+    for u, v in id_pairs_indexed:
+        dm[u, v] = metric(counts[u], counts[v], **kwargs)
+
+    return DistanceMatrix(dm + dm.T, ids)
+
+
 @experimental(as_of="0.4.0")
 def beta_diversity(metric, counts, ids=None, validate=True, pairwise_func=None,
-                   **kwargs):
+                   id_pairs=None, **kwargs):
```
gregcaporaso: This parameter is no longer used here, so should be dropped from the signature and doc.

skbio/diversity/__init__.py
```diff
@@ -377,15 +377,15 @@
    >>> plt.close('all') # not necessary for normal use
    >>> fig = sample_md.boxplot(column='Faith PD', by='body_site')
 
-We can also compute Spearman correlations between all pairs of columns in this
-``DataFrame``. Since our alpha diversity metrics are the only two numeric
-columns (and thus the only columns for which Spearman correlation is relevant),
-this will give us a symmetric 2x2 correlation matrix.
+    We can also compute Spearman correlations between all pairs of columns in
```
jairideout: Can you revert the indentation change to this section of the docstring? It was correct, the indented sections are necessary when plotting examples.

skbio/diversity/__init__.py
```diff
@@ -399,10 +399,10 @@
 
 from skbio.util import TestRunner
 
-from ._driver import (alpha_diversity, beta_diversity,
+from ._driver import (alpha_diversity, beta_diversity, partial_beta_diversity,
```
jairideout: list `partial_beta_diversity` under "Functions" section of docstring above, that way it will be linked in the built docs

CHANGELOG.md
jairideout: How about adding the following sentence:

"""
This function is immediately deprecated as its return type will change in the future and should be used with caution in its present form (see the function's documentation for details).
"""

skbio/diversity/_driver.py
```diff
@@ -180,9 +182,114 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
     return pd.Series(results, index=ids)
 
 
+@deprecated(as_of='0.4.2-dev', until='0.5.0',
```
jairideout: 0.5.0 is the next release (scheduled for 5/31). Let's bump this to 0.5.1

skbio/diversity/_driver.py
```diff
@@ -180,9 +182,114 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+            reason=('The return type is unstable. Developer caution is '
+                    'advised. The resulting DistanceMatrix object may include '
```
jairideout: *may* -> *will*

skbio/diversity/_driver.py
```diff
@@ -180,9 +182,114 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+                    'zeros when distance has not been calculated, and '
+                    'therefore can be misleading.'))
+def partial_beta_diversity(metric, counts, ids=None, validate=None,
```
jairideout: make `ids` a positional argument since it is always required

jairideout: change to `validate=True` to match `beta_diversity`

skbio/diversity/_driver.py
```diff
@@ -180,9 +182,114 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+    counts : 2D array_like of ints or floats
+        Matrix containing count/abundance data where each row contains counts
+        of OTUs in a given sample.
+    ids : iterable of strs
+        Identifiers for each sample in ``counts``.
+    id_pairs : iterable of tuple, optional
+        An iterable of tuples of IDs to compare (e.g., ``[('a', 'b'), ('a',
+        'c'), ...])``. If specified, the set of IDs described must be a subset
+        of ``ids``.
+    validate : ignored
```
jairideout: This parameter isn't ignored in the current implementation. Change to:

```
validate : bool, optional
    See ``skbio.diversity.beta_diversity`` for details.
```

wasade: It is in this method though...

skbio/diversity/_driver.py
wasade: I felt it was misleading to have `validate=True` as it isn't used. If we want to match the signature too then we should keep `ids` as kwarg

jairideout: make `id_pairs` a positional argument. if a user wants all pairs they should be using `beta_diversity` since this function is unstable/deprecated.

so the new function sig is:

```python
def partial_beta_diversity(metric, counts, ids, id_pairs, validate=True, **kwargs):
    ...
```

skbio/diversity/_driver.py
```diff
@@ -180,9 +182,114 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+                    'advised. The resulting DistanceMatrix object may include '
+                    'zeros when distance has not been calculated, and '
+                    'therefore can be misleading.'))
+def partial_beta_diversity(metric, counts, ids=None, validate=None,
+                           id_pairs=None, **kwargs):
+    """Compute distances only between specified ID pairs
+
+    Parameters
+    ----------
+    metric : callable
```
jairideout: `metric : str or callable`

skbio/diversity/_driver.py
```diff
@@ -180,9 +182,114 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+        The pairwise distance function to apply. ``metric`` must be resolvable
```
jairideout: *If ``metric`` is a string, must be resolvable by scikit-bio...*

skbio/diversity/_driver.py
wasade: Let's step back. Is it desirable for `beta_diversity(*args, **kwargs) == partial_beta_diversity(*args, **kwargs)` given a valid metric for `partial_beta_diversity` and an enumerated upper triangle of the distance matrix for `id_pairs`? In my opinion, that would be the most parsimonious API. However, I don't have strong feelings on this. 

(I realize that `validate=None` would make the above constraint not work)

skbio/diversity/_driver.py
wasade: oh, hah! sorry, didn't see the passing of it to the unifrac setup methods. my bad

skbio/diversity/_driver.py
```diff
@@ -180,9 +182,114 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+    validate : ignored
+        This kwarg is preserved to maintain consistency with the beta_diversity
+        signature.
+    kwargs : kwargs, optional
+        Metric-specific parameters.
+
+    Returns
+    -------
+    skbio.DistanceMatrix
+        Distances between all pairs of samples (i.e., rows). The number of
```
jairideout: change this section to:

*Distances between pairs of samples indicated by ``id_pairs``. Pairwise distances not defined by `id_pairs` will be 0.0. Use this resulting ``DistanceMatrix`` with caution as 0.0 is a valid distance.*

skbio/diversity/_driver.py
```diff
@@ -180,9 +182,114 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+
+        Warning: zeros in this matrix may represent a zero distance, or may
+        represent a pairwise distance that was not calculated.
+
+    Raises
+    ------
+    ValueError
+        If ``ids`` are not specified.
+        If ``id_pairs`` are not a subset of ``ids``.
+        If ``metric`` is not a callable.
```
jairideout: *or a string resolvable by scikit-bio.*

skbio/diversity/_driver.py
```diff
@@ -180,9 +182,114 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+        If duplicates are observed in ``id_pairs``.
+
+    See Also
+    --------
+    skbio.beta_diversity
```
jairideout: change to `skbio.diversity.beta_diversity`

skbio/diversity/_driver.py
```diff
@@ -180,9 +182,114 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+
+    if id_pairs is None:
+        # determine all pairs of the upper triangle.
+        # alternatively, could just call beta_diversity?
+        id_pairs = []
+        for row, i in enumerate(ids):
+            for j in ids[row+1:]:
+                id_pairs.append((i, j))
+
+    all_ids_in_pairs = set(itertools.chain.from_iterable(id_pairs))
```
jairideout: The current implementation will fail if providing an iterable because the code takes multiple passes through `id_pairs`. Can you add a test case for an iterable `id_pairs`? e.g., call `iter([('a','b'), ('c', 'd')])` or pass a generator

skbio/diversity/_driver.py
jairideout: @gregcaporaso and I don't think the signatures need to be the same since `partial_beta_diversity` is deprecated and slated for removal/change. A user shouldn't inadvertently use `partial_beta_diversity` where they meant to be using `beta_diversity`. Also, it doesn't make sense to adapt the `beta_diversity` `experimental` API to match the `deprecated` `partial_beta_diversity` API.

jairideout: > I felt it was misleading to have validate=True as it isn't used.

It is used, it's passed below in the UniFrac sections.

skbio/diversity/_driver.py
```diff
@@ -180,6 +182,111 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+    if not all_ids_in_pairs.issubset(ids):
+        raise ValueError("`id_pairs` are not a subset of `ids`")
+
+    hashes = {i for i in id_pairs}.union({i[::-1] for i in id_pairs})
+    if len(hashes) != len(id_pairs) * 2:
+        raise ValueError("Duplicate ID pairs observed.")
```
jairideout: This also raises if a "within" ID pair is passed (e.g., `('a', 'a')`). Can you update the error message to indicate that could also be causing the error?

skbio/diversity/_driver.py
```diff
@@ -180,6 +182,111 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+        If ``id_pairs`` are not a subset of ``ids``.
+        If ``metric`` is not a callable.
+        If duplicates are observed in ``id_pairs``.
+
+    See Also
+    --------
+    skbio.beta_diversity
+    skbio.diversity.get_beta_diversity_metrics
+
+    """
```
jairideout: Add these lines to the start of the function:

```python
if validate:
    counts = _validate_counts_matrix(counts, ids=ids)
```

wasade: I had a reason for not including it. But I can't remember it now.

skbio/diversity/tests/test_driver.py
```diff
@@ -529,6 +530,48 @@ def test_unweighted_unifrac(self):
                 npt.assert_almost_equal(dm1[id1, id2],
                                         expected_dm[id1, id2], 6)
 
+    def test_unweighted_unifrac_partial(self):
```
jairideout: Can you move all `partial_beta_diversity` test code into a new class `TestPartialBetaDiversity`?

skbio/diversity/tests/test_driver.py
```diff
@@ -635,5 +678,83 @@ def test_get_beta_diversity_metrics_sorted(self):
         n = sorted(list(m))
         self.assertEqual(m, n)
 
+
+class PartialPairwise(TestCase):
```
jairideout: merge these test cases into `TestPartialBetaDiversity`

skbio/diversity/tests/test_driver.py
```diff
@@ -635,5 +678,83 @@ def test_get_beta_diversity_metrics_sorted(self):
...
+        self.table2 = [[23, 64, 14, 0, 0, 3, 1],
+                       [0, 3, 35, 42, 0, 12, 1],
+                       [0, 5, 5, 0, 40, 40, 0],
+                       [44, 35, 9, 0, 1, 0, 0],
+                       [0, 2, 8, 0, 35, 45, 1],
+                       [0, 0, 25, 35, 0, 19, 0]]
+        self.table2 = np.array(self.table2)
+        self.sids2 = list('ABCDEF')
+
+    def test_duplicate_pairs(self):
```
jairideout: add a test case for "within" ID pair raising an error

jairideout:
Thanks for the updates @wasade! @gregcaporaso and I are done reviewing.

skbio/diversity/_driver.py
jairideout: It's important to include b/c it converts `array_like` into an `ndarray`. I think we have an open issue somewhere to separate validation from input munging but for now `_validate_counts_matrix` has to be called. It's also doing important validations like making sure the length of `ids` matches `counts`. Since we're exposing `validate=True` as part of the API and referring users to `beta_diversity`'s docs, validation needs to match here.

skbio/diversity/__init__.py
wasade: The reason I indented them was that they weren't indented and stood out w.r.t. the rest of the example text (see [here](https://github.com/biocore/scikit-bio/blob/master/skbio/diversity/__init__.py#L380))

jairideout: I agree it's a little weird, but the indentation is correct w.r.t. how Sphinx/numpydoc handle plotting

jairideout: If you want to keep this change, please build the docs locally and make sure they look correct (and that all plots are visible).

skbio/diversity/_driver.py
```diff
@@ -180,6 +182,106 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+    metric : str or callable
+        The pairwise distance function to apply. If ``metric`` is a string, it
+        must be resolvable by scikit-bio (e.g., UniFrac methods), or must be
+        callable.
+    counts : 2D array_like of ints or floats
+        Matrix containing count/abundance data where each row contains counts
+        of OTUs in a given sample.
+    ids : iterable of strs
+        Identifiers for each sample in ``counts``.
+    id_pairs : iterable of tuple, optional
```
jairideout: remove `optional`

skbio/diversity/tests/test_driver.py
```diff
@@ -635,5 +636,148 @@ def test_get_beta_diversity_metrics_sorted(self):
...
+            '((O1:0.25, O2:0.50):0.25, O3:0.75)root;'))
+        self.oids1 = ['O1', 'O2']
+
+        self.table2 = [[23, 64, 14, 0, 0, 3, 1],
+                       [0, 3, 35, 42, 0, 12, 1],
+                       [0, 5, 5, 0, 40, 40, 0],
+                       [44, 35, 9, 0, 1, 0, 0],
+                       [0, 2, 8, 0, 35, 45, 1],
+                       [0, 0, 25, 35, 0, 19, 0]]
+        self.table2 = np.array(self.table2)
```
jairideout: Can you remove this line so that `table2` is a nested Python list? That would have caught a bug where `counts` is `array_like` and gets converted in `_validate_counts_matrix`.

skbio/diversity/_driver.py
```diff
@@ -180,6 +182,106 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+    See Also
+    --------
+    skbio.diversity.beta_diversity
+    skbio.diversity.get_beta_diversity_metrics
+
+    """
+    if validate:
+        counts = _validate_counts_matrix(counts, ids=ids)
+
+    if ids is None:
```
jairideout: remove this check now that `ids` is a required parameter

skbio/diversity/_driver.py
```diff
@@ -180,6 +182,106 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+                counts, otu_ids=otu_ids, tree=tree, normalized=normalized,
+                validate=validate)
+        counts = counts_by_node
+    elif callable(metric):
+        metric = functools.partial(metric, **kwargs)
+        # remove all values from kwargs, since they have already been provided
+        # through the partial
+        kwargs = {}
+    else:
+        raise ValueError("partial_beta_diversity is only compatible with "
```
jairideout: Can you add a test for this case? That'll fix the coverage drop.

jairideout:
@gregcaporaso and I took a final pass and had a few comments (all are easy to address). Once that's done this is ready for merge!

jairideout:
Looks great, thanks @wasade!

ebolyen:
:fireworks:

wasade:
Woo!!!

Issue #1344, opened 2016-05-03
ANCOM documentation confusion


wasade:
The `notes` section for `ancom` describes statistical tests that the authors recommend, but I believe there is a typo. The specific line is [here](https://github.com/biocore/scikit-bio/blob/master/skbio/stats/composition.py#L724), where it indicates that the authors recommend method `w` and `x` when "the number of groups is greater than 2", and in the same sentence, it then recommends methods `y` and `z` when "the number of groups is greater than 2." My guess is that the first example is supposed to be "less than". 

cc @mortonjt

mortonjt:
Not quite - you can only run the `ttest_ind` and `mann-whitney` if there are two groups.

Fixing

mortonjt:
Actually, I'm looking over the wording, and am a little confused where the confusion is stemming from. 

The first two methods are parametric, and the second two methods are non-parametric.  Do you have any suggestions on improving this?

wasade:
Remove the repetition of "greater than 2"?

jairideout:
The current text is correct but hard to parse. I followed up on #1346 with a suggested rewording.

mortonjt:
Thanks!

wasade:
Thanks!

jairideout:
Fixed in #1346.

PR #1346, opened 2016-05-03
DOC: Resolving ANCOM documentation confusion

mortonjt:
Resolving confusion in ancom documentation

### DIFF
--- a/skbio/stats/composition.py
+++ b/skbio/stats/composition.py
@@ -721,14 +721,11 @@ def ancom(table, grouping,
     The developers of this method recommend the following significance tests
     ([2]_, Supplementary File 1, top of page 11): the standard parametric
     t-test (``scipy.stats.ttest_ind``) or one-way ANOVA
-    (``scipy.stats.f_oneway``) if the number of groups is greater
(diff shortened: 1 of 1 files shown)


mortonjt:
Addresses #1344

jairideout:
The previous wording is confusing but correct. It's listing parametric tests first (one for 2 groups, one for > 2 groups) and nonparametric tests (one for 2 groups, one for > 2 groups). I think that info is useful but needs rewording (it took me several reads to understand it). What about:

"""
The developers of this method recommend the following significance tests: if there are 2 groups, use the standard parametric t-test or non-parametric Wilcoxon rank sum test. If there are more than 2 groups, use parametric one-way ANOVA or nonparametric Kruskal-Wallis. Because one-way ANOVA is equivalent to...
"""

I left out the parenthetical bits but those would still be included in the text.

jairideout:
This looks good but can you add back in the scipy function names? That's what I meant by this sentence in my previous comment:

> I left out the parenthetical bits but those would still be included in the text.

mortonjt:
Sorry.  Added.

jairideout:
Thanks!

PR #1352, opened 2016-05-11
beta diversity via block decomposition 

wasade:
This PR includes #1337, and assumes that `beta_diversity` can return a `PartialDistanceMatrix`. Right now, this is an unexercised possible direction to take with doing the block decomposition. 

The design is such that it should allow for leveraging any parallel framework as the "map" and "reduce" functions can be arbitrary. The individual "computable" blocks are packed into `partial` functions so they can be executed by the `map_f` effectively deferring the heavy lifting (i.e., `_block_party` due to `shear` and `beta_diversity`) until the last possible moment.  

Note: this is _unexercised_ code, so it is likely that there are problems. It may also make sense to revisit how the methods themselves are decomposed. I am primarily interested at this time in whether the approach to parallelism is one that makes sense in the bigger picture of scikit-bio. It is critical that this method is usable in a parallel context such that execution can be distributed to independent execution environments. 

A notebook highlighting the approach to the parallelism can be found [here](https://gist.github.com/wasade/c3b3e8baa7c16fe3699d4371c8fffb53).


mortonjt:
👍 I think this is best route to go, since map/reduce is a standard for parallel applications.

wasade:
...and actually, I'm going to put together a notebook to see how this would work with `ipyparallel`

wasade:
It may be more appropriate to express `map_f` as mapping a function over arguments.

wasade:
[bam bam](https://gist.github.com/wasade/c3b3e8baa7c16fe3699d4371c8fffb53)

josenavas:
I just have to say that I love that last comment and the name of the gist xD

mortonjt:
Coming to think of it.  Would it be appropriate to open up an rfc on this?

wasade:
Would prefer not as this is defined and needed for EMP

skbio/diversity/_driver.py
```diff
@@ -180,9 +183,277 @@ def alpha_diversity(metric, counts, ids=None, validate=True, **kwargs):
...
+    """
+    ids = kwargs.get('ids')
+    blocksize = kwargs.get('blocksize')
+    for row_ids, col_ids, id_pairs in _block_and_id_pairs(ids, blocksize):
+        kw = kwargs.copy()
+        kw['id_pairs'] = id_pairs
+        kw['row_ids'] = row_ids
+        kw['col_ids'] = col_ids
+
+        yield functools.partial(_block_compute(*args, **kw))
```
wasade: if this `yield`s the args, then `_block_compute` can be used in a `map` within `block_beta_diversity`, which would then pull `map_f` into being an actual map as opposed to a dispatch. Removes possible obfuscation with `partial` as well

wasade:
This PR still depends on #1337 getting merged first. At this point, the method is minimally tested. A revised version of the POC notebook can be found [here](https://gist.github.com/wasade/c3b3e8baa7c16fe3699d4371c8fffb53) showing parallel execution via `ipyparallel` and `multiprocessing`. 

It's still rough around the edges and we may want to expand the test suite but I think it's good for a first pass review

wasade:
Short of conflicts, is there anything blocking this PR?

jairideout:
Not that I know of. The code needs to be reviewed, do you have devs on your end that can do that? We'll want a core dev to do the final pass/merge (likely @ebolyen, as I will be traveling the next week and unavailable). He should be available to review this and the Sequence interval metadata PR on or after October 10.

wasade:
Failures are unrelated to this PR, and suggest a possible CI issue

wasade:
@jairideout, I pinged @ElDeveloper and @RNAer for review

jairideout:
OK, we'll need someone to look into those CI failures and fix in a separate PR. Are you available to do that? We are swamped with QIIME 2 stuff and have zero time to commit to skbio right now. Even creating an issue would be a big help.

jairideout:
> @jairideout, I pinged @ElDeveloper and @RNAer for review

Great, thank you!

wasade:
@jairideout #1435 created. I'm taking a look but its in the dark parts of the io system which I'm not familiar with, unsure if a PR will stem from the investigation or not

jairideout:
Thanks for creating the issue! Perhaps @ebolyen can look into it next week if it is I/O related.

skbio/diversity/_block.py
```diff
@@ -0,0 +1,270 @@
...
+    counts : 2D array_like of ints or floats
+        Matrix containing count/abundance data where each row contains counts
+        of OTUs in a given sample.
+    row_ids : 1D np.ndarray of int
+        Block row IDs to keep in the counts matrix.
+    col_ids : 1D np.ndarray of int
+        Block column IDs to keep in the counts matrix. Note, these correspond
+        to rows in the counts matrix, but columns in a subsequent distance
+        matrix.
+
```
ElDeveloper: Can you add kwargs?

skbio/diversity/_block.py
```diff
@@ -0,0 +1,270 @@
...
+    ids_to_keep = np.unique(np.hstack([row_ids, col_ids]))
+
+    # create a view of the relevant samples
+    counts_block = counts[ids_to_keep]
+
+    # remove from the block any empty observations
+    # NOTE: this will perform an implicit copy
+    nonzero_cols = (counts_block != 0).any(axis=0)
+    counts_block = counts_block[:, nonzero_cols]
```
ElDeveloper: This seems a bit odd you are slicing with a True value? This should return a warning or am I misunderstanding something?

ElDeveloper: Ah, I see what this does.

skbio/diversity/_block.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+def _pairs_to_compute(rids, cids):
+    """Determine the pairs of samples to compute distances between
+
+    Parameters
+    ----------
+    rids : Iterable
+        The row IDs in the partial pairwise computation.
+
```
ElDeveloper: Can you remove this line?

skbio/diversity/_block.py
```diff
@@ -0,0 +1,270 @@
...
+        triangle, complain loudly.
+
+    Returns
+    -------
+    list of tuple
+        The ID pairs to compute distances between.
+    """
+    # if identical, gather the upper triangle
+    if len(rids) == len(cids) and (rids == cids).all():
+        return [(i, j) for idx, i in enumerate(rids) for j in rids[idx+1:]]
```
ElDeveloper: Could this be achieved with `np.triu_indices`?

skbio/diversity/_block.py
```diff
@@ -0,0 +1,270 @@
...
+
+    """
+    for kwargs in kw_gen:
+        yield func(**kwargs)
+
+
+def _reduce(blocks):
+    """Reduce an iterable of partial distance matrices into a full matrix"""
+    all_blocks = list(blocks)
+    n_ids = max(map(lambda x: max(x.ids), all_blocks)) + 1
```
ElDeveloper: Can you help me understand what's going on here?

skbio/diversity/_block.py
```diff
@@ -0,0 +1,270 @@
...
+
+    mat = np.zeros((n_ids, n_ids), dtype=float)
+
+    # TODO: something smarter.
+    for block in all_blocks:
+        n_blk_ids = len(block.ids)
+
+        # get the corresponding coordinates in the master matrix
+        master_idx = [(i, j) for row, i in enumerate(block.ids)
```
ElDeveloper: I'm unsure, but maybe np.triu_indices may work here ... maybe not 😕

skbio/diversity/_block.py
wasade: Adding

```python
    # Determine the maximum integer ID observed in the blocks. There exists a
    # 1-1 mapping between the integer ID and a sample ID. We increment by 1
    # as the integer ID space begins with zero, and we'll be using this value
    # to determine the size of the resulting full distance matrix.
```

skbio/diversity/_block.py
wasade: It does as before but at the moment, need to cram it through `zip(*foo)` so it appears to come out as 2x slower. Although, it potentially makes the code a little more readable so that may make the change worth it.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 * `DistanceMatrix.from_iterable` now accepts a `validate=True` parameter. ([#1343](https://github.com/biocore/scikit-bio/issues/1343)).
 * Added `IntervalMetadata` and `Interval` classes in `skbio.metadata` to store, query, and manipulate information of a sub-region of a sequence. ([#1414](https://github.com/biocore/scikit-bio/issues/1414)).
 
+* Added parallel beta diversity calculation support via `skbio.diversity.block_beta_diversity`. ([#1181](https://github.com/biocore/scikit-bio/issues/1181), [#1352](https://github.com/biocore/scikit-bio/pull/1352)).
```
wasade: Revised, thanks!

skbio/diversity/_block.py
```diff
@@ -0,0 +1,270 @@
...
+from skbio.stats.distance import DistanceMatrix
+from skbio.diversity._util import _validate_counts_matrix
+
+
+def _generate_id_blocks(ids=None, k=None, **kwargs):
+    """Generate blocks of IDs that map into a DistanceMatrix
+
+    Parameters
+    ----------
+    ids : Iterable
```
wasade: Revised, thanks!

skbio/diversity/_block.py
```diff
@@ -0,0 +1,270 @@
...
+        An iterable of IDs of whatever type.
+    k : int
```
wasade: Revised, thanks! I set a default value and am explicitly checking now that `ids is not None`.

skbio/diversity/_block.py
```diff
@@ -0,0 +1,270 @@
...
+        block_idx = [(i, j) for row, i in enumerate(range(n_blk_ids))
+                     for j in range(row+1, n_blk_ids)]
+
+        for (m_i, m_j), (b_i, b_j) in zip(master_idx, block_idx):
+            mat[m_i, m_j] += block.data[b_i, b_j]
+
+    return DistanceMatrix(mat + mat.T, list(range(n_ids)))
+
+
+@experimental(as_of="0.4.2-dev")
```
wasade: Fixed

skbio/diversity/_block.py
```diff
@@ -0,0 +1,270 @@
...
+def block_beta_diversity(metric, counts, ids, validate=True, k=64, **kwargs):
+    """Perform a block-decomposition beta diversity calculation
+
+    Parameters
+    ----------
+    reduce_f : function, optional
```
wasade: Fixed

wasade:
Thanks @ElDeveloper!

skbio/diversity/_block.py
```diff
@@ -0,0 +1,303 @@
...
+
+def _block_kwargs(**kwargs):
+    """Construct arguments describing a block to compute
+
+    Returns
+    -------
+    dict
+        The parameters for the block of the distance matrix to compute.
+    """
+    for row_ids, col_ids in _generate_id_blocks(**kwargs):
```
RNAer: oh, I see now why you need `kwargs` in `_generate_id_blocks` func param. however, would it be more clear to just pass `kwargs['ids']` and `kwargs['k']` explicitly?

skbio/diversity/_block.py
```diff
@@ -0,0 +1,303 @@
...
+
+    if 'tree' in kwargs and 'otu_ids' in kwargs:
+        block_kwargs['otu_ids'] = np.asarray(kwargs['otu_ids'])[nonzero_cols]
+        block_kwargs['tree'] = kwargs['tree'].shear(block_kwargs['otu_ids'])
+
+    return block_kwargs
+
+
+def _pairs_to_compute(rids, cids):
+    """Determine the pairs of samples to compute distances between
```
wasade: I think that reads right?

skbio/diversity/_block.py
wasade: cant' remember original reason so simplifying

wasade:
thanks @RNAer!

ElDeveloper:
I took this out for a spin, and it works as expected, so that's wonderful!

The one thing that I noticed was that when you run `partial_beta_diversity` there's noticeable running-time overhead for small datasets, I wonder if we can do something like warning the user about this  (but perhaps that's an overly preventive attitude). I was specially surprised when I set k=1 and the execution took a while to run, maybe this is something what we can warn against?

skbio/diversity/_block.py
```diff
@@ -0,0 +1,280 @@
...
+
+        for (m_i, m_j), (b_i, b_j) in zip(master_idx, block_idx):
+            mat[m_i, m_j] += block.data[b_i, b_j]
+
+    return DistanceMatrix(mat + mat.T, list(range(n_ids)))
+
+
+@experimental(as_of="0.5.0-dev")
+def block_beta_diversity(metric, counts, ids, validate=True, k=64,
+                         reduce_f=_reduce, map_f=_map, **kwargs):
```
ebolyen: We've been taking the approach of using `None` in these cases. `_reduce`/`_map` won't show nicely in the documentation.

skbio/diversity/_block.py
```diff
@@ -0,0 +1,280 @@
...
+        Matrix containing count/abundance data where each row contains counts
+        of OTUs in a given sample.
+    ids : iterable of strs
+        Identifiers for each sample in ``counts``.
+    validate : bool, optional
+        See ``skbio.diversity.beta_diversity`` for details.
+    reduce_f : function, optional
+        A method to reduce `PartialDistanceMatrix` objects into a single
+        `DistanceMatrix`. The expected signature is:
+            `DistanceMatrix <- f(Iterable of PartialDistanceMatrix)`
```
ebolyen: In typing lingo: `Callable[Iterable[PartialDistanceMatrix], DistanceMatrix]` up to you if you want to use that, but at least reverse the arrow `f(Iterable of PartialDistanceMatrix) -> DistanceMatrix`.

skbio/diversity/_block.py
```diff
@@ -0,0 +1,280 @@
...
+        must be resolvable by scikit-bio (e.g., UniFrac methods), or must be
+        callable.
+    counts : 2D array_like of ints or floats
+        Matrix containing count/abundance data where each row contains counts
+        of OTUs in a given sample.
+    ids : iterable of strs
+        Identifiers for each sample in ``counts``.
+    validate : bool, optional
+        See ``skbio.diversity.beta_diversity`` for details.
+    reduce_f : function, optional
```
ebolyen: This isn't actually a reduce function, which would be:
 `reduce(f(accumulator, element) -> accumulator, accumulator,  iterable) -> accumulator`.
Is there another word we could use? It seems to be basically a `sum` operation if I understand it correctly.

skbio/diversity/_block.py
```diff
@@ -0,0 +1,280 @@
...
+            `DistanceMatrix <- f(Iterable of PartialDistanceMatrix)`
+    map_f: function, optional
+        A method that accepts a `_block_compute`. The expected signature is:
+            `DistanceMatrix <- f(**kwargs)`
+        NOTE: ipyparallel's `map_async` will not work here as we need to be
+        able to pass around `**kwargs``.
+    k : int, optional
+        The blocksize used when computing distances
+    kwargs : kwargs, optional
+        Metric-specific parameters. See kwarg note below.
```
ebolyen: There is no note below.

skbio/diversity/_block.py
```diff
@@ -0,0 +1,280 @@
...
+
+    return partial_beta_diversity(**block_kw)
+
+
+def _map(func, kw_gen):
+    """Map a function over arguments
+
+    Note
+    ----
+    builtin map does not allow for mapping with kwargs.
```
ebolyen: Why not `map(lambda kwargs: func(**kwargs), kw_gen)`?

ebolyen: Actually I like your implementation better, and this is meant to be replaced by a parallel implementation typically right? I would just drop the note.

skbio/diversity/_block.py
```diff
@@ -0,0 +1,280 @@
...
+        # get the corresponding coordinates in the master matrix
+        master_idx = [(i, j) for row, i in enumerate(block.ids)
+                      for j in block.ids[row+1:]]
+
+        # get the corresponding coordinates within the current block
+        block_idx = [(i, j) for row, i in enumerate(range(n_blk_ids))
+                     for j in range(row+1, n_blk_ids)]
+
+        for (m_i, m_j), (b_i, b_j) in zip(master_idx, block_idx):
+            mat[m_i, m_j] += block.data[b_i, b_j]
```
ebolyen: Could this be vectorized so that you could just add `block.data` in one go? Since 0 represents an uncalculated element, I don't see why that wouldn't work.

skbio/diversity/_block.py
```diff
@@ -0,0 +1,280 @@
...
+    for kwargs in kw_gen:
+        yield func(**kwargs)
+
+
+def _reduce(blocks):
+    """Reduce an iterable of partial distance matrices into a full matrix"""
+    all_blocks = list(blocks)
+
+    # Determine the maximum integer ID observed in the blocks. There exists a
+    # 1-1 mapping between the integer ID and a sample ID. We increment by 1
```
ebolyen: If that's the case why not just use `len` on the ids?

wasade:
Thanks!!

I computed runtime curves a while ago (but can't find them right now...).
It is definitely the case that this method is, at least for
weighted/unweighted unifrac, not useful on small datasets (either number of
samples or number of OTUs). The default k=64 was based off of the
inflection point for weighted/unweighted unifrac relative to random subsets
from AG.

skbio/diversity/_block.py
```diff
@@ -0,0 +1,280 @@
...
+        kwargs that describe the block to compute. A filtered ``counts`` matrix
+        is stored in kwargs. If applicable, a filtered ``tree`` and ``otu_ids``
+        are also stored.
+    """
+    # filter kwargs to valid keys for downstream block compute
+    # explicit copies are not necessary as implicit copies or overwwrites are
+    # performed below.
+    valid_block_keys = {'counts', 'ids', 'tree', 'otu_ids', 'metric',
+                        'id_pairs', 'validate'}
+    block_kwargs = {k: v for k, v in kwargs.items() if k in valid_block_keys}
```
ebolyen: Actually it might be best to pull the validation out entirely as this gets run many times? Maybe `_block_kwargs` could inherit this? I would assume anything returned from `_block_kwargs` could be considered "safe".

skbio/diversity/_block.py
```diff
@@ -0,0 +1,280 @@
...
+
+    ids_to_keep = np.unique(np.hstack([row_ids, col_ids]))
+
+    # create a view of the relevant samples
+    counts_block = counts[ids_to_keep]
```
ebolyen: Why isn't `counts_block = counts[row_ids, col_ids]`? I must be missing something here. It looks like you are only slicing the columns by those which are nonzero.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 * `DistanceMatrix.from_iterable` now accepts a `validate=True` parameter. ([#1343](https://github.com/biocore/scikit-bio/issues/1343)).
 * Added `IntervalMetadata` and `Interval` classes in `skbio.metadata` to store, query, and manipulate information of a sub-region of a sequence. ([#1414](https://github.com/biocore/scikit-bio/issues/1414)).
```
ebolyen: extra newline

skbio/diversity/_block.py
ebolyen: Oh, these blocks all have partial id sets, that makes sense.

ebolyen: This function is very involved, is it reasonable to expect a user to be able to override it successfully?

skbio/diversity/_block.py
ebolyen: I think I'm starting to understand, this is the "reduce" step of map-reduce. Maybe point that out in the docs? I assume this exists so that results can be merged piecewise.

skbio/diversity/_block.py
```diff
@@ -0,0 +1,280 @@
...
+        Identifiers for each sample in ``counts``.
+    validate : bool, optional
+        See ``skbio.diversity.beta_diversity`` for details.
+    reduce_f : function, optional
+        A method to reduce `PartialDistanceMatrix` objects into a single
+        `DistanceMatrix`. The expected signature is:
+            `DistanceMatrix <- f(Iterable of PartialDistanceMatrix)`
+    map_f: function, optional
+        A method that accepts a `_block_compute`. The expected signature is:
+            `DistanceMatrix <- f(**kwargs)`
```
ebolyen: Same here, the arrow is backwards from typical convention. It looks like assignment notation.

skbio/diversity/_block.py
```diff
@@ -0,0 +1,280 @@
...
+
+def _reduce(blocks):
+    """Reduce an iterable of partial distance matrices into a full matrix"""
+    all_blocks = list(blocks)
+
+    # Determine the maximum integer ID observed in the blocks. There exists a
+    # 1-1 mapping between the integer ID and a sample ID. We increment by 1
+    # as the integer ID space begins with zero, and we'll be using this value
+    # to determine the size of the resulting full distance matrix.
+    n_ids = max(map(lambda x: max(x.ids), all_blocks)) + 1
```
ebolyen: Should this be a provided argument?

skbio/diversity/tests/test_block.py
```diff
@@ -0,0 +1,218 @@
...
+                'ids': [1, 2, 3, 4, 5],
+                'foo': 'bar', 'k': 2},
+
+               {'row_ids': np.array((2, 3)),
+                'col_ids': np.array((4,)),
+                'id_pairs': [(2, 4), (3, 4)],
+                'ids': [1, 2, 3, 4, 5],
+                'foo': 'bar', 'k': 2}]
+
+        obs = list(_block_kwargs(**kws))
```
ebolyen: Would it be better to compare sets? the precise order seems unimportant.

skbio/diversity/tests/test_block.py
```diff
@@ -0,0 +1,218 @@
...
+        exp = [1, 10]
+        obs = list(_map(func, kwargs))
+        self.assertEqual(obs, exp)
+
+    def test_reduce(self):
+        # note, the reduce doesn't actually care about what pairs are computed
+        # so if a distance between pairs exists multiple times, it'll get
+        # added. as such, this reduction is only safe to perform if by
+        # the block_beta_diversity method which assures that distances are not
+        # computed multiple times.
```
ebolyen: Worth moving this comment to `_reduce`?

skbio/diversity/tests/test_block.py
```diff
@@ -0,0 +1,218 @@
...
+               _generate_id_blocks(list(range(5)), 2)]
+        npt.assert_equal(obs, exp)
+
+    def test_block_party_tree(self):
+        counts = np.array([[1, 1, 1],
+                           [1, 0, 1],
+                           [1, 0, 1],
+                           [0, 0, 1],
+                           [0, 1, 1]])
+        tree = TreeNode.read(io.StringIO('(a:1,b:2,c:3);'))
```
ebolyen: Not important, but a shortcut is `TreeNode.read(['(a:1,b:2,c:3);'])`. (Lists of lines work in skbio as a "file" object).

wasade:
Thanks, @ebolyen! Responding now

CHANGELOG.md
wasade: Fixed

skbio/diversity/_block.py
wasade: I can do `pop` but I'd prefer to not put this logic outside of `_block_party` as the overhead is very small relative to the cost of computing a partial distance matrix, and at least with the current implementation, just shifts the burden up to `_block_compute`

skbio/diversity/_block.py
wasade: Because we are subsetting by sample not by observation

skbio/diversity/_block.py
wasade: Note added

skbio/diversity/_block.py
wasade: There are different ways a developer may choose to do a reduction, the implementation of which may be dependent on a parallel framework. I do not expect users to override it, but I expect developers to as necessary. In practice, the overhead for reducing partial distance matrices is very small relative to the walltime expended in mapping, so at least for the sized problems I've used this on, I haven't seen a benefit in redoing this method. That being said, there is room for a parallel reduction if someone so decided to implement it.

skbio/diversity/_block.py
wasade: I had that previously and was advised to change it... I'll revert

skbio/diversity/_block.py
wasade: It is a reduction from a map reduce standpoint as well as a functional standpoint although the accumulator is fixed (i.e., it is not a general reduction method). In my opinion, it would be confusing to name this something else as it is performing a reduction. I'll add a note.

skbio/diversity/_block.py
wasade: Fixed

skbio/diversity/_block.py
wasade: Fixed

skbio/diversity/_block.py
wasade: Fixed

skbio/diversity/tests/test_block.py
wasade: The order doesn't matter but it is also deterministic, so leaving this as is

skbio/diversity/tests/test_block.py
wasade: done

wasade:
I think that covers it... apologize if I missed something, in a rush at the moment.

skbio/diversity/_block.py
wasade: Also, per the `TODO`, I'm not married to how the default reduction is done particularly as it could be represented as a reduction tree and done largely in parallel. I'll be honest, I was surprised how fast this current approach was so just never felt too motivated to redo it.

wasade:
There was a sphinx build error. I think I got it but am not positive. I also added in a see also, returns and reference to `block_beta_diversity`. 

Regarding `_block_party`, I couldn't come up with a good name for it and would love a suggestion although it sounds like that confusion is mutual. It's scope is to a) subset the counts matrix to only the samples of relevance b) if it is a phylogenetic metric, to reduce the tree to only the observations still necessary as informed by the subset of counts. We need to manipulate `kwargs` as that's the current mechanism for passing on references to the subset data structures.

wasade:
...and, thanks for the review! :)

skbio/diversity/_block.py
ebolyen: `_block_compute` wouldn't need to handle this either, based on this line:
```python
dm = reduce_f(map_f(_block_compute, _block_kwargs(**kwargs)))
```
it looks like `_block_compute` should be able to assume the kwargs it receives are valid and not think twice about them.

`_block_kwargs` also seems like the sensible place to manage valid kwargs, and that validation could happen only once instead of for each block.

skbio/diversity/_block.py
ebolyen: @jairideout thoughts? I feel like the contract isn't really all that simple without something like `n_ids` being provided. Also it is easy to compute at the scope where `_reduce` is called while it isn't trivial within. Does a hypothetical parallel tree reduce inform this API in any way?

ebolyen:
I think this is basically ready to go, just two small API things to resolve: providing n_ids to reduce, and I still think we can move validation of kwargs into `_block_kwargs`.

ebolyen:
(both of which are open questions, to clarify)

skbio/diversity/_block.py
wasade: scratch that, you're right, it could happen in `_block_kwargs`. can shift there

skbio/diversity/_block.py
wasade: A parallel tree reduction wouldn't need to be concerned about determining `n_ids`, and instead could do something like `len(set(x.ids).union(y.ids)) + 1` or something. Thinking back on this, that was the motivation to perform this step within the default `reduce` as it is specific to this method.

wasade:
Thanks, @ebolyen. I am not in favor of providing `n_ids` to reduce as it is specific to the reduction being performed. I'll try and shift the kwargs to `_block_kwargs`

ebolyen:
@wasade sounds good, this is ready for merge after that then!

ebolyen:
@wasade flake8 is sad

wasade:
Sorry! new laptop and haven't set up a linter or whitespace removal... :/

wasade:
ping

gregcaporaso:
@wasade, @ebolyen is planning to get back to this today.

wasade:
thanks!

ebolyen:
Merged, thanks @wasade, sorry for the delay on this!

wasade:
No worries, thank you so much!!

PR #1354, opened 2016-05-12
ENH: partial pairwise beta diversity

### LINKED ISSUE #1221: Support partial beta diversity calculations

jairideout:
Contains commits from #1337.

Adds `skbio.stats.distance.PairwiseDistances` for storing distances between pairs of IDs.

Adds `DistanceMatrix.from_pairwise_distances` for constructing a full distance matrix from multiple `PairwiseDistances` objects.

`skbio.diversity.beta_diversity` returns a `PairwiseDistances` object when supplying `id_pairs` keyword argument.

Fixes #1221.

ping: @wasade @gregcaporaso @ebolyen @josenavas @ElDeveloper

### DIFF
--- a/skbio/stats/distance/_base.py
+++ b/skbio/stats/distance/_base.py
@@ -20,6 +20,8 @@ from skbio.stats._misc import _pprint_strs
 from skbio.util import find_duplicates
 from skbio.util._decorator import experimental, classonlymethod
 from skbio.util._misc import resolve_key
+from ._pairwise_distances import PairwiseDistances
+from ._util import is_id_pair
 
 
 class DissimilarityMatrixError(Exception):
@@ -453,6 +455,11 @@ class DissimilarityMatrix(SkbioObject):
             self.shape[0], self.shape[1], self._matrix_element_name,
             _pprint_strs(self.ids)) + str(self.data)
 
+    @experimental(as_of="0.4.2-dev")
+    def __repr__(self):
+        """String summary of this dissimilarity matrix."""
+        return str(self)
+
     @experimental(as_of="0.4.0")
     def __eq__(self, other):
         """Compare this dissimilarity matrix to another for equality.
@@ -587,7 +594,7 @@ class DissimilarityMatrix(SkbioObject):
         """
         if isinstance(index, str):
             return self.data[self.index(index)]
-        elif self._is_id_pair(index):
+        elif is_id_pair(index):
             return self.data[self.index(index[0]), self.index(index[1])]
         else:
             return self.data.__getitem__(index)
@@ -640,11 +647,6 @@ class DissimilarityMatrix(SkbioObject):
 
     def _index_list(self, list_):
         return {id_: idx for idx, id_ in enumerate(list_)}
-
-    def _is_id_pair(self, index):
-        return (isinstance(index, tuple) and
-                len(index) == 2 and
-                all(map(lambda e: isinstance(e, str), index)))
 
 
 class DistanceMatrix(DissimilarityMatrix):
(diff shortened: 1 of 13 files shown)


skbio/stats/distance/_pairwise_distances.py
```diff
@@ -0,0 +1,446 @@
...
+
+    @experimental(as_of="0.4.2-dev")
+    def __init__(self, distances):
+        if not isinstance(distances, dict):
+            raise TypeError("`distances` must be a dict, not type %r"
+                            % type(distances).__name__)
+
+        distances_ = {}
+        ids = set()
+        for id_pair, distance in distances.items():
```
wasade: It'd be great if the validation could be disabled

skbio/stats/distance/_base.py
```diff
@@ -738,6 +740,118 @@ def from_iterable(cls, iterable, metric, key=None, keys=None):
...
+        'A', 'B', 'C'
+        Data:
+        [[ 0.   0.3  0.1]
+         [ 0.3  0.   0.2]
+         [ 0.1  0.2  0. ]]
+
+        Note that the IDs are in sorted order.
+
+        """
+        distances = {}
```
wasade: Unfortunately, row-column-value using Python `dict`s will not scale to the size of datasets that we're developing #1337 for. Or at least, a mock test using 25k samples on my laptop killed it after consuming 16GB of memory. The fundamental issue is the memory allocation strategy that Python uses on dicts, and this parallels the RCV issues we had with BIOM. Would it be possible to instead populate a numpy array or sparse arrays? eg:

```python
distances = np.zeros([])
ids_in_distances = set()
for pwdist in pairwise_distances:
    # resize disances if necessary given IDs
    for i, j in pwdist:
        distances[i, j] = pwdist[(i, j)]
```

I'm wondering though if the simpliest thing is the best thing, were we just use dense DistanceMatrices at all blocks and then just provide a `DistanceMatrix.__add__`?

skbio/stats/distance/_pairwise_distances.py
```diff
@@ -0,0 +1,446 @@
...
+                return self._distances[id_pair]
+            elif (id_pair[1], id_pair[0]) in self._distances:
+                return self._distances[id_pair[1], id_pair[0]]
+            elif id_pair[0] == id_pair[1] and id_pair[0] in self.ids:
+                return 0.0
+            else:
+                raise KeyError(id_pair)
+        else:
+            raise TypeError(
+                "`id_pair` must be a tuple of exactly two strings, not %r"
```
wasade: For the block decomposition method, there is a large benefit in using integer IDs as the IDs map directly into the counts array for slicing.

skbio/stats/distance/_pairwise_distances.py
```diff
@@ -0,0 +1,446 @@
...
+        >>> pwdist = PairwiseDistances({})
+        >>> bool(pwdist)
+        False
+
+        """
+        return len(self) > 0
+
+    @experimental(as_of="0.4.2-dev")
+    def __len__(self):
+        """Number of "within" and "between" pairwise distances.
```
ebolyen: A philosophical question: why do we care about the "within" distances at all in this object? It seems like a lot of work for some relatively counter-intuitive behaviour, especially this method and `__iter__`.

I would kind of expect a partial strategy to skip the diagonal anyhow. Wouldn't it make just as much sense for `DistanceMatrix` to complain if it actually found a "within" distance that wasn't 0, and otherwise it hallucinates them in the alternate constructer?

skbio/stats/distance/_base.py
josenavas: There is actually a huge difference. I did some tests to actually check the performance of the solution presented here and the solution proposed by @wasade .

`chocker.py`:
```python
    d = {}
    for i in range(num_samples):
        for j in range(i+1, num_samples):
            print i, j
            if (i, j) in d or (j, i) in d:
                continue
            else:
                d[(i, j)] = 1

    print d
```

`no_chocker_numpy.py`
```python
    d = np.zeros([num_samples, num_samples])
    for i in range(num_samples):
        for j in range(i+1, num_samples):
            print i, j
            if (i, j) in d or (j, i) in d:
                continue
            else:
                d[i][j] = 1.0

    print d
```

I tried with `num_samples = 10000` and `num_samples = 20000`.

| solution    | memory (GB) | time  |
|-------------|-------------|-------|
| chocker 10k | ~34GB       | 4:32  |
| numpy 10k   | ~3GB        | 8:00  |
| chocker 20k | ~137GB      | 56:00 |
| numpy 20k   | ~12GB       | 39:04 |

@wasade this will explain why your laptop was really unhappy when you tried with 25k hahaha

As we can see, as we scale up, we no only use less memory, we are also faster. In this case, storing a dense representation is more efficient than relying on python dicts.

The memory values above are the ones reported by maxresident memory in /usr/bin/time and the time is elapsed time.

wasade: That's a lot more mem for numpy than expected. Weird. Thanks though!

jairideout:
Closing in favor of #1337 :) See @wasade's comment [here](https://github.com/biocore/scikit-bio/pull/1337#issuecomment-218917740) and my comment [here](https://github.com/biocore/scikit-bio/pull/1337#issuecomment-218918074) for the next steps for partial beta diversity.

jairideout:
Since this is closed, I'm not planning to follow up on specific comments/questions here. If anyone wants an answer please ping me and I'd be happy to follow up. I'd like to limit discussion here as this is a dead-end branch and partial beta diversity discussion is already fragmented across multiple PRs and an issue.

PR #1366, opened 2016-06-02
REV: remove numpy bootstrap hack in setup.py

### LINKED ISSUE #1296: remove numpy bootstrap hack in setup.py
This hack causes #997, which doesn't have a clear/clean fix. Since we're moving forward with conda, which doesn't have this problem, I think we should remove the hack and acknowledge that when installing scikit-bio with pip/setup.py, numpy must be installed first.

Update README.rst pip install instructions to be:

```
pip install numpy
pip install scikit-bio
```

jairideout:
Fixes #1296.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,13 +17,16 @@ scikit-bio is currently in beta. We are very actively developing it, and **backw
 Installing
 ----------
 
+The recommended way to install scikit-bio is via the ``conda`` package manager available in `Anaconda <http://continuum.io/downloads>`_ or `miniconda <http://conda.pydata.org/miniconda.html>`_.
+
 To install the latest release of scikit-bio::
 
-    pip install scikit-bio
-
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -17,13 +17,16 @@ scikit-bio is currently in beta. We are very actively developing it, and **backw
 Installing
 ----------
 
+The recommended way to install scikit-bio is via the ``conda`` package manager available in `Anaconda <http://continuum.io/downloads>`_ or `miniconda <http://conda.pydata.org/miniconda.html>`_.
+
 To install the latest release of scikit-bio::
 
-    pip install scikit-bio
+    conda install -c https://conda.anaconda.org/biocore scikit-bio
```
ebolyen: This can just be `-c biocore`

ebolyen:
One comment, otherwise looks good.

README.rst
jairideout: I'm using the longer form in case users have aliased `biocore` to something else in their condarc

Issue #1322, opened 2016-03-31
CCA issues


mortonjt:
There are actually three problems going on

1. The environmental labels are being dropped.  [Here](https://github.com/biocore/scikit-bio/blob/master/skbio/stats/ordination/_canonical_correspondence_analysis.py#L224), you can see that no column or row labels are being passed into biplot_scores
2. The column rows don't actually match up with what is given in vegan.  In the example from [SO](http://stackoverflow.com/questions/36208741/how-can-i-make-a-cca-triplot-using-scikit-bio-python), there are only 3 columns (i.e. 3 dimension). This probably should be consistent with the PC ids.  
3. Its also becoming apparent that there is a sore need for biplots.  CA, RDA and CCA are essentially unusable without this functionality present in scikit-bio.  Its not trivial to tune these plots by hand.

jairideout:
@mortonjt are you available to work on these bugs (items 1 and 2) for 0.5.0? We're shooting for June 7 to cut the release.

mortonjt:
The main reason why I haven't pushed a PR for this is because I'm still not
on the same page exactly is going on with the `biplot_scores`.  If @Jorge-C
is still available, I think it would be best if he could take a look.

jairideout:
I don't think @Jorge-C is available. Is there anyone else that can maintain the CCA code in scikit-bio?

mortonjt:
Alright. Let me take another look at it

jairideout:
Thanks!

mortonjt:
I think I figured out the index/column headers for `rda` - `index=env_ids` and `columns=pc_ids`.
The same is the same for `cca`, but for some reason, some of the rows (i.e environmental variables) are being excluded.  No idea why ...

mortonjt:
OK. I found the bloody bug.

Turns out to be in the [unittest](https://github.com/biocore/scikit-bio/blob/master/skbio/stats/ordination/tests/test_canonical_correspondence_analysis.py#L73)

We are ignoring the last column, which is why debugging this was really confusing.

jairideout:
Fixed in #1368, thanks @mortonjt!

PR #1368, opened 2016-06-03
ENH: Adding sensible row and column names to biplot_scores in CCA and…

mortonjt:
… RDA

Addresses #1322


jairideout:
Thanks for the fixes! I noticed that the unit tests are calling `assert_ordination_results_equal` with `ignore_biplot_scores_labels=True`, so I don't think the labels are actually being compared to the expected labels. I can't remember why we added that option to `assert_ordination_results_equal` at the sprint last year. Can it just be removed so that biplot score labels are always compared? You might need to search the codebase for `ignore_biplot_scores_labels` to see where else it's being used, and if we can drop it.

Does this also fix item 2 in #1322?

Can you add a changelog mention? Up to you whether this is tagged as a bug fix or feature enhancement.

Thanks again!

mortonjt:
Right on.

Yeah.  Turns out that only the canonical axes are reported for this.  So
this also fixes part 2.

jairideout:
Awesome! Let me know when you're ready for me to take another pass through.

mortonjt:
Should be good to go!

jairideout:
Thanks @mortonjt!

PR #1373, opened 2016-06-07
BUG: fix performance regression by reinstating empty metadata optimizations

### LINKED ISSUE #1361: Large regression on master
In the process of getting our benchmarking pipeline up and running, I noticed there was apparently a large regression (4x-5x performance loss across about half our current benchmarks) introduced on master starting at commit [`f9016283`](https://github.com/biocore/scikit-bio/commit/f9016283638ef49ffccb3bb5f79e5a421462cfd1). asv plots can be examined [here](https://s3-us-west-2.amazonaws.com/scikit-bio.org/benchmarks/master/index.html).

jairideout:
Fixes #1361 by reinstating the empty/missing metadata optimizations that were removed in #1321.

This is not a commit revert because upstream changes have happened since #1321, causing many merge conflicts. We also want to keep some of the changes in #1321 and discard others, so a manual revert was necessary.

The unit tests testing private state to ensure performance have not been reinstated, as this didn't scale (from a development/codebase perspective) and was fragile to implementation details. Now, private state unit tests exist only for the metadata mixin classes, where that private state is defined. Instead of writing private state unit tests for performance, users and developers are encouraged to submit ASV benchmarks if a method is determined to have suboptimal performance when handling missing/empty metadata.

Also improved coverage for abstract properties/methods (in a separate commit). Use `raise NotImplementedError` instead of defining abstract properties/methods with `pass`. Based on http://stackoverflow.com/a/9212387/3776794

### DIFF
--- a/skbio/alignment/_tabular_msa.py
+++ b/skbio/alignment/_tabular_msa.py
@@ -767,9 +767,10 @@ class TabularMSA(MetadataMixin, PositionalMetadataMixin, SkbioObject):
     def __init__(self, sequences, metadata=None, positional_metadata=None,
                  minter=None, index=None):
         if isinstance(sequences, TabularMSA):
-            if metadata is None:
+            if metadata is None and sequences.has_metadata():
                 metadata = sequences.metadata
-            if positional_metadata is None:
+            if (positional_metadata is None and
+                    sequences.has_positional_metadata()):
                 positional_metadata = sequences.positional_metadata
             if minter is None and index is None:
                 index = sequences.index
@@ -800,9 +801,16 @@ class TabularMSA(MetadataMixin, PositionalMetadataMixin, SkbioObject):
         override values.
         """
         if metadata is NotImplemented:
-            metadata = self.metadata
+            if self.has_metadata():
+                metadata = self.metadata
+            else:
+                metadata = None
+
         if positional_metadata is NotImplemented:
-            positional_metadata = self.positional_metadata
+            if self.has_positional_metadata():
+                positional_metadata = self.positional_metadata
+            else:
+                positional_metadata = None
 
         if index is NotImplemented:
             if isinstance(sequences, pd.Series):
@@ -1192,7 +1200,7 @@ class TabularMSA(MetadataMixin, PositionalMetadataMixin, SkbioObject):
     def _get_position_(self, i):
         seq = Sequence.concat([s[i] for s in self._seqs], how='outer')
         # TODO: change for #1198
-        if len(self):
+        if len(self) and self.has_positional_metadata():
             seq.metadata = dict(self.positional_metadata.iloc[i])
         return seq
 
@@ -1200,7 +1208,7 @@ class TabularMSA(MetadataMixin, PositionalMetadataMixin, SkbioObject):
         seqs = self._seqs.apply(lambda seq: seq[i])
         # TODO: change for #1198
         pm = None
-        if len(self):
+        if len(self) and self.has_positional_metadata():
             pm = self.positional_metadata.iloc[i]
         return self._constructor_(seqs, positional_metadata=pm)
     # end of helpers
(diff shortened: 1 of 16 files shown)


jairideout:
autobencher [has benchmarks](https://s3-us-west-2.amazonaws.com/scikit-bio.org/benchmarks/pull_requests/jairideout/issue-1361/index.html), thanks @anderspitman! Performance is back to where it was before the regression.

The only benchmark I'm unsure about is [`benchmarks.BenchmarkSuite.time_search_for_motif_in_gapped`](https://s3-us-west-2.amazonaws.com/scikit-bio.org/benchmarks/pull_requests/jairideout/issue-1361/index.html#benchmarks.BenchmarkSuite.time_search_for_motif_in_gapped). In this PR I fixed the benchmark to use `consume_iterator`. The benchmarks show a huge spike at [f6273152](https://github.com/biocore/scikit-bio/commit/f6273152c63c1595d6b8136fba6f568fd54a0599), but that was just a doc fix on master so I can't explain why we're seeing that. The latest commits from this PR show an expected "regression" now that the benchmark is doing work. @anderspitman do you have any ideas about the performance spike I noted above?

@ebolyen this is ready for review/merge.

skbio/alignment/_tabular_msa.py
```diff
@@ -2234,8 +2244,17 @@ def join(self, other, how='strict'):
                 [self.positional_metadata, other.positional_metadata],
                 ignore_index=True, **concat_kwargs)
 
+            if not self.has_positional_metadata():
+                del self.positional_metadata
```
ebolyen: There's an edge case here I just thought of, I don't know if we care:
```python
s1 = Sequence("")
s2 = Sequence("")
setattr(s1.positional_metadata, 'foo', 42)

s1.positional_metadata.foo  # 42

s3 = s1.join(s2)

s1.positional_metadata.foo  # undefined
```

skbio/io/format/stockholm.py
```diff
@@ -695,64 +695,69 @@ def _tabular_msa_to_stockholm(obj, fh):
...
+                        raise StockholmFormatError(
+                            "Expected reference information to be stored as a "
+                            "dictionary, found reference %d stored as %r." %
+                            (ref_num, type(dictionary).__name__))
+
+                    fh.write("#=GF RN [%d]\n" % ref_num)
+                    for feature in dictionary:
+                        if feature not in _REFERENCE_TAGS:
+                            formatted_reference_tags = ', '.join(
+                                [tag for tag in _REFERENCE_TAGS])
```
ebolyen: This could be a generator comprehension (sorry I know you were just re-flowing what was there)

anderspitman:
@jairideout if you look at the plot, the last commit shown before the spike is [fd987ec7](https://github.com/biocore/scikit-bio/commit/fd987ec77b640099076af03f7a68c6b39fdad01a), which is quite out of date since things are moving fast on master. Let me bring the master benchmarks up to date and re-run and see how this looks. It's probably nothing to worry about though.

skbio/alignment/_tabular_msa.py
ebolyen: Talked offline with @jairideout, we aren't going to support this. 

 1. Users should use `metadata` for this kind of situation. There are so many ways to create properties which may not be propagated, it is the entire reason `metadata` exists.
 2. If we are monkeypatching methods, create a subclass and then post an issue about how the subclass doesn't propogate.
 3. This only happens if it's empty
 4. If after all of the above, it is still done, users get an AttributeError which is noisy and easy to detect.

ebolyen:
@jairideout This looks good to merge, let me know when we are done testing ASV on this PR.

anderspitman:
It's good to go as far as I'm concerned. I've rerun it several times but the spike is still there. I can't explain it. I agree there's nothing in that commit that should cause a regression.

jairideout:
Thanks for reviewing @ebolyen, and thanks for investigating the benchmarks @anderspitman! I think this is ready for merge.

PR #1374, opened 2016-06-07
ANCOM returns percentile abundances of each feature in each group

### LINKED ISSUE #1293: ancom should report mean/std abundance of OTUs
This information helps in interpreting results, and computing it outside of ANCOM isn't good, as it requires the user to reproduce the grouping. For example, a dataframe like this would be ideal (building on the docstring example):

screenshot 2016-02-26 14 13 47

I'll share some code that will help with this as I'm adapting ANCOM for an analysis that I'm working on. (See the "proof-of-concept" commit below.)

### LINKED ISSUE #1375: ancom reject column is misleading
Currently, "significant" values are True in the reject column, perhaps the column should be called "significant" or invert the values so False are those that shouldn't be rejected ...

gregcaporaso:
fixes #1293 
fixes #1375


gregcaporaso:
@mortonjt, could you do a first review of this one?

skbio/stats/composition.py
```diff
@@ -838,6 +884,12 @@ def ancom(table, grouping,
     if (table.isnull()).any().any():
         raise ValueError('Cannot handle missing values in `table`.')
 
+    for percentile in percentiles:
+        if not 0.0 <= percentile <= 100.0:
+            raise ValueError('Percentiles must be in the range [0, 100], %r '
+                             'was provided.' % percentile)
+
```
mortonjt: Now, what if the user gives the following percentiles

```python
percentiles = [0, 50, 10, 100]
```
Should a sort be performed, or should an error be thrown?

skbio/stats/composition.py
```diff
@@ -696,6 +697,10 @@ def ancom(table, grouping,
         classes.  This function must be able to accept at least two 1D
         array_like arguments of floats and returns a test statistic and a
         p-value. By default ``scipy.stats.f_oneway`` is used.
+    percentiles : list of floats, optional
+        Percentile abundances to return for each feature in each group. By
+        default, will return the minimum, 25th percentile, median, 75th
+        percentile, and maximum abundances for each feature in each group.
```
mortonjt: May also want to specify that the ordering matters

skbio/stats/composition.py
```diff
@@ -896,9 +948,27 @@ def ancom(table, grouping,
...
-    return pd.DataFrame({'W': pd.Series(W, index=labs),
-                         'reject': pd.Series(reject, index=labs)})
+
+    feat_ids = mat.columns
+    ancom_df = pd.DataFrame({'W': pd.Series(W, index=feat_ids),
+                             'reject': pd.Series(reject, index=feat_ids)})
+
+    if len(percentiles) == 0:
+        return ancom_df, pd.DataFrame()
+    else:
```
mortonjt: What do you think about separating this helper function into its own private function?  I could see this functionality becoming useful for other tests (i.e. anosim, permanova).

skbio/stats/tests/test_composition.py
```diff
@@ -550,7 +550,121 @@ def test_ancom_basic_counts(self):
...
+        self.assertAlmostEqual(result[(0.0, 'a')]['b2'], 11.0)
+        self.assertAlmostEqual(result[(0.0, 'b')]['b2'], 1.0)
+        self.assertAlmostEqual(result[(25.0, 'a')]['b2'], 11.0)
+        self.assertAlmostEqual(result[(25.0, 'b')]['b2'], 27.0)
+        self.assertAlmostEqual(result[(50.0, 'a')]['b2'], 11.0)
+        self.assertAlmostEqual(result[(50.0, 'b')]['b2'], 53.0)
+        self.assertAlmostEqual(result[(75.0, 'a')]['b2'], 11.0)
+        self.assertAlmostEqual(result[(75.0, 'b')]['b2'], 76.5)
+        self.assertAlmostEqual(result[(100.0, 'a')]['b2'], 11.0)
+        self.assertAlmostEqual(result[(100.0, 'b')]['b2'], 100.0)
```
mortonjt: Since we are on a tight deadline here, this is more of a future note.

We probably should be using assert_dataframe_equals here.  I'm not sure if it can handle MultiIndex objects at the moment.

skbio/stats/composition.py
```diff
@@ -838,6 +884,12 @@ def ancom(table, grouping,
     if (table.isnull()).any().any():
         raise ValueError('Cannot handle missing values in `table`.')
 
+    for percentile in percentiles:
```
mortonjt: Worthwhile to type check to see if percentiles is a `list`?

skbio/stats/composition.py
mortonjt: While we are on the subject, how should we handle duplicate percentiles?

mortonjt:
First pass through - looking solid.  I don't see any problems once these comments are addressed.

skbio/stats/composition.py
gregcaporaso: I think both are fine - the values are clearly defined in the resulting dataframe, so neither case would give the user incorrect information.

skbio/stats/composition.py
gregcaporaso: It doesn't, unless I'm missing something. Can you clarify?

skbio/stats/composition.py
gregcaporaso: It actually only needs to be an iterable. I'll update that in the docs.

skbio/stats/tests/test_composition.py
gregcaporaso: I was running into issues with building that dataframe and getting it compare equal, but I agree that that's preferable. I'll try again for a few minutes.

skbio/stats/composition.py
gregcaporaso: Let me think about this one a little more this morning. I agree that this would be ideal (and it's the plan for the future as we discussed in #1293), but it's really important that we don't duplicate any logic for splitting samples into groups that could get out of sync. In the future, the splitting would be a operation performed by the feature/contingency table object that is provided to this function.

gregcaporaso:
Thanks for the review @mortonjt!

skbio/stats/composition.py
gregcaporaso: I confirmed that order doesn't matter with a new unit test.

skbio/stats/composition.py
gregcaporaso: You're right that duplicates aren't handled well in the resulting data frame, so those are explicitly disallowed (and new test added). Thanks for catching that!

skbio/stats/composition.py
gregcaporaso: After reviewing the code, I think we should leave this as-is for now rather than duplicate the creation of the ``mat`` variable. We'll factor that aspect correctly where we adapt this API for the new contingency/feature table object. Sound ok?

skbio/stats/composition.py
```diff
@@ -644,14 +646,15 @@ def ancom(table, grouping,
           tau=0.02,
           theta=0.1,
           multiple_comparisons_correction=None,
-          significance_test=None):
+          significance_test=None,
+          percentiles=[0.0, 25.0, 50.0, 75.0, 100.0]):
```
jairideout: lists are mutable and shouldn't be used as default values in Python. Can you change this to `None` or a tuple?

skbio/stats/composition.py
```diff
@@ -708,6 +715,10 @@ def ancom(table, grouping,
 
         `"reject"` indicates if feature is significantly different or not.
 
+    pd.DataFrame
```
jairideout: Does this render correctly? I'm unsure about multiple returns formatted this way.

skbio/stats/composition.py
```diff
@@ -696,6 +699,10 @@ def ancom(table, grouping,
         classes.  This function must be able to accept at least two 1D
         array_like arguments of floats and returns a test statistic and a
         p-value. By default ``scipy.stats.f_oneway`` is used.
+    percentiles : list of floats, optional
```
jairideout: iterable of floats

skbio/stats/composition.py
```diff
@@ -708,6 +715,10 @@ def ancom(table, grouping,
 
         `"reject"` indicates if feature is significantly different or not.
 
+    pd.DataFrame
+        A table of features and their percentile abundances in each group. If
+        ``percentiles = []``, this will be an empty data frame.
```
jairideout: `percentiles` can be an iterable, I'd change the second sentence to:

"""
If `percentiles` is empty, this will be an empty data frame.
"""

skbio/stats/composition.py
```diff
@@ -838,6 +886,18 @@ def ancom(table, grouping,
     if (table.isnull()).any().any():
         raise ValueError('Cannot handle missing values in `table`.')
 
+    for percentile in percentiles:
+        if not 0.0 <= percentile <= 100.0:
+            raise ValueError('Percentiles must be in the range [0, 100], %r '
+                             'was provided.' % percentile)
+
+    if len(percentiles) != len(set(percentiles)):
```
jairideout: Use `skbio.util.find_duplicates`, check out how it's used [here](https://github.com/biocore/scikit-bio/blob/master/skbio/stats/distance/_base.py#L629). Note that the code is taking multiple passes over `percentiles` at this point, so will fail on an `iterable`.

skbio/stats/tests/test_composition.py
jairideout: `skbio.util.assert_data_frame_almost_equal` should handle `MultiIndex` just fine, we're using it in `TabularMSA` tests with `MultiIndex`. @gregcaporaso let me know if you're having trouble getting it to work and I can take a look.

skbio/stats/composition.py
jairideout: Can you also describe here what the overall structure of the data frame is? I can't tell from the doctest examples below because only pieces of the data frame are displayed. Is it a data frame with features as index and MultiIndex on columns, where groups are nested under percentiles?

skbio/stats/tests/test_composition.py
```diff
@@ -550,7 +550,131 @@ def test_ancom_basic_counts(self):
...
+                              [20],
+                              [23]],
+                             index=['s1', 's2', 's3', 's4', 's5', 's6'],
+                             columns=['b1'])
+        grouping = pd.Series(['a', 'a', 'a', 'b', 'b', 'b'],
+                             index=['s1', 's2', 's3', 's4', 's5', 's6'])
+        result = ancom(table, grouping, percentiles=[])[1]
+        self.assertEqual(result.shape, (0, 0))
+
+    def test_ancom_invalid_percentiles(self):
```
jairideout: I know this is an annoying request, but can you please split this test case into several smaller test cases that each test one thing? Arrange, Act, Assert is the strategy we're using for new tests.

skbio/stats/composition.py
mortonjt: I think that is fine.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 
 ### Backward-incompatible changes [experimental]
 * `TabularMSA.append` and `TabularMSA.extend` now require one of `minter`, `index`, or `reset_index` to be provided when incorporating new sequences into an MSA. Previous behavior was to auto-increment the index labels if `minter` and `index` weren't provided and the MSA had a default integer index, otherwise error. Use `reset_index=True` to obtain the previous behavior in a more explicit way.
+* `skbio.stats.ancom` now returns two `pd.DataFrame` objects, where it previously returned one. The first contains the ANCOM test results, as before, and the second contains percentile abundances of each feature in each group. The specific percentiles that are computed and returned is controlled by the new `percentiles` parameter to `skbio.stats.ancom`. In the future, this second `pd.DataFrame` will not be returned by this function, but will be available through the [contingency table API](https://github.com/biocore/scikit-bio/issues/848). ([#1293](https://github.com/biocore/scikit-bio/issues/1293))
```
jairideout: `skbio.stats.ancom` -> `skbio.stats.composition.ancom`

jairideout:
@gregcaporaso done reviewing. Don't worry about the ASV benchmark failure, @anderspitman is looking into it, this code shouldn't have affected any of the existing benchmarks.

skbio/stats/composition.py
gregcaporaso: This seems to render ok: 

screenshot 2016-06-08 13 14 38

jairideout: Sweet, thanks for checking!

skbio/stats/composition.py
gregcaporaso: Done. Fixed the latter, and added a test. Thanks for catching that.

skbio/stats/tests/test_composition.py
gregcaporaso: Got it sorted out.

gregcaporaso:
This should address all of your comments @mortonjt and @jairideout. I also addressed #1375 in this latest push. Thanks for the review.

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@
 
 ### Backward-incompatible changes [experimental]
 * `TabularMSA.append` and `TabularMSA.extend` now require one of `minter`, `index`, or `reset_index` to be provided when incorporating new sequences into an MSA. Previous behavior was to auto-increment the index labels if `minter` and `index` weren't provided and the MSA had a default integer index, otherwise error. Use `reset_index=True` to obtain the previous behavior in a more explicit way.
+* `skbio.stats.composition.ancom` now returns two `pd.DataFrame` objects, where it previously returned one. The first contains the ANCOM test results, as before, and the second contains percentile abundances of each feature in each group. The specific percentiles that are computed and returned is controlled by the new `percentiles` parameter to `skbio.stats.composition.ancom`. In the future, this second `pd.DataFrame` will not be returned by this function, but will be available through the [contingency table API](https://github.com/biocore/scikit-bio/issues/848). ([#1293](https://github.com/biocore/scikit-bio/issues/1293))
```
jairideout: Can you also note the "reject" -> "Reject null hypothesis" column name change in the first data frame that's returned?

skbio/stats/composition.py
```diff
@@ -708,6 +714,13 @@ def ancom(table, grouping,
 
         `"reject"` indicates if feature is significantly different or not.
```
jairideout: Update these docs to use new column name

skbio/stats/composition.py
```diff
@@ -751,33 +766,37 @@ def ancom(table, grouping,
...
-    Then create a grouping vector.  In this scenario, there
-    are only two classes, and suppose these classes correspond to the
-    treatment due to a drug and a control.  The first three samples
-    are controls and the last three samples are treatments.
+    Then create a grouping vector. In this example, there is a treatment group
+    and a placebo group.
 
-    >>> grouping = pd.Series([0, 0, 0, 1, 1, 1],
-    ...                      index=['s1','s2','s3','s4','s5','s6'])
+    >>> grouping = pd.Series(['treatment', 'treatment', 'treatment',
```
jairideout: 👍

skbio/stats/composition.py
```diff
@@ -751,33 +766,37 @@ def ancom(table, grouping,
...
+    feature in each group.
 
-    >>> results = ancom(table, grouping)
-    >>> results['W']
+    >>> ancom_df, percentile_df = ancom(table, grouping)
+    >>> ancom_df['W']
     b1    0
     b2    4
-    b3    1
+    b3    0
```
jairideout: Why did this result change? From what I can tell you're using equivalent input, just different labels.

gregcaporaso: The test became more conservative now that we're correcting for multiple comparisons by default.

CHANGELOG.md
gregcaporaso: Done, see below.

skbio/stats/composition.py
```diff
@@ -896,9 +961,27 @@ def ancom(table, grouping,
...
+        {'W': pd.Series(W, index=feat_ids),
+         'Reject null hypothesis': pd.Series(reject, index=feat_ids)})
+
+    if len(percentiles) == 0:
+        return ancom_df, pd.DataFrame()
+    else:
+        data = []
+        columns = []
+        for group in groups:
+            feat_dists = mat[grouping == group]
```
jairideout: This should be `mat[cats == group]` because `cats` is index-aligned to `mat` and `grouping` isn't. I don't think there's a bug here because pandas will index-align the boolean `Series` with `mat` before slicing, but we get a warning in that case:

```python
UserWarning: Boolean Series key will be reindexed to match DataFrame index.
```

Using `cats` should avoid this warning and an unnecessary alignment step.

CHANGELOG.md
mortonjt: Hold that thought.

skbio/stats/composition.py
```diff
@@ -706,7 +712,15 @@ def ancom(table, grouping,
         `"W"` is the W-statistic, or number of features that a single feature
         is tested to be significantly different against.
 
-        `"reject"` indicates if feature is significantly different or not.
+        `"Reject null hypothesis"` indicates if feature is differentially
+        abundance across groups (`True`) or not (`False`).
```
jairideout: abundance -> abundant

skbio/stats/composition.py
jairideout: Makes sense, thanks!

skbio/stats/tests/test_composition.py
```diff
@@ -546,12 +546,197 @@ def test_ancom_basic_counts(self):
...
-                                                False, False, False],
-                                               dtype=bool)})
+        exp = pd.DataFrame(
+            {'W': np.array([5, 5, 2, 2, 2, 2, 2]),
+             'Reject null hypothesis': np.array([True, True, False, False,
+                                                 False, False, False],
+                                                dtype=bool)})
+        assert_data_frame_almost_equal(result[0], exp)
+
+    def test_ancom_basic_counts_shuffled_order(self):
```
jairideout: This test basically exists as `test_ancom_basic_counts_swapped`. Instead of testing the first data frame (which `test_ancom_basic_counts_swapped` is testing), can you change this test to assert the second data frame is correct?

jairideout: Actually, it'd be better to just update `test_ancom_basic_counts_swapped` to also assert the second data frame is correct and remove this test

skbio/stats/tests/test_composition.py
```diff
@@ -546,12 +546,197 @@ def test_ancom_basic_counts(self):
...
+        tuples = [(p, g) for g in groups for p in percentiles]
+        exp_mi = pd.MultiIndex.from_tuples(tuples,
+                                           names=['Percentile', 'Group'])
+        exp_data = np.array(
+            [[1.0, 11.0], [5.0, 11.0], [9.0, 11.0], [10.5, 11.0], [12.0, 11.0],
+             [20.0, 1.0], [21.0, 27.0], [22.0, 53.0], [22.5, 76.5],
+             [23.0, 100.0]])
+        exp = pd.DataFrame(exp_data.T, columns=exp_mi, index=['b1', 'b2'])
+
+        result = ancom(table, grouping, percentiles=percentiles)[1]
```
jairideout: Can you have this test use the default percentiles (i.e. don't provide percentiles)? There currently isn't a test for that.

skbio/stats/composition.py
jairideout: I was playing around with tests locally and wasn't receiving the `UserWarning` as I expected (we have a test case that should hit it). The reason is that `table` is aligned to whatever order `grouping` is in, so `cats` is just a copy of `grouping`. Thus, using `grouping` vs. `cats` here is equivalent and we don't hit the pandas codepath I suspected.

`DataFrame.align` doesn't document how the index alignment will be performed. It appears to align the data frame (i.e. the object the `align` method is called on) to the input series, returning the aligned data frame and a copy of the input series. Since this is undocumented behavior and might change across pandas versions or in the future, it's safer to use `cats` here, even though `cats` and `grouping` are currently equivalent.

jairideout:
@gregcaporaso done with second pass

skbio/stats/composition.py
mortonjt: +1 I think this is clear.

CHANGELOG.md
mortonjt: 👍

skbio/stats/tests/test_composition.py
gregcaporaso: If I understand correctly, that'd be a big data frame to construct for the test (7 samples * 2 groups * 5 percentiles = 70 values). I'll do this in a new test that will result in a smaller data frame.

gregcaporaso:
@jairideout, pushed up changes. I think these address your comments (though see my comment about the "swapped" tested - I'm not sure I understand what you're asking for).

jairideout:
Looks great, thanks @gregcaporaso!

PR #1378, opened 2016-06-08
Adds more descriptive error message to registry

### LINKED ISSUE #1326: skbio.io.read: improve error message when format has no generator reader
When reading a file using the procedural `skbio.io.read` interface, the error message could be improved when `into` isn't provided and there isn't a generator reader. For example:

```python
import skbio.io
msa = skbio.io.read('my-stockholm-file', format='stockholm')
...
UnrecognizedFormatError: Cannot read 'stockholm' from <_io.BufferedReader name='my-stockholm-file'>, no generator reader found.
```

It'd be helpful to provide a hint about passing `into`, including valid classes for the provided `format`. In this example (Stockholm), `TabularMSA` is the only class the file can be read into.

Kleptobismol:
Fixes #1326

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **IMPORTANT**: scikit-bio is no longer compatible with Python 2. scikit-bio is compatible with Python 3.4 and later.
 
 ### Features
+* Added more descriptive error message to `skbio.io.registry` when attempting to read without specifying `into` and when there is no generator reader. ([#1326](https://github.com/biocore/scikit-bio/issues/1326))
 * Added support for reference tags to `skbio.io.format.stockholm` reader and writer. ([#1348](https://github.com/biocore/scikit-bio/issues/1348))
 * Expanded error message in `skbio.io.format.stockholm` reader when `constructor` is not passed, in order to provide better explanation to user. ([#1327](https://github.com/biocore/scikit-bio/issues/1327))
 * Added `skbio.sequence.distance.kmer_distance` for computing the kmer distance between two sequences. ([#913](https://github.com/biocore/scikit-bio/issues/913))
--- a/skbio/io/registry.py
+++ b/skbio/io/registry.py
@@ -558,10 +558,24 @@ class IORegistry:
 
         reader = self.get_reader(fmt, into)
         if reader is None:
+            possible_intos = [r.__name__ for r in
+                              self._get_possible_readers(fmt)]
+            message = ''
+            if possible_intos:
+                message = ("Possible values for `into` include: %s"
+                           % ', '.join(possible_intos))
+            into_message = '`into` also not provided.' if not into else ''
             raise UnrecognizedFormatError(
-                "Cannot read %r from %r, no %s reader found." %
-                (fmt, file, into.__name__ if into else 'generator'))
+                "Cannot read %r from %r, no %s reader found. %s %s" %
+                (fmt, file, into.__name__ if into else 'generator',
+                 into_message, message))
         return reader, kwargs
+
+    def _get_possible_readers(self, fmt):
+        for lookup in self._lookups:
+            if fmt in lookup:
+                return list(lookup[fmt].readers)
+        return []
 
     @stable(as_of="0.4.0")
     def write(self, obj, format, into, **kwargs):
--- a/skbio/io/tests/test_registry.py
+++ b/skbio/io/tests/test_registry.py
@@ -872,6 +872,21 @@ class TestRead(RegistryTest):
         self.assertEqual(TestClass([1, 2, 3, 4]), instance)
         fh.close()
 
+    def test_into_is_none_and_no_generator_reader(self):
+        format1 = self.registry.create_format('format1')
+
+        fh = StringIO('1\n2\n3\n4')
+
+        @format1.reader(TestClass)
+        def reader(fh):
+            self.assertIsInstance(fh, io.TextIOBase)
+            return
+
+        with self.assertRaisesRegex(UnrecognizedFormatError,
+                                    "Cannot read 'format1'.*Possible.*include"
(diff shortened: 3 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 **IMPORTANT**: scikit-bio is no longer compatible with Python 2. scikit-bio is compatible with Python 3.4 and later.
 
 ### Features
+* Added more descriptive error message to `skbio.io.registry` when attempting to read without specifying `into` and when there is
+no generator reader. ([#1326](https://github.com/biocore/scikit-bio/issues/1326))
```
ebolyen: single line

skbio/io/registry.py
```diff
@@ -558,11 +558,28 @@ def _init_reader(self, file, fmt, into, verify, kwargs, io_kwargs):
...
+                " not found. %s" %
+                (fmt, file, into.__name__ if into else 'generator',
+                 message))
         return reader, kwargs
 
+    def _get_possible_readers(self, fmt):
+        for lookup in self._lookups:
+            for format in lookup.values():
+                if format.name == fmt:
+                    return list(format.readers.keys())
```
ebolyen: This can be simplified to 
```python
for lookup in self._lookups:
    if fmt in lookupt:
        return list(lookup[fmt].readers)
```

skbio/io/registry.py
```diff
@@ -558,11 +558,28 @@ def _init_reader(self, file, fmt, into, verify, kwargs, io_kwargs):
 
         reader = self.get_reader(fmt, into)
         if reader is None:
+            possible_readers = self._get_possible_readers(fmt)
+            if possible_readers:
+                possible_intos = []
```
ebolyen:
```python
possible_intos = [r.__name__ for r in self._get_possible_readers(fmt)]
if possible_intos:
    <set message>
else:
    message = ''
```

skbio/io/registry.py
```diff
@@ -558,11 +558,28 @@ def _init_reader(self, file, fmt, into, verify, kwargs, io_kwargs):
...
+                for reader in possible_readers:
+                        possible_intos.append(reader.__name__)
+                message = """Possible values for `into` include: %s
+                          """ % ', '.join(possible_intos)
+            else:
+                message = ''
```
ebolyen: Add a test to cover this branch of the code

skbio/io/registry.py
```diff
@@ -558,11 +558,28 @@ def _init_reader(self, file, fmt, into, verify, kwargs, io_kwargs):
...
             raise UnrecognizedFormatError(
-                "Cannot read %r from %r, no %s reader found." %
-                (fmt, file, into.__name__ if into else 'generator'))
+                "Cannot read %r from %r, no %s reader found. `into` also"
+                " not found. %s" %
+                (fmt, file, into.__name__ if into else 'generator',
```
ebolyen: This ternary suggests that sometimes `into` is set, so the error will be incorrect sometimes.

ebolyen:
@Kleptobismol sorry for the delayed review, done now. Have some comments that need to be addressed.

skbio/io/registry.py
Kleptobismol: fixed

skbio/io/registry.py
Kleptobismol: fixed

skbio/io/registry.py
Kleptobismol: fixed

Kleptobismol:
@ebolyen addressed your comments, except for the test one, as I'm not sure what you meant. I ran coverage and all branches of the new code are being tested. Also, I had to tweak one of your changes a bit.

skbio/io/registry.py
```diff
@@ -558,11 +558,25 @@ def _init_reader(self, file, fmt, into, verify, kwargs, io_kwargs):
 
         reader = self.get_reader(fmt, into)
         if reader is None:
+            readers = self._get_possible_readers(fmt)
+            possible_intos = [r.__name__ for r in readers] if readers else None
+            if possible_intos:
```
ebolyen: Oh I see, let's make it so that `_get_possible_readers` always returns a list (empty if there isn't anything). That way we can just chain it with the list comprehension. Then `possible_intos` will be empty (false) or it will have the available class names.

skbio/io/registry.py
```diff
@@ -558,11 +558,25 @@ def _init_reader(self, file, fmt, into, verify, kwargs, io_kwargs):
...
+                message = """Possible values for `into` include: %s
+                          """ % ', '.join(possible_intos)
```
ebolyen: Not a huge fan of using triple-quotes here. Could you change to 
```python
message = ("possible values for `into include: %s" 
           % ', '.join(possible_intos))
```
That would match most of our code-base a little better.

skbio/io/registry.py
```diff
@@ -558,11 +558,25 @@ def _init_reader(self, file, fmt, into, verify, kwargs, io_kwargs):
...
+            else:
+                message = ''
+            into_message = '`into` also not found.' if not into else ''
```
ebolyen: found -> provided 

might be a better wording

ebolyen:
RE: coverage, I think that makes sense.

skbio/io/registry.py
Kleptobismol: fixed

skbio/io/registry.py
Kleptobismol: fixed

skbio/io/registry.py
Kleptobismol: fixed

ebolyen:
Looks good @Kleptobismol! Just squash your commits and it should be good for merge.

PR #1381, opened 2016-06-10
MAINT: TabularMSA.conservation: better use of skbio API, fix a couple of typos

jairideout:


ebolyen:
Nice catches! @jairideout, much better use of the API.

Issue #1393, opened 2016-06-23
ILR misbehaving with one dimension vectors


ltoran:
Hello,

I'm trying to use the `ilr` module under `skbio.stats.composition`.
I'm having a problem with one dimensional basis.

Here's my code
```python
from skbio.stats.composition import ilr
import numpy as np
table = np.array([[  1.        ,  10.        ],
                  [  1.14141414,   9.90909091],
                  [  1.28282828,   9.81818182],
                  [  1.42424242,   9.72727273],
                  [  1.56565657,   9.63636364]])
basis = np.array([[ 0.80442968,  0.19557032]])
ilr_coords = ilr(table, basis=basis)
```

It seems to work when I cast the basis to 2 dimensions.  Any thoughts on this?

jsevo:
I am having difficulty understanding what the analysis of 2D data is
supposed to yield. What's your use case for the ILR here?

mortonjt:
This problem is popping up when there is a single basis vector (its actually a typo)

```python
basis = np.array([ 0.80442968,  0.19557032])
```
There is a problem with the `_check_orthogonality` function.
Its an easy fix, putting in a PR right now.

jairideout:
Fixed in #1396.

PR #1396, opened 2016-06-24
ENH: Addressing #1393

mortonjt:
Fixing _check_orthogonality
Addressing #1393

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Backward-incompatible changes [stable]
 
 ### Backward-incompatible changes [experimental]
+* Modifying basis handling in `skbio.stats.composition.ilr_inv` prior to checking for orthogonality.  Now the basis is strictly assumed to be in the Aitchison simplex. 
 
 ### Bug fixes
 
--- a/skbio/stats/composition.py
+++ b/skbio/stats/composition.py
@@ -538,7 +538,10 @@ def ilr(mat, basis=None, check=True):
 
     basis: numpy.ndarray, float, optional
         orthonormal basis for Aitchison simplex
-        defaults to J.J.Egozcue orthonormal basis
+        defaults to J.J.Egozcue orthonormal basis.
+
+    check: bool
+        Specifies if the basis is orthonormal.
 
     Examples
     --------
@@ -548,12 +551,24 @@ def ilr(mat, basis=None, check=True):
     >>> ilr(x)
     array([-0.7768362 , -0.68339802,  0.11704769])
 
+    Notes
+    -----
+    If the `basis` parameter is specified, it is expected to be a basis in the
+    Aitchison simplex.  If there are `D-1` elements specified in `mat`, then
+    the dimensions of the basis needs be `D-1 x D`, where rows represent
+    basis vectors, and the columns represent proportions.
     """
     mat = closure(mat)
     if basis is None:
         basis = clr_inv(_gram_schmidt_basis(mat.shape[-1]))
-    elif check:
-        _check_orthogonality(basis)
+    else:
+        if len(basis.shape) != 2:
+            raise ValueError("Basis needs to be a 2D matrix, "
+                             "not a %dD matrix." %
+                             (len(basis.shape)))
+        if check:
+            _check_orthogonality(basis)
+
     return inner(mat, basis)
 
 
@@ -575,7 +590,7 @@ def ilr_inv(mat, basis=None, check=True):
 
     where :math:`[e_1,\ldots, e_{D-1}]` is an orthonormal basis in the simplex.
 
-    If an orthornormal basis isn't specified, the J. J. Egozcue orthonormal
+    If an orthonormal basis isn't specified, the J. J. Egozcue orthonormal
     basis derived from Gram-Schmidt orthogonalization will be used by
     default.
 
@@ -591,6 +606,10 @@ def ilr_inv(mat, basis=None, check=True):
         orthonormal basis for Aitchison simplex
         defaults to J.J.Egozcue orthonormal basis
 
+    check: bool
+        Specifies if the basis is orthonormal.
+
+
     Examples
     --------
     >>> import numpy as np
@@ -599,12 +618,27 @@ def ilr_inv(mat, basis=None, check=True):
     >>> ilr_inv(x)
     array([ 0.34180297,  0.29672718,  0.22054469,  0.14092516])
 
+    Notes
+    -----
+    If the `basis` parameter is specified, it is expected to be a basis in the
+    Aitchison simplex.  If there are `D-1` elements specified in `mat`, then
+    the dimensions of the basis needs be `D-1 x D`, where rows represent
+    basis vectors, and the columns represent proportions.
     """
 
     if basis is None:
         basis = _gram_schmidt_basis(mat.shape[-1] + 1)
-    elif check:
-        _check_orthogonality(basis)
+    else:
+        if len(basis.shape) != 2:
+            raise ValueError("Basis needs to be a 2D matrix, "
+                             "not a %dD matrix." %
+                             (len(basis.shape)))
+        if check:
+            _check_orthogonality(basis)
+        # this is necessary, since the clr function
+        # performs np.squeeze()
+        basis = np.atleast_2d(clr(basis))
+
     return clr_inv(np.dot(mat, basis))
 
 
@@ -1081,6 +1115,7 @@ def _check_orthogonality(basis):
     basis: numpy.ndarray
         basis in the Aitchison simplex
     """
+    basis = np.atleast_2d(basis)
     if not np.allclose(inner(basis, basis), np.identity(len(basis)),
                        rtol=1e-4, atol=1e-6):
         raise ValueError("Aitchison basis is not orthonormal")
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -9,7 +9,7 @@
 ### Backward-incompatible changes [experimental]
 
 ### Bug fixes
-
+* Improving useability for `ilr` and `ilr_inv` for 2D compositions.  ([#1393](https://github.com/biocore/scikit-bio/issues/1393))
```
jairideout: Can you expand the description of the bug? From this description, I can't tell what input(s) caused the bug and what the buggy behavior was (was it a silent bug producing incorrect results, or did it error noisily?). It appears the `basis` parameter didn't work with 1-D arrays for `ilr` and `ilr_inv`, but I'm unsure about the `clr_inv` transformation to `basis` that you added to `ilr_inv`. If there's more than one issue with `basis`, please describe that here.

The markdown heading below this section doesn't render correctly, please add a blank line after this bullet point.

Please also indicate the full public import path for these functions, i.e. `skbio.stats.composition.ilr` and `skbio.stats.composition.ilr_inv`.

jairideout:
Thanks for the fix @mortonjt! I just had one comment related to informing users about the bug (changelog).

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Backward-incompatible changes [experimental]
 
 ### Bug fixes
+* Modified `skbio.stats.composition.ilr` and `skbio.stats.composition.ilr_inv` to suppress errors when inputing a 1D basis.  ([#1393](https://github.com/biocore/scikit-bio/issues/1393))
```
jairideout: Thanks for expanding, this looks great! Just to confirm, what is the rationale for changing `ilr_inv` handling of `basis` from:

```python
_check_orthogonality(basis)
```

to:

```python
_check_orthogonality(clr_inv(basis))
```
It seems unrelated to handling 1-D basis. Is this a separate bug? If so please add that to the changelog.

mortonjt: Sorry about that.  Yeah, that was a separate bug.  Added.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Backward-incompatible changes [experimental]
 
 ### Bug fixes
+* Modified `skbio.stats.composition.ilr` and `skbio.stats.composition.ilr_inv` to suppress errors when inputing a 1D basis.  ([#1393](https://github.com/biocore/scikit-bio/issues/1393)).  Also fixing how the basis is handled in `skbio.stats.composition.ilr_inv`.
```
jairideout: Please update the second bug description to include what the previous behavior was (was it a silent bug that produced incorrect results, did it fail noisily, or something else?) and what the new behavior is.

In general, when a bug fix is submitted it is essential that the changelog entry includes these details as this is how we notify our users. This is analogous to QIIME 1's blog posts describing bugs (they include the same information for users).

jairideout: Also, can you squash your commits when that's addressed?

mortonjt: Alrighty.  Here's a run down of what happened.

- `_check_orthogonality` only accepts a basis that is within the simplex.  The basis passed into `ilr_inv` lives only in `ilr` space, so in order to make sure that the basis is correct (i.e. orthogonal), it needs to be transformed into the simplex space to match `_check_orthogonality`
- So basically, you need to be consistent between these two spaces (`ilr` space, and simplex space).  And I just fixed this mistake to enforce this consistency.  I didn't catch this before, because we didn't have a test case for passing in a basis for `ilr_inv`.

We now have that fix.  Let me know if there is anything else that I need to clarify.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Backward-incompatible changes [experimental]
 
 ### Bug fixes
+* Modified `skbio.stats.composition.ilr` and `skbio.stats.composition.ilr_inv` to suppress errors when inputing a 1D basis.  ([#1393](https://github.com/biocore/scikit-bio/issues/1393)).  Also fixing how the basis is handled in `skbio.stats.composition.ilr_inv`, the basis is now transformed back into the simplex space before checking to see if the basis is orthogonal.
```
ebolyen: Instead of "suppress errors" how about "coerces rank to 2D" or something similar?

`Also ...` should be a second bullet with past tense. Examples would be good.

mortonjt:
I don't think this is a good use of time to be spending on perfecting a 2 sentence bug report.

If there is still misunderstanding on exactly what this functionality is doing, let's schedule a call after Scipy.  Otherwise, I recommend merging this fix as is.

mortonjt:
Had a discussion offline with @ebolyen about the functionality that has been modified.

Here's exactly what is going on with the `ilr_inv` method.

There is no bug in the existing interface.  But there is some potential confusion for user concerning the input basis for `ilr` and `ilr_inv`.   The `ilr` method takes in a `basis` that is `D-1 x D` matrix (i.e. proportions by balances).  The `ilr_inv` method takes in a `basis` that is `D x D-1` matrix (i.e. balances by proportions).  This can be confusing for users, because this would require that the basis needs to be also transformed when applying `ilr` and `ilr_inv`.

In other words, if you wanted to run the following code
```
>>> from skbio.stats.composition import ilr, ilr_inv, clr
>>> import numpy as np
>>> basis = np.array([[0.80442968, 0.19557032]])
>>> table = np.array([[np.log(1/10)*np.sqrt(1/2),
...                            np.log(1.14141414 / 9.90909091)*np.sqrt(1/2),
...                            np.log(1.28282828 / 9.81818182)*np.sqrt(1/2),
...                            np.log(1.42424242 / 9.72727273)*np.sqrt(1/2),
...                            np.log(1.56565657 / 9.63636364)*np.sqrt(1/2)]]).T
>>> res = ilr(ilr_inv(table, basis=basis), basis=basis)  
>>> res
array([-0.70097524, -0.65792874, -0.61956576, -0.58489877, -0.55322134])
>>> table
array([-1.62817353, -1.52818832, -1.4390816 , -1.35855968, -1.2849817 ])      
```

`res` and `table` are different, even though they should be the same, since `ilr` and `ilr_inv` are inverses of each other.

In this PR, I have made this easier to run.  More specifically, I am making sure that `ilr` and `ilr_inv` can take in the exact same basis.  And that basis lives in the Aitchison simplex.  

In other words. The functions have the following signatures

```
ilr (x, e) :: (P, P) -> B
ilr_inv(y, e) :: (B, P) -> P
```

Where `x` is a `n x D` matrix of proportions (`n` samples and `D` proportions),
`y` is a `n x D-1` matrix of balances (`n` samples and `D-1` balances),
`e` is a `D-1 x D` matrix representing a basis of proportions (`D-1` vectors of `D` proportions),
`P` represents the space of proportions (i.e. Aitchison simplex), and `B` represents the space of balances (i.e. ilr space).  

An additional test under`test_ilr_basis_isomorphism` explicitly ensures that `ilr` and `ilr_inv` are inverses of each other if the same basis is applied to both functions.

Hopefully this makes this function a little more clear, and easier to use.  Let me know if you have any more questions about this. Thanks!

mortonjt:
Also note that we won't be worrying about the issue brought up in #1393.  This method should not take in 1D matrices, since the `basis` parameter should be a 2D matrix.  A ValueError is now raised to catch that.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Backward-incompatible changes [experimental]
 
 ### Bug fixes
+* Modifying basis handling in `skbio.stats.composition.ilr_inv` prior to checking for orthogonality.  Now the basis is strictly assumed to be in the Aitchison simplex. ([#1393](https://github.com/biocore/scikit-bio/issues/1393))
```
ebolyen: Won't have a chance to really review until the morning, but real quick comment:
This should be in the section just above (e.g. in `Backward-incompatible changes [experimental]`) as code that was already accounting for the inverse basis will need to be adjusted.
Thanks!

ebolyen:
Thanks for the explanation @mortonjt!

skbio/stats/composition.py
```diff
@@ -548,6 +551,12 @@ def ilr(mat, basis=None, check=True):
...
 
+    Notes
+    -----
+    If the `basis` parameter is specified, it is expected to be a basis in the
+    Aitchison simplex.  If there are `D-1` elements specified in `mat`, then
+    the dimensions of the basis needs be `D-1 x D`, where rows represent
+    basis vectors, and the columns represent proportions.
     """
     mat = closure(mat)
     if basis is None:
```
ebolyen: Commenting here because GitHub...

Just after line 563 we should perform the same check for 2D as in `ilr_inv`.

CHANGELOG.md
ebolyen: Also the issue number/link is no longer relevant to this change, so you can just drop it entirely.

ebolyen:
@mortonjt Done reviewing, two little things and then this should be good to go.

mortonjt:
Addressed both comments.  Thanks @ebolyen !

ebolyen:
Thanks @mortonjt, this should be ready for merge when tests pass.

ebolyen:
@mortonjt looks like we have  a PEP8 error:
```
skbio/stats/composition.py:574:1: E302 expected 2 blank lines, found 1
```

ebolyen:
Thanks @mortonjt!

mortonjt:
🎆 Thanks!!

PR #1398, opened 2016-06-27
Treenode prune bug

wasade:
Found a bug when using `TreeNode.shear` where `TreeNode.prune` could not handle a resulting subtree in which the root was not included. `TreeNode.prune` is called implicitly by `shear`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Bug fixes
 
+* `skbio.tree.TreeNode.prune` was not handling a situation in which a parent was validly removed during pruning operations as may happen if the resulting subtree does not include the root.
+
 ### Deprecated functionality [stable]
 
 ### Deprecated functionality [stable]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Bug fixes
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 
 ### Bug fixes
 
+* `skbio.tree.TreeNode.prune` was not handling a situation in which a parent was validly removed during pruning operations as may happen if the resulting subtree does not include the root.
```
jairideout: Can you please note the previous behavior (silent bug with incorrect results, or errors noisily?)

jairideout: Also note that this bug applies to `shear`, since users won't necessarily know that `shear` calls `prune` internally.

skbio/tree/tests/test_tree.py
```diff
@@ -303,6 +303,13 @@ def test_remove_node(self):
         self.assertEqual(len(n.children), 2)
         self.assertNotIn(n, self.simple_t.children)
 
+    def test_shear_prune_parent_dropped(self):
```
jairideout: Can you please add a test for `prune` specifically, in case `shear` doesn't use `prune` internally in the future?

jairideout:
Thanks for the fix @wasade! Had a couple comments related to changelog and unit tests.

jairideout:
Thanks @wasade! Ready for merge once tests finish.

PR #1399, opened 2016-06-27
PERF: 25% reduction in runtime for TreeNode.shear

wasade:
A smarter shearing algorithm. Observed ~25% reduction in runtimes when benchmarking on Greengenes 85% and 88% tree randomly pulling 10% of the tips.

### DIFF
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -450,10 +450,19 @@ class TreeNode(SkbioObject):
         if not ids.issubset(all_tips):
             raise ValueError("ids are not a subset of the tree.")
 
-        while len(list(tcopy.tips())) != len(ids):
-            for n in list(tcopy.tips()):
-                if n.name not in ids:
-                    n.parent.remove(n)
+        marked = set()
+        for tip in tcopy.tips():
+            if tip.name in ids:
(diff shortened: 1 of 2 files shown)


skbio/tree/_tree.py
```diff
@@ -450,10 +450,19 @@ def shear(self, names):
...
+                marked.add(tip)
+                for anc in tip.ancestors():
+                    if anc in marked:
+                        break
+                    else:
+                        marked.add(anc)
+
+        for node in list(tcopy.traverse()):
+            if node not in marked:
+                node.parent.remove(node)
```
ebolyen: :+1: This uses so fewer iterations it's almost not funny!

ebolyen:
@wasade Looks great, can you add a changelog indicating the performance increase for `shear`?

jairideout:
The 0.5.0-dev changelog entry is missing a `### Performance enhancements` section, can you add it there please?

skbio/tree/_tree.py
wasade: Not sure why we didn't just do this simpler method in the first place...

jairideout:
Thanks @wasade!

PR #1400, opened 2016-07-06
PERF: leverage pointer arithmetic for unifrac cython code

wasade:
Borrowed the idea from scikit-learn's decision [tree](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/tree/_tree.pyx). The idea is to create a pointer to the underlying data buffer of the numpy struct, and index directly into the C buffer. This approach avoids a cast and a few numerical operations which stem from the `__Pyx_BufPtrStrided2d` macro on each indexing operation. The reduction of operations and casting can get large on inner loops. Specifically, 2 additions, 2 products and 2 casts down to a single addition. To be specific, replacing

```c
#define __Pyx_BufPtrStrided2d(type, buf, i0, s0, i1, s1) (type)((char*)buf + i0 * s0 + i1 * s1)
```

...with:

```python
# assumes:
foo_ptr = <DTYPE_t*> foo.data

# 1D index and set
foo_ptr[i] = bar

# 2D index and set
foo_ptr[i * offset + j] = bar
```

I haven't benched this to assess performance differences -- my guess is a few percent at best but could be surprising particularly for nested forloops. I'd be happy to bench if necessary for review/merge.


wasade:
On reading a little more, it might be better to use `&foo[0,0]` as `foo.data` is marked as deprecated (years ago). Benefit is better API resilience. So still need some exploration here. 

One other benefit is this allows us to release the GIL on the nested loops as well

wasade:
Unexpectedly, I'm not seeing a difference in performance when benching against a 512 sample table. I am a bit confused about that, although goes to show the importance of gauging performance. I'm closing this PR as a result, however it may still be beneficial to do direct addressing in the future.

ebolyen:
It is probably because Cython was already optimizing in this way 
(there's an import ` cimport numpy as np` which would give it basically the same information)

wasade:
The generated C is tighter with fewer primitive operations performed,
although this suggests that those operations were the (very) minor
component of the runtime. On further testing, it isn't necessary for
releasing the GIL.

ebolyen:
Interesting, so the optimizations did make it through the Cython compiler, they just didn't matter? Thanks for looking into it. (Sorry for the late reply)

wasade:
Yes. Not sure if I linked before, but [this](https://jakevdp.github.io/blog/2012/08/08/memoryview-benchmarks/) is relevant. In thinking further, I think we do want to switch this code to memoryviews but I haven't had time to do so yet.

josenavas:
Thanks for the link @wasade I just took a quick look and it is really interesting!

PR #1404, opened 2016-07-19
ENH: Add DissimiliarityMatrix.from_iterable

### LINKED ISSUE #1343: Request: DissimiliarityMatrix.from_iterable
Right now, only the DistanceMatrix only has a `from_iterable` function.

It would be really nice if the DissimiliarityMatrix has an equivalent function.  This can be useful when calculating stuff like correlation matrices, where the coefficients don't need to be symmetric.

thermokarst:
Moves `from_iterable` definition from `DistanceMatrix` to `DissimilarityMatrix` (`DistanceMatrix` inherits from `DissimilarityMatrix`).

Copied existing `DistanceMatrix` `from_iterable` tests.

Fixes #1343.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,15 @@
 ## Version 0.5.0-dev  (changes since 0.5.0 go here)
 
 ### Features
+* `DissimilarityMatrix` now has a new constructor method called `from_iterable`. ([#1343](https://github.com/biocore/scikit-bio/issues/1343)).
+* `DissimilarityMatrix` now allows non-hollow matrices. ([#1343](https://github.com/biocore/scikit-bio/issues/1343)).
+* `DistanceMatrix.from_iterable` now accepts a `validate=True` parameter. ([#1343](https://github.com/biocore/scikit-bio/issues/1343)).
 
 ### Backward-incompatible changes [stable]
 
 ### Backward-incompatible changes [experimental]
 * Modifying basis handling in `skbio.stats.composition.ilr_inv` prior to checking for orthogonality.  Now the basis is strictly assumed to be in the Aitchison simplex. 
+* `DistanceMatrix.from_iterable` default behavior is now to validate matrix by computing all pairwise distances. Pass `validate=False` to get the previous behavior (no validation, but faster execution).([#1343](https://github.com/biocore/scikit-bio/issues/1343)).
 
 ### Performance enhancements
 * `TreeNode.shear` was rewritten for approximately a 25% performance increase. ([#1399](https://github.com/biocore/scikit-bio/pull/1399))
--- a/skbio/stats/distance/_base.py
+++ b/skbio/stats/distance/_base.py
@@ -109,6 +109,57 @@ class DissimilarityMatrix(SkbioObject):
         self._ids = ids
         self._id_index = self._index_list(self._ids)
 
+    @classonlymethod
+    @experimental(as_of="0.5.0-dev")
+    def from_iterable(cls, iterable, metric, key=None, keys=None):
+        """Create DissimilarityMatrix from an iterable given a metric.
+
+        Parameters
+        ----------
+        iterable : iterable
+            Iterable containing objects to compute pairwise dissimilarities on.
+        metric : callable
+            A function that takes two arguments and returns a float
+            representing the dissimilarity between the two arguments.
+        key : callable or metadata key, optional
+            A function that takes one argument and returns a string
+            representing the id of the element in the dissimilarity matrix.
+            Alternatively, a key to a `metadata` property if it exists for
+            each element in the `iterable`. If None, then default ids will be
+            used.
+        keys : iterable, optional
+            An iterable of the same length as `iterable`. Each element will be
+            used as the respective key.
+
+        Returns
+        -------
+        DissimilarityMatrix
+            The `metric` applied to all pairwise elements in the `iterable`.
+
+        Raises
+        ------
+        ValueError
+            If `key` and `keys` are both provided.
+
+        """
+        iterable = list(iterable)
+        if key is not None and keys is not None:
+            raise ValueError("Cannot use both `key` and `keys` at the same"
+                             " time.")
+
+        keys_ = None
+        if key is not None:
+            keys_ = [resolve_key(e, key) for e in iterable]
+        elif keys is not None:
+            keys_ = keys
+
+        dm = np.empty((len(iterable),) * 2)
+        for i, a in enumerate(iterable):
+            for j, b in enumerate(iterable):
+                dm[i, j] = metric(a, b)
+
+        return cls(dm, keys_)
+
     @property
     @experimental(as_of="0.4.0")
     def data(self):
@@ -623,9 +674,6 @@ class DissimilarityMatrix(SkbioObject):
         if data.dtype != np.double:
             raise DissimilarityMatrixError("Data must contain only floating "
                                            "point values.")
-        if np.trace(data) != 0:
-            raise DissimilarityMatrixError("Data must be hollow (i.e., the "
-                                           "diagonal can only contain zeros).")
         duplicates = find_duplicates(ids)
         if duplicates:
             formatted_duplicates = ', '.join(repr(e) for e in duplicates)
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Version 0.5.0-dev  (changes since 0.5.0 go here)
 
 ### Features
+* DissimilarityMatrix now has a new constructor method called from_iterable.
```
jairideout: Can you add the issue number that this fixes? Would also be nice to backtick DissimilarityMatrix and from_iterable for readability.

skbio/stats/distance/_base.py
```diff
@@ -109,6 +109,63 @@ def __init__(self, data, ids=None):
         self._ids = ids
         self._id_index = self._index_list(self._ids)
 
+    @classonlymethod
+    @experimental(as_of="0.5.0")
```
jairideout: change to 0.5.0-dev, we'll bump the version number at our next release

skbio/stats/distance/_base.py
```diff
@@ -109,6 +109,63 @@ def __init__(self, data, ids=None):
...
+        Parameters
+        ----------
+        iterable : iterable
+            Iterable containing objects to compute pairwise distances on.
+        metric : callable
+            A function that takes two arguments and returns a float
+            representing the distance between the two arguments.
+        key : callable or metadata key, optional
+            A function that takes one argument and returns a string
+            representing the id of the element in the distance matrix.
```
jairideout: *distance* -> *dissimilarity*

thermokarst:
I probably need to also update the metrics used in the tests that were added (right now these are distance metrics, but I need some guidance on what to replace with). Thanks!

mortonjt:
Awesome!  Thanks for knocking this out @thermokarst 

For good dissimilarity metrics, I would test [pearson correlation](http://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.stats.pearsonr.html) and [spearman correlation](http://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.stats.spearmanr.html) (without the pvalues)

thermokarst:
@jairideout: update based on our in-person discussion today.

- Builds the `DissimilarityMatrix` using all pairs from the iterable
- Drops the hollowness check from `DissimilarityMatrix._validate`
- Adds the hollowness check to `DistanceMatrix._validate`
- Introduces a `validate` kwarg to `DissimilarityMatrix.from_iterable` (default `True`), which causes `from_iterable` to work the same as before when True, or else opt-in to the `super` implementation.
- Added some new tests to handle the `validate` kwarg
- Updated docstrings

Thanks!

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Version 0.5.0-dev  (changes since 0.5.0 go here)
 
 ### Features
+* `DissimilarityMatrix` now has a new constructor method called `from_iterable`. ([#1343](https://github.com/biocore/scikit-bio/issues/1343)).
```
jairideout: Can you note that `DissimilarityMatrix` now allows non-hollow matrices, and the addition of `validate=True` parameter to `DistanceMatrix.from_iterable`? Also note in the "backwards-incompatible changes (experimental)" section that `DistanceMatrix.from_iterable` default behavior is now to validate matrix by computing all pairwise distances. Instruct users to pass `validate=False` to get the previous behavior (no validation, but faster execution).

thermokarst: No problem. Did I do the `validate` kwarg backwards? Right now `True` provides the previous behavior (default).

skbio/stats/distance/_base.py
```diff
@@ -109,6 +109,57 @@ def __init__(self, data, ids=None):
...
+            raise ValueError("Cannot use both `key` and `keys` at the same"
+                             " time.")
+
+        keys_ = None
+        if key is not None:
+            keys_ = [resolve_key(e, key) for e in iterable]
+        elif keys is not None:
+            keys_ = keys
+
+        dm = np.zeros((len(iterable),) * 2)
```
jairideout: Now that the entire matrix is filled in it's a bit faster to use `np.empty` instead of `np.zeros`. Can you update this call?

skbio/stats/distance/_base.py
```diff
@@ -702,6 +751,9 @@ def from_iterable(cls, iterable, metric, key=None, keys=None):
         keys : iterable, optional
             An iterable of the same length as `iterable`. Each element will be
             used as the respective key.
+        validate : boolean, optional
+            If ``True``, assume symmetry and hollowness when constructing the
```
jairideout: Can you add the following text to give the user more detail:

```
If ``True``, all pairwise distances are computed, including upper and lower triangles and the diagonal,
and the resulting matrix is validated for symmetry and hollowness. If ``False``, `metric` is assumed to
be hollow and symmetric and only the lower triangle (excluding the diagonal) is computed.
Pass ``validate=False`` if you are sure `metric` is hollow and symmetric for improved performance.
```

skbio/stats/distance/_base.py
```diff
@@ -715,28 +767,31 @@ def from_iterable(cls, iterable, metric, key=None, keys=None):
 
         Notes
         -----
-        Symmetry and hollowness are assumed when calculating the distances via
-        `metric`. Therefore, distances are only computed for the strictly
+        When `validate=True`, distances are only computed for the strictly
```
jairideout: Remove this "Notes" section since the details are covered above in the description for `validate`.

jairideout: I can't comment on the lines above this diff, but can you remove "all" from the "Returns" section text? i.e. "The `metric` applied to pairwise elements in the `iterable`."

skbio/stats/distance/_base.py
```diff
@@ -715,28 +767,31 @@ def from_iterable(cls, iterable, metric, key=None, keys=None):
...
         upper/lower triangle.
 
         """
-        iterable = list(iterable)
-        if key is not None and keys is not None:
-            raise ValueError("Cannot use both `key` and `keys` at the same"
-                             " time.")
+        if validate:
```
jairideout: This `if` statement is inverted from what it should be, we want to call `super` if `validate=True`, else only compute lower triangle.

skbio/stats/distance/tests/test_base.py
```diff
@@ -113,10 +113,101 @@ def test_init_invalid_input(self):
         with self.assertRaises(DissimilarityMatrixError):
             DissimilarityMatrix(data, [])
 
-        # Non-hollow.
-        data = [[0.0, 1.0], [1.0, 0.01]]
```
jairideout: Can you add the following test cases?

- [ ] test that `DissimilarityMatrix` constructor supports non-hollow data
- [ ] test that `DissimilarityMatrix.from_iterable` supports non-hollow data
- [ ] test that `DissimilarityMatrix.from_iterable` supports asymmetric data
- [ ] test that `DistanceMatrix.from_iterable(validate=True)` is equal to `DistanceMatrix.from_iterable(validate=False)` when passed a valid (i.e., symmetric and hollow) metric
- [ ] test that `DistanceMatrix.from_iterable(validate=False)` "works" with a non-hollow/asymmetric metric (i.e. doesn't raise an error)

CHANGELOG.md
jairideout: Yeah `validate` is backwards

jairideout:
Thanks @thermokarst, done reviewing!

thermokarst:
Thanks @jairideout! I think I have addressed all of your comments.

CHANGELOG.md
```diff
@@ -3,8 +3,12 @@
...
+* `DissimilarityMatrix` now allows non-hollow matrices. ([#1343](https://github.com/biocore/scikit-bio/issues/1343)).
+* `DistanceMatrix.from_iterable` now accepts a `validate=True` parameter. ([#1343](https://github.com/biocore/scikit-bio/issues/1343)).
 
 ### Backward-incompatible changes [stable]
+* `DistanceMatrix.from_iterable` default behavior is now to validate matrix by computing all pairwise distances. Pass `validate=False` to get the previous behavior (no validation, but faster execution).([#1343](https://github.com/biocore/scikit-bio/issues/1343)).
```
jairideout: move this under "experimental"

jairideout:
Thanks @thermokarst!!

PR #1423, opened 2016-09-06
mixin for intervalmetadata

RNAer:
This is a followup PR of #1414 .

@jairideout and @ebolyen , could you take a quick look to see if the mixin class API and the way how the repr represents the object are ok? don't worry about the tests. I will add more docstring and fix tests.

The next task is to modify all the sequence classes to adapt the addition of `IntervalMetadata`.

The next next task is to modify GenBank parser.

### DIFF
--- a/skbio/metadata/_mixin.py
+++ b/skbio/metadata/_mixin.py
@@ -410,7 +410,6 @@ class PositionalMetadataMixin(metaclass=abc.ABCMeta):
 
 
 class IntervalMetadataMixin(metaclass=abc.ABCMeta):
-    @abc.abstractmethod
     def _interval_metadata_axis_len_(self):
         ''''''
         raise NotImplementedError
--- a/skbio/metadata/_mixin.py
+++ b/skbio/metadata/_mixin.py
@@ -410,27 +410,40 @@ class PositionalMetadataMixin(metaclass=abc.ABCMeta):
 
 
 class IntervalMetadataMixin(metaclass=abc.ABCMeta):
+    @abc.abstractmethod
+    def _interval_metadata_axis_len_(self):
+        ''''''
+        raise NotImplementedError
+
+    @abc.abstractmethod
     def __init__(self, interval_metadata=None):
         raise NotImplementedError
 
     def _init_(self, interval_metadata=None):
-        self.interval_metadata = interval_metadata
+        if interval_metadata is None:
+            # Could use deleter but this is less overhead and needs to be fast.
+            self._interval_metadata = None
+        else:
+            # Use setter for validation and copy.
+            self.interval_metadata = interval_metadata
 
     @property
     @experimental(as_of="0.5.0-dev")
     def interval_metadata(self):
         '''``IntervalMetadata`` object containing info about interval features.
 
+        Examples
+        --------
         >>> from skbio import DNA
         >>> from skbio.metadata import IntervalMetadata
-        >>> im = IntervalMetadata()
+        >>> im = IntervalMetadata(4)
         >>> intvl = im.add([(0, 3)], metadata={'gene': 'sagA'})
         >>> seq = DNA('ACGT', interval_metadata=im)
         >>> seq
         DNA
         --------------------------
         Interval metadata:
-            1 interval features
+            1 interval feature
         Stats:
             length: 4
             has gaps: False
@@ -442,16 +455,22 @@ class IntervalMetadataMixin(metaclass=abc.ABCMeta):
         '''
         if self._interval_metadata is None:
             # Not using setter to avoid copy.
-            self._interval_metadata = IntervalMetadata()
+            self._interval_metadata = IntervalMetadata(
+                self._interval_metadata_axis_len_())
         return self._interval_metadata
 
     @interval_metadata.setter
     def interval_metadata(self, interval_metadata):
-        if interval_metadata is None:
-            interval_metadata = IntervalMetadata()
         if isinstance(interval_metadata, IntervalMetadata):
+            upper_bound = interval_metadata.upper_bound
+            axis_len = self._interval_metadata_axis_len_()
+            if upper_bound != axis_len:
+                raise ValueError(
+                    'The upper bound for the interval features (%d) '
+                    'must match the interval metadata axis length (%d)'
+                    % (upper_bound, axis_len))
             # copy all the data to the mixin
-            self._interval_metadata = copy.deepcopy(interval_metadata)
+            self._interval_metadata = copy.copy(interval_metadata)
         else:
             raise TypeError('You must provide `IntervalMetadata` object.')
 
@@ -484,7 +503,7 @@ class IntervalMetadataMixin(metaclass=abc.ABCMeta):
         >>> seq.has_interval_metadata()
         False
         >>> from skbio.metadata import IntervalMetadata
-        >>> im = IntervalMetadata()
+        >>> im = IntervalMetadata(4)
         >>> seq = DNA('ACGT', interval_metadata=im)
         >>> seq.has_interval_metadata()
         False
@@ -494,14 +513,27 @@ class IntervalMetadataMixin(metaclass=abc.ABCMeta):
         True
         """
         return (self._interval_metadata is not None and
-                len(self.interval_metadata._intervals) > 0)
+                self.interval_metadata.num_interval_features > 0)
 
     @abc.abstractmethod
     def __eq__(self, other):
         raise NotImplementedError
 
     def _eq_(self, other):
-        return self.interval_metadata == other.interval_metadata
+        # We're not simply comparing self.interval_metadata to
+        # other.interval_metadata in order to avoid creating "empty"
+        # interval metadata representations on the objects if they don't have
+        # interval metadata.
+        if self.has_interval_metadata() and other.has_interval_metadata():
+            return self.interval_metadata == other.interval_metadata
+        elif not (self.has_interval_metadata() or
+                  other.has_interval_metadata()):
+            # Both don't have interval metadata.
+            return (self._interval_metadata_axis_len_() ==
+                    other._interval_metadata_axis_len_())
+        else:
+            # One has interval metadata while the other does not.
+            return False
 
     @abc.abstractmethod
     def __ne__(self, other):
--- a/skbio/metadata/_mixin.py
+++ b/skbio/metadata/_mixin.py
@@ -410,8 +410,16 @@ class PositionalMetadataMixin(metaclass=abc.ABCMeta):
 
 
 class IntervalMetadataMixin(metaclass=abc.ABCMeta):
+    @abc.abstractmethod
     def _interval_metadata_axis_len_(self):
-        ''''''
+        '''Return length of axis that positional metadata applies to.
+
+        Returns
+        -------
+        int
+            Positional metadata axis length.
+
+        '''
         raise NotImplementedError
 
     @abc.abstractmethod
@@ -430,6 +438,11 @@ class IntervalMetadataMixin(metaclass=abc.ABCMeta):
     @experimental(as_of="0.5.0-dev")
     def interval_metadata(self):
         '''``IntervalMetadata`` object containing info about interval features.
+
+        Notes
+        -----
+        This property can be set and deleted. When setting new interval metadata, a
+        shallow copy of the `IntervalMetadata` object is made.
         '''
         if self._interval_metadata is None:
             # Not using setter to avoid copy.
@@ -442,7 +455,7 @@ class IntervalMetadataMixin(metaclass=abc.ABCMeta):
         if isinstance(interval_metadata, IntervalMetadata):
             upper_bound = interval_metadata.upper_bound
             axis_len = self._interval_metadata_axis_len_()
-            if upper_bound != axis_len:
+            if upper_bound != axis_len or :
                 raise ValueError(
                     'The upper bound for the interval features (%d) '
                     'must match the interval metadata axis length (%d)'
--- a/skbio/metadata/_mixin.py
+++ b/skbio/metadata/_mixin.py
@@ -454,8 +454,13 @@ class IntervalMetadataMixin(metaclass=abc.ABCMeta):
     def interval_metadata(self, interval_metadata):
         if isinstance(interval_metadata, IntervalMetadata):
             upper_bound = interval_metadata.upper_bound
+            lower_bound = interval_metadata.lower_bound
             axis_len = self._interval_metadata_axis_len_()
-            if upper_bound != axis_len or :
+            if lower_bound != 0:
+                raise ValueError(
+                    'The lower bound for the interval features (%d) '
+                    'must be zero.' % lower_bound)
+            if upper_bound != axis_len:
                 raise ValueError(
                     'The upper bound for the interval features (%d) '
                     'must match the interval metadata axis length (%d)'
--- a/skbio/metadata/_mixin.py
+++ b/skbio/metadata/_mixin.py
@@ -441,8 +441,10 @@ class IntervalMetadataMixin(metaclass=abc.ABCMeta):
 
         Notes
         -----
-        This property can be set and deleted. When setting new interval metadata, a
-        shallow copy of the `IntervalMetadata` object is made.
+        This property can be set and deleted. When setting new
+        interval metadata, a shallow copy of the ``IntervalMetadata``
+        object is made.
+
         '''
         if self._interval_metadata is None:
             # Not using setter to avoid copy.
--- a/skbio/metadata/_mixin.py
+++ b/skbio/metadata/_mixin.py
@@ -11,7 +11,8 @@ import copy
 
 import pandas as pd
 
-from skbio.util._decorator import stable
+from skbio.util._decorator import stable, experimental
+from skbio.metadata import IntervalMetadata
 
 
 class MetadataMixin(metaclass=abc.ABCMeta):
@@ -406,3 +407,125 @@ class PositionalMetadataMixin(metaclass=abc.ABCMeta):
         """
         return (self._positional_metadata is not None and
                 len(self.positional_metadata.columns) > 0)
+
+
+class IntervalMetadataMixin(metaclass=abc.ABCMeta):
+    def __init__(self, interval_metadata=None):
+        raise NotImplementedError
+
+    def _init_(self, interval_metadata=None):
+        self.interval_metadata = interval_metadata
+
+    @property
+    @experimental(as_of="0.5.0-dev")
+    def interval_metadata(self):
+        '''``IntervalMetadata`` object containing info about interval features.
+
+        >>> from skbio import DNA
+        >>> from skbio.metadata import IntervalMetadata
+        >>> im = IntervalMetadata()
+        >>> intvl = im.add([(0, 3)], metadata={'gene': 'sagA'})
+        >>> seq = DNA('ACGT', interval_metadata=im)
+        >>> seq
+        DNA
+        --------------------------
+        Interval metadata:
+            1 interval features
+        Stats:
+            length: 4
+            has gaps: False
+            has degenerates: False
+            has definites: True
+            GC-content: 50.00%
+        --------------------------
+        0 ACGT
+        '''
+        if self._interval_metadata is None:
+            # Not using setter to avoid copy.
+            self._interval_metadata = IntervalMetadata()
+        return self._interval_metadata
+
+    @interval_metadata.setter
+    def interval_metadata(self, interval_metadata):
+        if interval_metadata is None:
+            interval_metadata = IntervalMetadata()
+        if isinstance(interval_metadata, IntervalMetadata):
+            # copy all the data to the mixin
+            self._interval_metadata = copy.deepcopy(interval_metadata)
+        else:
+            raise TypeError('You must provide `IntervalMetadata` object.')
+
+    @interval_metadata.deleter
+    def interval_metadata(self):
+        self._interval_metadata = None
+
+    @experimental(as_of="0.5.0-dev")
+    def has_interval_metadata(self):
+        """Determine if the object has interval metadata.
+
+        An object has interval metadata if its ``interval_metadata``
+        has at least one ```Interval`` objects.
+
+        Returns
+        -------
+        bool
+            Indicates whether the object has interval metadata.
+
+        Examples
+        --------
+        .. note:: scikit-bio objects with interval metadata share a common
+           interface for accessing and manipulating their interval metadata.
+           The following examples use scikit-bio's ``DNA`` class for
+           demonstration. These examples apply to all other
+           scikit-bio objects storing interval metadata.
+
+        >>> from skbio import DNA
+        >>> seq = DNA('ACGT')
+        >>> seq.has_interval_metadata()
+        False
+        >>> from skbio.metadata import IntervalMetadata
+        >>> im = IntervalMetadata()
+        >>> seq = DNA('ACGT', interval_metadata=im)
+        >>> seq.has_interval_metadata()
+        False
+        >>> intvl = im.add([(0, 1)], metadata={'gene': 'sagA'})
+        >>> seq = DNA('ACGT', interval_metadata=im)
+        >>> seq.has_interval_metadata()
+        True
+        """
+        return (self._interval_metadata is not None and
+                len(self.interval_metadata._intervals) > 0)
+
+    @abc.abstractmethod
+    def __eq__(self, other):
+        raise NotImplementedError
+
+    def _eq_(self, other):
+        return self.interval_metadata == other.interval_metadata
+
+    @abc.abstractmethod
+    def __ne__(self, other):
+        raise NotImplementedError
+
+    def _ne_(self, other):
+        return not (self == other)
+
+    @abc.abstractmethod
+    def __copy__(self):
+        raise NotImplementedError
+
+    def _copy_(self):
+        if self.has_interval_metadata():
+            return copy.copy(self.interval_metadata)
+        else:
+            return None
+
+    @abc.abstractmethod
+    def __deepcopy__(self, memo):
+        raise NotImplementedError
+
+    def _deepcopy_(self, memo):
+        if self.has_interval_metadata():
+            return copy.deepcopy(self.interval_metadata, memo)
+        else:
+            return None
--- a/skbio/metadata/_mixin.py
+++ b/skbio/metadata/_mixin.py
@@ -468,13 +468,6 @@ class IntervalMetadataMixin(metaclass=abc.ABCMeta):
         bool
             Indicates whether the object has interval metadata.
 
-        Examples
-        --------
-        .. note:: scikit-bio objects with interval metadata share a common
-           interface for accessing and manipulating their interval metadata.
-           The following examples use scikit-bio's ``DNA`` class for
-           demonstration. These examples apply to all other
-           scikit-bio objects storing interval metadata.
         """
         return (self._interval_metadata is not None and
                 self.interval_metadata.num_interval_features > 0)
--- a/skbio/metadata/_mixin.py
+++ b/skbio/metadata/_mixin.py
@@ -468,7 +468,8 @@ class IntervalMetadataMixin(metaclass=abc.ABCMeta):
             # copy all the data to the mixin
             self._interval_metadata = copy.copy(interval_metadata)
         else:
-            raise TypeError('You must provide `IntervalMetadata` object.')
+            raise TypeError('You must provide `IntervalMetadata` object, '
+                            'not type %s.' % type(interval_metadata).__name__)
 
     @interval_metadata.deleter
     def interval_metadata(self):
--- a/skbio/metadata/_mixin.py
+++ b/skbio/metadata/_mixin.py
@@ -412,12 +412,12 @@ class PositionalMetadataMixin(metaclass=abc.ABCMeta):
 class IntervalMetadataMixin(metaclass=abc.ABCMeta):
     @abc.abstractmethod
     def _interval_metadata_axis_len_(self):
-        '''Return length of axis that positional metadata applies to.
+        '''Return length of axis that interval metadata applies to.
 
         Returns
         -------
         int
-            Positional metadata axis length.
+            Interval metadata axis length.
 
         '''
         raise NotImplementedError
--- a/skbio/metadata/_mixin.py
+++ b/skbio/metadata/_mixin.py
@@ -430,27 +430,6 @@ class IntervalMetadataMixin(metaclass=abc.ABCMeta):
     @experimental(as_of="0.5.0-dev")
     def interval_metadata(self):
         '''``IntervalMetadata`` object containing info about interval features.
-
-        Examples
-        --------
-        >>> from skbio import DNA
-        >>> from skbio.metadata import IntervalMetadata
-        >>> im = IntervalMetadata(4)
-        >>> intvl = im.add([(0, 3)], metadata={'gene': 'sagA'})
-        >>> seq = DNA('ACGT', interval_metadata=im)
-        >>> seq
-        DNA
-        --------------------------
-        Interval metadata:
-            1 interval feature
-        Stats:
-            length: 4
-            has gaps: False
-            has degenerates: False
-            has definites: True
-            GC-content: 50.00%
-        --------------------------
-        0 ACGT
         '''
         if self._interval_metadata is None:
             # Not using setter to avoid copy.
@@ -496,20 +475,6 @@ class IntervalMetadataMixin(metaclass=abc.ABCMeta):
            The following examples use scikit-bio's ``DNA`` class for
            demonstration. These examples apply to all other
            scikit-bio objects storing interval metadata.
-
-        >>> from skbio import DNA
-        >>> seq = DNA('ACGT')
-        >>> seq.has_interval_metadata()
-        False
-        >>> from skbio.metadata import IntervalMetadata
-        >>> im = IntervalMetadata(4)
-        >>> seq = DNA('ACGT', interval_metadata=im)
-        >>> seq.has_interval_metadata()
-        False
-        >>> intvl = im.add([(0, 1)], metadata={'gene': 'sagA'})
-        >>> seq = DNA('ACGT', interval_metadata=im)
-        >>> seq.has_interval_metadata()
-        True
         """
         return (self._interval_metadata is not None and
                 self.interval_metadata.num_interval_features > 0)
--- a/skbio/metadata/_repr.py
+++ b/skbio/metadata/_repr.py
@@ -15,8 +15,8 @@ from skbio._base import ElasticLines
 
 
 class _MetadataReprBuilder(metaclass=ABCMeta):
-    """Abstract base class for building a repr for an object containing
-    metadata, positional metadata, and/or interval metadata.
+    """Abstract base class for building  a repr for an object containing
+    metadata and/or positional metadata.
 
     Parameters
     ----------
@@ -50,7 +50,6 @@ class _MetadataReprBuilder(metaclass=ABCMeta):
         self._process_header()
         self._process_metadata()
         self._process_positional_metadata()
-        self._process_interval_metadata()
         self._process_stats()
         self._process_data()
 
@@ -126,16 +125,6 @@ class _MetadataReprBuilder(metaclass=ABCMeta):
         key_fmt = self._format_key(key)
         dtype_fmt = '<dtype: %s>' % str(dtype)
         return self._wrap_text_with_indent(dtype_fmt, key_fmt, 1)
-
-    def _process_interval_metadata(self):
-        if (hasattr(self._obj, "has_interval_metadata") and
-           self._obj.has_interval_metadata()):
-            self._lines.add_line('Interval metadata:')
-            n = self._obj.interval_metadata.num_interval_features
-            line = self._indent + '%d interval feature' % n
-            if n > 1:
-                line += 's'
-            self._lines.add_line(line)
 
     def _format_key(self, key):
         """Format metadata key.
--- a/skbio/metadata/_repr.py
+++ b/skbio/metadata/_repr.py
@@ -128,7 +128,8 @@ class _MetadataReprBuilder(metaclass=ABCMeta):
         return self._wrap_text_with_indent(dtype_fmt, key_fmt, 1)
 
     def _process_interval_metadata(self):
-        if self._obj.has_interval_metadata():
+        if (hasattr(self._obj, "has_interval_metadata") and
+           self._obj.has_interval_metadata()):
             self._lines.add_line('Interval metadata:')
             n = self._obj.interval_metadata.num_interval_features
             line = self._indent + '%d interval feature' % n
--- a/skbio/metadata/_repr.py
+++ b/skbio/metadata/_repr.py
@@ -15,8 +15,8 @@ from skbio._base import ElasticLines
 
 
 class _MetadataReprBuilder(metaclass=ABCMeta):
-    """Abstract base class for building  a repr for an object containing
-    metadata and/or positional metadata.
+    """Abstract base class for building a repr for an object containing
+    metadata, positional metadata, and/or interval metadata.
 
     Parameters
     ----------
@@ -50,6 +50,7 @@ class _MetadataReprBuilder(metaclass=ABCMeta):
         self._process_header()
         self._process_metadata()
         self._process_positional_metadata()
+        self._process_interval_metadata()
         self._process_stats()
         self._process_data()
 
@@ -125,6 +126,13 @@ class _MetadataReprBuilder(metaclass=ABCMeta):
         key_fmt = self._format_key(key)
         dtype_fmt = '<dtype: %s>' % str(dtype)
         return self._wrap_text_with_indent(dtype_fmt, key_fmt, 1)
+
+    def _process_interval_metadata(self):
+        if self._obj.has_interval_metadata():
+            self._lines.add_line('Interval metadata:')
+            self._lines.add_line(
+                    self._indent + '%d interval features' % len(
+                        self._obj.interval_metadata._intervals))
 
     def _format_key(self, key):
         """Format metadata key.
--- a/skbio/metadata/_repr.py
+++ b/skbio/metadata/_repr.py
@@ -130,9 +130,11 @@ class _MetadataReprBuilder(metaclass=ABCMeta):
     def _process_interval_metadata(self):
         if self._obj.has_interval_metadata():
             self._lines.add_line('Interval metadata:')
-            self._lines.add_line(
-                    self._indent + '%d interval features' % len(
-                        self._obj.interval_metadata._intervals))
+            n = self._obj.interval_metadata.num_interval_features
+            line = self._indent + '%d interval feature' % n
+            if n > 1:
+                line += 's'
+            self._lines.add_line(line)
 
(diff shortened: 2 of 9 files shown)


RNAer:
@wasade , @josenavas, @ekopylova, it would be great if you can take a look too.

skbio/metadata/_mixin.py
```diff
@@ -406,3 +407,125 @@ def has_positional_metadata(self):
...
+
+    def _init_(self, interval_metadata=None):
+        self.interval_metadata = interval_metadata
+
+    @property
+    @experimental(as_of="0.5.0-dev")
+    def interval_metadata(self):
+        '''``IntervalMetadata`` object containing info about interval features.
+
+        >>> from skbio import DNA
```
mortonjt: Need to have `Examples` header

skbio/metadata/_mixin.py
```diff
@@ -406,3 +407,125 @@ def has_positional_metadata(self):
         """
         return (self._positional_metadata is not None and
                 len(self.positional_metadata.columns) > 0)
+
+
+class IntervalMetadataMixin(metaclass=abc.ABCMeta):
+    def __init__(self, interval_metadata=None):
```
jairideout: Decorate with `@abc.abstractmethod`

skbio/metadata/_mixin.py
```diff
@@ -406,3 +407,125 @@ def has_positional_metadata(self):
...
+        raise NotImplementedError
+
+    def _init_(self, interval_metadata=None):
```
jairideout: Please structure methods in this class like they are in the positional metadata mixin. I'm trying to keep the 3 mixin classes as consistent as possible so it's easier to maintain and reason about them, and certain design decisions were made for safety and performance. It'll also make writing the tests easier because you can model those off of `metadata` or `positional_metadata`. For example, `_init_` should be:

```python
def _init_(self, interval_metadata=None):
    if interval_metadata is None:
        # Could use deleter but this is less overhead and needs to be fast.
        self._interval_metadata = None
    else:
        # Use setter for validation and copy.
        self.interval_metadata = interval_metadata
```

I'll mark any other methods that need updating to match positional metadata mixin.

skbio/metadata/_mixin.py
```diff
@@ -406,3 +407,125 @@ def has_positional_metadata(self):
...
+            has gaps: False
+            has degenerates: False
+            has definites: True
+            GC-content: 50.00%
+        --------------------------
+        0 ACGT
+        '''
+        if self._interval_metadata is None:
+            # Not using setter to avoid copy.
+            self._interval_metadata = IntervalMetadata()
```
jairideout: We'll need the `IntervalMetadata` constructor to accept a length, and validate that interval features all fall within that coordinate space when they are added. This is necessary to ensure that interval metadata spans the length of the sequence, similar to how the number of rows in positional metadata (`pd.DataFrame`) match the sequence length.

This doesn't have to be called "length", since that could be confused with the number of interval features in the interval metadata. I'm not sure what to call it, do you have any ideas?

Once `IntervalMetadata` has a "length", `IntervalMetadata._reverse` can be renamed `.reverse()` and will no longer need to accept `length` as a parameter (since it will be instance state).

This mixin will need an abstract method similar to `_positional_metadata_axis_len_` that objects must implement to define their length. Probably makes sense to call it `_interval_metadata_axis_len_` (or some variation if we have a better name than "length").

skbio/metadata/_mixin.py
```diff
@@ -406,3 +407,125 @@ def has_positional_metadata(self):
...
+        return self._interval_metadata
+
+    @interval_metadata.setter
+    def interval_metadata(self, interval_metadata):
+        if interval_metadata is None:
```
jairideout: Remove support for `None` since the constructor, getter, and deleter will handle this (see positional metadata's setter or metadata's setter for examples).

skbio/metadata/_mixin.py
```diff
@@ -406,3 +407,125 @@ def has_positional_metadata(self):
...
+            interval_metadata = IntervalMetadata()
+        if isinstance(interval_metadata, IntervalMetadata):
+            # copy all the data to the mixin
+            self._interval_metadata = copy.deepcopy(interval_metadata)
```
jairideout: We want a shallow copy here instead of a deep copy to match the other metadata mixins.

`IntervalMetadata` doesn't define a `__copy__` method, which is what `copy.copy` calls (`copy.copy` performs a shallow copy). It *may* work as-is, but can you please add an `IntervalMetadata.__copy__` method with those semantics defined explicitly? I tend to not trust that custom classes will work with `copy.copy` (for example, numpy arrays and pandas dataframes don't always work with `copy.copy`, though that may have been fixed in more recent versions).

Also, above this line is where we'd want to validate that the supplied `interval_metadata` object's length matches `self._interval_metadata_axis_len_()`. This validation is similar to the positional metadata setter.

skbio/metadata/_mixin.py
```diff
@@ -406,3 +407,125 @@ def has_positional_metadata(self):
...
+
+    @interval_metadata.deleter
+    def interval_metadata(self):
+        self._interval_metadata = None
+
+    @experimental(as_of="0.5.0-dev")
+    def has_interval_metadata(self):
+        """Determine if the object has interval metadata.
+
+        An object has interval metadata if its ``interval_metadata``
```
jairideout: 👍

skbio/metadata/_mixin.py
```diff
@@ -406,3 +407,125 @@ def has_positional_metadata(self):
...
+        """
+        return (self._interval_metadata is not None and
+                len(self.interval_metadata._intervals) > 0)
+
+    @abc.abstractmethod
+    def __eq__(self, other):
+        raise NotImplementedError
+
+    def _eq_(self, other):
+        return self.interval_metadata == other.interval_metadata
```
jairideout: See the logic for positional metadata mixin's `_eq_` method -- there's more complicated logic to handle cases where positional metadata is missing.

skbio/metadata/_mixin.py
```diff
@@ -406,3 +407,125 @@ def has_positional_metadata(self):
...
+    def _ne_(self, other):
+        return not (self == other)
+
+    @abc.abstractmethod
+    def __copy__(self):
+        raise NotImplementedError
+
+    def _copy_(self):
+        if self.has_interval_metadata():
+            return copy.copy(self.interval_metadata)
```
jairideout: See [my comment](https://github.com/biocore/scikit-bio/pull/1423/files#r77926433) about defining `IntervalMetadata.__copy__`. This code won't need updating because `copy.copy` will handle calling the magic method.

skbio/metadata/_mixin.py
```diff
@@ -406,3 +407,125 @@ def has_positional_metadata(self):
...
+        else:
+            return None
+
+    @abc.abstractmethod
+    def __deepcopy__(self, memo):
+        raise NotImplementedError
+
+    def _deepcopy_(self, memo):
+        if self.has_interval_metadata():
+            return copy.deepcopy(self.interval_metadata, memo)
```
jairideout: Related to my comment [here](https://github.com/biocore/scikit-bio/pull/1423/files#r77926433). Can you define `IntervalMetadata.__deepcopy__` to ensure this works as we intend? This code won't need updating because `copy.deepcopy` will handle calling the magic method once it exists.

skbio/metadata/_repr.py
```diff
@@ -126,6 +127,13 @@ def _format_positional_metadata_column(self, key, dtype):
         dtype_fmt = '<dtype: %s>' % str(dtype)
         return self._wrap_text_with_indent(dtype_fmt, key_fmt, 1)
 
+    def _process_interval_metadata(self):
+        if self._obj.has_interval_metadata():
+            self._lines.add_line('Interval metadata:')
+            self._lines.add_line(
+                    self._indent + '%d interval features' % len(
```
jairideout: Can you handle the singular case? i.e. "1 interval feature"

jairideout:
@RNAer overall this looks great! Most of my comments are related to matching the behavior of the other metadata mixin classes, and defining additional methods on `IntervalMetadata` that the mixin will use. I also had some comments about validation, which requires `IntervalMetadata` to have a length. The repr looks good.

I didn't look at the tests, will do that when you're ready and these larger comments have been addressed. Please let me know if you have any questions about my review comments.

skbio/metadata/_mixin.py
```diff
@@ -406,3 +407,125 @@ def has_positional_metadata(self):
...
+        >>> seq = DNA('ACGT', interval_metadata=im)
+        >>> seq.has_interval_metadata()
+        False
+        >>> intvl = im.add([(0, 1)], metadata={'gene': 'sagA'})
+        >>> seq = DNA('ACGT', interval_metadata=im)
+        >>> seq.has_interval_metadata()
+        True
+        """
+        return (self._interval_metadata is not None and
+                len(self.interval_metadata._intervals) > 0)
```
jairideout: Maybe it makes sense for `IntervalMetadata` to have a read-only property, something like `num_intervals` or `num_interval_features` so that this mixin doesn't rely on private API? This would avoid the repr builder from accessing that private state too.

skbio/metadata/_mixin.py
RNAer: This sounds good to me. how about `upper_bound` or `sequence_length` for the nameing?

skbio/metadata/_repr.py
RNAer: sure. so the seq repr looks ok to show just the number of interval features as in line 430?

skbio/metadata/_mixin.py
RNAer: what about implementing `__len__` and `__iter__`? this is more pythonic?

RNAer:
Thanks, @jairideout . I have replied your 3 comments and if you think ok, I can add those to `IntervalMetadata` in another PR.

skbio/metadata/_mixin.py
jairideout: I like `upper_bound`. It seems interval metadata *is* an interval from `[0, N]`, where `N` is the `upper_bound`, and interval metadata contains regions (contiguous or non-contiguous) that are identified to have particular metadata attached to them.

`sequence_length` would make sense if interval metadata was only attached to sequences, but it can be applied to any other scikit-bio object that has a notion of length (e.g. `TabularMSA`).

@ebolyen do you have thoughts about naming here?

skbio/metadata/_repr.py
jairideout: Yeah I think the repr looks great!

skbio/metadata/_mixin.py
jairideout: It'd be more Pythonic, but we had originally deferred defining `__len__`, `__iter__`, `__getitem__`, `__setitem__`, etc., which all must agree on what "length" means for `IntervalMetadata`. We weren't able to come to a conclusion in the RFC because we weren't sure how/if `Sequence.__getitem__` would use `IntervalMetadata.__getitem__`. `__getitem__`'s behavior will determine the behavior of `__len__`, `__iter__`, etc. so we can't define those yet.

Defining these methods depends on whether we see `IntervalMetadata` as a Python "sequence" or a "mapping". For details on sequence and mapping protocols, see [[1]](https://docs.python.org/3/reference/datamodel.html#emulating-container-types) and [[2]](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes). I think we should defer defining these methods until we have a clearer idea of what makes the most sense / what is useful for them to do (and that'll require some discussion). To move this forward, I think we should define an attribute with a clear name that isn't tied to the magic methods listed above. That way we are free to define those later in a backwards-compatible way. I like `num_interval_features` or `interval_feature_count` since they're very descriptive names, and that's how they're described in the repr and `has_interval_metadata`. What do you think?

jairideout: @ebolyen do you have thoughts on this?

jairideout:
Thanks @RNAer! I replied to your comments, also soliciting feedback from @ebolyen. Let's finalize those discussions and then it'd be great if you could make those `IntervalMetadata` changes in another PR that is merged before this one. Thanks!

skbio/metadata/_mixin.py
ebolyen: `_interval_metadata_axis_len_` makes sense to me... We aren't saying anything about the interval metadata so much as the limitation of the axis it exists on (as defined by the mixin-ee/mixee).

jairideout: This discussion is about what the name `IntervalMetadata` constructor's parameter defining its "length".

skbio/metadata/_mixin.py
ebolyen: :+1: for delaying the builtins, I have no strong feelings on the name though.

skbio/metadata/_mixin.py
ebolyen: Sorry I misunderstood. What about `axis_length`? then it matches the "sunder" method and doesn't imply too much.

skbio/metadata/_mixin.py
jairideout: Thanks @ebolyen! @RNAer, I'm leaning towards `num_interval_features` or `interval_feature_count`. Do you think either of those names makes sense here?

skbio/metadata/_mixin.py
jairideout: "axis length" makes sense for the object this mixin is applied to, since some objects have more than one axis (e.g. for `TabularMSA`, it could be attached to rows or columns. `Sequence` is less interesting.). For `IntervalMetadata` though, it really is a closed interval `[0, N]`. `N` is what we're trying to name here (which will be the constructor parameter). Using "axis" in `IntervalMetadata` doesn't seem as clear to me, that's why I was thinking something like the `upper_bound` of the interval would make more sense. Does that distinction make sense, or is "axis" really applicable to `IntervalMetadata` standalone? Note that this is similar to how positional metadata mixin defines `_positional_metadata_axis_len_`, and the `pd.DataFrame`'s "length" is the number of rows (the two don't directly correspond, since a data frame has two axes).

ebolyen: I suppose the name does kind of have "interval" in it, so `upper_bound` does make sense. But I think it can also be said that it *has* a continuous axis (which does not necessarily need to be bounded, but is in practice). I think any name has the potential to look weird in the context of how we decide to implement `iter` and friends, which is of course unknown.

I think I see this object as closer to a line with references on it, which makes me intuitively feel like the line comes first and the features second (i.e. not a mapping) which is why I like `axis_length`, and `length` feels even more natural. **BUT** that assumes something really important about the object (and makes `__len__` pretty pointless), so I'll defer to you on naming.

skbio/metadata/_mixin.py
RNAer: I was thinking only defining `__len__` and `__iter__`, not `__getitem__` and `__setitem__`. But what you said makes sense. Let's defer it. I will choose `num_interval_features` because it is one char shorter :)

skbio/metadata/_mixin.py
RNAer: ok. Let's settle on `upper_bound` then.

jairideout: Thanks for your input @ebolyen, that makes a lot of sense. The parameter name seems to depend on how we're conceptually thinking of the `IntervalMetadata`'s coordinate space (bounded vs unbounded). Given that I'm fine with either name. Let's go with `upper_bound`, @RNAer is the tie-breaker :)

skbio/metadata/_mixin.py
jairideout: Sounds good, thanks!

jairideout:
@RNAer, I think we've settled on the API changes to `IntervalMetadata`. I'll be on the lookout for that PR, the review should be pretty quick.

RNAer:
@mortonjt , could you take a pass thru this PR if you are bored at east coast? : )

@jairideout , this is ready for review, the test failed because the various places calls `has_interval_metadata` in `__repr__` in `Sequence` and its child classes. Adding interval metadata mixin to all the classes is gonna be a big PR. do you think it is ok? or there is good way to decompose it?

skbio/metadata/_testing.py
```diff
@@ -904,3 +905,166 @@ def test_has_positional_metadata(self):
...
+class IntervalMetadataMixinTests:
+    def _set_up(self):
+        self.upper_bound = 9
+        self.im = IntervalMetadata(self.upper_bound)
+        self.intvls = [
+            {'locations': [(0, 1)], 'metadata': {'gene': 'sagA'}},
+            # repeat it
+            {'locations': [(0, 1)], 'metadata': {'gene': ['a'],
+                                                 'product': 'foo'}}]
+
```
mortonjt: Would it be worthwhile to have a different interval other than (0, 1) or multiple intervals? It would be nice to double check if the equality works as anticipated.

mortonjt:
Concerning the failing tests,  I wouldn't merge this branch into master, but merge it into another branch,  so that the changes for Sequence and other subclasses can be reviewed separately

skbio/metadata/_mixin.py
```diff
@@ -406,3 +407,157 @@ def has_positional_metadata(self):
...
         return (self._positional_metadata is not None and
                 len(self.positional_metadata.columns) > 0)
+
+
+class IntervalMetadataMixin(metaclass=abc.ABCMeta):
+    @abc.abstractmethod
+    def _interval_metadata_axis_len_(self):
+        ''''''
+        raise NotImplementedError
+
```
RNAer: pls c.f. `PositionalMetadataMixin`. basically the child class inheriting the mixins handle that in its own polymorphismi way

RNAer:
Thanks, @mortonjt . @jairideout , do you think it is a good idea to PR agains a non-master branch, review code there, accumulate changes there, and merge all to master from there?

jairideout:
@RNAer, I like how you've been splitting up these PRs into smaller, atomic chunks that get merged in -- it helps a lot with reviewing and keeps master branch moving forward with incremental improvements. I'd prefer to keep using this strategy of merging smaller PRs into master branch when it's possible. If not, we can do what @mortonjt is suggesting by using a separate feature branch.

I'll follow up with an inline comment showing how we can fix the Sequence/TabularMSA repr issue so that those classes can be updated in a subsequent PR (it's an easy fix in this case).

skbio/metadata/_repr.py
```diff
@@ -126,6 +127,15 @@ def _format_positional_metadata_column(self, key, dtype):
         dtype_fmt = '<dtype: %s>' % str(dtype)
         return self._wrap_text_with_indent(dtype_fmt, key_fmt, 1)
 
+    def _process_interval_metadata(self):
+        if self._obj.has_interval_metadata():
```
jairideout: To work around the Sequence/TabularMSA repr issues, add this:

```python
if hasattr(self._obj, "has_interval_metadata") and self._obj.has_interval_metadata():
    ...
```

This is a little gross, but this repr class (`_MetadataReprBuilder`) already needs some refactoring so that classes can mix and match which types of metadata they have and need reprs for (`.metadata`, `.positional_metadata`, `.interval_metadata`). It probably makes sense to remove this class and move each of the metadata repr helpers into their respective metadata mixin classes (e.g. as `_repr_` methods). This is a larger change that's not super important to make (it's all private API) so I think using the `hasattr` workaround is a practical way to move forward here.

It's also fine if you want to remove the repr updates (and mixin doctests) from this PR, and add them in the upcoming Sequence PR. I think we'll still need the `hasattr` workaround.

skbio/metadata/_testing.py
```diff
@@ -904,3 +905,166 @@ def test_has_positional_metadata(self):
...
+    def test_eq_basic(self):
+        im1 = IntervalMetadata(self.upper_bound)
+        im1.add(**self.intvls[0])
+        obj1 = self._interval_metadata_constructor_(self.upper_bound, im1)
+
+        im2 = IntervalMetadata(self.upper_bound)
+        im2.add(**self.intvls[0])
+        obj2 = self._interval_metadata_constructor_(self.upper_bound, im2)
+
+        self.assertReallyEqual(obj1, obj2)
```
RNAer: @jairideout , I think we can ditch `skbio.util.ReallyEqualMixin` here, right? I read the webpage linked in its docstring (https://ludios.org/testing-your-eq-ne-cmp/). Basically, the justification for this is gone in Python 3. `__cmp__` is deprecated and the fallback of `__ne__` returns the negated result of` __eq__` in Python3.

jairideout: We should continue using `skbio.util._testing.ReallyEqualMixin` because it performs other equality sanity checks, such as making sure `A == B` and `B == A`, and `A != B` and `B != A`. This can catch subtle bugs in equality implementations. The mixin in scikit-bio doesn't include the `cmp` checks as those are no longer necessary in Python 3.

jairideout:
👍 for adding reverse complement tests. Also I noticed coverage dropped, can you run coverage locally and add tests as needed? I'll take a pass thru when you're done with that.

RNAer:
The reverse method is under `IntervalMetadata` not `IntervalMetadataMixin` - why adding a test here?

mortonjt:
This is mainly to make sure that `_axis_len` is being interacted with properly, since there is a significant amount of length tinkering in this PR.  Otherwise, I don't see any additional issues (outside of @jairideout 's comments) that need to be addressed.

jairideout:
I agree, 👍 for some basic reverse tests to make sure length is being passed thru correctly

RNAer:
While I added the reverse test, I still dont see this is needed. the reverse() is isolated in `IntervalMetadata` and there is no variable passing between it and anything else. Anyway.

@jairideout , the test coverage drop is mainly due to the remove of repr doctests. Now I removed repr modification and will add it back in next PR where the mixin will be added to `Sequence`. I think this is ready for you to take a pass.

jairideout:
Great, thanks @RNAer! I'll review tomorrow morning.

skbio/metadata/_mixin.py
```diff
@@ -406,3 +407,114 @@ def has_positional_metadata(self):
...
+            # Could use deleter but this is less overhead and needs to be fast.
+            self._interval_metadata = None
+        else:
+            # Use setter for validation and copy.
+            self.interval_metadata = interval_metadata
+
+    @property
+    @experimental(as_of="0.5.0-dev")
+    def interval_metadata(self):
+        '''``IntervalMetadata`` object containing info about interval features.
```
jairideout: Can you add a `Notes` section, similar to metadata/positional metadata? Something like:

```
Notes
-----
This property can be set and deleted. When setting new interval metadata, a
shallow copy of the `IntervalMetadata` object is made.
```

We'll need an `Examples` section too but will defer that until this class is hooked up to `Sequence` in another PR.

skbio/metadata/_mixin.py
```diff
@@ -406,3 +407,114 @@ def has_positional_metadata(self):
...
+            axis_len = self._interval_metadata_axis_len_()
+            if upper_bound != axis_len:
+                raise ValueError(
+                    'The upper bound for the interval features (%d) '
+                    'must match the interval metadata axis length (%d)'
+                    % (upper_bound, axis_len))
+            # copy all the data to the mixin
+            self._interval_metadata = copy.copy(interval_metadata)
+        else:
+            raise TypeError('You must provide `IntervalMetadata` object.')
```
jairideout: More informative error message to help users, something like:

```python
raise TypeError("Interval metadata must be an `IntervalMetadata` object, not type %r" % type(interval_metadata).__name__)
```

skbio/metadata/_mixin.py
```diff
@@ -406,3 +407,114 @@ def has_positional_metadata(self):
...
+            self._interval_metadata = IntervalMetadata(
+                self._interval_metadata_axis_len_())
+        return self._interval_metadata
+
+    @interval_metadata.setter
+    def interval_metadata(self, interval_metadata):
+        if isinstance(interval_metadata, IntervalMetadata):
+            upper_bound = interval_metadata.upper_bound
+            axis_len = self._interval_metadata_axis_len_()
+            if upper_bound != axis_len:
```
jairideout: Can you also check that `.lower_bound` is zero? Currently it will always be zero but that may not be true in the future, and we'd want to catch that here.

jairideout: There is an off-by-one error here. IntervalMetadata lower/upper bounds are inclusive, so this code forces the IntervalMetadata length to be one position greater than the object's axis length. For example, if you have a Sequence with 4 bases, it's axis length will be 4, so the IntervalMetadata's bounds should be `[0, 3]`. The check should be:

```python
if upper_bound != (axis_len - 1):
    ...
```

Can you add unit tests that would have caught this bug?

skbio/metadata/_testing.py
```diff
@@ -904,3 +905,245 @@ def test_has_positional_metadata(self):
...
+                                    'You must provide `IntervalMetadata` '
+                                    'object.'):
+            self._interval_metadata_constructor_(0, '')
+
+    def test_constructor_interval_metadata_len_mismatch(self):
+        for i in [0, 1, 3, 100]:
+            with self.assertRaisesRegex(ValueError, '\(9\).*\(%d\)' % i):
+                self._interval_metadata_constructor_(i, self.im)
+
+    def test_constructor_no_interval_metadata(self):
```
jairideout: You have tests for missing interval metadata but no tests for interval metadata that is present. Can you add the following test cases?

- test that a length-1 object can have interval metadata set thru the constructor (e.g. similar to `test_constructor_with_positional_metadata_len_one`). This is the base case (i.e. the smallest object that can have interval metadata)
- test that a length-greater-than-1 object can have interval metadata set thru the constructor (e.g. similar to `test_constructor_with_positional_metadata_len_greater_than_one`)

skbio/metadata/_testing.py
```diff
@@ -904,3 +905,245 @@ def test_has_positional_metadata(self):
...
+        for i, im in [(0, None), (self.upper_bound, self.im)]:
+            obj = self._interval_metadata_constructor_(i, im)
+            self.assertFalse(obj.has_interval_metadata())
+            self.assertIsInstance(obj.interval_metadata, IntervalMetadata)
+
+    def test_constructor_makes_shallow_copy_of_interval_metadata(self):
```
jairideout: Can you add a test case similar to `test_constructor_handles_missing_positional_metadata_efficiently`?

skbio/metadata/_testing.py
```diff
@@ -904,3 +905,245 @@ def test_has_positional_metadata(self):
...
+        obj2 = self._interval_metadata_constructor_(self.upper_bound)
+        obj2.interval_metadata.add(**self.intvls[0])
+
+        self.assertReallyEqual(obj1, obj2)
+
+    def test_eq_both_empty(self):
+        obj1 = self._interval_metadata_constructor_(self.upper_bound)
+        obj2 = self._interval_metadata_constructor_(self.upper_bound)
+        self.assertReallyEqual(obj1, obj2)
+
```
jairideout: add a test similar to `test_eq_handles_missing_positional_metadata_efficiently`

skbio/metadata/_testing.py
```diff
@@ -904,3 +905,245 @@ def test_has_positional_metadata(self):
...
+        im1.add(**self.intvls[0])
+        obj1 = self._interval_metadata_constructor_(self.upper_bound, im1)
+
+        im2 = IntervalMetadata(self.upper_bound)
+        im2.add(**self.intvls[1])
+        obj2 = self._interval_metadata_constructor_(self.upper_bound, im2)
+
+        self.assertReallyNotEqual(obj1, obj2)
+
+    def test_copy_interval_metadata_empty(self):
```
jairideout: add more assertions similar to those in `test_copy_positional_metadata_empty`

jairideout: also add a test case similar to `test_copy_positional_metadata_none`

skbio/metadata/_testing.py
```diff
@@ -904,3 +905,245 @@ def test_has_positional_metadata(self):
...
         self.assertTrue(obj.has_positional_metadata())
+
+
+class IntervalMetadataMixinTests:
+    def _set_up(self):
+        self.upper_bound = 9
+        self.im = IntervalMetadata(self.upper_bound)
+        self.intvls = [
+            {'locations': [(0, 1), (2, 9)], 'metadata': {'gene': 'sagA'}},
+            # repeat it
```
jairideout: outdated comment?

skbio/metadata/_testing.py
```diff
@@ -904,3 +905,245 @@ def test_has_positional_metadata(self):
...
+
+    def test_copy_interval_metadata_empty(self):
+        obj = self._interval_metadata_constructor_(self.upper_bound, self.im)
+        obj_copy = copy.copy(obj)
+
+        self.assertEqual(obj, obj_copy)
+        self.assertIsNot(obj, obj_copy)
+
+    def test_copy_interval_metadata(self):
+        self.im.add(**self.intvls[0])
```
jairideout: can you use `self.intvls[1]` instead so that the mutable list value in `.metadata` can be asserted to be the same object? that's where the behavior differs between shallow vs. deep copy

skbio/metadata/_testing.py
```diff
@@ -904,3 +905,245 @@ def test_has_positional_metadata(self):
...
+        self.assertIsNot(obj.interval_metadata._intervals,
+                         obj_copy.interval_metadata._intervals)
+        for i, j in zip(obj.interval_metadata._intervals,
+                        obj_copy.interval_metadata._intervals):
+            self.assertIsNot(i, j)
+            self.assertIsNot(i.metadata, j.metadata)
+            self.assertIsNot(i.metadata['gene'], j.metadata['gene'])
+            self.assertIs(i.metadata['product'], j.metadata['product'])
+
+    def test_deepcopy_interval_metadata_empty(self):
```
jairideout: add more assertions similar to those in `test_deepcopy_positional_metadata_empty`

jairideout: also add a test case similar to `test_deepcopy_positional_metadata_none`

skbio/metadata/_testing.py
```diff
@@ -904,3 +905,245 @@ def test_has_positional_metadata(self):
...
+        obj.interval_metadata._intervals[0].metadata['gene'] = 'sagB'
+        self.assertNotEqual(obj.interval_metadata, self.im)
+        self.im._intervals[0].metadata['gene'] = 'sagB'
+        self.assertEqual(obj.interval_metadata, self.im)
+
+        # Add new interval feature.
+        obj.interval_metadata.add(**self.intvls[1])
+        self.im.add(**self.intvls[1])
+        self.assertEqual(obj.interval_metadata, self.im)
+
```
jairideout: Can you add a test similar to `test_positional_metadata_getter_no_positional_metadata`?

skbio/metadata/_testing.py
```diff
@@ -904,3 +905,245 @@ def test_has_positional_metadata(self):
...
+        obj = self._interval_metadata_constructor_(self.upper_bound, self.im)
+
+        for i in [2, None, '', {}, []]:
+            with self.assertRaisesRegex(
+                    TypeError,
+                    'You must provide `IntervalMetadata` object'):
+                obj.interval_metadata = i
+
+        self.assertEqual(self.im, obj.interval_metadata)
+
```
jairideout: Can you add a test case similar to `test_positional_metadata_setter_len_mismatch`?

skbio/metadata/_testing.py
jairideout: Also add a test case that asserts the setter works with valid IntervalMetadata. It's being implicitly tested in some of these other test cases but it's good to have an explicit test for this, in case the other tests change in the future. e.g. similar to `test_positional_metadata_setter_data_frame`

skbio/metadata/_testing.py
```diff
@@ -904,3 +905,245 @@ def test_has_positional_metadata(self):
...
+        self.assertFalse(obj.has_interval_metadata())
+
+        obj = self._interval_metadata_constructor_(self.upper_bound, self.im)
+        self.assertFalse(obj.has_interval_metadata())
+
+        self.im.add([(0, 1)])
+        obj = self._interval_metadata_constructor_(self.upper_bound, self.im)
+        self.assertTrue(obj.has_interval_metadata())
+
+    def test_reverse(self):
```
jairideout: Sorry for requesting this test, it's unnecessary b/c we're already asserting that the IntervalMetadata object is constructed correctly in several different situations, and that the object's axis length is respected in the IntervalMetadata object. We're not testing specific methods on IntervalMetadata here, and `reverse` isn't a special case. Feel free to remove this test.

skbio/metadata/tests/test_mixin.py
```diff
@@ -76,5 +79,40 @@ def __deepcopy__(self, memo):
         self._positional_metadata_constructor_ = ExamplePositionalMetadataMixin
 
 
+class TestIntervalMetadataMixin(unittest.TestCase, ReallyEqualMixin,
+                                IntervalMetadataMixinTests):
+    def setUp(self):
+        super()._set_up()
```
jairideout: This is a little confusing -- can you just define a `setUp` method in `IntervalMetadataMixinTests`, and call `super().setUp` here? That's usually how I've seen this handled.

skbio/metadata/_mixin.py
RNAer: @jairideout , the current code is correct. The bounds are working like Python index. The `[(0, 4)]` means nucleotide from 0 to 3, not including 4. and the `upper_bound` should be 4 for a sequence of 4.

skbio/metadata/_testing.py
RNAer: this test doesn't make sense to me for either interval or positional metadata mixins, because even the `_init_` implementation uses inefficient deleter, this test will still pass. If it is for testing of empty interval metadata, this is tested in other tests. So I think thies is a wast code.

skbio/metadata/tests/test_mixin.py
RNAer: I view it the opposite way: note that `IntervalMetadataMixinTests` is not inherited from `TestCase`, so  naming it `setUp` may cause more confusion. More importantly, call `super().setUp` will call `TestCase.setUp()` rather than `IntervalMetadataMixinTests.setUp` if I name `_set_up` to `setUp`.

skbio/metadata/_testing.py
RNAer: I added this test anyway, in case you think otherwise. We can delete them in a future PR if I am correct.

skbio/metadata/_mixin.py
```diff
@@ -406,3 +407,114 @@ def has_positional_metadata(self):
...
+
+    @property
+    @experimental(as_of="0.5.0-dev")
+    def interval_metadata(self):
+        '''``IntervalMetadata`` object containing info about interval features.
+        '''
+        if self._interval_metadata is None:
+            # Not using setter to avoid copy.
+            self._interval_metadata = IntervalMetadata(
+                self._interval_metadata_axis_len_())
```
RNAer: I disagree on that. We shouldn't make this an exceptional case. Plus, positional metadata can have 0 rows. interval metadata should behave similarly. And this should be perfectly legal code (although it is meaningless):
```
im = IntervalMetadata(0)
im.add([(0, 0)])  # Again, behaving like Python index, this feature has zero length.
```

RNAer:
@jairideout , I replied on a few of your comments. Other than those, all are addressed. could you comment and take another pass? Thanks!

skbio/metadata/_mixin.py
jairideout: I am pretty certain there is an off-by-one error. IntervalMetadata lower/upper bounds are inclusive, so having a lower bound of 0 and an upper bound of 4 (in this example) is five positions, where the number of positions in the Sequence is 4. Here is an example using the `ExampleIntervalMetadataMixin` test object:

```python
In [1]: from skbio.metadata._mixin import IntervalMetadataMixin
In [2]: from skbio.metadata import IntervalMetadata
In [3]: from skbio.util._decorator import overrides
In [4]: # copy and paste the ExampleIntervalMetadataMixin class definition
        # from the unit tests, not included here for brevity

# Create an object of length 4 (i.e. 4 positions). The current code requires
# its IntervalMetadata to have an upper bound of 4, which is inclusive.
In [5]: e = ExampleIntervalMetadataMixin(4, IntervalMetadata(4))

# Confirm the object's axis length is 4 (i.e. 4 positions).
In [6]: e._interval_metadata_axis_len_()
Out[6]: 4

# The IntervalMetadata's lower bound is 0 (inclusive).
In [7]: e.interval_metadata.lower_bound
Out[7]: 0

# The IntervalMetadata's upper bound is 4 (inclusive).
In [8]: e.interval_metadata.upper_bound
Out[8]: 4

# Confirm IntervalMetadata's bounds are inclusive by adding metadata
# that includes the lower and upper bounds (0 and 4, respectively).
In [9]: e.interval_metadata.add(locations=[(0, 4)], metadata={'foo': 42})
Out[9]: Interval(interval_metadata=<4592643040>, locations=[(0, 4)], boundaries=[(True, True)], metadata={'foo': 42})

# Confirm that IntervalMetadata disallows an interval with an upper bound
# of 5.
In [10]: e.interval_metadata.add(locations=[(0, 5)], metadata={'bar': 43})
<ipython-input-10-2867edfcc7cf> in <module>()
----> 1 e.interval_metadata.add(locations=[(0, 5)], metadata={'bar': 43})

/Users/jairideout/dev/scikit-bio/skbio/metadata/_interval.py in add(self, locations, boundaries, metadata)
    588                         locations=locations,
    589                         boundaries=boundaries,
--> 590                         metadata=metadata)
    591
    592     @_rebuild_tree

/Users/jairideout/dev/scikit-bio/skbio/metadata/_interval.py in __init__(self, interval_metadata, locations, boundaries, metadata)
     80         self._interval_metadata = interval_metadata
     81
---> 82         self._locations_boundaries_setter(locations, boundaries)
     83
     84         # Metadata

/Users/jairideout/dev/scikit-bio/skbio/metadata/_interval.py in _locations_boundaries_setter(self, locations, boundaries)
    228             if boundaries is None:
    229                 locations.sort()
--> 230                 self._check_bounds(locations)
    231                 self._locations = locations
    232                 # reset all the boundaries to True!!

/Users/jairideout/dev/scikit-bio/skbio/metadata/_interval.py in _check_bounds(self, locations)
    252                              'larger than upper bound (%r) or smaller than '
    253                              'lower bound (%r).' %
--> 254                              (locations, upper_bound, lower_bound))
    255
    256     @property

ValueError: Cannot set `locations` ([(0, 5)]) with coordinate larger than upper bound (4) or smaller than lower bound (0).

In [11]:
```

jairideout: I understand the source of my confusion. `IntervalMetadata` doesn't explicitly define the upper bounds of `locations` or `upper_bound` itself to be inclusive or exclusive, and from the API/usage of the object it isn't clear either. I understood upper bounds to be inclusive for two reasons:

1) The constructor parameter `upper_bound` is documented as:

```
upper_bound : int
        Defines the upper bound of the interval features. No coordinate can
        be greater than it.
```

"No coordinate can be greater than it" implies (to me at least) that a coordinate equal to the upper bound is acceptable.

2) An interval's range(s) are specified as `locations`. `locations` implies coordinates/positions in the coordinate space, so it wasn't clear that the upper bound of a location follows Python's interpretation of slices and ranges.

I understand now that upper bounds are *always* exclusive, and lower bounds are always inclusive, in order to match Python's notion of slices/ranges. There isn't an off-by-one error given this interpretation of bounds.

I can see this potentially causing confusion with users (for example, I've played with the object and reviewed the code, and had an entirely different interpretation of the upper bound). @ebolyen and I talked about this and think it is important at this stage to make the following changes to `IntervalMetadata`/`Interval` in a separate PR in order to make this clearer:

1) Change all occurrences of `locations` to `bounds`. `locations` implies specific positions in the coordinate space, while the name `bounds` is more akin to Python's notion of slices and ranges, which has lower and (exclusive) upper bounds.

2) Change all occurrences of `boundaries` to `fuzzy`. `boundaries` could be misinterpreted to mean inclusive/exclusive for the bounds in `locations`, but it means something entirely different (unsure where the start and/or end really is). "fuzzy" matches Biopython's nomenclature too for their equivalent data structure. "fuzzy" is also a distinctive name that won't get confused with defining inclusive/exclusive bounds, and most users will see the name and ignore it (which covers most use-cases). When `fuzzy` is used, the booleans should be inverted, such that `True` means a lower/upper bound is "fuzzy", and `False` means we have a non-fuzzy (precise) understanding of the lower/upper bound.

3) With inclusive lower bound and exclusive upper bound, an interval with the same start and end (e.g. `[1, 1)`) doesn't represent a position in the coordinate space. These types of intervals should be disallowed to avoid potential confusion (and it isn't clear what exactly it means in the context of interval metadata). If a user wants an interval spanning a single position, use `[1, 2)`.

4) Document `bounds` to always be inclusive lower and exclusive upper. Document `upper_bound` in a similar way.

I think these changes will help make the object more easily understood by users. A couple of my comments on this PR don't apply given the interpretation of upper bounds, I'll follow up on those now.

skbio/metadata/_testing.py
jairideout: This test is important because it asserts `__eq__` doesn't turn an internal `None` representation of interval metadata into an empty `IntervalMetadata` object. The test also makes sense for `metadata` and `positional_metadata`. If the `__eq__` method had been implemented like you had it before (i.e. without the `has_interval_metadata` guards), this test would fail. Basically we don't want to turn `None` into an empty `IntervalMetadata` object for efficiency. If you don't use `has_interval_metadata` to guard that and just use the getter, it will turn `None` into an empty `IntervalMetadata` object.

This has a very real impact on performance when creating many of these objects (e.g. reading a FASTQ file into Sequence objects). We tried removing the `None` optimization and there was a huge degradation in performance. Thus, it is important to test it here and in the other places I requested.

skbio/metadata/tests/test_mixin.py
jairideout: That makes sense, thanks for clarifying why it is called `_set_up`.

skbio/metadata/_mixin.py
jairideout: My comment here no longer makes sense given the interpretation of exclusive upper bounds. It's great that an empty object (e.g. Sequence) can have empty positional and interval metadata!

I don't think we should allow start/end bounds that are the same (e.g. `(0, 0)`) because it doesn't represent any place in the coordinate space and users won't be able to query one of these intervals with `.query(locations=[(0,0)]`. Seems clearer to disallow these types of intervals entirely. It also matches how positional metadata works -- you can't have metadata attached to an "empty" position.

jairideout:
@RNAer, I replied to your comments and I think we've cleared up the confusion with inclusive vs. exclusive upper bounds. I made a comment about issuing a separate PR to make this clearer in `IntervalMetadata`/`Interval`, I'll review/merge that one when it's ready and we can come back to this PR (which is close to merge).

skbio/metadata/_mixin.py
RNAer: I can imagine the confusion given current doc and naming. will change it as you request here.

jairideout: Great, thanks @RNAer!

mortonjt:
👍 @jairideout what do you think?

jairideout:
Looks good, thanks for making those changes @RNAer!

RNAer:
yay!

mortonjt:
🍻 Thanks guys for getting this in!

Issue #702, opened 2014-10-13
Windows support


quiqua:
Hey there!
Did you successfully compile the cython extension on windows?
I'm using Python 2.7 running in a virtual Windows 7 32 bit machine, inlcuded a `"stdint.h"` file like the Pandas project does ([like here](https://github.com/pydata/pandas/tree/master/pandas/src/headers))(since python 2.7. relies on MSVC 2008 and this version does not include this header). 
Yet there seems to be a problem with the `__m128i` type provided by `<emmintrin.h>`
```
...
skbio/alignment/_lib/ssw.c(196) : error C2275: '__m128i' : illegal use of this type as an expression
            C:\Program Files\Microsoft Visual Studio 9.0\VC\INCLUDE\emmintrin.h(45) : see declaration of '__m128i'
```

mattfullerton:
For those wanting to play around with a source tree that does actually compile (pip install -e <folderpath>, you can find one here: https://github.com/mattfullerton/scikit-bio/commit/731b133db87867e4f2ca09e787991ca1fb194d30

I had to do a lot of deciphering here and in #780 to figure out what actually needs changing (that pull request in and of itself does not fix the compiling problem). I have no reason to believe this compile will be any better than the reported binaries that don't work, but who knows. Hence, I'm not submitting a pull request right now.

ebolyen:
Oops! Trying to comment. Thanks for the exploration @mattfullerton! This is really cool. Does anyone have access to a i686 machine to make sure that this doesn't also break that build?

anderspitman:
We discussed setting up a windows VM on powertrip a while back, but that sort of got pushed to the side. I can open an issue for that if it's still desirable.

ebolyen:
:+1: I think we do need something like that if we are ever going to really fix this.

jairideout:
:+1:

methodds:
I just tried @mattfullerton 's windows version and while compiling worked just fine I cannot import skbio due to the error:

````
Python 3.5.0 |Anaconda 2.4.0 (64-bit)| (default, Oct 20 2015, 07:26:33) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import skbio
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #1424, opened 2016-09-07
Fix compiler error on Windows

hotdogee:
I was able to build a bdist_wheel on Win10 64 and it didn't crash when trying to use the ssw extension (https://github.com/biocore/scikit-bio/issues/702#issuecomment-60054711).
```
Python 3.5.2 (v3.5.2:4def2a2901a5, Jun 25 2016, 22:18:55) [MSC v.1900 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import skbio
>>> from skbio.alignment._ssw_wrapper import StripedSmithWaterman as ssw
>>> x = ssw('AATTTNGNTTTANTTNGTA')
>>> x('AATTGNTNANTNGNTAATNGT')
{
    'optimal_alignment_score': 10,
    'suboptimal_alignment_score': 0,
    'query_begin': 9,
    'query_end': 18,
    'target_begin': 6,
    'target_end_optimal': 15,
    'target_end_suboptimal': 0,
    'cigar': '10M',
    'query_sequence': 'AATTTNGNTTTANTTNGTA',
    'target_sequence': 'AATTGNTNANTNGNTAATNGT'
}
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,6 +64,8 @@ ext = '.pyx' if USE_CYTHON else '.c'
 # details. This acts as a workaround until the next Python 3 release -- thanks
 # Wolfgang Maier (wolma) for the workaround!
 ssw_extra_compile_args = ['-Wno-error=declaration-after-statement']
+if sys.platform == 'win32':
+    ssw_extra_compile_args = []
 
 # Users with i686 architectures have reported that adding this flag allows
 # SSW to be compiled. See https://github.com/biocore/scikit-bio/issues/409 and


RNAer:
@hotdogee , did all the tests pass on windows? For continuous integration test, we should also set it up on AppVeyor.

Just FYI, @hotdogee is a collaborator who are excited to contribute. 

@ebolyen , there is no opposition to have Windows support for skbio, right?

jairideout:
@hotdogee thanks for contributing this! It's a minimal change so I think this is safe to merge. scikit-bio still won't officially support Windows because we'd need all tests passing and CI set up with AppVeyor, but this gets us moving in the right direction.

Note that we have official Windows support on the horizon, but when that will happen is unclear (perhaps early 2017). We are planning to remove SSW in favor of a different alignment API that supports cross-platform compatible aligners. SSW has been the culprit of other platform-specific issues, not always related to Windows, so we are in favor of dropping it as a dependency. This should make it much easier to support Windows.

We currently do not have expertise in Windows development. Is this something you'd be interested in taking the lead on?

Issue #1309, opened 2016-03-18
include subpackages in __init__.py files


jairideout:
Importing a module should have its sub-modules/packages available. For example, the following paths should be accessible:

```python
import skbio
skbio.stats.ordination
skbio.diversity
...
```

These examples currently fail with an `AttributeError`.

To fix this, sub-modules/packages need to be imported in the package's `__init_.py` and added to `__all__`.

See #1258 and #1307 for original discussion.

jolespin:
Is this fixed on the current stable version? If not, is there a way around it atm? 

```
$ conda install -c https://conda.anaconda.org/biocore scikit-bio
Fetching package metadata .................
Solving package specifications: .

# All requested packages already installed.
# packages in environment at /Users/jespinoz/anaconda:
#
scikit-bio                0.4.2               np112py36_0    bioconda
jespinozlt-osx:pt_I jespinoz$ python
Python 3.6.1 |Anaconda custom (x86_64)| (default, Mar 22 2017, 19:25:17)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.57)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from skbio import stats
>>> stats.ordination
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'skbio.stats' has no attribute 'ordination'
```

jairideout:
The fix is available in scikit-bio 0.5.1 (it was merged after 0.5.0 release went out).

jolespin:
`conda install -c https://conda.anaconda.org/biocore scikit-bio` installed 0.4.2.  Is there a way to directly install 0.5.1 via conda or pip?

jairideout:
scikit-bio 0.5.0 and later are only compatible with Python 3. Is your conda environment using Python 2?

jolespin:
I'm using Python 3.6.1 but I think it might be an issue with PyQT5 because my ete3 and ScikitBio broke at the same time when I updated some other modules.

jairideout:
Can you try creating a new environment with the latest version of scikit-bio? I tested this locally and I get Python 3.5.3 and scikit-bio 0.5.1 (that's the latest release):

```shell
conda create -n scikit-bio -c https://conda.anaconda.org/biocore scikit-bio
```

jolespin:
[...]

PR #1425, opened 2016-09-07
Include subpackages in __init__.py #1309

hotdogee:
Fixes https://github.com/biocore/scikit-bio/issues/1309. the following example no longer fails with an `AttributeError`, all other sub-modules are already accessible:
```
import skbio
skbio.diversity
skbio.stats.evolve
```
Not sure if we want to add to `__all__`, since wildcard imports ( from skbio import * ) should be avoided (PEP8).


jairideout:
Thanks @hotdogee!

PR #1430, opened 2016-09-28
Hook ``IntervalMetadataMixin`` to `Sequence` and `GrammaredSequence` classes.

RNAer:
This is ready for review, @mortonjt and @wasade . Thanks!

### DIFF
--- a/skbio/metadata/_interval.py
+++ b/skbio/metadata/_interval.py
@@ -11,7 +11,7 @@ import copy
 import functools
 
 from ._intersection import IntervalTree
-from skbio.util._decorator import experimental
+from skbio.util._decorator import experimental, classonlymethod
 
 
 class Interval:
@@ -558,6 +558,80 @@ fuzzy=[(False, False)], metadata={'gene': 'sagB'})
 
         # DONT' forget this!!!
         self._is_stale_tree = True
+
+    @classonlymethod
+    @experimental(as_of="0.5.0-dev")
+    def concat(cls, interval_metadata):
+        '''Concatenate an iterable of ``IntervalMetadata`` objects.
+
+        It concatenates the multiple ``IntervalMetadata`` objects into
+        one coordinate space. The order of the objects in the input
+        iterable matters. The coordinate of the second
+        ``InterableMetadata`` will be shifted up with the length of
+        the first ``IntervalMetadata`` object.
+
+        This function is useful when you concatenate multiple sequences.
+
+        Parameters
+        ----------
+        interval_metadata : Iterable (IntervalMetadata)
+            The interval metadata to concatenate.
+
+        Returns
+        -------
+        IntervalMetadata
+            Concatenated interval metadata.
+
+        Examples
+        --------
+        >>> from skbio.metadata import IntervalMetadata
+
+        Create two ``IntervalMetadata`` objects:
+
+        >>> im1 = IntervalMetadata(3)
+        >>> _ = im1.add([(0, 2)], [(True, False)], {'gene': 'sagA'})
+        >>> im2 = IntervalMetadata(4)
+        >>> _ = im2.add([(1, 4)], [(True, True)], {'gene': 'sagB'})
+
+        Concatenate them into a single coordinate space. The second
+        ``IntervalMetadata``'s interval features are all shifted
+        up. The resulting ``IntervalMetadata``'s upper bound is the
+        sum of upper bounds of concatenated objects:
+
+        >>> im = IntervalMetadata.concat([im1, im2])
+        >>> im   # doctest: +ELLIPSIS
+        2 interval features
+        -------------------
+        Interval(interval_metadata=<...>, bounds=[(0, 2)], \
+fuzzy=[(True, False)], metadata={'gene': 'sagA'})
+        Interval(interval_metadata=<...>, bounds=[(4, 7)], \
+fuzzy=[(True, True)], metadata={'gene': 'sagB'})
+        >>> im.upper_bound
+        7
+
+        '''
+        interval_metadata = list(interval_metadata)
+
+        if len(interval_metadata) == 0:
+            return cls(0)
+
+        upper_bound = 0
+        for im in interval_metadata:
+            upper_bound += im.upper_bound
+        new = cls(upper_bound)
+
+        length = 0
+        for i, im in enumerate(interval_metadata):
+            for intvl in im._intervals:
+                bounds = intvl.bounds
+                fuzzy = intvl.fuzzy
+                if i != 0:
+                    bounds = [(start + length, end + length)
+                              for start, end in bounds]
+                new.add(bounds, fuzzy, intvl.metadata)
+            length += im.upper_bound
+
+        return new
 
     @experimental(as_of='0.5.0-dev')
     def sort(self, ascending=True):
--- a/skbio/metadata/_repr.py
+++ b/skbio/metadata/_repr.py
@@ -15,8 +15,8 @@ from skbio._base import ElasticLines
 
 
 class _MetadataReprBuilder(metaclass=ABCMeta):
-    """Abstract base class for building  a repr for an object containing
-    metadata and/or positional metadata.
+    """Abstract base class for building a repr for an object containing
+    metadata, positional metadata, and/or interval metadata.
 
     Parameters
     ----------
@@ -50,6 +50,7 @@ class _MetadataReprBuilder(metaclass=ABCMeta):
         self._process_header()
         self._process_metadata()
         self._process_positional_metadata()
+        self._process_interval_metadata()
         self._process_stats()
         self._process_data()
 
@@ -125,6 +126,16 @@ class _MetadataReprBuilder(metaclass=ABCMeta):
         key_fmt = self._format_key(key)
         dtype_fmt = '<dtype: %s>' % str(dtype)
         return self._wrap_text_with_indent(dtype_fmt, key_fmt, 1)
+
+    def _process_interval_metadata(self):
+        if (hasattr(self._obj, "has_interval_metadata") and
+           self._obj.has_interval_metadata()):
+            self._lines.add_line('Interval metadata:')
+            n = self._obj.interval_metadata.num_interval_features
+            line = self._indent + '%d interval feature' % n
+            if n > 1:
+                line += 's'
+            self._lines.add_line(line)
 
     def _format_key(self, key):
         """Format metadata key.
--- a/skbio/metadata/_repr.py
+++ b/skbio/metadata/_repr.py
@@ -15,8 +15,11 @@ from skbio._base import ElasticLines
 
 
 class _MetadataReprBuilder(metaclass=ABCMeta):
-    """Abstract base class for building a repr for an object containing
-    metadata, positional metadata, and/or interval metadata.
+    """An ABC for building a repr from an object containing metadata.
+
+    This abstract base class constructs a repr string for an
+    object which contains metadata, positional metadata and/or
+    interval metadata.
 
     Parameters
     ----------
@@ -26,6 +29,7 @@ class _MetadataReprBuilder(metaclass=ABCMeta):
         Maximum width of the repr.
     indent : int
         Number of spaces to use for indented lines.
+
     """
     def __init__(self, obj, width, indent):
         self._obj = obj
@@ -128,6 +132,8 @@ class _MetadataReprBuilder(metaclass=ABCMeta):
         return self._wrap_text_with_indent(dtype_fmt, key_fmt, 1)
 
     def _process_interval_metadata(self):
+        # TODO: this hasattr check can be removed once all the relevant
+        # classes have interval_metadata added to it.
         if (hasattr(self._obj, "has_interval_metadata") and
            self._obj.has_interval_metadata()):
             self._lines.add_line('Interval metadata:')
--- a/skbio/metadata/tests/test_interval.py
+++ b/skbio/metadata/tests/test_interval.py
@@ -468,6 +468,34 @@ class TestIntervalMetadata(unittest.TestCase, ReallyEqualMixin):
         self.assertEqual(len(intvls), 1)
         self.assertTrue(intvl is intvls[0])
 
+    def test_concat_empty(self):
+        for i in 0, 1, 2:
+            obs = IntervalMetadata.concat([self.im_empty] * i)
+            exp = IntervalMetadata(self.upper_bound * i)
+            self.assertEqual(obs, exp)
+
+        obs = IntervalMetadata.concat([])
+        self.assertEqual(obs, IntervalMetadata(0))
+
+    def test_concat(self):
+        im1 = IntervalMetadata(3)
+        im2 = IntervalMetadata(4)
+        im3 = IntervalMetadata(5)
+        im1.add([(0, 2)], [(True, True)])
+        im2.add([(0, 3)], [(True, False)], {'gene': 'sagA'})
+        im2.add([(2, 4)], metadata={'gene': 'sagB'})
+        im3.add([(1, 5)], [(False, True)], {'gene': 'sagC'})
+        obs = IntervalMetadata.concat([im1, im2, im3])
+
+        exp = IntervalMetadata(12)
+        exp.add(bounds=[(0, 2)], fuzzy=[(True, True)])
+        exp.add(bounds=[(3, 6)], fuzzy=[(True, False)],
+                metadata={'gene': 'sagA'})
+        exp.add(bounds=[(5, 7)], metadata={'gene': 'sagB'})
+        exp.add(bounds=[(8, 12)], fuzzy=[(False, True)],
+                metadata={'gene': 'sagC'})
+        self.assertEqual(obs, exp)
+
     def test_sort(self):
         interval = Interval(
             self.im_2,
--- a/skbio/sequence/_dna.py
+++ b/skbio/sequence/_dna.py
@@ -29,6 +29,10 @@ class DNA(GrammaredSequence, NucleotideMixin,
         Arbitrary per-character metadata. For example, quality data from
         sequencing reads. Must be able to be passed directly to the Pandas
         DataFrame constructor.
+    interval_metadata : IntervalMetadata
+        Arbitrary interval metadata which applies to intervals within
+        a sequence to store interval features (such as genes on the
+        DNA sequence).
     lowercase : bool or str, optional
         If ``True``, lowercase sequence characters will be converted to
         uppercase characters in order to be valid IUPAC DNA characters. If
@@ -52,6 +56,7 @@ class DNA(GrammaredSequence, NucleotideMixin,
     values
     metadata
     positional_metadata
+    interval_metadata
     alphabet
     gap_chars
     default_gap_char
@@ -169,8 +174,8 @@ class DNA(GrammaredSequence, NucleotideMixin,
 
         Notes
         -----
-        DNA sequence's metadata and positional metadata are included in the
-        transcribed RNA sequence.
+        DNA sequence's metadata, positional, and interval
+        metadata are included in the transcribed RNA sequence.
 
         Examples
         --------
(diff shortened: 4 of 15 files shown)


skbio/metadata/_interval.py
```diff
@@ -559,6 +559,58 @@ def _reverse(self):
...
 
+    @classonlymethod
+    @experimental(as_of="0.5.0-dev")
+    def concat(cls, intvl_metadata):
+        '''Concatenate an iterable of ``IntervalMetadata`` objects.
+
+        Parameters
+        ----------
+        intvl_metadata : Iterable (IntervalMetadata)
+
```
wasade: Possible to add one-liners to the parameters and returns? eg: "The interval metadata to concatenate"

skbio/metadata/_interval.py
```diff
@@ -559,6 +559,58 @@ def _reverse(self):
...
+            upper_bound += im.upper_bound
+        new = cls(upper_bound)
+
+        length = 0
+        for i, im in enumerate(intvl_metadata):
+            for intvl in im._intervals:
+                bounds = intvl.bounds
+                fuzzy = intvl.fuzzy
+                if i != 0:
+                    bounds = [(start + length, end + length)
```
wasade: perhaps there's something critical I don't understand, but why do the bounds shift?

skbio/metadata/_repr.py
```diff
@@ -15,8 +15,8 @@
 
 
 class _MetadataReprBuilder(metaclass=ABCMeta):
-    """Abstract base class for building  a repr for an object containing
-    metadata and/or positional metadata.
+    """Abstract base class for building a repr for an object containing
```
wasade: Can this be a single line please?

skbio/metadata/_repr.py
```diff
@@ -126,6 +127,16 @@ def _format_positional_metadata_column(self, key, dtype):
         dtype_fmt = '<dtype: %s>' % str(dtype)
         return self._wrap_text_with_indent(dtype_fmt, key_fmt, 1)
 
+    def _process_interval_metadata(self):
+        if (hasattr(self._obj, "has_interval_metadata") and
```
wasade: is there a valid situation when the obj would not have the `has_interval_metadata` attr?

skbio/metadata/tests/test_interval.py
```diff
@@ -468,6 +468,31 @@ def test_duplicate_bounds(self):
         self.assertEqual(len(intvls), 1)
         self.assertTrue(intvl is intvls[0])
 
+    def test_concat_empty(self):
+        for i in 0, 1, 2:
+            obs = IntervalMetadata.concat([self.im_empty] * i)
+            exp = IntervalMetadata(self.upper_bound * i)
+            self.assertEqual(obs, exp)
+
+    def test_concat(self):
```
wasade: it doesn't seem like intervals can overlap when being concatenated?

skbio/sequence/_dna.py
```diff
@@ -212,9 +217,14 @@ def transcribe(self):
         if self.has_positional_metadata():
             positional_metadata = self.positional_metadata
 
+        interval_metadata = None
```
wasade: This method is marked as `stable`. These changes are not at the API level but can result in side effects. Just marking as I do not know if this is or is not a concern from the perspective of the API decorators

skbio/sequence/_sequence.py
```diff
@@ -517,7 +563,7 @@ def concat(cls, sequences, how='strict'):
                                  " how='inner' or how='outer'")
         seq_data = []
         pm_data = []
-        for seq in seqs:
+        for i, seq in enumerate(seqs):
```
wasade: is `enumerate` needed here?

skbio/sequence/_sequence.py
```diff
@@ -538,9 +586,13 @@ def _assert_can_cast_to(cls, target):
...
         return len(self)
 
+    @overrides(IntervalMetadataMixin)
+    def _interval_metadata_axis_len_(self):
+        return len(self)
+
     @stable(as_of="0.4.0")
     def __init__(self, sequence, metadata=None, positional_metadata=None,
-                 lowercase=False):
+                 interval_metadata=None, lowercase=False):
```
wasade: Similar question here with respect to `stable`. This is an API change, although it is an expansion in which the default behavior should remain the same (as far as I can tell) in that a valid API call against 0.4.0 is still valid here for parameters and the returned instance.

skbio/sequence/_sequence.py
```diff
@@ -1088,8 +1155,10 @@ def __repr__(self):
         ...     'quality': [3, 10, 11, 10],
         ...     'exons': [True, True, False, True]
         ... }
-        >>> Sequence('ACGT', metadata=metadata,
+        >>> seq = Sequence('ACGT', metadata=metadata,
         ...          positional_metadata=positional_metadata)
+        >>> _ = seq.interval_metadata.add([(0, 2)], metadata={'gene': 'sagA'})
```
wasade: I think the `_ =` could be removed and it would just `repr`, right?

skbio/sequence/_sequence.py
```diff
@@ -1230,21 +1303,55 @@ def copy(self, deep=False):
...
+fuzzy=[(False, False)], metadata={'gene': ['BCC7']})
+
         Since only a *shallow* copy was made, updates to mutable objects stored
         as metadata affect the original sequence's metadata:
 
+        >>> seq_copy = seq.copy()
         >>> seq_copy.metadata['authors'].append('Bob')
         >>> pprint(seq_copy.metadata)
-        {'authors': ['Alice', 'Bob'], 'id': 'new-id'}
+        {'authors': ['Alice', 'Bob'], 'id': 'seq-id'}
```
wasade: why did the value associated with `id` change?

skbio/sequence/_sequence.py
```diff
@@ -1302,16 +1432,21 @@ def _copy(self, deep, memo):
...
         seq_copy = self._constructor(sequence=bytes_, metadata=None,
-                                     positional_metadata=None)
+                                     positional_metadata=None,
+                                     interval_metadata=None)
 
         if deep:
             seq_copy._metadata = MetadataMixin._deepcopy_(self, memo)
             seq_copy._positional_metadata = \
                 PositionalMetadataMixin._deepcopy_(self, memo)
+            seq_copy._interval_metadata = IntervalMetadataMixin._deepcopy_(
```
wasade: not `__deepcopy__`? See notes at the bottom [here](https://docs.python.org/3/library/copy.html#copy.deepcopy). Similar question w.r.t. `__copy__` below

skbio/metadata/_interval.py
RNAer: it is concatenating the coordinate space, rather than just putting all the interval features together. I will explain it more clearly in docstring

skbio/sequence/_sequence.py
RNAer: because it is a new copy, not the copy that was modified. pls see the change in the doctest.

skbio/sequence/_sequence.py
RNAer: yes. just to avoid doctest fail or adding the `repr` print here. I can use `# noqa` to skip doctest

skbio/sequence/_sequence.py
RNAer: good question. not sure either.

skbio/metadata/_repr.py
RNAer: need to cut the text. will this cause problem?

skbio/metadata/_repr.py
RNAer: yes. For the classes that haven't been modified to adapt interval metadata, eg `TabularMSA`, some of their tests would fail if this is not checked.

skbio/metadata/_interval.py
wasade: Okay, thanks. So it isn't possible to concatenate overlapping intervals?

skbio/metadata/_repr.py
wasade: What about:

```python
class _MetadataReprBuilder(metaclass=ABCMeta):
    """An ABC for building a repr from an object containing metadata

    This abstract baseclass is will construct a repr string for an object which contains metadata, positional metadata and/or interval metadata.

    ...snip...
    """

File: skbio/sequence/_sequence.py (hunk 10 above)
Comment (wasade) [inline skbio/sequence/_sequence.py]: i see it now, thanks and sorry for not looking closer

Comment (RNAer) [inline skbio/sequence/_sequence.py]: no worries

**Comment by RNAer** [comment]
You can always add overlapping interval features into a interval metadata
object. I think we are still talking on different pages. Hope this
docstring I will add helps clarify your confusion?

```
        Concatenate an iterable of ``IntervalMetadata`` objects.

        It concatenates the multiple ``IntervalMetadata`` objects into
        one coordinate space. The order of the objects in the input
        iterable matters. The coordinate of the second
        ``InterableMetadata`` will be shifted up with the length of
        the first ``IntervalMetadata`` object.

        This function is useful when you concatenate multiple sequences.
```

File: skbio/metadata/_interval.py (hunk 1 above)
Comment (mortonjt) [inline skbio/metadata/_interval.py]: 👍 If you could add some clarification (i.e. an example of concatenating 2 difference sequence objects), that would also help.

File: skbio/metadata/_interval.py (hunk 2 above)
Comment (mortonjt) [inline skbio/metadata/_interval.py]: It should be.  You can have intervals that do overlap.

**Comment by wasade** [comment]
Oh, I see now re: concatenation, that makes sense especially for MSAs and concatenating gene family MSAs

File: skbio/metadata/_interval.py [hunk 12]
```diff
@@ -559,6 +559,68 @@ def _reverse(self):
...
+        -------
+        IntervalMetadata
+            Concatenated interval metadata.
+
+        Examples
+        --------
+        >>> from skbio.metadata import IntervalMetadata
+        >>> im1 = IntervalMetadata(3)
+        >>> _ = im1.add([(0, 2)], [(True, False)], {'gene': 'sagA'})
+        >>> im2 = IntervalMetadata(4)
```
mortonjt: If we could have some commentary on what exactly is going on here could help improve readability of this example

skbio/sequence/_sequence.py
mortonjt: 👍 on repr

skbio/metadata/_repr.py
```diff
@@ -15,8 +15,11 @@
 
 
 class _MetadataReprBuilder(metaclass=ABCMeta):
-    """Abstract base class for building  a repr for an object containing
-    metadata and/or positional metadata.
+    """An ABC for building a repr from an object containing metadata.
+
+    This abstract baseclass is will construct a repr string for an
```
mortonjt: `is will construct` -> `constructs`

skbio/sequence/_sequence.py
mortonjt: Because there is some additional logic required for deepcopy.  See [here](https://github.com/biocore/scikit-bio/blob/master/skbio/metadata/_mixin.py#L140)

mortonjt: Same with [copy](https://github.com/biocore/scikit-bio/blob/master/skbio/metadata/_mixin.py#L130)

wasade: I guess I don't understand the need for the indirection as `copy.deepcopy(None) -> None`

skbio/metadata/_interval.py
wasade: What's the method for merging or concatenating (unsure of the best verb here) for overlapping intervals?

skbio/sequence/_sequence.py
wasade: @gregcaporaso @jairideout @ebolyen, flagging this comment for comment :) Basically, we're unsure about subtleties of the implications of API stability

skbio/metadata/_repr.py
ebolyen: :+1: for a TODO

skbio/metadata/tests/test_interval.py
mortonjt: That is correct, since the actual coordinate spaces are being concatenated together.

for example,  if `len(seq1)=10` and `len(seq2)=20`, then `len(concat((seq1, seq2)))==30`.
So by default, the coordinates should not be overlapping, unless negative coordinates, or coordinates larger than the sequence length are allowed.

skbio/sequence/_sequence.py
mortonjt: Its the way that the API is setup at the moment.  Its because the metadata objects have their own copy methods.  So a check needs to be done to see if the metadata exists in the first place, before calling the copy command.  I don't have a strong reference for replacing this, though keep in mind that will require refactoring a bit of legacy code.

@ebolyen do you think it is worthwhile to do that refactor here?  Or is it better to submit an issue and address this down the road?

mortonjt:
It looks like there are a few minor comments bringing brought to attention here, suggesting that this PR is getting very close to being merge-able (thanks @RNAer).  @ebolyen @jairideout would you like to take a pass at this?

skbio/sequence/_sequence.py
```diff
@@ -1230,21 +1303,55 @@ def copy(self, deep=False):
...
+        >>> seq.interval_metadata  # doctest: +ELLIPSIS
+        1 interval feature
+        ------------------
+        Interval(interval_metadata=<...>, bounds=[(1, 3)], \
+fuzzy=[(False, False)], metadata={'gene': ['BCC7']})
+
         Since only a *shallow* copy was made, updates to mutable objects stored
         as metadata affect the original sequence's metadata:
 
+        >>> seq_copy = seq.copy()
```
ebolyen: What was the reason for adding this line? `seq_copy` did already exist and it seems like the prose would still make sense without this line.

skbio/sequence/_nucleotide_mixin.py
```diff
@@ -161,6 +161,12 @@ def complement(self, reverse=False):
 
         if reverse:
             complement = complement[::-1]
+
+        if self.has_interval_metadata():
+            interval_metadata = self.interval_metadata
+            if reverse:
+                interval_metadata._reverse()
```
ebolyen: Add a TODO noting that this can be revised to match `positional_metadata` when `__getitem__` supports `interval_metadata`.

skbio/sequence/_sequence.py
```diff
@@ -1484,11 +1619,15 @@ def replace(self, where, character):
...
+        interval_metadata = None
+        if self.has_interval_metadata():
+            interval_metadata = self.interval_metadata
         # Use __class__ instead of _constructor so that validations are
         # performed for subclasses (the user could have introduced invalid
         # characters).
         return self.__class__(seq_bytes, metadata=metadata,
-                              positional_metadata=positional_metadata)
+                              positional_metadata=positional_metadata,
+                              interval_metadata=interval_metadata)
```

skbio/metadata/_interval.py
mortonjt: Right, there cannot be overlapping intervals between the two interval metadata objects.  Sorry about the confusion.

skbio/sequence/_sequence.py
RNAer: I would argue put `interval_metadata` immiediately after `positional_metadata` because it is natural and more logical

RNAer:
Thanks, @ebolyen ! what would you like me to do with your 1st general comment? add a clarification in the docstring of these methods (__iter__, __reversed__, iter_contiguous, iter_kmers, find_motifs)?

ebolyen:
Sorry, the first general comment doesn't need action, just confirming the situation.

skbio/sequence/_sequence.py
```diff
@@ -526,7 +572,9 @@ def concat(cls, sequences, how='strict'):
         pm = pd.concat(pm_data, join=how, ignore_index=True)
         bytes_ = np.concatenate(seq_data)
 
-        return cls(bytes_, positional_metadata=pm)
+        im = IntervalMetadata.concat(i.interval_metadata for i in seqs)
```
RNAer: there is nothing special about the sequence without `interval_metadata`, because the getter will return the empty `IntervalMetadata` with proper upper_bound. The existing code works alright.

skbio/sequence/_sequence.py
RNAer: to get refresh object to facility the following test codes

RNAer:
@ebolyen , all the requested tests are added and changes are made. it passed `make test` on my machine. only waiting for the @jwdebelius 's fix (thanks).

skbio/sequence/_sequence.py
jairideout: This is stable API so I don't think the parameter order should be changed. @ebolyen, IIRC you argued strongly earlier this year to avoid stable API changes (unless it results in fixing a bug) so I think we should hold to that here.

It obviously is more natural to have it immediately after positional metadata but will break code that isn't using keyword arguments. Too bad they aren't defined as py3 keyword-only args.

RNAer: do you guys all agree put it at the end? Although I still much prefer to keep it in the current order logically, I can make the change.

jairideout: @ebolyen and I chatted some more and are on the fence with this one. We are okay with leaving the order as-is (`sequence, metadata=None, positional_metadata=None, interval_metadata=None, lowercase=False`) because it is unlikely in practice to break user code. A user's code would break only if they had supplied `metadata`, `postional_metadata`, *and* `lowercase` parameters positionally. In the unlikely event that this happens, users will get an error telling them a `bool` isn't a valid `IntervalMetadata` type, so it won't silently produce buggy behavior.

Can you add a note under the `Backward-incompatible changes [stable]` heading for 0.5.0-dev changelog? Please provide a detailed explanation of this change, including the previous constructor signature and the new one, why the change was made, and why we are okay with breaking stable API (just transcribe what I wrote above if you want). Thanks!

skbio/sequence/_nucleotide_mixin.py
```diff
@@ -161,6 +161,15 @@ def complement(self, reverse=False):
 
         if reverse:
             complement = complement[::-1]
+
+        if self.has_interval_metadata():
+            interval_metadata = self.interval_metadata
```
jairideout: `IntervalMetadata._reverse` mutates the object in-place. This method produces a new `Sequence` with copies of the reversed metadata, so we'd want to first do a shallow-copy of `self.interval_metadata` and reverse the copy.

To achieve this, I think the code can be simplified while also matching the above pattern used for `metadata` and `positional_metadata`. It will also be more efficient by avoiding an extra shallow-copy operation:

```python
interval_metadata = None
if self.has_interval_metadata():
    interval_metadata = self.interval_metadata

complement = self._constructor(
    sequence=result,
    metadata=metadata,
    positional_metadata=positional_metadata,
    interval_metadata=interval_metadata)

if reverse:
    complement = complement[::-1]

    # TODO: this can be revised to match
    # positional_metadata when __getitem__
    # supports interval_metadata
    if complement.has_interval_metadata():
        complement.interval_metadata._reverse()
```

Can you also add a unit test that would have caught the bug?

skbio/sequence/_sequence.py
RNAer: great! will update changelog

skbio/sequence/_nucleotide_mixin.py
RNAer: I will fix it. but your code won't work, because `complement = complement[::-1]` drops `interval_metadata`, making the following line never executed.

jairideout: Ah I see. Thanks for fixing!

RNAer:
@ebolyen and @jairideout , I addressed all the comments. could you do a final pass? thanks!

jairideout:
Thanks @RNAer!

RNAer:
yay! Thanks to all the reviewers! I will put on genbank PR soon.

Issue #1436, opened 2016-10-13
Issue in skbio.stats.power code, probably just a broken test?


ebolyen:
I'm afraid I don't know enough about this code to know if it is just a broken test, or if an update to pandas actually found something interesting. @jwdebelius or @wasade would you be able to investigate? This is blocking #1435.

```pytb
======================================================================
ERROR: test__identify_sample_groups_not_strict (skbio.stats.tests.test_power.PowerAnalysisTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/evan/biocore/scikit-bio/skbio/stats/tests/test_power.py", line 491, in test__identify_sample_groups_not_strict
    sorted(test_pairs.values()))
  File "/home/evan/.conda/envs/skbio/lib/python3.4/site-packages/pandas/indexes/base.py", line 3317, in _evaluate_compare
    raise ValueError('Lengths must match to compare')
ValueError: Lengths must match to compare
```

wasade:
I don't know either here either, sorry

RNAer:
@jwdebelius , is this your code? could you take a look?

jwdebelius:
I'm committed all day tommorrow, but I will go through and follow up over the weekend.

jairideout:
Fixed in https://github.com/biocore/scikit-bio/pull/1439

PR #1439, opened 2016-10-17
fixes issue with paired test

jwdebelius:
cleans up code (slightly)
corrects for pandas data typing and improves testing based on arrays
rather than lists

addresses #1436; this looks like a broken test following pandas compatibility

...I did have a bunch of test break associated with `skbio.io.format.tests.test_blast7.TestBlast7Reader`

### DIFF
--- a/skbio/stats/power.py
+++ b/skbio/stats/power.py
@@ -1017,7 +1017,7 @@ def _identify_sample_groups(meta, cat, control_cats, order, strict_match):
         m_ids = meta.loc[ids].groupby(cat).groups
         # Checks if samples from the cat groups are represented in those
         # Samples
-        id_vecs = [m_ids[o].values for o in order if o in
+        id_vecs = [m_ids[o] for o in order if o in
                    m_ids]
         # If all groups are represented, the index and results are retained
         if len(id_vecs) == len(order):
--- a/skbio/stats/power.py
+++ b/skbio/stats/power.py
@@ -1017,10 +1017,10 @@ def _identify_sample_groups(meta, cat, control_cats, order, strict_match):
         m_ids = meta.loc[ids].groupby(cat).groups
         # Checks if samples from the cat groups are represented in those
         # Samples
-        ids_vec = id_vecs = [m_ids[o] for o in order if o in
-                             m_ids]
+        id_vecs = [m_ids[o].values for o in order if o in
+                   m_ids]
         # If all groups are represented, the index and results are retained
-        if len(ids_vec) == len(order):
+        if len(id_vecs) == len(order):
             min_vec = np.array([len(v) for v in id_vecs])
             loc_vec = np.arange(0, min_vec.min())
             meta_pairs[i1] = id_vecs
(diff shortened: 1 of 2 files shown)


skbio/stats/power.py
```diff
@@ -1017,10 +1017,10 @@ def _identify_sample_groups(meta, cat, control_cats, order, strict_match):
         m_ids = meta.loc[ids].groupby(cat).groups
         # Checks if samples from the cat groups are represented in those
         # Samples
-        ids_vec = id_vecs = [m_ids[o] for o in order if o in
-                             m_ids]
+        id_vecs = [m_ids[o].values for o in order if o in
```
ebolyen: Could you expand on why this change was made?  Was it a result of the pandas update? If so we may need to bump our minimum version of pandas.

skbio/stats/tests/test_power.py
```diff
@@ -477,18 +477,23 @@ def test__identify_sample_groups(self):
 
     def test__identify_sample_groups_not_strict(self):
         # Defines the know values
-        known_pairs = {0: [['PP'], ['CD', 'NR']],
-                       1: [['MM', 'WM'], ['MH']],
-                       2: [['GW'], ['CB']]}
+        known_pairs = {1: [np.array(['PP'], dtype=object),
+                           np.array(['CD', 'NR'], dtype=object)],
+                       0: [np.array(['MM', 'WM'], dtype=object),
```
ebolyen: It looks like `0` and `1` are transposed from their previous mapping.

ebolyen: Should this be a list? it looks as if you are depending on iteration order of a dictionary.

skbio/stats/power.py
ebolyen: Sorry I mean adding `.values` to the call, should have clarified.

skbio/stats/tests/test_power.py
jwdebelius: The 0 and 1 are proxies for a group description. The change in order reflects a change in the order returned by groupby (again, I think this is a pandas update result). The groups have remained consistent.

RNAer:
@ebolyen , is there anything else blocking this PR?

skbio/stats/power.py
ebolyen: Just tested this PR with a couple of different pandas versions. If we want to avoid bumping our minimum pandas version, we need to drop `.values`. Older versions of pandas return lists here apparently, and index objects are essentially ndarrays as it is, so it works fine without.

skbio/stats/tests/test_power.py
ebolyen: Reverting the order with older versions causes the tests to fail. I honestly have no idea how this was _ever_ passing... Good to leave as is, thanks.

jwdebelius:
@ebolyen Pinging the updated review request

ebolyen:
Merged, thanks @jwdebelius!

Issue #1159, opened 2015-10-28
Sequence.read fails for E.coli genbank record (0.4.0-dev)


ekopylova:
Python exists with `Killed: 9` when attempting to load the E. coli GenBank record on local computer.
By rough calculation (thanks @RNAer ), there are ~260000 features in the record, and each feature is stored as a column of genome length (4686137) in positional_metadata. That will take 260000 \* 4686137 / 1024**3 = 1134GB.

file source: ftp://ftp.ncbi.nlm.nih.gov//genomes/Bacteria/Escherichia_coli_K_12_substr__DH10B_uid58979/NC_010473.gbk

```
(scikit-bio)hutgw-70-72-dhcp:jenya$ python
Python 2.7.10 (default, Jul 14 2015, 19:46:27) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from skbio import Sequence
>>> seq = Sequence.read("./NC_010473.gbk", format='genbank')
Killed: 9
```

mortonjt:
Talking to @RNAer , it seems that the heart of the problem is the usage of dense pandas dataframes.
While it using sparse pandas dataframes may alleviate the problem, I'm not sure how scalable it is.

One alternative to this issue is using [Interval Trees](https://en.wikipedia.org/wiki/Interval_tree).  This allows for O(log n) access time to get all overlapping intervals.  Plus, it deals with the sparsity problem nicely, since it is only storing intervals.  One potential package to use is [bxpython](https://bitbucket.org/james_taylor/bx-python/wiki/Home).  This is personally what I use whenever I'm querying from genbank, gff and what-not.

RNAer:
Thanks, @mortonjt . Interval Trees seems a very good option. It is not as convenient as pandas dataframe, but close. I haven't come up with a use case that can't be done with Interval Trees easily. Anyone else?

I think it all boils down the memory usage even with pandas `SparseDataFrame` and how big the genome we want to support?

RNAer:
It should also solve the case (e) in #1137

ebolyen:
[...]

PR #1440, opened 2016-10-18
update genbank parser to use `IntervalMetadata`

RNAer:
also cleans up the code a bit.


RNAer:
@mortonjt , tests pass! it would be great for you to take a first pass. Thanks!

skbio/io/format/genbank.py
```diff
@@ -609,77 +610,86 @@ def _parse_features(lines, length):
...
+    ----------
+    intvl_md : dict
+        Interval.metadata
+    '''
     padding = ' ' * 8
     qualifiers = []
-    for k in sorted(obj):
-        if k.endswith('_') or k in ('location', 'type'):
+    for k in sorted(intvl_md):
+        if k.endswith('__') and k.startswith('__'):
```
mortonjt: Since we are explicitly searching for keywords with `__` in front, shouldn't we add a note in the docs about this?  The malevolent user may try using keywords with `__`.  

If we are extremely paranoid about this, maybe we should even raise an error that words that being and end with `__` are reserved?

mortonjt:
One comment.  I think this is looking quite solid.

RNAer:
Thanks, @mortonjt ! @ebolyen and @jairideout , would you mind taking a pass?

CHANGELOG.md
```diff
@@ -8,13 +8,12 @@
 * `DistanceMatrix.from_iterable` now accepts a `validate=True` parameter. ([#1343](https://github.com/biocore/scikit-bio/issues/1343)).
 * Added `IntervalMetadata` and `Interval` classes in `skbio.metadata` to store, query, and manipulate information of a sub-region of a sequence. ([#1414](https://github.com/biocore/scikit-bio/issues/1414)).
 * `Sequence` and its child classes (including `GrammaredSequence`, `RNA`, `DNA`, `Protein`) now takes `IntervalMetadata` in their contructor API. Some of their relevant methods are also updated acoordingly. ([#1430](https://github.com/biocore/scikit-bio/pull/1430))
+* GenBank parser now reads and writes `Sequence` or its subclass objects with `IntervalMetadata`. ([#1440](https://github.com/biocore/scikit-bio/pull/1440))
```
jairideout: It'd be nice to note the performance gains from this change (they are huge) and reference the original issue that prompted the creation of interval metadata (#1159). Probably makes sense in the `Performance enhancements` section below.

There were also some backward-incompatible changes made here (for example, `positional_metadata` is no longer used in favor of `interval_metadata`, and the keys used in `metadata`/`positional_metadata` have changed). Can you note all backward-incompatible changes under `Backward-incompatible changes [experimental]` section? Basically anything that could break existing user code relying on the GenBank reader/writer.

skbio/io/format/genbank.py
```diff
@@ -435,20 +422,38 @@ def _serialize_single_genbank(obj, fh):
...
         if header in md:
             serializer = _SERIALIZER_TABLE.get(
                 header, _serialize_section_default)
-            out = serializer(header, md[header])
-            # test if 'out' is a iterator.
-            # cf. Effective Python Item 17
-            if iter(out) is iter(out):
-                for s in out:
+            if header == 'FEATURES':
+                indent = 21
```
jairideout: Can you note where this magic number came from, or (better) remove the magic number in favor of something more self-documenting?

skbio/io/format/genbank.py
```diff
@@ -609,77 +613,86 @@ def _parse_features(lines, length):
...
-            feature['type_'] = type
-            feature['location'] = location
-            loc, loc_pmd = _parse_loc_str(location, length)
-            feature.update(loc)
+    section_iter = section_splitter(lines)
+
+    # 1st section is location
+    section = next(section_iter)
+    feature_type, feature_loc = _parse_section_default(
+        section, join_delimitor='', return_label=True)
```
jairideout: Minor: `delimitor` should be `delimiter`

skbio/io/format/genbank.py
```diff
@@ -721,47 +736,80 @@ def _parse_loc_str(loc_str, length):
     (e) a remote entry identifier followed by a local location
         descriptor (i.e., a-d). e.g. J00194.1:100..202
 
-    TODO:
-    handle (b), (c), (e) cases correctly
+    Notes
+    -----
+    This does not fully handle (e) case. It will discard the remote
```
jairideout: Do you think this documentation makes more sense to put in the user-facing docstring at the top of this module? Seems like a caveat/note that may be important to a user, and they won't see it here unless they're browsing the source code.

jairideout:
For reference, this fixes #1159.

RNAer:
Thanks, @jairideout ! I will address your comments together with @ebolyen 's once he is done

skbio/io/format/genbank.py
RNAer: the variable name is pretty self-documenting? can't think of a better way to do it. it is basically the number of indents I counted from a genbank file. the number could be 20 or 22 (not strict).

jairideout: Maybe add a comment explaining where the number came from? i.e. what you just explained to me about it anecdotally being used in genbank files, and that the number isn't strict (can be 20 or 22, maybe other values?). I could see a future maintainer wondering about why 21 is used as an indent.

skbio/io/format/genbank.py
RNAer: sure. added.

RNAer:
wondering if there is ETA on this. have a code review coming up on Monday. would like to have it merged and show it, if possible

jairideout:
@ebolyen ping on reviewing this

jairideout:
@RNAer out of curiosity, why do you need this merged before your code review on Monday? Maybe it is better to have this PR reviewed by your group then? Seems backwards to me :)

RNAer:
that makes sense after a second thought. I am also taking the opportunity to refactor out the feature table parser to a separate module so it can be used by EMBL and DDBJ formats (http://www.insdc.org/documents/feature_table.html) as well

CHANGELOG.md
mortonjt: I think those performance gains are addressed below? (Line 24)

RNAer:
@ebolyen and @jairideout , could you take a look at the latest commit (5927659)? would be great to have your input before pursuing further on it. 

Mainly I created a new module _feature_table.py, where I define the common vocabulary for GFF3 and GenBank/ENA (and probably other format as well) for interval feature metadata. do you agree the way how I handle it?

Thanks!

CHANGELOG.md
jairideout: It is now, but wasn't when I reviewed and added this comment :P

CHANGELOG.md
```diff
@@ -8,21 +8,21 @@
...
 
 ### Backward-incompatible changes [experimental]
 * Modifying basis handling in `skbio.stats.composition.ilr_inv` prior to checking for orthogonality.  Now the basis is strictly assumed to be in the Aitchison simplex.
 * `DistanceMatrix.from_iterable` default behavior is now to validate matrix by computing all pairwise distances. Pass `validate=False` to get the previous behavior (no validation, but faster execution).([#1343](https://github.com/biocore/scikit-bio/issues/1343)).
 
 ### Performance enhancements
 * `TreeNode.shear` was rewritten for approximately a 25% performance increase. ([#1399](https://github.com/biocore/scikit-bio/pull/1399))
+* The `IntervalMetadata` allows dramatic decrease in memory usage for feature rich sequences. ([#1159](https://github.com/biocore/scikit-bio/issues/1159))
```
jairideout: Can you add more detail, i.e. indicating this decreases memory usage in GenBank I/O? It isn't clear what memory usage is decreased.

I also requested this earlier:

> There were also some backward-incompatible changes made here (for example, positional_metadata is no longer used in favor of interval_metadata, and the keys used in metadata/positional_metadata have changed). Can you note all backward-incompatible changes under `Backward-incompatible changes [experimental]` section? Basically anything that could break existing user code relying on the GenBank reader/writer.

skbio/io/format/_feature_table.py
```diff
@@ -0,0 +1,90 @@
...
+# Copyright (c) 2013--, scikit-bio development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+'''INSDC defined a common "feature table" specification for
+DDBJ/ENA/GenBank formats It is described
+`here<http://www.ebi.ac.uk/ena/WebFeat/>`_. The feature types and
```
jairideout: does this link and the one below render in html? doesn't look like valid rst

jairideout:
note that tests are failing

jairideout:
@RNAer are there any pending comments from your group review on Monday, or are you ready for @ebolyen to take a final pass through?

skbio/io/format/_feature_table.py
RNAer: it is valid. cf: http://docutils.sourceforge.net/docs/user/rst/quickref.html#hyperlink-targets

but it is certainly less readable. I will change it later.

jairideout: specifically I am wondering if a space character is necessary between `here` and the opening `<`

RNAer:
@jairideout , there are no pending comments from code review.

1. I name it `_feature_table.py` because I plan to put the parsers and serializers of genbank feature table section into it, because feature table section are the same in ENA (also known as EMBL) format, which we plan to add to skbio as well. does it make sense?

2. how about putting `_vocabulary_*` functions from `_feature_table.py` into `io.format._base`? or another module `io.format._vocabulary`?

3. I have a particular question about docstring. If we move the vocabulary functions into `io.format._base`, where should we put the table in the current `_feature_table.py` docstring? it is tightly linked to the vocabulary functions and should be able to be rendered by sphinx.

@ebolyen , do you have any opinion on these question as well? I would ask you for the final review after we sort out the above question. try to avoid wasting your time.

jairideout:
> I name it _feature_table.py because I plan to put the parsers and serializers of genbank feature table section into it, because feature table section are the same in ENA (also known as EMBL) format, which we plan to add to skbio as well. does it make sense?

What I'm noting here is that there is a name clash between two completely unrelated things: sequence feature table vocabulary and a "feature table", AKA BIOM Table / contingency table. What about renaming `_feature_table.py` to `_sequence_feature_vocabulary.py`?

> how about putting _vocabulary_* functions from _feature_table.py into io.format._base? or another module io.format._vocabulary?

I'd prefer to avoid adding more code to `io.format._base` because it is already a pretty big file and it should be renamed to something more specific like `_sequence_base.py` since it consists of FASTA/FASTQ/QSEQ functionality.

> I have a particular question about docstring. If we move the vocabulary functions into io.format._base, where should we put the table in the current _feature_table.py docstring? it is tightly linked to the vocabulary functions and should be able to be rendered by sphinx.

The docs will need to be in a public module that is linked to somewhere from `skbio.io` docs. We haven't come up with a great solution to this. For now I would put all those docs in the Genbank I/O module docstring, and then reference/link to that table from the GFF3/ENA/etc formats when they exist. This is the strategy we took with FASTA, where it has the more detailed docs and FASTQ links to those docs in several places. @ebolyen do you have any other ideas?

RNAer:
@jairideout and @ebolyen , it is ready for the final review pass. Travis fails because sphinx warning:

```
Warning, treated as error:

/home/travis/miniconda3/envs/env_name/lib/python3.5/site-packages/skbio/io/format/genbank.py:docstring of skbio.io.format.genbank:70: WARNING: Block quote ends without a blank line; unexpected unindent.
```

I can't figure out what's wrong with line 70. do you have any idea?

ebolyen:
Re: points 1, 2, and 3.

1. I like `_sequence_feature_vocabulary.py`

2. Agree with @jairideout, `_base.py` is getting gross.

3. I think matching the way FASTQ references FASTA is the best solution for now. At some point we do need arbitrary prose-style documentation. That's probably where I would see these kinds of things documented eventually. Then the relevant formats can just refer users to these stand-alone articles which can be shared between many formats.

jairideout:
@RNAer, regarding the sphinx warning, I tried debugging it for a few minutes without success. The lines numbers (e.g. line 70) are usually wrong in the error message so the issue could be anywhere in the docstring. What I recommend doing is creating a standalone .rst file with the GenBank RST and just build from that, so that you can quickly iterate from source change to build. Then it's a game of removing RST source until you can get it to build successfully. Sorry I can't be of more help, it's very difficult to debug sphinx/rst issues and that's how I usually fix these issues as a last resort.

RNAer:
Thanks, @jairideout . how to run sphinx on a single RST file? tried `sphinx-build -c source -b html -d _build/doctrees . _build/html /tmp/genbank.rst` in the `doc` dir without luck.

is it ok for you guys to review while fix this sphinx warning? as the travis passes for Python=3.5.

RNAer:
I fixed the sphinx error. It turns out that issue is because sphinx can't process the tables that center justified - it will recognize them as block quote due to the padding spaces before the text.

This is for review for real now.

jairideout:
Great, glad you were able to clear up the Sphinx issue! @ebolyen will review (not sure what the ETA is).

RNAer:
@ebolyen , do you have ETA on review this? thanks!

ebolyen:
Hey @RNAer, I won't be able to get to this until this Monday. Sorry about that!

RNAer:
Understand. Thanks for letting me know!

skbio/io/format/genbank.py
```diff
@@ -39,47 +37,113 @@
...
+|           |           |           |         |be used for features whose    |
+|           |           |           |         |strandedness is relevant, but |
+|           |           |           |         |unknown.                      |
++-----------+-----------+-----------+---------+------------------------------+
+|codon_start|phase      |phase      |int (0,  |the offset at which the first |
+|           |(column 8) |           |1, or 2) |complete codon of a coding    |
+|           |           |           |         |feature can be found, relative|
+|           |           |           |         |to the first base of that     |
+|           |           |           |         |feature. It is 0, 1, or 2 in  |
+|           |           |           |         |GFF3 or 1, 2, or 3 in GenBank |
```
ebolyen: Is this normalized?

ebolyen: I assume it is given the `Value type stored`, but the description has so far been used to describe the `Value type stored`.

skbio/io/format/genbank.py
```diff
@@ -39,47 +37,113 @@
...
-The information of the ``'FEATURES'`` is stored in both ``metadata`` and
-``positional_metadata`` of ``Sequence`` or its sub-class. For each feature,
-its location is stored as boolean column in ``positional_metadata``; other
-qualifiers are stored as a ``dict`` in the ``list`` of
-``metadata['FEATURES']``. In the ``dict`` of qualifiers, there are a few
-extra keys, which end with ``'_'``, including:
+``FEATURES`` section
+^^^^^^^^^^^^^^^^^^^^
+The International Nucleotide Sequence Database Collaboration (INSDC
+[2]_) foundational initiative between the DDBJ, EMBL, and
```
ebolyen: Sentence fragment.

skbio/io/format/genbank.py
```diff
@@ -39,47 +37,113 @@
...
+       ``(101, 110)``.
+
+    4. the base numbers delimiting a sequence span. e.g. 340..565. This
+       is parsed to ``(339, 565)``.
+
+    5. a remote entry identifier followed by a local location
+       descriptor (i.e., a-d). e.g. J00194.1:100..202. This will be
+       discarded because it is not on the current sequence. When it is
+       combined with local descriptor like J00194.1:100..202,200..209,
+       the local part will be kept to be ``(199, 209)``.
```
ebolyen: :+1:

skbio/io/format/_sequence_feature_vocabulary.py
```diff
@@ -0,0 +1,351 @@
...
+    RPAREN = r'(?P<RPAREN>\))'
+    COMMA = r'(?P<COMMA>,)'
+    WS = r'(?P<WS>\s+)'
+    a = r'(?P<A>\d+)'
+    b = r'(?P<B>\d+\^\d+)'
+    c = r'(?P<C>\d+\.\d+)'
+    d = r'(?P<D><?\d+\.\.>?\d+)'
+    e_left = r'(?P<EL><?[a-zA-Z_0-9\.]+:\d+\.\.>?\d+)'
+    e_right = r'(?P<ER><?\d+\.\.>?[a-zA-Z_0-9\.]+:\d+)'
+    illegal = r'(?P<ILLEGAL>.+)'
```
ebolyen: This is very cool!

skbio/io/format/_sequence_feature_vocabulary.py
```diff
@@ -0,0 +1,351 @@
...
+        if k in voca_change:
+            k = voca_change[k]
+        if isinstance(v, list):
+            for vi in v:
+                qualifiers.append(_serialize_qualifier(k, vi))
+        else:
+            qualifiers.append(_serialize_qualifier(k, v))
+
+    if '__location' in md:
+        loc = md['__location']
```
ebolyen: What if someone were to mutate `__location`? Do we care?

RNAer: It is out of our control, but any key starting with "___" is private and should be modified with care. Add explanation in line 139.

RNAer:
@ebolyen , thanks! I addressed all you comments.

ebolyen:
:tada: :cake::tada: :cake::tada: :cake::tada: :cake: :exclamation:

Issue #1441, opened 2016-10-19
`IntervalMetadata` should allow the same start and end in bounds


RNAer:
currently, we disallow it. For example,

```
from skbio.metadata import IntervalMetadata
im = IntervalMetadata(3)
im.add([(1, 1)])
```

```
scikit-bio/skbio/metadata/_interval.py in _assert_valid_bound(bound)
    932                 isinstance(end, int)) or start >= end:
    933             raise ValueError('`start` (%r) must be a smaller int '
--> 934                              'than `end` (%r).' % (start, end))
    935     else:
    936         raise TypeError("Each `bound` must be a tuple, not {!r}".format(

ValueError: `start` (1) must be a smaller int than `end` (1).
```

@jairideout , we talked this over the `IntervalMetadata` PR and decided then not to allow it. However, I think we should allow this for annotation of `a site between two indicated adjoining bases. e.g. 123^124`, which is one type of location specified in genbank and other formats. For this example, it can be represented by `[(123, 123)]`.

what do you think?

jairideout:
Sounds good, I think we disallowed b/c there wasn't a clear use-case for a non-position. @ebolyen and I played around with querying those types of intervals and the behavior was kinda odd, though I don't have specific details about that. When you're implementing this, can you make sure these types of intervals are query-able?

RNAer:
sure. I play with it and make sure it queryable.

jairideout:
Thanks!

RNAer:
The current behavior is like this:
```
n [2]: a = IntervalMetadata(9)

In [3]: a.add([(1,1)])
Out[3]: Interval(interval_metadata=<4580885280>, bounds=[(1, 1)], fuzzy=[(False, False)], metadata={})

In [9]: list(a.query([(0, 1)]))
Out[9]: []
In [6]: list(a.query([(1, 2)]))
Out[6]: []
In [7]: list(a.query([(1, 1)]))
Out[7]: []

In [8]: list(a.query([(0, 2)]))
Out[8]: [Interval(interval_metadata=<4580885280>, bounds=[(1, 1)], fuzzy=[(False, False)], metadata={})]
```

It is a little surprising but not totally unexpected, because to determine if the query overlaps with a interval, it uses `self.end > start  and  self.start < end` (where `self` is the interval). Thus, you have to give `(p-1, p+1)` to retrieve the interval of position `p`. 

[...]

PR #1445, opened 2016-10-28
allow interval bounds with equal start and end

RNAer:
addressing #1441 . The demo for the edge cases are in the changes in `skbio/metadata/tests/test_intersection.py`. This is ready for review.


RNAer:
@jairideout and @ebolyen , this is ready for you to review. just wanna make sure you guys don't miss it. :)

jairideout:
@RNAer, apologies for the delay, one of us should have time to review on Monday.

skbio/metadata/_intersection.pyx
```diff
@@ -186,7 +186,15 @@ cdef class IntervalNode:
         if self.cleft is not EmptyNode and self.cleft.maxend > start:
             self.cleft._intersect( start, end, results )
         # This interval
-        if ( self.end > start ) and ( self.start < end ):
+        if self.end == self.start:
+            send = self.end
+        else:
+            send = self.end - 1
```
ebolyen: Add a comment indicating this is to mimic  the original behavior of `>` which should be enough to put a maintainer on the right track.

skbio/metadata/tests/test_intersection.py
```diff
@@ -189,6 +189,19 @@ def test_find(self):
...
+        iv.insert(1, 1, 'foo')
+        iv.insert(3, 7, 'spam')
+        self.assertEqual(iv.find(0, 1), [])
+        self.assertEqual(iv.find(1, 1), ['foo'])
+        self.assertEqual(iv.find(1, 2), ['foo'])
+        self.assertEqual(iv.find(2, 3), [])
+        self.assertEqual(iv.find(3, 3), ['spam'])
+        self.assertEqual(iv.find(3, 4), ['spam'])
+        self.assertEqual(iv.find(6, 7), ['spam'])
+        self.assertEqual(iv.find(7, 7), [])
```
ebolyen: Add a test that `iv.find(0, 8)` finds `['foo', 'spam']` just to make sure that "point ranges" interact as expected with normal ranges.

skbio/metadata/_intersection.pyx
RNAer: I don't get this. what comment you would like me to add here? Originally I didn't think about mimicing the old behaviour...

RNAer:
@ebolyen , thanks for reviewing! updated according to your points except the comment you suggest adding.

skbio/metadata/_intersection.pyx
ebolyen: The original code was `self.end > start`, but when that changed to `self.end >= start` you needed to subtract one from `self.end` to match the new check (when start != end). Or at least I think that is why this code is written that way...

RNAer: Actually nope. I tried other approaches. This is the most clear, working approach I can think of...

RNAer: If you know the comment or other cleaner code, I would be happy to update with it. :)

ebolyen: Ok let's not worry about this then.

ebolyen:
@RNAer merging this broke master, tried rerunning this build to see what's up, but now it can't find the ref to pull the code...

PR #1446, opened 2016-11-08
Revert "allow interval bounds with equal start and end"

ebolyen:
Reverts biocore/scikit-bio#1445


ebolyen:
What does travis do??? Who knows.

Issue #1448, opened 2016-11-08
Master broken on Linux (strange)


ebolyen:
Master is currently broken after merging both #1440 and #1445. Based on #1447 and #1446 it appears that either of these PRs in isolation work fine. So there is something *really* strange happening.

We are able to reproduce this runtime split locally in a couple of linux (which fail) and mac (which succeed) environments.

RNAer:
@wasade and @mortonjt , you guys have any idea?

mortonjt:
Taking a look now!

ebolyen:
Fixed by #1449

PR #1449, opened 2016-11-09
fix a bug in _intersection.pyx

RNAer:
This addresses #1448. The bug is that I need to change the greater/less-than signs to greater/less-than-or-equal-to sign to search for left/right subtree.

The fact that tests passes in some environment was confusing and misleading. The reason for that behavior is that OSX and linux starts with different interval nodes to search intersection. If it started with the lucky node (as in OSX), it would find the expected nodes; if it is not (as in some cases in linux), it ignored left or right subtree and returned empty.

### DIFF
--- a/skbio/metadata/_intersection.c
+++ b/skbio/metadata/_intersection.c
@@ -494,7 +494,7 @@ struct __pyx_obj_5skbio_8metadata_13_intersection_IntervalTree;
 struct __pyx_opt_args_5skbio_8metadata_13_intersection_12IntervalNode_left;
 struct __pyx_opt_args_5skbio_8metadata_13_intersection_12IntervalNode_right;
 
-/* "skbio/metadata/_intersection.pyx":261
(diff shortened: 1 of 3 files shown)


RNAer:
@ebolyen and @jairideout , this should fix the bug. Thanks for all the help with debugging! Also big thanks to @wasade for suggesting useful approaches to identify the bug!

wasade:
Wow... good find! Is there an additional test or two which would be beneficial to add?

mortonjt:
Damn.  That is sneaky.  👍 for adding a test.

RNAer:
The root node (the staring node for search) is a bit stochastic because it
tries to do smart things to re-balance the tree so that the interval search
is optimized. Added extra tests - ccording to the code, both left and right
subtrees are tested now.

wasade:
:+1:

mortonjt:
Awesome!

ebolyen:
Amazing. Awesome work @RNAer!

PR #1450, opened 2016-11-09
GFF3 IO

RNAer:
@mortonjt , would you mind giving it a pass of review?


jairideout:
I'm not reviewing this right now, but please remember the changelog :)

RNAer:
ah, thanks for reminding! updated changelog.

jairideout:
Thanks! Just a heads up: I'm planning a scikit-bio 0.5.1 release this weekend (Nov. 12-13), so code will need to be merged by Friday afternoon to be included in the release.

RNAer:
Thanks for the heads up! Is there any chance you or @ebolyen can review this now? I tried to grab people for first passes, but most are tied up currently.

jairideout:
I can review tomorrow. It would be helpful to have at least one person on your end take a pass through before that if someone is available. Thanks!

skbio/io/format/tests/test_gff3.py
```diff
@@ -0,0 +1,280 @@
...
+                fh,
+                seq_ids=['Chromosome',
+                         'gi|556503834|ref|NC_000913.3|',
+                         'NC_7'],
+                skip=False)
+            obs = [i for i in fh.getvalue().splitlines()
+                   if not i.startswith('#')]
+
+        with open(self.multi_fp) as f:
+            exp = [i[:-1] for i in f.readlines() if not i.startswith('#')]
```
mortonjt: This is for the newline right?  Why not `i.rstrip()`?

skbio/io/format/tests/test_gff3.py
```diff
@@ -0,0 +1,280 @@
...
+                _gff3_to_interval_metadata(
+                    self.single_fp,
+                    interval_metadata=IntervalMetadata(self.upper_bound1)),
+                fh,
+                seq_id='Chromosome')
+            obs = [i for i in fh.getvalue().splitlines()
+                   if not i.startswith('#')]
+
+        with open(self.single_fp) as f:
+            exp = [i[:-1] for i in f.readlines() if not i.startswith('#')]
```
mortonjt: Same here about `i.rstrip()`

skbio/io/format/tests/test_gff3.py
```diff
@@ -0,0 +1,280 @@
...
+    def test_interval_metadata_to_gff3_sub_region(self):
+        with open(self.multi_fp) as f:
+            lines = [i.strip() for i in f if not i.startswith('#')]
+
+        with io.StringIO() as fh:
+            _serialize_interval_metadata(
+                self.imd3, seq_id='NC_7', fh=fh, skip=False)
+            obs = [i for i in fh.getvalue().splitlines()
+                   if not i.startswith('#')]
+        exp = lines[-3:]
```
mortonjt: What's this magic number?

skbio/io/format/tests/test_gff3.py
```diff
@@ -0,0 +1,280 @@
...
+        self.assertEqual(exp, obs)
+
+        with io.StringIO() as fh:
+            _serialize_interval_metadata(self.imd3, seq_id='NC_7', fh=fh)
+            obs = [i for i in fh.getvalue().splitlines()
+                   if not i.startswith('#')]
+        exp = [lines[-3]]
```
mortonjt: Same here

skbio/io/format/tests/test_gff3.py
```diff
@@ -0,0 +1,280 @@
...
+    def test_interval_metadata_to_gff3(self):
+        with io.StringIO() as fh:
+            _interval_metadata_to_gff3(self.imd1, fh, seq_id='Chromosome')
+            # only compare the uncommented lines because the comments are not
+            # stored in IntervalMetadata
+            obs = [i for i in fh.getvalue().splitlines()
+                   if not i.startswith('#')]
+
+        with open(self.single_fp) as f:
+            exp = [i[:-1] for i in f.readlines() if not i.startswith('#')]
```
mortonjt: Same thing re `i.rstrip()`

skbio/io/format/tests/test_gff3.py
```diff
@@ -0,0 +1,280 @@
...
+            self.assertEqual(_gff3_sniffer(fp), (False, {}))
+
+
+class ReaderTests(GFF3IOTests):
+    def test_yield_record(self):
+        obs = [('seqid1', ['seqid1\txxx', 'seqid1\tyyy']),
+               ('seqid2', ['seqid2\tzzz'])]
+        s = ('seqid1\txxx\n'
+             'seqid1\tyyy\n'
+             'seqid2\tzzz\n')
```
mortonjt: It is possible to replace this with joins, but that is optional

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,469 @@
...
+    escape = str.maketrans({';': '%3B',
+                            '=': '%3D',
+                            '&': '%26',
+                            ',': '%2C'})
+
+    for interval in interval_metadata._intervals:
+        md = interval.metadata
+        bd = interval.bounds
+        start = str(bd[0][0] + 1)
+        end = str(bd[-1][-1])
```
mortonjt: It would be nice to have a comment to clarify this.  I'm assuming that this is converting to 1 based indexing for the first and last coordinate in the

RNAer:
@mortonjt , thanks for your comments! will fix them together after @jairideout 's comments today

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,469 @@
...
+    '''Serialize an IntervalMetadata to GFF3.
+
+    Parameters
+    ----------
+    interval_metadata : IntervalMetadata
+    skip : bool
+        whether to skip outputting each sub region as a line in GFF3.
+    '''
+    # write file header
+    print('##gff-version 3', file=fh)
```
jairideout: Remove print statements, there's four in this file.

mortonjt: Just to clarify, you recommend using

`fh.write('##gff-version 3')` 

instead correct?  I think the file header is necessary.

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,469 @@
...
+text, 9-column, tab-delimited files [1]_.
+
+Format Support
+--------------
+**Has Sniffer: Yes**
+
++------+------+---------------------------------------------------------------+
+|Reader|Writer|                          Object Class                         |
++======+======+===============================================================+
+|Yes   |Yes   |:mod:`skbio.metadata.IntervalMetadata`                         |
```
mortonjt: Just to note, this needs to be done by the code freeze tomorrow right?

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,469 @@
...
+Column 9 (ATTR) is list of feature attributes in the format
+tag=value. Multiple tag=value pairs are separated by
+semicolons. Multiple values of the same tag are indicated by
+separating the values with the comma ",". The following tags have
+predefined meanings:
+
+* ID. Indicates the unique identifier of the feature. IDs must be
+  unique within the scope of the GFF file.
+
+* Name. Display name for the feature. This is the name to be displayed
```
RNAer: Can we just delete that and just link to the reference?

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,469 @@
...
+some sequences in the GFF3 doesn't match any sequence IDs in the input
+dict of ``interval_metadata_dict``, those annotations will be skipped
+and not parsed.
+
+``interval_metadata`` is an ``IntervalMetadata`` object required for
+``IntervalMetadata`` GFF3 reader. It reads the GFF3 file and add the
+annotations to the input ``interval_metadata``.
+
+``rec_num`` is an ``int` parameter used with the ``IntervalMetadata``,
+``DNA``, and ``Sequence`` GFF3 readers. It specifies which GFF3 record
```
RNAer: 1. multiple rows with the same seq id.

2. read the rows of the first seq id seen.

3. that's a good idea. providing a `seq_id` is good. will change to that.

skbio/io/format/gff3.py
RNAer: @jairideout , as @mortonjt pointed out, this is writing to file.

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,469 @@
...
+-----------------
+
+Reader-specific Parameters
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+``interval_metadata_dict`` is a ``dict`` parameter required for
+``IntervalMetadata`` generator. Its key is the ``str`` of sequence ID
+and its value is the ``IntervalMetadata`` object for the sequence. The
+generator will read the GFF3 file and add the annotation of sequences
+to the corresponding ``IntervalMetadata`` objects. If the annotations of
+some sequences in the GFF3 doesn't match any sequence IDs in the input
```
wasade: "...of some of the sequences..."

"...in the GFF3 file..."

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,469 @@
...
+dict of ``interval_metadata_dict``, those annotations will be skipped
+and not parsed.
```
wasade: Should this be parameterized (i.e., `strict`)?

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,469 @@
...
+'strand': '+', 'score': '.', 'Parent': 'gene00001'})
+Interval(interval_metadata=<4601272528>, bounds=[(1049, 9000)], fuzzy=\
+[(False, False)], metadata={'source': '.', 'type': 'mRNA', 'strand': '+', \
+'score': '.', 'ID': 'mRNA00001', 'Parent': 'gene00001'})
+>>> im == im_return
+True
+
+References
+----------
+.. [1] https://github.com/The-Sequence-Ontology/Specifications/blob/master/gff3.md  # noqa
```
wasade: Is there a specific tag or commit this can be tied too?

skbio/io/format/gff3.py
jairideout: Oh, I didn't see this was printing to a filehandle. Recommend using `fh.write()` since that's typically the API used to write to a filehandle (I didn't even know there was a `file` parameter with `print`!).

RNAer:
@jairideout , thanks for reviewing! Responded to a few of your points when neccessary.

1. I am ok to leave this out from tomorrow's release if we don't feel comfortable yet, although I would really want to.

2. on the reader/writer behavior, I strongly argue for the current behavior - reading annotations to an existing `IntervalMetadata`. In many use cases, the sequence is already have some annotation in its `interval_metadata`, it makes much more sense and is much less expensive to add annotations on the existing one than creating a new object and merge it to the existing one. Secondly, even  if a sequence don't have an existing `interval_metatadata`, we have to pass in the seq length, which is required to create a new one. In summary, It seems clearer to pass `IntervalMetadata` for both cases.

skbio/io/format/gff3.py
jairideout: Thanks for the details. Based on that I think it makes sense to require passing a `seq_id` when reading a single sequence's annotations. This will avoid odd cases where annotations from different sequences are interleaved, and it isn't clear what annotations you'll get from the "first" sequence/record. Requiring a `seq_id` on read also nicely mirrors the required `seq_id` when writing. Does that sound good to you?

RNAer: yep

jairideout:
> I am ok to leave this out from tomorrow's release if we don't feel comfortable yet, although I would really want to.

Thanks for being flexible on this. I'd like to wait on merging this until after the release, to have more time to test it and to get these larger changes worked out (time is a limiting factor here).

> on the reader/writer behavior, I strongly argue for the current behavior - reading annotations to an existing IntervalMetadata. In many use cases, the sequence is already have some annotation in its interval_metadata, it makes much more sense and is much less expensive to add annotations on the existing one than creating a new object and merge it to the existing one. Secondly, even if a sequence don't have an existing interval_metatadata, we have to pass in the seq length, which is required to create a new one. In summary, It seems clearer to pass IntervalMetadata for both cases.

I strongly argue against the current API because it is an API anti-pattern and the rest of scikit-bio I/O doesn't work this way. If efficiency concerns are going to determine whether we use a less-than-ideal API over the one I'm proposing, please include performance benchmarks that highlight the problem.

If merging `IntervalMetadata` objects is desirable, and `IntervalMetadata.add`/`concat` aren't performant enough, maybe it makes sense to have a `IntervalMetadata.merge` or `concat` instance method that merges an interval metadata object into another, in-place.

Regarding coordinate space and sequence length, we know the sequence length if reading into a `Sequence` object. If we're just reading annotations, GFF3 includes the coordinate space in a `sequence-region` directive.

skbio/io/format/gff3.py
RNAer: The thing I copied directly in genbank is among line 118-130 and some examples in those lines. Is this a problem? I can try to rephrase it.

RNAer:
Ok. Side-effect is not ideal, although I wouldn't say it as anti-pattern. @jairideout , could your confirm you agree with the current APIs?

skbio/io/format/gff3.py
jairideout: I think it'd be good to remove/rephrase it. Would it be possible to have that ready before the freeze this afternoon?

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,466 @@
...
+        return False, {}
+
+    if re.match(r'##gff-version\s+3', line):
+        return True, {}
+    else:
+        return False, {}
+
+
+@gff3.reader(None)
+def _gff3_to_generator(fh, lengths):
```
RNAer: only if the length is specified there. In most cases, it is not.

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,466 @@
...
+    obj : Iterable of IntervalMetadata
+    fh : file handler
+    seq_ids : Iterable of seq id (str)
+    '''
+    for obj_i, seq_id in zip(obj, seq_ids):
+        _serialize_interval_metadata(obj_i, seq_id, fh, skip)
+
+
+@gff3.reader(Sequence)
+def _gff3_to_sequence(fh, rec_num=1):
```
RNAer: `rec_num` is better in this reader. this is read into `Sequence` object, so presumably users don't have sequence ahead of time, not to say seq id. this mirrors fasta/fastq/genbank reader. I can change `rec_num` to `seq_num` as used in those readers.

RNAer:
@jairideout , saw your 0.5.1 release PR. It looks great!

This PR is ready for another round of review. One thing that need a bit explanation: the reader to generator requires a parameter of dict `id_lengths` (instead of a list), because it is not easy to match the lengths to gff records in the file. For example, when we annotate a set of sequences and output to GFF3, one sequence may don't have any annotation in the GFF3 file and we won't know this ahead of time. With `id_lengths`, this is not a problem any more.

skbio/io/format/gff3.py
jairideout: Bummer!

skbio/io/format/gff3.py
jairideout: So this `rec_num` would refer to the record number in the FASTA portion of the file? I think that makes sense, probably rename to `seq_num` to consistency (and it's more explicit). This writer could also accept a `seq_id` (that's mutually exclusive with `rec_num`) if you think that's useful.

jairideout:
@ebolyen and I chatted about the unknown length problem. Here's an idea:

- use the sequence length in FASTA record if there is sequence data for an annotation, or use `sequence-region` directive if it's present
- if not, there's a `unknown_length` parameter that gives users control over what to do with annotations of unknown length. It could have 3 modes (similar to Python stdlib encoding/decoding str API):
    - `unknown_length='error'` is the default, it raises an error if an annotation has an unknown length
    - `'ignore'` will ignore those annotations, user will only get annotations with known lengths. Or maybe `'filter'` is a better name?
    - user provides a callback function that accepts a sequence ID and returns the length

This `unknown_length` parameter would be used with the generator reader and single `IntervalMetadata` reader. `Sequence`-related readers wouldn't need it.

What do you think about this API?

RNAer:
I am not convinced about `unknown_length` parameter. 

> user provides a callback function that accepts a sequence ID and returns the length

If there is no `sequence-region` pragma, there is no way to obtain the length from the GFF3 file. This callback function seems useless to me.

And guys may not get my previous comment on `id_lengths`. Let me give an example. Let's say we have  a sequence file `foo.fasta`:

````
>seq1
ATG...
>seq2
ACG...
>seq3
TGC...
```

And after gene annotation, I get the following gff3 file. Because `seq2` doesn't have genes on its sequence, there are no lines about `seq2`:

```
##gff-version 3
seq1	\tprodigal\t...
seq1\tprodigal\t...
seq3\tprodigal\t...
...
```

In this case, we need to match the seq id and the lengths provided or obtained from whatever source outside the GFF3 file.

My current reader API is this:

```
# to sequence or dna
read(fp, format='gff3', seq_num=2)
# to intervalmetadata
read(fp, format='gff3', seq_id='seq3', length=300)
# to intervalmetadata generator
read(fp, format='gff3', id_lengths={'seq1':100, 'seq3':300})
```

ebolyen:
I think the callback would work exactly like your `id_lengths` parameters
in this example (and allow more complex things). An alternate idea, what if
`gff3` was available as an optional parameter to `fasta`, just like `qual`?

jairideout:
@RNAer, the `unknown_length` parameter allows you to accomplish the same things as the `dict` API you're proposing, but it is more flexible. The `dict` is simply a mapping and the callback can also act as a mapping (e.g. using closure).

@ebolyen, that's an interesting idea but I think we'd need GFF3 as its own format because the FASTA portion is included in the file format. If GFF3 is an optional parameter to FASTA readers that would be cool but I don't think it solves this immediate problem where there is no sequence data to infer length from.

ebolyen:
@jairideout RE: GFF3/FASTA; agreed, but it might make the ergonomics of
this slightly less important (if fasta + gff3 is the typical way of
interacting with this data).

RNAer:
Maybe just because I don't get your callback fully, could you elaborate a bit more on `allow more complex things` and `more flexible`? An example of callback would be like: check `sequence-region` pragma first, if not finding length there, then return the provided length in the callback? 

````
def callback(seqid):
      l = {'seq1': 100,  'seq3':300}
      # get the length from `sequence-region`
      try:
           length = get_from_sequence_region...
      except Errror:
           length = l[seqid]
      return length
```

It seems making API complicated? Because we have to write a callback (either from scratch or call a function dispatcher) every time you read a GFF3?

ebolyen:
@RNAer, the callback would only ever be invoked if the length was unknown. So in the example you could write:
```python
l = {'seq1': 100,  'seq3':300}
read(fp, format='gff3', unknown_length=lambda id: l[id])
```

Or you could get fancy and do something like:

```python
def lookup_from_sequences(seqs):
    mapping = {s.metadata['id']: s for s in seqs}
    def callback(id):
        return mapping[id]
    return callback

# Then use the factory:
read(fp, format='gff3', unknown_length=lookup_from_sequences(my_sequences))
```

This does add a little bit of boilerplate, but it opens the door for us to provide arbitrary factories where useful, and it lets the user customize the behaviour a little more.

If we paired this with the `fasta` readers then those could just define a particular factory (which could expect same iteration order) and we would be able yield sequences without parsing all of the records first (i.e. we don't need a completed dict before reading the GFF3 metadata).

RNAer:
OK. So the default APIs become the following?

````
# to intervalmetadata
read(fp, format='gff3', seq_id, unknown_length='error')
# to intervalmetadata generator
read(fp, format='gff3', unknown_length='error')
````

**Comment by jairideout** [comment]
@RNAer 👍 on that API. Do you like `unknown_seq_length` or `unknown_seq_len` better than `unknown_length`?

**Comment by RNAer** [comment]
Oh gosh, this reminds me the quote I saw before:

> There are only two hard things in Computer Science: cache invalidation and naming things. 

I will pick `unknown_seq_len`, but if there is a better name, I would be very happy to hear.

**Comment by RNAer** [comment]
@jairideout and @ebolyen , I made the change to the generator at line 256. However, it makes me think the "ignore" or "error" values for `unknown_seq_len` cumbersome and unnecessary. 

if it is set to "ignore", basically it ignores the annotations of no `sequence-region` pragma defined for its sequence. This is not the user case. We often have a set of seq IDs in mind to read the annotation, but seldom read annotations based on whether the pragma is there or not. Same for "error".

Another question is at line 340. It is a little bit silly to pass a callback in the reader and call it with seq_id to get length rather than obtaining the length outside the reader and passing it in.

These set me back to think about the advantage of have a callback. What's the pros of obtaining the length inside the reader from the callback? comparing to calling the same callback function to get the id-length mapping and passing the mapping to the reader.

**Comment by RNAer** [comment]
One alternative is to modify `IntervalMetadata` and allow `IntervalMetadata.upper_bound` to be `None`. It is reset to the seq length when an `IntervalMetadata` object is assigned to a sequence object. If I don't miss anything, this shouldn't change the behaviour of those objects and should work alright. I like this, because this will make the GFF3 (and potentially other formats to include in the future) reader/writer APIs so much cleaner.

@jairideout or @ebolyen , what do you think?

**Comment by ebolyen** [comment]
Talked with @RNAer offline about this. @jairideout we're thinking that allowing `IntervalMetadata` to have an `upper_bound` of `None` is probably the cleanest way to handle this. We can probably make `IntervalMetadata.concat` private and so it shouldn't really impact the API. When attached to a sequence object the `upper_bound` will be set to `len(seq)`. Which probably implies that `upper_bound` is mutable.

@RNAer was also able to explain the relationship of these files with something like a fasta file, and in practice the relationships are crazy messy. So it may not always be easy to know ahead of time what the `upper_bound` would be set as, so allowing it to be set at a later point offers a bit more flexibility.

**Comment by RNAer** [comment]
Thanks, @ebolyen ! 

To be clear, the relationship between GFF3 and fasta sequence is complex:

1) the sequences are in the GFF3 file;

2) a GFF3 file have annotations for multiple sequences whose actual DNA sequence exists in different fasta files; 

3) a sequence has multiple annotations in different GFF3 files.

Conceptually, allowing `None` for `IntervalMetadata.upper_bound` means: we are agnostic about the coordinate space and assume all the interval features added to IntervalMetadata object are within the coordinate space, which can be verified later when it is attached to a sequence and/or `upper_bound` is reset to the seq length. This is in parallel to our use cases: we do not necessarily know the sequence or its length at the moment of reading a GFF3 file. That's why we think this is best solution.

**Comment by jairideout** [comment]
This sounds like a good plan. A few comments:

### Proposed API

The latest API you proposed for single `IntervalMetadata` reader/writer is a little different than what I was thinking:

```python
read(fp, format='gff3', seq_id=seq_id)  -> intervalmetadata

write(intervalmetadata, fp, format='gff3', seq_id=seq_id)
```

When reading a single `IntervalMetadata`, the `seq_id` must be specified. The result is an `IntervalMetadata` object, not a tuple of `(seq_id, IntervalMetadata)`. The writer accepts a `seq_id` parameter (required) and an `IntervalMetadata` object to write.

The reasoning for this change: if the reader/writer accept a `tuple` type, we can't dynamically add methods `IntervalMetadata.read` and `IntervalMetadata.write` to the `tuple` type. By having it read into a single scikit-bio object, we can add the OO I/O interface to `IntervalMetadata`. Also, it doesn't make sense to force the user to provide a `seq_id` when reading, and then have that same `seq_id` yielded by the reader.

What do you think about the proposed API?

### Optional upper bound

I like this idea. I think the public `upper_bound` property should be read-only (like it is right now), and when adding `IntervalMetadata` to a `Sequence`, the private `_upper_bound` property is updated to the sequence's length. If the max coordinate in the `IntervalMetadata` is greater than the sequence length, an error is raised.

The reasoning for this change: it's safer to have the `IntervalMetadata` `upper_bound` immutable so that it can't be changed after the `IntervalMetadata` is attached to a `Sequence`. We need to maintain the invariant of coordinate space matching the sequence length. For example, we'd want to avoid this situation:

```python
seq = Sequence('ACGT', interval_metadata=IntervalMetdata(4)) # coordinate space matches sequence length
seq.interval_metadata.upper_bound = 5 # lengths do not match anymore
```

Does that make sense?

RNAer:
Thanks, @jairideout . I am ok with the API and keeping `upper_bound` immutable. @ebolyen argued it for mutable because public API is preferred over private `_upper_bound` attribute (like we discussed over call yesterday). If both of you agree with keeping it immutable, I can change it back.

jairideout:
Chatted with @ebolyen about this. He agrees it is ideal to have `upper_bound` be immutable for the safety reasons I described.

The downside to making it immutable is that users cannot set an upper bound manually after an `IntervalMetadata` object is created. To solve the safety problem *and* support changing the upper bound, we were thinking of modifying the `IntervalMetadata` constructor to also accept an `IntervalMetadata` object, and to keep `upper_bound` as a required parameter. If an `IntervalMetadata` object is provided to the constructor, the newly constructed object would be a shallow copy of the input `IntervalMetadata`, and its upper bound would be set to the `upper_bound` passed to the constructor. This allows users to create a new `IntervalMetadata` object from an existing one and set an upper bound. `upper_bound` would support `None` as a valid value, but needs to be a required parameter so that users can create an unbounded upper bound from an object that is bounded.

This solution will also avoid using private API from within the mixin -- the mixin setter would use the constructor I described to create a new (shallow copy) of the input with the correct upper bound.

RNAer:
This new constructor is a good idea balancing the safety and API usage. will make changes accordingly in #1460

jairideout:
@RNAer, now that #1460 is merged I think this PR is ready to move forward. @wasade, since we have hammered out the API, can you please take the lead on reviewing and merging this PR? Note that I haven't reviewed any of the code, I was only looking at what the public API will be. I am out of bandwidth on scikit-bio reviews and need to prioritize QIIME 2 development now. I think this is a good opportunity for you guys to try out the code review doc I put together now that you are core maintainers.

wasade:
Will aim for review by the end of the week.

RNAer:
Thanks, @wasade !

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,469 @@
+r'''
+GFF3 format (:mod:`skbio.io.format.gff3`)
+=========================================
+
+.. currentmodule:: skbio.io.format.gff3
+
+GFF3 is a standard file format for storing genomic features in a text
```
jairideout: Ping on this comment

jairideout:
Thanks @wasade! When you're reviewing, can you please verify that previous review comments (from @mortonjt, yourself, and I) have been addressed? I noticed a couple of mine weren't addressed, but haven't done an extensive check across the previous reviews.

Checking that previous review comments have been addressed is part of my review process since an outdated comment (based on diff) doesn't necessarily mean the comment was addressed. Here's a bookmarklet to expand all outdated comments on a PR:

```javascript
javascript: Array.from(document.getElementsByClassName('outdated-diff-comment-container')).concat(Array.from(document.getElementsByClassName('outdated-comment'))).forEach(l => l.classList.add('open'));
```

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,508 @@
...
+            if k in voca_skip:
+                # skip the metadata don't go to attribute column
+                continue
+            if k in voca_change:
+                k = voca_change[k]
+            v = md[k]
+            attr.append('%s=%s' % (k.translate(escape), v.translate(escape)))
+        columns.append(';'.join(attr))
+
+        print('\t'.join(columns), file=fh)
```
RNAer: Didn't change that because I need another line to add a linefeed if using `fh.write()`. Thought there is no preference using `write` or `print` to output to file. Can change that if `write` is appealing more to you.

RNAer:
I usually address in the order of comments brought up, trying not to miss any. Sorry if there is any one missed - I'll fix it.

jairideout:
No problem, and thanks for checking that! There's been a lot of discussion on this PR and the reviews are kinda old, so I just wanted to make sure they weren't forgotten.

skbio/io/format/gff3.py
jairideout: I'd prefer `.write()` even if it requires another line of code. The current code is relying on `print()` to supply a newline, whereas an explicit `fh.write('\n')` makes it clear what's happening. Seeing a `print` statement is also confusing b/c it looks like debug statements were left in the code. I chatted with @ebolyen and he also prefers `.write()` over `print`.

RNAer: Thanks for clarifying that. will change.

RNAer:
@wasade , just a friendly reminder - this is still waiting for your review. All the comments so far are addressed

skbio/io/format/gff3.py
wasade: Yes, this is critical to address.

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,508 @@
...
+
+>>> with io.StringIO() as fh:    # doctest: +NORMALIZE_WHITESPACE
+...     print(im.write(fh, format='gff3', seq_id='seq_1').getvalue())
+##gff-version 3
+seq_1	.	gene	10	90	.	+	0	ID=gen1
+seq_1	.	exon	10	30	.	+	.	Parent=gen1
+seq_1	.	exon	50	90	.	+	.	Parent=gen1
+<BLANKLINE>
+
+If the GFF3 file does not have the sequence ID, it will return empty object:
```
wasade: "will return an empty object"

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,508 @@
...
+
+>>> gff = io.StringIO(gff_str)
+>>> im = read(gff, format='gff3', into=IntervalMetadata,
+...           seq_id='foo')
+>>> im
+0 interval features
+-------------------
+
+We can also read the GFF3 file into generator:
```
wasade: "into a generator"

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,508 @@
...
+    for attr in s.split(';'):
+        k, v = attr.split('=')
+        if k in voca_change:
+            k = voca_change[k]
+        md[k] = v
+    return md
+
+
+def _serialize_interval_metadata(
+        interval_metadata, seq_id, fh, skip_subregion=True):
```
wasade: can these parameters be put on the first line of the definition first?

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,508 @@
...
+    '''Serialize an IntervalMetadata to GFF3.
+
+    Parameters
+    ----------
+    interval_metadata : IntervalMetadata
```
wasade: `seq_id` and `fh` aren't described. I realize this is a private method, but if the docstring goes into this level of detail it would be good to be complete

skbio/io/format/gff3.py
wasade: It looks like this is pending still

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,508 @@
...
+                        'the parent interval feature: %r' % interval)
+                columns[8] = 'Parent=%s' % parent
+                print('\t'.join(columns), file=fh)
+
+
+def _serialize_seq(seq, fh, skip_subregion=True):
+    '''Serialize a sequence to GFF3.'''
+    _serialize_interval_metadata(
+        seq.interval_metadata, seq.metadata['id'], fh, skip_subregion)
+    print('##FASTA', file=fh)
```
wasade: note re: `print`

mortonjt:
The gff documentation has quite a few backspaces.  I'm getting that there were quite a few escape characters.  I'm not sure how this renders (will need to double check that), but an easy way around this problem is to use [python raw strings](http://stackoverflow.com/questions/4780088/what-does-preceding-a-string-literal-with-r-mean).  Its up to you if you want to change this, but it'll make updating the docstrings easier in the future.

skbio/io/format/tests/data/gff3_multi_record
```diff
@@ -0,0 +1,15 @@
...
+#!genome-build-accession GCA_000005845.2
+#!genebuild-last-updated 2014-08
+Chromosome	European Nucleotide Archive	chromosome	1	4641652	.	.	.	Alias=U00096.3;ID=chromosome:Chromosome;Is_circular=true
+###
+Chromosome	regulondb_feature	biological_region	148	148	.	+	.	external_name=Promoter thrLp (RegulonDB:ECK120010236);logic_name=regulondb_promoter
+gi|556503834|ref|NC_000913.3|	Prodigal_v2.60	gene	337	2799	1.8	+	0	ID=1_1;gc_cont=0.427
+gi|556503834|ref|NC_000913.3|	Prodigal_v2.60	CDS	337	2799	333.8	+	0	ID=1_2;Parent=1_1;rbs_motif=GGAG/GAGG;rbs_spacer=5-10bp
+NC_7	Prodigal_v2.60	gene	1	100	1.8	+	0	ID=1_1;gene=FXR receptor
+NC_7	Prodigal_v2.60	exon	1	50	1.8	+	0	Parent=1_1
+NC_7	Prodigal_v2.60	exon	56	100	1.8	+	0	Parent=1_1
```
mortonjt: Can we add another row, where one of the entries has a space in it? I know that these scenarios do happen in reality, and they messed up my first round of gff parsers back in the day. I just can't think of an example off of the top of my head ...

skbio/io/format/gff3.py
RNAer: the line would be too long

skbio/io/format/tests/data/gff3_multi_record
RNAer: sure.

RNAer:
@mortonjt , could you indicate where in the documentation you recommend to remove the "backspaces"?

RNAer:
@wasade and @mortonjt , comments are addressed. btw, the doc renders very well.

CHANGELOG.md
```diff
@@ -1,8 +1,10 @@
 # scikit-bio changelog
 
-## Version 0.5.1-dev (changes since 0.5.1 go here)
+## Version 0.5.1-dev  (changes since 0.5.1 go here)
```
wasade: why the added space?

skbio/io/format/gff3.py
wasade: @RNAer this is critical to address

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,534 @@
...
+@gff3.writer(Sequence)
+def _sequence_to_gff3(obj, fh, skip_subregion=True):
+    # write file header
+    fh.write('##gff-version 3\n')
+    _serialize_seq(obj, fh, skip_subregion)
+
+
+@gff3.reader(DNA)
+def _gff3_to_dna(fh, seq_num=1):
+    ''''''
```
wasade: why empty doc strings?

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,534 @@
...
+    # write file header
+    fh.write('##gff-version 3\n')
+    _serialize_seq(obj, fh, skip_subregion)
+
+
+@gff3.reader(IntervalMetadata)
+def _gff3_to_interval_metadata(fh, seq_id):
+    '''Read a GFF3 record into the specified interval metadata.
+
+    Parameters
```
wasade: If a `Parameters` section is to be included, please describe all parameters

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,534 @@
...
+                    'Unknown section in the input GFF3 file: '
+                    '%r %r %r' % (data_type, sid, data))
+    # return an empty instead of None
+    return IntervalMetadata(None)
+
+
+@gff3.writer(IntervalMetadata)
+def _interval_metadata_to_gff3(obj, fh, seq_id, skip_subregion=True):
+    '''
+    Parameters
```
wasade: same. also, if a docstring is to be included, please provide a 1-liner

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,534 @@
...
+    _serialize_interval_metadata(obj, seq_id, fh, skip_subregion=True)
+
+
+def _construct_seq(fh, constructor=DNA, seq_num=1):
+    lines = []
+    for i, (data_type, seq_id, l) in enumerate(_yield_record(fh), 1):
+        if data_type == 'data' and seq_num == i:
+            lines = l
+    # you can't read directly from fh because in registry.py line 543
+    # file.tell() will fail "telling position disabled by next() call".
```
wasade: @ebolyen just wanted to draw your attention

skbio/io/format/gff3.py
```diff
@@ -0,0 +1,534 @@
...
+                lines.append(line)
+            else:
+                if current is not False:
+                    yield 'data', current, lines
+                lines = [line]
+                current = seq_id
+    if current is False:
+        # if the input file object is empty, it should return
+        # an empty generator
+        return
```
wasade: is the `return` necessary? somewhat surprised this is valid to have within a function that has `yield`

skbio/io/format/gff3.py
wasade: eg 

```python
def _serialize_interval_metadata(interval_metadata, seq_id, fh,
                                 skip_subregion=True):
```

CHANGELOG.md
RNAer: darn, this is an accident I missed.

skbio/io/format/gff3.py
RNAer: thought it was just a comment rather than a request of change, because somehow it was and is still only showing your comments out of context without Jai's first comment. now I am able to see them all with some tricks. rewording.

skbio/io/format/gff3.py
RNAer: residues from previous modification. deleting

skbio/io/format/gff3.py
RNAer: yes! Without `return`, the function yield `None` (which is not what we want) instead nothing. `return\nyield` is the python syntax to return a empty generator.

RNAer:
@wasade , thanks for your comments! sorry for missing a couple of ones, especially the word-by-word copy of 1st paragraph. surprised on how easy to miss things while doing a large PR. it should be ready for a round, hopefully the last.

RNAer:
Thanks, @wasade. do I need any more approval to merge?

wasade:
@RNAer it looks like @jairideout still needs to approve

RNAer:
@jairideout , do you need to review and approve? or the approval of @wasade is enough to merge?

jairideout:
Approved, thanks for the ping! Note: I didn't take a final pass through since @mortonjt and @wasade took over the PR awhile back. Thanks everyone!

skbio/io/format/gff3.py
ebolyen: That sounds right, `.tell()` is disable in python 3 after iteration (which it should be!) because `tell` reflects the C read-ahead buffer which is wholly unrelated to what the iterator of the `for` loop is up to.

Since we are composing filehandles in non-trivial ways, it's not too strange to see this happening. The registry is invoked to make `fh` in the first place, and then invoked again on the `read` call. It is using `tell` and `seek` to try and preserve the `fh` as best it can, this is supposed to be happening before the file-handle is really touched. It is kind of an unfortunate situation as `tell` is only accurate if you seek without reading. Ideally the registry would reset the position to `0` instead of using `tell`, but then this kind of composition would be impossible anyway as it wouldn't ever be able to start in the middle of a file. Meaning you'd still need to split (just like `StringIO` below). Really the only way to avoid this is to not compose sniffing, or invent our own file-handle object in `C` which reflects the Python pointer position instead of the buffer.

Given that, I think we can avoid this if we were to disable format verification, so if you set `verify=False`, you should be able to use `fh` directly because skbio would have no reason to reset the handle's position (because it never sniffed).

ebolyen: Another alternative would be to use the underlying functions that can consume fasta, I think they are in `_base.py` skipping the registry altogether (might be the best option).

ebolyen: Sorry I keep thinking of things: we could also probably use `read(fh.readlines(), ...)` (although runtime will be identical because it has to make a `StringIO` object underneath the hood). That would be better though, as it doesn't appear `stream` is closed, so it's technically leaking a "fake" resource right now.

RNAer:
Thanks, @ebolyen! Following your thoughts, I changed it to use the
underlying functions to parse fasta seq. Would you mind taking a look to
see if you agree?

skbio/io/format/gff3.py
```diff
@@ -362,10 +362,8 @@ def _construct_seq(fh, constructor=DNA, seq_num=1):
     for i, (data_type, seq_id, l) in enumerate(_yield_record(fh), 1):
         if data_type == 'data' and seq_num == i:
             lines = l
-    # you can't read directly from fh because in registry.py line 543
-    # file.tell() will fail "telling position disabled by next() call".
-    stream = StringIO(fh.read())
-    seq = read(stream, format='fasta', into=constructor, seq_num=seq_num)
+    seq = _get_nth_sequence(_fasta_to_generator(fh, constructor=constructor),
```
ebolyen: So technically `_fasta_to_generator` is wrapped by the decoration and will invoke the `IO` registry again. But it doesn't have any reason to sniff now which is why it is working. Regardless, I do think this is better (even though it is still re-entrant).

Looking at `fasta.py` it looks like the only way to really get the functionality you'd want without re-entering the registry would be to pull out: 
```python
for seq, id_, desc in _parse_fasta_raw(fh, _parse_sequence_data,
                                       FASTAFormatError):
    yield constructor(seq, metadata={'id': id_, 'description': desc},
                      **kwargs)
```
from `_fasta_to_generator` which probably isn't worth it, but that's up to you.

It may even be argued that decorated IO functions shouldn't actually be wrapped, which would make your code work better... I don't think there's a reason they need to be wrapped as they aren't really public functions. The registry shouldn't care either way. 

As I think about this, that is probably the way to go... So I'm good with this implementation and in the future we can just return the original function which will save some time.

wasade:
:party: woo! @RNAer super awesome!

RNAer:
yay! Can't be done without your contributions, @jairideout , @ebolyen , @wasade and @mortonjt !

PR #1451, opened 2016-11-11
add merge method to IntervalMetadata

RNAer:


skbio/metadata/_interval.py
```diff
@@ -633,6 +633,28 @@ def concat(cls, interval_metadata):
...
+        ``self``. Note this will not check if there are any duplicates
+        of interval features after merge.
+
+        Parameters
+        ----------
+        other : ``IntervalMetadata``
+            The other ``IntervalMetadata`` to be merged.
+
+        '''
+        if self.upper_bound != other.upper_bound:
```
jairideout: Can you check `lower_bound` too?

skbio/metadata/tests/test_interval.py
```diff
@@ -494,6 +494,28 @@ def test_concat(self):
...
+        self.assertEqual(self.im_empty, im)
+        # empty + non-empty
+        self.im_empty.merge(self.im_1)
+        self.assertEqual(self.im_empty, self.im_1)
+        # non-empty + non-empty
+        self.im_empty.merge(self.im_2)
+        self.im_2.merge(self.im_1)
+        self.assertEqual(self.im_empty, self.im_2)
+
+    def test_merge_unequal_upper_bounds(self):
```
jairideout: Can you add a test for the lower bound case? Coverage will drop otherwise

RNAer:
is it neccessary? it is set internally and unchangable. isn't the test just
"white elephant"?

RNAer:
better to leave the test for future use cases of lower_bound?

jairideout:
See my previous comment:

> Coverage will drop otherwise

And in case the value changes in the future.

RNAer:
The coverage is only marginally dropped (0.01%)? And it is essentially tested in the current test, ie test no exception raised. If you have idea to test of exception raising, I'd be happy to add

jairideout:
To test this you could make a `IntervalMetadata._lower_bound` attribute that is returned by the `lower_bound` getter, and then modify that attribute in the unit test. Probably too much work for 1 line of coverage though.

Thanks for your work on this!

RNAer:
Thanks, @jairideout !

PR #1406, opened 2016-07-26
Export condensed distance matrix to a pandas object #1397

virginiasaulnier:
I've implemented to_series to create a Pandas series for DissimilarityMatrix in stats.distance _base.py.  I've also created 3 tests in test_base.py.  All tests pass.

PR #1453, opened 2016-11-11
ENH: add `DistanceMatrix.to_series` method

### LINKED ISSUE #1397: Export condensed distance matrix to a pandas object
It would be awesome if distance matrices could be converted between a sparce format (condensed with sample ids) and the square format, possibly using the `condensed` function. Ideally, rather than a numpy array, the condensed format could be a pandas Series or DataFrame with a multidimensional index.

jairideout:
This commit is based on the work by @virginiasaulnier in
https://github.com/biocore/scikit-bio/pull/1406 with final review changes made.

Fixes #1397.


jairideout:
I'm merging this because the code had final review in #1406 and I addressed those changes here (all minor).

PR #1457, opened 2016-11-13
MAINT: remove broken/legacy doctests from _intersection.pyx

jairideout:
These doctests were written to work with the bx-python intersection code before it was ported to scikit-bio (import paths are incorrect). After fixing imports, the doctests don't work in Python 3 due to unordered `dict` repr.

I'm removing the doctests instead of fixing because they are documenting package-private Cython code. For some reason, doctests in .pyx files aren't executed in Travis-CI tests, nor in source installs from PyPI. They *are* executed when running tests on a conda-built package, so I only noticed the failing doctests after I had built the 0.5.1 conda release.

If doctests in .pyx code is desirable in the future, we'll need to make sure those tests are executed by Travis-CI in order to catch any failing tests before releasing.


jairideout:
@RNAer or @ebolyen, could one of you guys review?

RNAer:
Thanks for looking into this! Did those doctests cause any travis error/test failure? I view them as valuable to demo the usage and they helped me understand the code a lot.

BTW, I did a quick google and see the work around for pyx doctest here: https://github.com/cython/cython/wiki/FAQ#how-can-i-run-doctests-in-cython-code-pyx-files

jairideout:
> Did those doctests cause any travis error/test failure? I view them as valuable to demo the usage and they helped me understand the code a lot.

See what I wrote above:

> For some reason, doctests in .pyx files aren't executed in Travis-CI tests, nor in source installs from PyPI. They are executed when running tests on a conda-built package, so I only noticed the failing doctests after I had built the 0.5.1 conda release.

Thanks for the link to getting Cython doctests to execute. I don't have bandwidth to get these tests running in Travis-CI and PyPI source installs. For now I recommend removing the doctests b/c they are failing and we don't have a way to verify them until the conda package has been built (and at that point in the release process, it is a *major* pain to go back and fix the tests). IMO it's better to remove them until the problem is solved generally, in case there's another release in the meantime.

It'd be great to solve this generally so .pyx doctests are discovered and verified like the other .py doctests. I created an issue to track progress on this: https://github.com/biocore/scikit-bio/issues/1458

RNAer:
yeah, I had seen you said it failed in conda build. I assume you get the doctest error when checking the build in the `test` section in the recipe? As long as all the travis and test pass, to check the conda build success, I think it is safe to run nosestests without `--with-doctest` to skip the doctest? Anyway, this is minor and I am ok with deleting them now.

jairideout:
> I assume you get the doctest error when checking the build in the test section in the recipe?

No, the conda build (including whatever tests conda runs) is successful. The failing tests are only discovered and executed after installing scikit-bio from conda and running `python -m skbio.test`.

> As long as all the travis and test pass, to check the conda build success, I think it is safe to run nosestests without --with-doctest to skip the doctest? 

Travis *is not* running those doctests, so we don't see failures until the conda package is built and tested with the command I posted above. `python -m skbio.test` doesn't have an option to turn off doctests. Even if it did, I don't think that's the right solution here -- all doctests in the codebase (defined in .py and .pyx files) need to be executed during the Travis-CI build, otherwise broken doctests will make it into the codebase and be discovered much later on, when it is too late to go back and fix them (i.e. like what happened last night when I cut the release).

> Anyway, this is minor and I am ok with deleting them now.

I will leave this up to you -- if the broken tests are left in the codebase, we'll need to fix them and resolve #1458 before the next release. I don't have bandwidth to do that but if you have time it would be nice to get that fixed.

RNAer:
I can try to fix it before the next release, which is ~5 months away? from what I can see from the past release cycle.

jairideout:
We don't really have a scheduled release cycle. I relabeled #1458 as "blocking" because those issues always get fixed before releases. Do you want me to close this PR and assign #1458 to you?

RNAer:
That sounds good.

jairideout:
Done, thanks!

PR #1460, opened 2016-11-18
allow None for IntervalMetadata.upper_bound

RNAer:
This is to address #1450 , where the upper bound of the interval features (or seq length) is unknown at the time the object is created.

It adds a get_right_bound() function in the cython code to get the max value from the interval tree. Setting upper_bound to `None` is now more efficient than setting it to a actual value because it avoids the repetitive bound check for adding interval to the intervalmetadata. It only checks once when upper_bound is set.

@ebolyen or @jairideout , this is ready for review.

### DIFF
--- a/skbio/metadata/_intersection.pyx
+++ b/skbio/metadata/_intersection.pyx
@@ -202,6 +202,15 @@ cdef class IntervalNode:
         if self.cright is not EmptyNode and self.start <= qend:
             self.cright._intersect( start, end, results )
 
+    cpdef int _right_bound(IntervalNode self):
+        cdef IntervalNode node = self
+        while True:
+            i = node.cright
+            if i is not EmptyNode:
+                node = i
+            else:
+                return node.end
+
     cpdef void update(IntervalNode self, int start, int end,
                       object old_feature, object new_feature):
         """
@@ -416,6 +425,12 @@ cdef class IntervalTree:
         root = None
 
     # ---- Position based interfaces -----------------------------------------
+
+    def get_right_bound( self ):
+        '''Get the right bound'''
+        if self.root is None:
+            return None
+        return self.root._right_bound()
 
     def insert( self, int start, int end, object value=None ):
         """
--- a/skbio/metadata/_intersection.pyx
+++ b/skbio/metadata/_intersection.pyx
@@ -202,15 +202,6 @@ cdef class IntervalNode:
         if self.cright is not EmptyNode and self.start <= qend:
             self.cright._intersect( start, end, results )
 
-    cpdef int _right_bound(IntervalNode self):
-        cdef IntervalNode node = self
-        while True:
-            i = node.cright
-            if i is not EmptyNode:
-                node = i
-            else:
-                return node.end
-
     cpdef void update(IntervalNode self, int start, int end,
                       object old_feature, object new_feature):
         """
@@ -425,12 +416,6 @@ cdef class IntervalTree:
         root = None
 
     # ---- Position based interfaces -----------------------------------------
-
-    def get_right_bound( self ):
-        '''Get the right bound'''
-        if self.root is None:
-            return None
-        return self.root._right_bound()
 
     def insert( self, int start, int end, object value=None ):
         """
--- a/skbio/metadata/_interval.py
+++ b/skbio/metadata/_interval.py
@@ -386,6 +386,21 @@ fuzzy=[(False, False), (False, False)], metadata={'name': 'genA'})
         return self._interval_metadata is None
 
 
+def _rebuild_tree(method):
+    """Rebuild the IntervalTree."""
+    @functools.wraps(method)
+    def inner(interval_metadata, *args, **kwargs):
+        if interval_metadata._is_stale_tree is False:
+            return method(interval_metadata, *args, **kwargs)
+        interval_metadata._interval_tree = IntervalTree()
+        for f in interval_metadata._intervals:
+            for start, end in f.bounds:
+                interval_metadata._interval_tree.add(start, end, f)
+        interval_metadata._is_stale_tree = False
+        return method(interval_metadata, *args, **kwargs)
+    return inner
+
+
 class IntervalMetadata():
     """Stores the interval features.
 
@@ -402,11 +417,12 @@ class IntervalMetadata():
     ----------
     upper_bound : int or None
         Defines the exclusive upper bound of the interval features. No
-        coordinate can be greater than it. It is ``None`` by default,
-        meaning there is no upper bound to the coordinate space.
-    copy_from : IntervalMetadata or None
+        coordinate can be greater than it. ``None``
+        means that the coordinate space is unbounded.
+    copy_from : IntervalMetadata or None, optional
         Create a new object from the input ``IntervalMetadata`` object by
-        shallow copying if it is not ``None``.
+        shallow copying if it is not ``None``. The upper bound of the new
+        object will be updated with the ``upper_bound`` parameter specified.
 
     Notes
     -----
@@ -522,27 +538,11 @@ fuzzy=[(False, False)], metadata={'gene': 'sagB'})
 
         if copy_from is not None:
             for interval in copy_from._intervals:
-                # Only need to shallow-copy `bounds` and `fuzzy`
-                # because their elements are immutable.
                 bounds_cp = interval.bounds[:]
                 fuzzy_cp = interval.fuzzy[:]
                 metadata_cp = copy.copy(interval.metadata)
 
                 self.add(bounds_cp, fuzzy=fuzzy_cp, metadata=metadata_cp)
-
-    def _rebuild_tree(method):
-        """Rebuild the IntervalTree."""
-        @functools.wraps(method)
-        def inner(interval_metadata, *args, **kwargs):
-            if interval_metadata._is_stale_tree is False:
-                return method(interval_metadata, *args, **kwargs)
-            interval_metadata._interval_tree = IntervalTree()
-            for f in interval_metadata._intervals:
-                for start, end in f.bounds:
-                    interval_metadata._interval_tree.add(start, end, f)
-            interval_metadata._is_stale_tree = False
-            return method(interval_metadata, *args, **kwargs)
-        return inner
 
     @property
     @experimental(as_of='0.5.1')
@@ -583,7 +583,8 @@ fuzzy=[(False, False)], metadata={'gene': 'sagB'})
         self._is_stale_tree = True
 
     @classonlymethod
-    def _concat(cls, interval_metadata):
+    @experimental(as_of="0.5.1")
+    def concat(cls, interval_metadata):
         '''Concatenate an iterable of ``IntervalMetadata`` objects.
 
         It concatenates the multiple ``IntervalMetadata`` objects into
@@ -620,7 +621,7 @@ fuzzy=[(False, False)], metadata={'gene': 'sagB'})
         up. The resulting ``IntervalMetadata``'s upper bound is the
         sum of upper bounds of concatenated objects:
 
-        >>> im = IntervalMetadata._concat([im1, im2])
+        >>> im = IntervalMetadata.concat([im1, im2])
         >>> im   # doctest: +ELLIPSIS
         2 interval features
         -------------------
@@ -667,17 +668,31 @@ fuzzy=[(True, True)], metadata={'gene': 'sagB'})
         ``self``. Note this will not check if there are any duplicates
         of interval features after merge.
 
+        Notes
+        -----
+        It will raise error if you merge an unbounded
+        ``IntervalMetadata`` object to the current object if it is
+        bounded. This avoids partially updating the current object if
+        the merge fails in the middle of the process due to the
+        possibility that some interval features to be merged may live
+        outside the current defined upper bound.
+
         Parameters
         ----------
         other : ``IntervalMetadata``
             The other ``IntervalMetadata`` to be merged.
 
         '''
-        if self.upper_bound != other.upper_bound:
-            raise ValueError(
-                'The upper bounds of the two IntervalMetadata objects '
-                'are not equal (%d != %d)' % (
-                    self.upper_bound, other.upper_bound))
+        if self.upper_bound is not None:
+            if other.upper_bound is None:
+                raise ValueError(
+                    'Cannot merge an unbound IntervalMetadata object '
+                    'to a bounded one')
+            elif self.upper_bound != other.upper_bound:
+                raise ValueError(
+                    'The upper bounds of the two IntervalMetadata objects '
+                    'are not equal (%r != %r)' % (
+                        self.upper_bound, other.upper_bound))
         if self.lower_bound != other.lower_bound:
             raise ValueError(
                 'The lower bounds of the two IntervalMetadata objects '
--- a/skbio/metadata/_interval.py
+++ b/skbio/metadata/_interval.py
@@ -518,41 +518,6 @@ fuzzy=[(False, False)], metadata={'gene': 'sagB'})
         # Indicates if the IntervalTree needs to be rebuilt.
         self._is_stale_tree = False
 
-    @property
-    @experimental(as_of='0.5.1')
-    def upper_bound(self):
-        '''The exclusive upper bound of interval features.'''
-        return self._upper_bound
-
-    @upper_bound.setter
-    @experimental(as_of='0.5.1')
-    def upper_bound(self, value):
-        '''Set exclusive upper bound of interval features.
-
-        It checks all the coordinates of the interval features are not
-        bigger than the new value.
-
-        '''
-        for intvl in self._intervals:
-            if intvl.bounds[-1][-1] > value:
-                raise ValueError(
-                    'Cannot set upper bound to %r because it is smaller '
-                    'than the bounds of interval feature: %r' %
-                    (value, intvl))
-        self._upper_bound = value
-
-    @property
-    @experimental(as_of='0.5.1')
-    def lower_bound(self):
-        '''The inclusive lower bound of interval features.'''
-        return 0
-
-    @property
-    @experimental(as_of='0.5.1')
-    def num_interval_features(self):
-        '''The total number of interval features.'''
-        return len(self._intervals)
-
     def _rebuild_tree(method):
         """Rebuild the IntervalTree."""
         @functools.wraps(method)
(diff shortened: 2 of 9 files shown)


jairideout:
Thanks @RNAer! See my [comment](https://github.com/biocore/scikit-bio/pull/1450#issuecomment-262592865) on #1450 about a potential change to this API. Also, please add a changelog note.

skbio/metadata/_interval.py
```diff
@@ -396,9 +400,10 @@ class IntervalMetadata():
 
     Parameters
     ----------
-    upper_bound : int
+    upper_bound : int or None
         Defines the exclusive upper bound of the interval features. No
-        coordinate can be greater than it.
+        coordinate can be greater than it. It is ``None`` by default,
```
jairideout: I'd rephrase this to something like:

```
Defines the exclusive upper bound of the interval features. No coordinate can be greater than it. It is ``None`` by default, meaning there is no upper bound to the coordinate space.
```

skbio/metadata/_interval.py
```diff
@@ -530,38 +569,29 @@ def num_interval_features(self):
...
         in the forward strand to coordinates in the reversal strand.
         """
+
         for f in self._intervals:
-            intvls = [(self.upper_bound - x[1], self.upper_bound - x[0])
-                      for x in reversed(f.bounds)]
+            try:
+                intvls = [(self.upper_bound - x[1], self.upper_bound - x[0])
+                          for x in reversed(f.bounds)]
+            except TypeError:
```
wasade: can an explicit test for `upper_bound is None` be performed prior to the for loop? That would remove the need for the try/except.

skbio/metadata/_interval.py
```diff
@@ -530,38 +569,29 @@ def num_interval_features(self):
...
+                                'when the upper bound is `None`')
             f.bounds = intvls
 
         # DONT' forget this!!!
         self._is_stale_tree = True
 
     @classonlymethod
     @experimental(as_of="0.5.1")
-    def concat(cls, interval_metadata):
+    def _concat(cls, interval_metadata):
```
wasade: since it is now private, should the `@experimental` decorator be removed?

RNAer: good catch!

skbio/metadata/_interval.py
RNAer: I feel like this is more pythonic because of [EAFP](https://docs.python.org/3/glossary.html#term-eafp)?

skbio/metadata/_interval.py
```diff
@@ -264,11 +264,15 @@ def _check_bounds(self, bounds):
...
-        if bounds[-1][-1] > upper_bound or bounds[0][0] < lower_bound:
+        if upper_bound is not None and bounds[-1][-1] > upper_bound:
             raise ValueError('Cannot set `bounds` (%r) with coordinate '
-                             'larger than upper bound (%r) or smaller than '
-                             'lower bound (%r).' %
-                             (bounds, upper_bound, lower_bound))
+                             'larger than upper bound (%r)' %
+                             (bounds, upper_bound))
+
+        if bounds[0][0] < lower_bound:
```
RNAer: lower_bound is essentially a stub currently, leaving it there for future use cases. would prefer to leave it as it is.

skbio/metadata/_interval.py
```diff
@@ -530,38 +569,29 @@ def num_interval_features(self):
...
-                      for x in reversed(f.bounds)]
+            try:
+                intvls = [(self.upper_bound - x[1], self.upper_bound - x[0])
+                          for x in reversed(f.bounds)]
+            except TypeError:
+                raise TypeError('You cannot reverse the coordinates '
+                                'when the upper bound is `None`')
             f.bounds = intvls
 
         # DONT' forget this!!!
```
RNAer: good catch on the typo!

RNAer:
@jairideout and @ebolyen , added the alternative `IntervalMetadata` constructor for upper bound update purpose (from line 535-563). One thing I am not sure is the API, whether it should be a static method (as currently defined) or a normal method (`def clone(self, upper_bound) -> new IntervalMetadata`). It seems very subtle to me. any suggestion?

skbio/metadata/_interval.py
wasade: kk

wasade:
@RNAer, a few test failures btw

RNAer:
Thanks for looking at it, @wasade ! I am aware of the test failure. wanna get feedback on the alternative constructor before fixing them.

jairideout:
What about adding an optional `from=None` parameter to the existing constructor, where `from` accepts an `IntervalMetadata` object?

```python
def __init__(self, upper_bound, from=None):
    ...
```

RNAer:
That sounds good too - does the `copy` works in `__init__`? Anyway, I still need a function to do the heavy-lifting construction because the `_rebuild_tree` decorator needs to be applied to it (line 536, 553). Prone to define it as normal class method instead of a static

jairideout:
I was thinking the constructor would behave similar to `__copy__` by 1) setting `self._upper_bound` and 2) looping over the `Interval` objects and calling `self.add()`. `add()` already takes care of the error-checking and can be reused in the constructor.

RNAer:
ah, sorry, you are right. I was overthinking and forgetting `add()` does checking.

RNAer:
I think this is ready for final review. thanks for all the comments!

skbio/metadata/_testing.py
```diff
@@ -922,6 +922,15 @@ def test_constructor_invalid(self):
                                     'object.'):
             self._interval_metadata_constructor_(0, '')
 
+    def test_constructor_interval_metadata_upper_bound_is_none(self):
```
jairideout: Please add these test cases:

- successfully constructing from a non-empty `IntervalMetadata` object with `upper_bound=None`
- failing to construct from a non-empty `IntervalMetadata` object with `upper_bound=None` and coordinates that fall outside the mixin's coordinate space

Please also add similar test cases for the mixin setter (including `test_constructor_interval_metadata_upper_bound_is_none`, that would be 3 test cases for the setter).

skbio/metadata/_interval.py
```diff
@@ -530,38 +562,28 @@ def num_interval_features(self):
...
             f.bounds = intvls
 
-        # DONT' forget this!!!
+        # DON'T forget this!!!
         self._is_stale_tree = True
 
     @classonlymethod
-    @experimental(as_of="0.5.1")
-    def concat(cls, interval_metadata):
+    def _concat(cls, interval_metadata):
```
jairideout: Does `merge()` need a similar update to account for `upper_bound=None`?

jairideout: Why is `concat` private now?

skbio/metadata/_interval.py
```diff
@@ -396,9 +400,13 @@ class IntervalMetadata():
 
     Parameters
     ----------
-    upper_bound : int
+    upper_bound : int or None
         Defines the exclusive upper bound of the interval features. No
-        coordinate can be greater than it.
+        coordinate can be greater than it. It is ``None`` by default,
+        meaning there is no upper bound to the coordinate space.
+    copy_from : IntervalMetadata or None
```
jairideout: `copy_from : IntervalMetadata or None, optional`

skbio/metadata/_interval.py
```diff
@@ -512,6 +520,30 @@ def __init__(self, upper_bound):
...
+            for interval in copy_from._intervals:
+                # Only need to shallow-copy `bounds` and `fuzzy`
+                # because their elements are immutable.
+                bounds_cp = interval.bounds[:]
+                fuzzy_cp = interval.fuzzy[:]
+                metadata_cp = copy.copy(interval.metadata)
+
+                self.add(bounds_cp, fuzzy=fuzzy_cp, metadata=metadata_cp)
+
+    def _rebuild_tree(method):
```
RNAer: You are talking about renaming from `self` to `interval_metadata`? That's the only change I made. It can stay it was. But I feel like it makes more sense with the new name. BTW, I think this function better live outside the class? Just to be clear, it can't be `@staticmethod` decorated.

RNAer:
Thanks, @jairideout ! I replied to a few of your comments - would you mind responding? I will address all your other comments.

skbio/metadata/_interval.py
jairideout: I'd prefer to leave `_rebuild_tree` as it was since it was working fine

skbio/metadata/_interval.py
jairideout: > because @ebolyen and I think the only use case of _concat is inside the sequence objects.

I thought you were arguing earlier against using private APIs in other parts of the codebase? Following that logic, the mixin would prefer to use `concat` over a private `_concat`.

> I don't see use case for merge() with upper bound of None yet. hesitate to add it. let's add it when necessary?

You updated `concat` to explicitly handle `upper_bound=None`. We need to make sure `merge` handles it appropriately too (they are similar operations), even if you don't have a use-case for it. Other users can try to `merge` with `upper_bound=None` and it needs to work correctly.

skbio/metadata/_intersection.pyx
```diff
@@ -417,6 +426,12 @@ cdef class IntervalTree:
 
     # ---- Position based interfaces -----------------------------------------
 
+    def get_right_bound( self ):
```
jairideout: I think it should be removed. The method isn't necessary for this PR, and adding unused functionality increases maintenance burden.

jairideout:
Yep, I just followed up

skbio/metadata/_interval.py
RNAer: > I thought you were arguing earlier against using private APIs in other parts of the codebase? Following that logic, the mixin would prefer to use concat over a private _concat.

I argued for public over private APIs if both exist, which is a bit different from the situation here. That being said, I don't have strong feeling keep it private or public. will revert it to public.

> You updated concat to explicitly handle upper_bound=None. We need to make sure merge handles it appropriately too (they are similar operations), even if you don't have a use-case for it. Other users can try to merge with upper_bound=None and it needs to work correctly.

ok. will add it, although I think it's fine to raise error if `upper_bound=None`.

skbio/metadata/_interval.py
RNAer: The more strongly I feel like putting it outside the class, the more I look at it, because it is more logic. "it was working fine" should not stop us improving it?

RNAer:
@jairideout , does changelog really need update? Because he API changes is not new and compatible backwards. If so, then which section should I put it?

RNAer:
@jairideout , all comments are addressed except for changelog

jairideout:
> @jairideout , does changelog really need update? Because he API changes is not new and compatible backwards. If so, then which section should I put it?

Yes. As detailed in the [code review guidelines](https://github.com/biocore/scikit-bio/blob/master/REVIEWING.md#changelog) and in the [pull request template](https://github.com/biocore/scikit-bio/blob/master/PULL_REQUEST_TEMPLATE.md), **all public-facing changes** are documented in the changelog. There are several sections in the changelog to categorize changes; for example, "features", "bug fixes", "performance enhancements", "backwards-incompatible changes", etc. *Any* change visible to a user should be documented in the changelog, not just backwards-incompatible changes. This is why we have several sections to categorize the types of changes present in a release.

For this PR, there are two public-facing changes. Both are new features, so they would go under the "features" section of the changelog:

- `skbio.interval.IntervalMetadata` constructor has a new keyword argument, `copy_from`, for creating an `IntervalMetadata` object from an existing `IntervalMetadata` object with specified `upper_bound`.

- `skbio.interval.IntervalMetadata` constructor allows `None` as a valid value for `upper_bound`. An `upper_bound` of `None` means that the `IntervalMetadata` object has no upper bound.

skbio/metadata/_interval.py
jairideout: I don't see how moving this outside the class is an improvement. Now you have a private function that modifies private state on an object, which is not good practice.

RNAer:
`concat` is not parallel to `merge` function in the sense that `concat`
returns a new object while `merge` modify the current object and that
`concat` requires `upper_bound` defined to add up coordinate spaces while
`merge` works fine with unbounded upper bound. I added check in `merge` in
the previous comment. It should not raise an error neccessarily when
`upper_bound=None`. I explained in line 673. could you check if it makes
sense to you?

RNAer:
Note `_rebuild_tree` is a decorator and only apply to the function that
operate on `IntervalMetadata` object. It is not modifies the private state
of an object directly, only when it is used to decorate a function that
access interval tree. The motivation leaving it outside is, I think,
because it can be used on other functions outside the class.

RNAer:
I am clear now. Thanks for explanation! will add.

skbio/metadata/tests/test_interval.py
```diff
@@ -446,6 +446,34 @@ def test_init_upper_bound_lt_lower_bound(self):
         with self.assertRaises(ValueError):
             IntervalMetadata(-1)
 
+    def test_upper_bound_is_none(self):
```
jairideout: Can you address this comment? `IntervalMetadata` defines an `__eq__`, so you can create an expected `IntervalMetadata` object and use `self.assertEqual` to compare it to the observed `IntervalMetadata` object. It is safer and shorter than the current implementation.

RNAer: I thought I had addressed this comment? I can't use `__eq__` to directly assert equality of `IntervalMetadata` in these cases (eg line 471) because `__eq__` also compares `upper_bound`, which is not the same between the expected and observed objects. Thus, I have to use `_intervals`.

RNAer: ...unless you are ok not to check `upper/lower_bound` in `__eq__`, which probably makes more sense.

jairideout:
> `concat` is not parallel to `merge` function in the sense that `concat`
returns a new object while `merge` modify the current object and that
`concat` requires `upper_bound` defined to add up coordinate spaces while
`merge` works fine with unbounded upper bound. I added check in `merge` in
the previous comment. It should not raise an error neccessarily when
`upper_bound=None`. I explained in line 673. could you check if it makes
sense to you?

Makes sense, thanks.

> Note `_rebuild_tree` is a decorator and only apply to the function that
operate on `IntervalMetadata` object. It is not modifies the private state
of an object directly, only when it is used to decorate a function that
access interval tree. The motivation leaving it outside is, I think,
because it can be used on other functions outside the class.

I disagree -- the decorator is tightly coupled with `IntervalMetadata` and accesses private state on those objects (`_is_stale_tree`, `_interval_tree`, and `_intervals`), so it should continue to exist as a method on `IntervalMetadata` to avoid breaking encapsulation.

The decorator is only being applied to `IntervalMetadata` methods right now, but if it were to be applied to other functions in the future, it is possible to use the decorator as it is right now (i.e. defined as a method instead of a function):

```python
@IntervalMetadata._rebuild_tree
def foo():
    ...
```

skbio/metadata/tests/test_interval.py
jairideout: `__eq__` should continue to check `upper_bound` because bounded vs unbounded interval metadata objects shouldn't compare equal.

You can define an *expected* `IntervalMetadata` object that has the expected `upper_bound` (in addition to whatever other state is expected), and use `self.assertEqual` to compare observed to expected. Is there a reason this strategy won't work? This is a really common pattern in unit testing.

RNAer:
One issue with putting `_rebuild_tree` inside class is: you can prevent people to use it as a normal method `instance._rebuild_tree(xxx)` (although this is a private func). It causes error because it can't not be decorated with `staticmethod`.

skbio/metadata/tests/test_interval.py
RNAer: In those tests, the private attributes have to be used any way and the code is longer to create expected objects, but I agree with you that it'd be better to use `__eq__` in case the internal attributes change in the future. I reorganize and modify the tests to accommodate your request.

jairideout:
> One issue with putting _rebuild_tree inside class is: you can not prevent people to use it as a normal method like instance._rebuild_tree(xxx) (although this is a private func). It causes error because it can't be decorated with staticmethod to avoid this.

This is a problem with any private method, because nothing is truly private in Python. Even if you refactor it into a private function, users can still call that private function.

If a tree-rebuilding API is desired, it would be better to promote this API to a public method instead of a decorator. Then `self.rebuild_tree()` can be called appropriately within `IntervalMetadata` methods, and users outside the class can use it if desired. It sounds like it could be a useful public API in its own right, as a user may want to explicitly rebuild the tree before querying (or other operations).

I don't think the current refactoring of the private method into a private function has any benefits, and it has drawbacks that I noted in earlier comments. If you think refactoring the API into a public instance method is a good idea, it would be better to include that API change in a separate PR b/c it is orthogonal to the changes in this PR.

skbio/metadata/tests/test_interval.py
jairideout: Thank you for reworking the tests to use `__eq__`. By using this strategy, private attributes will only be used within `IntervalMetadata.__eq__` and not relied upon in the unit tests.

RNAer:
You misunderstood me. I was not talking about Python private. I was saying `_rebuild_tree` can't be decorated with `staticmethod` as I mentioned previously and its 1st parameter is not `self`, thus calling `instance._rebuild_tree` (which is perfectly legitimate syntax-wise) will cause error.

Multiple persons liked the idea of decorator before, because it isolates the tree build logic out and syntax-wise it is clean to just decorate the methods that access interval tree. I don't think we should reverse this design. And `_interval_tree` is private, we shouldn't expose this tree rebuild API to users; Otherwise, users may rebuild tree duplicately  or get confused whether they should rebuild tree before query or other operations accessing interval tree.

With that being said, if you still insist putting this back into the class, I can reverse this function to the previous state.

ebolyen:
@RNAer I see your concern about the `_rebuild_tree` method. It definitely isn't a "method" in its present context, so pulling it out as a function does initially seem logical to keep the decorator syntax. However I would argue that it is actually pretty simple to disambiguate the first argument so that we have a method which can act both as a decorator in a static calling context (inside class definition) and as proper method (still private) for rebuilding a tree in a bound context. This would preserve the encapsulation of the class which is something we work very hard to maintain in scikit-bio (not to mention having very intentional API boundaries). 

That being said, it seems like this code is frankly a little complicated for what is does, and a simple instance method like @jairideout suggested would suffice, and even make good sense as a part of the public API. 

In either case, it seems like this method isn't hurting any public API and it probably isn't worth discussing this much further in the context of this PR. Perhaps making a new PR specifically targeted around refactoring this decorator makes the most sense for moving forward.

RNAer:
agree with leaving it for future. I will reverse the change wrt it.

RNAer:
ready for final review? Thanks!

RNAer:
Thanks!

PR #1461, opened 2016-11-19
Updating references for compositional functions

mortonjt:
Replacing the old reference with a much more up to date reference.

The book can be found [here](http://onlinelibrary.wiley.com/book/10.1002/9781119003144) and has quite a few more updated samples.


jairideout:
Thanks @mortonjt!

Issue #1483, opened 2017-01-31
feature addition: add a parser for EMBL format


bunop:
Dear all, 
I think that adding a EMBL parser would be useful to deal with data stored in ENA. I think that I could use `skbio.io.format.genbank` and `skbio.io.format.tests.test_genbank` as a model to develop two new modules like `skbio.io.format.embl` and `skbio.io.format.tests.test_embl`. If I'm able, I will propose a pull requests with such features.
Regards,
Paolo

jairideout:
cc @RNAer

RNAer:
cc @hotdogee , are you working on this? If you already have it, could you issue a PR?

bunop:
Dear all,

regarding the embl parser, I have some consideration to make and would like to have some suggestion if I'm working properly or not.
* **Record sniffing**: First of all, for what concerns DNA and RNA data present in ENA, embl record have the first line like this:
```
ID   M14399; SV 1; linear; mRNA; STD; PRO; 63 BP.
```
That's look like to genbank record:
```
LOCUS       ECOALKP   63 bp   mRNA   linear   BCT   26-APR-1993
```
However, uniprot records look different, even they share the same embl format:
```
ID   001R_FRG3G              Reviewed;         256 AA.
```
For the moment, I've decided to handle only embl records and throw error when sniffing uniprot record (So, no protein support for the moment). I think that should be easy to write a uniprot sniffer in a different module and to import common functions from embl; However, the type of information parsed from the first column will be different (the example is the same RNA segment, took from embl and genbank - the same used in unittest):
```python
embl.metadata['LOCUS']
{'accession': 'M14399',
 'class': 'STD',
 'division': 'PRO',
 'mol_type': 'mRNA',
 'shape': 'linear',
 'size': 63,
 'unit': 'bp',
 'version': 1}
genbank.metadata['LOCUS']
{'date': '26-APR-1993',
 'division': 'BCT',
 'locus_name': 'ECOALKP',
 'mol_type': 'mRNA',
 'shape': 'linear',
 'size': 63,
 'unit': 'bp'}
```
There are informations, like `date` that can't be derived from the first line of embl file; Record like `division` are different since they represent a different type of grouping data by embl and genbank. I think I can't convert one record into another, for instance the `locus_name` attribute is not present in embl record. 
* **Record keys**: embl data have a record key on each line, using a two chars notation, ie:
```
ID   M14399; SV 1; linear; mRNA; STD; PRO; 63 BP.
[...]

PR #1490, opened 2017-02-07
BUG: fixed bug in _parse_loc_str - fixes #1489

bunop:
Now `_parse_loc_str` can deal with `loc_str` like `AB000684.1:<1..>275`

Please complete the following checklist:

* [x] I have read the guidelines in [CONTRIBUTING.md](https://github.com/biocore/scikit-bio/blob/master/CONTRIBUTING.md).

* [x] I have documented all public-facing changes in [CHANGELOG.md](https://github.com/biocore/scikit-bio/blob/master/CHANGELOG.md).

* [ ] **This pull request includes code, documentation, or other content derived from external source(s).** If this is the case, ensure the external source's license is compatible with scikit-bio's [license](https://github.com/biocore/scikit-bio/blob/master/COPYING.txt). Include the license in the `licenses` directory and add a comment in the code giving proper attribution. Ensure any other requirements set forth by the license and/or author are satisfied. **It is your responsibility to disclose code, documentation, or other content derived from external source(s).** If you have questions about whether something can be included in the project or how to give proper attribution, include those questions in your pull request and a reviewer will assist you.

* [x] **This pull request does not include code, documentation, or other content derived from external source(s).**

**Note:** [REVIEWING.md](https://github.com/biocore/scikit-bio/blob/master/REVIEWING.md) may also be helpful to see some of the things code reviewers will be verifying when reviewing your pull request.

PR #1495, opened 2017-02-27
Add embl support - References #1483

bunop:


bunop:
Add embl file support. Here are the key aspects:

* Added reader and writer method for DNA, Sequence, RNA. No support for protein
* Keys are converted into genbank when possible
* Keys are grouped in sections
* Feature metadata were stored as genbank features
* Minor fixes and upgrades

More information could be found at [issue-1483](https://github.com/biocore/scikit-bio/issues/1483)

bunop:
Uhm, seems that travis fails. Even if in configuration is specified `$PYTHON_VERSION=3.5` the command `conda create --yes -n env_name python=$PYTHON_VERSION --file ci/conda_requirements.txt` install the default python version (3.6 in travis - 2.7 in my environment - see the file location of failing modules). By forcing installation with:

```bash
$ conda create -n travis python=3.6
$ source activate travis
$ conda install --file ci/conda_requirements.txt
$ pip install -r ci/pip_requirements.txt
$ pip install . --no-deps
$ WITH_COVERAGE=TRUE make test
```
I can reproduce error in local

bunop:
I've opened another issue in which I solved (I hope) the failing tests. If you could give me a feedback, I can include this fixes in this pull request.
Thank you for your support,
Regards,
Paolo

wasade:
Thank you @bunop!

wasade:
Note for reviewers, please read through the context in this [comment] prior to review (https://github.com/biocore/scikit-bio/issues/1483#issuecomment-282692986).

skbio/io/__init__.py
```diff
@@ -57,8 +58,11 @@
    FileFormatError
    BLAST7FormatError
    ClustalFormatError
+   EMBLFormatError
    FASTAFormatError
    FASTQFormatError
+   GenBankFormatError
+   GFF3FormatError
```
RNAer: 👍

skbio/io/format/_sequence_feature_vocabulary.py
```diff
@@ -255,6 +255,18 @@ def _parse_loc_str(loc_str):
...
+            start, end = interval.split('..')
+            fuzzy_s = fuzzy_e = False
+            if start.startswith('<'):
+                start = start[1:]
+                fuzzy_s = True
+            if end.startswith('>'):
+                end = end[1:]
+                fuzzy_e = True
+            bounds.append((int(start)-1, int(end)))
+            fuzzy.append((fuzzy_s, fuzzy_e))
```
RNAer: I see you incorporated you previous PR #1490 here. Pls see my comment in that PR and remove it from this PR.

skbio/io/format/genbank.py
```diff
@@ -45,6 +45,8 @@
 section, its value is stored as a list, as there are often multiple
 reference sections in one GenBank record.
 
+.. _genbank_feature_section:
+
```
RNAer: why adding these lines here?

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+
+Feature-level products contain nucleotide sequence and related annotation
+derived from submitted ENA assembled and annotated sequences. Data are
+distributed in flatfile format, similar to that of parent ENA records,
+with each flatfile representing a single feature. While only the sequence
+of the feature is included in such entries, features are derived from the
+parent entry, and can't be applied as interval metadata. For such reason,
+interval metatdata are ignored from Feature-level products, as they will be
+ignored by subsetting a generic Sequence object. More information on
+Feature-level product can be found here [3]_.
```
wasade: Some of this text appears to be copied from `_3`. Would it be possible to revise the text, or to provide an explicit statement indicating the text is reproduced from `_3`?

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+
+Format Support
+--------------
+**Has Sniffer: Yes**
+
+**NOTE: No protein support at the moment**
```
wasade: Can a corresponding issue be created and linked too in this note?

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+|Yes   |Yes   |:mod:`skbio.sequence.RNA`                                      |
++------+------+---------------------------------------------------------------+
+|No    |No    |:mod:`skbio.sequence.Protein`                                  |
++------+------+---------------------------------------------------------------+
+|Yes   |Yes   | generator of :mod:`skbio.sequence.Sequence` objects           |
++------+------+---------------------------------------------------------------+
+
+Format Specification
+--------------------
+**State: Experimental as of 0.5.1.**
```
wasade: 0.5.1-dev I believe

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+
+Sections before ``FH (Feature Header)``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+All the sections before ``FH (Feature Header)`` will be read into the attribute
+of ``metadata``. The header and its content of a section is stored as
+a pair of key and value in ``metadata``. For the ``RN (Reference Number)``
```
wasade: a pair of key and value -> key-value pairs

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
++------+------+---------------------------------------------------------------+
+
+Format Specification
+--------------------
+**State: Experimental as of 0.5.1.**
+
+Sections before ``FH (Feature Header)``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+All the sections before ``FH (Feature Header)`` will be read into the attribute
+of ``metadata``. The header and its content of a section is stored as
```
wasade: is -> are

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+                   'FT': 'FEATURES',
+                   'SQ': 'ORIGIN',
+                   }
+
+# the original genbank _yield_section divides entries in sections relying on
+# spaces (the same section has the same level of indentation). Uniprot entries
+# have a key for each line, so to divide record in sections I need to define a
+# corresponance for each key to section, then I will divide a record in
+# sections using these section name. Some keys are commented out since I don't
+# find them in example. What I have to do if I find them?
```
wasade: I'm not sure. @RNAer, do you know?

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+                   # sequence
+                   'SQ': 'ORIGIN',
+                   '  ': 'ORIGIN',
+                   # 'CO': 'ORIGIN,'
+                   # spacer (discarded)
+                   'XX': 'SPACER'
+                  }
+
+
+# for convenience: I think such functions are more readadble whiel accessing
```
wasade: whiel -> while

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+# valuies in lambda functions
```
wasade: valuies -> values

skbio/io/format/embl.py
wasade: I've always been on the fence with lambdas here. The justification can be found [here](https://www.python.org/dev/peps/pep-0008/). Part of the motivation is improved clarity on tracebacks.

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+def _get_embl_key(line):
+    """Return first part of a string as a embl key (ie 'AC M14399;' -> 'AC')"""
+
```
wasade: Remove please

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+# get embl key from value
+# http://stackoverflow.com/questions/8023306/get-key-by-value-in-dictionary
+def _get_embl_key_by_value(value, mydict=KEYS_TRANSLATOR):
+    """Return the key(s) associated to a value from a dictionary, or the
+    value if no key is defined"""
+
+    # try to get a key (keys) from a value in dictionary
+    try:
+        return list(mydict.keys())[list(mydict.values()).index(value)]
```
wasade: What about storing the inverse dict explicitly and just indexing in?

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+def _get_embl_key(line):
+    """Return first part of a string as a embl key (ie 'AC M14399;' -> 'AC')"""
+
+    # embl keys have a fixed size of 2 chars
+    return line[:2]
+
+
+# get embl key from value
+# http://stackoverflow.com/questions/8023306/get-key-by-value-in-dictionary
+def _get_embl_key_by_value(value, mydict=KEYS_TRANSLATOR):
```
wasade: it is worth being careful with default arguments:

```python
>>> d = {'foo': 'bar'}
>>> def f(x=d):
...   x['foo'] = 'baz'
...
>>> f()
>>> d
{'foo': 'baz'}
```

...effectively changing the default

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+    """Return the embl section from uniprot key(ie 'RA' -> 'REFERENCE')"""
+
+    # get embl key
+    key = _get_embl_key(line)
+
+    # get embl section from key
+    section = KEYS_2_SECTIONS[key]
+
+    # debug
+    # print("_get_embl_section line: >%s<" % (line))
```
wasade: remove please

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+
+    # a new dictionary of results
+    new_data = {}
+
+    # I can't replace keys in original values, sometimes test will fails. So, I
+    # create a new copy. This is a strange behaviour, I don't understand
+    for old_key in old_keys:
+        new_key = _translate_key(old_key)
+        new_data[new_key] = data[old_key]
+
```
wasade: can this be replaced with

```python
new_data = {_translate_key(k): v for k, v in data.items()}
```

?

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+    return _construct(record, RNA, **kwargs)
+
+
+# No protein support at the moment
+@embl.reader(Protein)
+def _embl_to_protein(fh, seq_num=1, **kwargs):
+    # no protein support, at the moment
+    # record = _get_nth_sequence(_parse_embls(fh), seq_num)
+    # return _construct(record, Protein, **kwargs)
+    raise EMBLFormatError("There's no protein support for EMBL record")
```
wasade: I think this is great to be explicit. When the issue is created (see comment above on the NOTE) would it be possible to add a mention of the issue into the exception message?

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+def _embl_to_rna(fh, seq_num=1, **kwargs):
+    record = _get_nth_sequence(_parse_embls(fh), seq_num)
+    return _construct(record, RNA, **kwargs)
+
+
+# No protein support at the moment
+@embl.reader(Protein)
+def _embl_to_protein(fh, seq_num=1, **kwargs):
+    # no protein support, at the moment
+    # record = _get_nth_sequence(_parse_embls(fh), seq_num)
```
wasade: remove please

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+
+@embl.writer(RNA)
+def _rna_to_embl(obj, fh):
+    _serialize_single_embl(obj, fh)
+
+
+@embl.writer(Protein)
+def _protein_to_embl(obj, fh):
+    # no protein support, at the moment
+    # _serialize_single_embl(obj, fh)
```
wasade: remove please

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+            # RNA mol type has T instead of U for genbank from from NCBI
+            constructor = DNA
+
+        elif unit == 'aa':
+            # no protein support, at the moment
+            # constructor = Protein
+            raise EMBLFormatError("There's no protein support for EMBL record")
+
+    if constructor == RNA:
+        return DNA(
```
wasade: It's not possible to construct `RNA` explicitly?

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+            parser = partial(
+                parser, join_delimiter="\n")
+
+        # call function on section
+        parsed = parser(section)
+
+        # reference can appear multiple times
+        if embl_section == 'REFERENCE':
+            # genbank data hasn't CROSS_REFERENCE section, To have a similar
+            # metatadata object, I chose to remove CCROSS_REFERENCE from
```
wasade: metatadata -> metadata

CCROSS_REFERENCE -> CROSS_REFERENCE

skbio/io/format/_sequence_feature_vocabulary.py
bunop: I saw your comment and I agree. I will remove this piece of code from this PR

skbio/io/format/genbank.py
bunop: Sorry, I couldn't comment this piece of code in sphinx documentation. This is a [inline markup](http://www.sphinx-doc.org/en/stable/markup/inline.html#cross-referencing-arbitrary-locations) to refer to an arbitrary location, and let me to reference to a precise point of another document in text by using
```
See :ref:`Genbank FEATURES section<genbank_feature_section>`
```
(note the absence of the initial `_` in the link). The markup need to be placed immediately before the title I need to refer to. My idea was, since features were read in the same way as genbank, to refer to genbank documentation to explain `FT section`. Feel free to ask me to modify this aspect and describe `FT` in embl module

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Version 0.5.1-dev (changes since 0.5.1 go here)
 
+* Added ``skbio.io.format.embl`` for reading and writing EMBL files for ``DNA``, ``RNA``, ``Protein`` and ``Sequence`` classes.
```
bunop: Yes. I will fix it. I wrote this line before deciding to concentrate myself on DNA

skbio/alignment/_indexing.py
```diff
@@ -180,7 +180,7 @@ def _slice_sequences(self, obj, indexable):
         self._assert_tuple_rules(indexable)
         if (self._has_fancy_index() and
                 type(indexable) is not tuple and
-                pd.core.common.is_list_like(indexable) and
+                pd.api.types.is_list_like(indexable) and
```
bunop: I made those changes because `pd.core.common.is_list_like` was deprecated, and this warning appeared many times in my tests. I found that the recommended module was `pd.core.common.is_list_like` (in my `pandas` library)

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
+"""
+EMBL format (:mod:`skbio.io.format.embl`)
+=========================================
+
+.. currentmodule:: skbio.io.format.embl
+
+EMBL format stores sequence and its annotation together. The start of the
+annotation section is marked by a line beginning with the word "ID". The start
+of sequence section is marked by a line beginning with the word "SQ" and the
+end of the section is marked by a line with only "//". More information on
```
bunop: more precisely, `//` denote the end of embl record. Maybe it need to be described also for genbank objects?

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+EMBL file format can be found here [1]_.
+
+The EMBL file may ends with .embl or .txt extension. An example of EMBL file
```
bunop: fixed

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+can be seen here [2]_.
+
+Feature level products
```
bunop: Ok

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1497 @@
...
+^^^^^^^^^^^^^^^^^^^^^^
+
+Feature-level products contain nucleotide sequence and related annotation
```
bunop: fixed

skbio/io/format/embl.py
bunop: I think that I can't be clearer than ENA, so I will render the text as a citation

skbio/io/format/embl.py
bunop: Do you mean that I need to open a ussue like "add uniprot support" and then cite it in this point?

skbio/io/format/embl.py
bunop: So I will point to last version of changelog

skbio/io/format/embl.py
bunop: I don't understand. Should I remove docstring?

skbio/io/format/embl.py
bunop: Ok removed. But the problems persists:
```python
In [1]: d = {'foo': 'bar'}

In [2]: def f(x):
   ...:     x['foo'] = 'baz'
   ...:     

In [3]: f(d)

In [4]: d
Out[4]: {'foo': 'baz'}
```
This is because all objects (not strings) are passed by reference, so `d1=d` will point `d1` to the same memory location of `d`. This is why I modify the reference dictionary in `_parse_single_embl` but I create a copy of this dictionary in `_serialize_reference`, otherwise my data changes after the first serialization

skbio/alignment/_indexing.py
jairideout: Sounds good, thanks @bunop!

skbio/io/format/embl.py
wasade: That would be great, thank you. The intent is that, if a user wants to know more about the issue, they can track it down easily. And any development conversation surrounding this support has a home.

skbio/io/format/embl.py
wasade: Oh, I thought the EMBL records stored RNA as RNA? The skbio constructor does work assuming the sequence is actually RNA:

```python
In [1]: from skbio import RNA

In [2]: RNA('AUCG')
Out[2]:
RNA
--------------------------
Stats:
    length: 4
    has gaps: False
    has degenerates: False
    has definites: True
    GC-content: 50.00%
--------------------------
0 AUCG

In [3]: ?RNA

In [4]: RNA('ATCG')
<ipython-input-4-02895b4a2413> in <module>()
----> 1 RNA('ATCG')

/Users/mcdonadt/miniconda3/envs/sk050-biom215/lib/python3.5/site-packages/skbio/sequence/_grammared_sequence.py in __init__(self, sequence, metadata, positional_metadata, lowercase, validate)
    334
    335         if validate:
--> 336             self._validate()
    337
    338     def _validate(self):

/Users/mcdonadt/miniconda3/envs/sk050-biom215/lib/python3.5/site-packages/skbio/sequence/_grammared_sequence.py in _validate(self)
    358                    [str(b.tostring().decode("ascii")) for b in bad] if
    359                    len(bad) > 1 else bad[0],
--> 360                    list(self.alphabet)))
    361
    362     @stable(as_of='0.4.0')

ValueError: Invalid character in sequence: b'T'.
Valid characters: ['H', '-', 'A', 'N', 'M', 'G', 'R', 'Y', 'K', 'V', 'W', 'U', '.', 'S', 'C', 'B', 'D']
Note: Use `lowercase` if your sequence contains lowercase characters not in the sequence's alphabet.
```

skbio/io/format/tests/test_embl.py
```diff
@@ -0,0 +1,773 @@
...
+
+class SnifferTests(TestCase):
+    def setUp(self):
+        self.positive_fps = list(map(get_data_path, [
+            'embl_single_record',
+            'embl_multi_records']))
+
+        self.negative_fps = list(map(get_data_path, [
+            'empty',
+            'whitespace_only',
```
wasade: would it be possible to add in an entry that has extra whitespace at the start of the file? in other words, it has valid embl content, but the file is malformed, which should trigger a `False` on parse

skbio/io/format/tests/test_embl.py
```diff
@@ -0,0 +1,773 @@
...
+        exp = DNA(exp[0], metadata=exp[1], lowercase=True,
+                  interval_metadata=exp[2])
+
+        self.assertEqual(exp, obs)
+
+    def test_embl_to_protein(self):
+        # TODO: add protein support
+        i = 0
+        # there is no support for protein at the moment
+#        exp = self.multi[i]
```
wasade: can the `#` marks be indented properly?

In general, we advise removing commented out code as it can be misleading. However, in this case, I think it is appropriate. Would it be possible to add to the comment a mention explicitly stating that the following code should be a valid test once protein support is added?

skbio/io/format/tests/test_embl.py
```diff
@@ -0,0 +1,773 @@
...
+
+
+class Convertertest(EMBLIOTests):
+    def test_gbtoembl(self):
+        genbank = DNA.read(self.genbank_fp, format="genbank")
+
+        with io.StringIO() as fh:
+            DNA.write(genbank, format="embl", file=fh)
+
+            # EMBL can't deal with genbank version (ie M14399.1  GI:145229)
```
wasade: What happens in this case?

Is there really no means to differentiate versions with EMBL? That is unfortunate as basically anything can change in a record between versions, including substantial changes to the sequence itself...

skbio/io/format/embl.py
bunop: Embl store rRNA data as DNA (ex [LK021130.1:74067..75610:rRNA](http://www.ebi.ac.uk/ena/data/view/Non-coding:LK021130.1:74067..75610:rRNA&display=text)). Sorry, you can instantiate a RNA object directly but it will be converted in DNA in writing, for example:
```python
import io
import skbio

# load test object for simplicity
from skbio.io.format.tests.test_embl import ReaderTests

test = ReaderTests()
test.setUp()

# this tuple contains, a RNA sequence, metadata, IntervalMetadata, and RNA class
seq, md, imd, constructor = test.single_rna

# even if constructor is skbio.sequence._rna.RNA, do
obs = skbio.RNA(seq, metadata=md, lowercase=True, interval_metadata=imd)

# obj is skbio.sequence._rna.RNA
type(obs)

# Write this RNA obj into file
handle = io.StringIO()
skbio.io.write(obs, format="embl", into=handle)

# but data is written in DNA
print(handle.getvalue())
```
genbank module works in a similat way (tests and code for embl were inspired from test and code of genbank)

skbio/io/format/tests/test_embl.py
bunop: I can deal with genbank access id since is composed by `<accession>.<version_number>` (I'm integrating it). However I don't know at the moment how to do with `GI` (is a db xref?). Moreover. NCBI says in september that `GI` will be [dismissed](https://www.ncbi.nlm.nih.gov/news/03-02-2016-phase-out-of-GI-numbers/) (and all my scripts using `GI` numbers have a lot of problems now); Maybe we need to update also genbank tests and code?

skbio/io/format/tests/test_embl.py
bunop: Added a test with extra space

skbio/io/format/genbank.py
RNAer: ah, I see. I will double check it by regenerate the doc. thanks for comments!

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1558 @@
...
+    has degenerates: False
+    has definites: True
+    GC-content: 35.99%
+----------------------------------------------------------------------
+0    AAACAAACCA AATATGGATT TTATTGTAGC CATATTTGCT CTGTTTGTTA TTAGCTCATT
+60   CACAATTACT TCCACAAATG CAGTTGAAGC TTCTACTCTT CTTGACATAG GTAACCTGAG
+...
+1740 AGAAGCTATG ATCATAACTA TAGGTTGATC CTTCATGTAT CAGTTTGATG TTGAGAATAC
+1800 TTTGAATTAA AAGTCTTTTT TTATTTTTTT AAAAAAAAAA AAAAAAAAAA AAAAAAAAA
+
```
RNAer: it is nice to add example of reading into `RNA` too (considering you imported `RNA` here). The goal in the doc here is to document most of the public facing functionalities. It is beneficial to demo most usage.

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1558 @@
...
+        'REFERENCE_COMMENT': 'RC',
+        'TITLE': 'RT',
+        'organelle': 'OG',
+        'taxonomy': 'OC'}
+
+
+# the original genbank _yield_section divides entries in sections relying on
+# spaces (the same section has the same level of indentation). Uniprot entries
+# have a key for each line, so to divide record in sections I need to define a
+# corresponance for each key to section, then I will divide a record in
```
RNAer: typo `corresponance`?

skbio/io/format/embl.py
RNAer: For the keys not found in examples, I recommend keeping them. I am sure we'll come across them. Plus, it doesn't cost anything to keep them, does it?

bunop: I agree. When I wrote this comment, there are keys that I can't handle. Non I can handle almost any of such keys. The last key I added is 'CO' that stand for constructed sequences like [LT357133](http://www.ebi.ac.uk/ena/data/view/LT357133&display=text). Such entries don't have sequences (so no intervalmetadata is possible) but but *rather the assembly information on all accession.versions and sequence locations relevant to building the constructed sequence*. This is an example of referencing external sequences. At the moment, I read data (metadata) in order to not raise an exception

skbio/io/format/tests/test_embl.py
```diff
@@ -0,0 +1,858 @@
...
+            _sequence_to_embl(_embl_to_sequence(self.single_rna_fp), fh)
+            obs = fh.getvalue()
+
+        with open(self.single_rna_fp) as fh:
+            exp = fh.read()
+
+        self.assertEqual(obs, exp)
+
+
+class Convertertest(EMBLIOTests):
```
RNAer: I like these tests. To make it more robust, is it possible to do this?

1. do embl file -> `DNA` object dna1 -> gbk file -> `DNA` object dna2 -> embl file 2
2. and then check `dna1 == dna2` and `embl file == embl file2`

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+                   'OG': 'SOURCE',
+                   # reference keys
+                   'RA': 'REFERENCE',
+                   'RP': 'REFERENCE',
+                   'RC': 'REFERENCE',
+                   'RX': 'REFERENCE',
+                   'RG': 'REFERENCE',
+                   'RT': 'REFERENCE',
+                   'RL': 'REFERENCE',
+                   'RN': 'SPACER',
```
RNAer: Is this a spacer? looks like reference number to me.

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+                   'SQ': 'ORIGIN',
+                   }
+
+
+# the inverse of KEYS_TRANSLATOR, for semplicity
+REV_KEYS_TRANSLATOR = {v: k for k, v in KEYS_TRANSLATOR.items()}
+
+
+# the original genbank _yield_section divides entries in sections relying on
+# spaces (the same section has the same level of indentation). Uniprot entries
```
bunop: Changed in `EMBL  entries`

skbio/io/format/embl.py
bunop: Yes, it's Reference Number. However, in order to split between reference with `_embl_yield_section`, I need to change this value to `spacer`. This dictionary is needed only to process group of keys with the same function. The reference number information will be stored as reference index in the final list of references. I wrote such things as comments

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+        # call function on section
+        parsed = parser(section)
+
+        # reference can appear multiple times
+        if embl_section == 'REFERENCE':
+            # genbank data hasn't CROSS_REFERENCE section, To have a similar
+            # metatadata object, I chose to remove CROSS_REFERENCE from
+            # each single reference and put them in metadata. Since I could
+            # have more references, I need a list of CROSS_REFERENCE, with
+            # None values when CROSS_REFERENCE are not defined
```
bunop: I think not. In testing examples. I have `LK021130.1:74067..75610:rRNA` which has two references. The first has no cross reference, the second yes. So I need a list with `[None, `<cross_reference>`] to model this case. With a empty value, I couldn't know which referene has cross reference

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+            '{mol_type}; {class}; {division}; {size} {unit}.\n').format(
+                header=header, indent=indent, **kwargs)
+
+
+# For non-coding, rRNA and spacer records, where a feature-level ID has not
+# previously existed, the ID, e.g. AB012758.1:1..40:tRNA, has a complex
+# format to ensure that it is unique and unambiguous. The structure of
+# the ID may be represented as:
+# <locus_name>.<version>:<feature location>:<feature name>[:ordinal]
+def _parse_locus_name(locus_dict):
```
bunop: removed.

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+    # LK021130.1:74067..75610:rRNA to model a table feature like
+    # FT   rRNA            LK021130.1:74067..75610
+    # FT                   /gene="16S"
+    # FT                   /product="16S rRNA"
+    # FT                   /note="16S rRNA subunit (checked and believed
+    # FT                   to be right,based on 454- and PacBio-sequencing)"
+    # I could express features by resizing sequence, however the resulting
+    # sequence will be different than data read. I decide to mantain
+    # sequences indentical to read values, and to discard features
+    # for such entries
```
bunop: Removed. This comment and the previous one were added when I dealt with feature-level-products,  I was thinking to resize sequence to define interval metadata for such records. Then I chose to ignore completely interval metadata.

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+        strip=False)
+
+    # process each section, like genbank does
+    for section in section_splitter(chunks):
+        # embl_key is line type (eg ID, AC, ...)
+        embl_key = _get_embl_key(section[0])
+
+        # even section have different keys, (RA, RP, ...), I need to get the
+        # correct section to call the appropriate method (eg REFERENCE)
+        embl_section = _get_embl_section(embl_key)
```
bunop: changed `_embl_yield_section`. I've renamed `embl_section` in `section_name`, it seems clearer to me

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+        # SQ   Sequence 275 BP; 64 A; 73 C; 88 G; 50 T; 0 other;
+        if line.startswith('SQ'):
+            continue
+
+        # remove the numbers inside strings. revome spaces around string
+        line = ''.join([i for i in line.strip() if not i.isdigit()])
+
+        # remove space from sequence
+        line = line.replace(" ", "")
+
```
bunop: The problem is that embl record can have numbers or not for example, see [uniprot](https://github.com/bunop/scikit-bio/blob/issue-1483/skbio/io/format/tests/data/embl_uniprot_record#L43) and [embl](https://github.com/bunop/scikit-bio/blob/issue-1483/skbio/io/format/tests/data/embl_single_record#L105) records in unittest: The first has no numbers, the second yes. I don't remeber if I found another record in ENA with no line numbers, but I wrote this line to deal with both record types.

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+def _serialize_sequence(obj, indent=5):
+    '''Serialize seq to SQ.
+
+    Parameters
+    ----------
+    obj : DNA, RNA, Sequence Obj
+    '''
+
+    # magic numbers
```
bunop: :-) I added a few comments

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+         "{n_t} T; {n_others} other;\n"
+
+    # apply format
+    SQ = SQ.format(size=len(obj), unit=obj.metadata["LOCUS"]["unit"].upper(),
+                   n_a=n_a, n_c=n_c, n_g=n_g, n_t=n_t, n_others=n_others)
+
+    for i in range(0, len(seq), line_size):
+        line = seq[i:i+line_size]
+        s = '{key:<{indent}}{s}'.format(
+            key='', indent=indent, s=chunk_str(line, frag_size, ' '))
```
bunop: fixed

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+        # pad string left and right
+        s = '{:<70}'.format(s) + '{:>10}\n'.format(i+len(line))
```
bunop: Ok. I changed those strings

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+        if n == 1:
+            # Add SQ header to sequence
+            s = SQ + s
+        n = n + line_size
```
bunop: fixed

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+    line_size = 60
+    frag_size = 10
+
+    # count bases in sequence
+    seq = str(obj).lower()
+    n_a = seq.count("a")
+    n_c = seq.count("c")
+    n_g = seq.count("g")
+    n_t = seq.count("t")
+    n_others = len(obj) - (n_a + n_c + n_g + n_t)
```
bunop: yes, fixed

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+        # pad string left and right
+        s = '{:<70}'.format(s) + '{:>10}\n'.format(i+len(line))
+        if n == 1:
+            # Add SQ header to sequence
+            s = SQ + s
+        n = n + line_size
+        yield s
+
+
+def _embl_parse_feature_table(lines, metadata):
```
bunop: fixed

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+    """Parse embl feature tables"""
+
+    # get size of the feature
+    length = metadata["LOCUS"]["size"]
+
+    if "PARENT_ACCESSION" in metadata:
```
bunop: check done in `_parse_single_embl`

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+    # define interval metadata
+    imd = IntervalMetadata(length)
+
+    # remove feature header table
+    idxs = [i for i, line in enumerate(lines) if line.startswith("FH")]
+    lines = [line for i, line in enumerate(lines) if i not in idxs]
+
+    # remove FT from the header
+    lines = [line.replace("FT", "  ", 1) for line in lines]
+
```
bunop: Ok. And magik number shuld be 19, since I removed (with no replacement) two chars from the string

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+
+def _serialize_feature_table(intervals, indent=21):
+    '''
+    Parameters
+    ----------
+    intervals : list of ``Interval``
+    '''
+
+    # define a embl wrrapper object. I need to replace only the first two
```
bunop: fix

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+    # characters from _serialize_single_feature output
+    wrapper = _get_embl_wrapper("FT", indent=2, subsequent_indent=21)
+
+    for intvl in intervals:
+        tmp = _serialize_single_feature(intvl, indent)
+        output = []
+
+        # I need to remove two spaces, cause I will add a FT key
+        for line in tmp.split("\n"):
+            output += wrapper.wrap(re.sub("^  ", "", line))
```
bunop: changed

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+
+        # re add newline between elements, and a final "\n"
+        yield "\n".join(output) + "\n"
+
+
+def _parse_date(lines, label_delimiter=None, return_label=False):
+    """Parse embl data records"""
```
bunop: fixed

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+    line = lines[0]
+
+    # take the first line, divide the key from the text
+    label, section = line.split(label_delimiter, 1)
+
+    # add section to data to return
+    data += [section]
+
+    # read all the others dates and append to data array
+    data.extend(line.split(label_delimiter, 1)[-1] for line in lines[1:])
```
bunop: changed

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+    # serialize data and return it
+    return _serialize_list(wrapper, data)
+
+
+def _serialize_dbsource(embl_key, obj, indent=5):
+    """Serialize DBSOURCE"""
+
+    # deal with rx pattern
+    DR = re.compile("([^;\s]*); ([^\s]*)")
+
```
bunop: I've modified this method to be more generic

skbio/io/format/embl.py
```diff
@@ -0,0 +1,1568 @@
...
+    return _serialize_list(wrapper, dbsource)
+
+
+def _parse_assembly(lines):
+    """Parse embl assembly records"""
+
+    output = []
+
+    # assembly pattern
+    pattern = re.compile("AS   ([^\s]+)\s+([^\s]+)\s+([^\s]+)\s+(\w?)")
```
bunop: simplified

skbio/io/format/tests/test_embl.py
bunop: In Genbank parser, GI are read since the are in [VERSION](https://github.com/bunop/scikit-bio/blob/issue-1483/skbio/io/format/tests/data/genbank_single_record#L4) record. To test it with embl, I could download a record with no GI from gb

skbio/io/format/tests/test_embl.py
```diff
@@ -0,0 +1,858 @@
...
+    def test_gbtoembl(self):
```
bunop: fixed

skbio/io/format/tests/test_embl.py
bunop: Is not so simple. When I write a `genbank` record, I put a locus like this `LOCUS       M14399   63 bp   mRNA   linear   PRO   02-SEP-1999`. However the embl `PRO` division is not contemplated by [genbank parser](https://github.com/bunop/scikit-bio/blob/issue-1483/skbio/io/format/genbank.py#L565). I need to translate division in genbank division (not sure if it will be always possible). Could I modify `genbank` to read any three letter code?

skbio/io/format/embl.py
bunop: added

skbio/io/format/tests/test_embl.py
bunop: Or I can remove it from `skbio/io/format/tests/data/genbank_single_record`. However, I need to fix `test_genbank` module

skbio/io/format/embl.py
bunop: created as #1499

skbio/io/format/embl.py
RNAer: ah, I see. didn't realize the number is optional. thanks for pointing it out! then the following should be better:

items = [i for i in line.split() if not i.isdigit()]

skbio/io/format/embl.py
```diff
@@ -1352,17 +1320,27 @@ def _serialize_sequence(obj, indent=5):
...
+
+    # get sequence as a string with lower letters (uniprot will be upper!)
     seq = str(obj).lower()
-    n_a = seq.count("a")
-    n_c = seq.count("c")
-    n_g = seq.count("g")
-    n_t = seq.count("t")
+
+    # count bases in sequence
+    n_a = sum(obj.frequencies({"a", "A"}).values())
```
RNAer: `DNA` or `RNA` store all capital case. no need to use both. suggesting doing this:

````
freq = obj.frequencies()
n_a = freq.pop('A')
...
n_others = sum(freq.values())
````

File: skbio/io/format/tests/test_embl.py (hunk 33 above)
Comment (RNAer) [inline skbio/io/format/tests/test_embl.py]: yep. It is fine to improve genbank parser as well. there shouldn't be an API change, right?

File: skbio/io/format/embl.py [hunk 57]
```diff
@@ -1352,17 +1320,27 @@ def _serialize_sequence(obj, indent=5):
...
+    # magic numbers: there will be 60 letters (AA, bp) on each line
+    chunk_size = 60
+
+    # letters (AA, bp) will be grouped by 10: each group is divided by
+    # one space from each other
     frag_size = 10
 
-    # count bases in sequence
+    # this is the final line length
+    line_size = 80
```
RNAer: is the line size fixed? what if the sequence is really long and `60 + number of spaces + number of digits (index on the right)` larger than 80?

skbio/io/format/embl.py
```diff
@@ -1374,45 +1352,46 @@ def _serialize_sequence(obj, indent=5):
...
-        line = seq[i:i+line_size]
-        s = '{key:<{indent}}{s}'.format(
-            key='', indent=indent, s=chunk_str(line, frag_size, ' '))
+    # fasta sequence will have indent spaces on the left, chunk_size/frag_size
+    # groups of frag_size letters separated by n-1 groups of single spaces,
+    # then the sequence length aligned on the right to get a string of
+    # line_size. Calculate padding for semplicity
+    n_frags = int(chunk_size / frag_size)
+    pad_right = chunk_size + n_frags-1
+    pad_left = line_size - pad_right - indent
```
RNAer: should `pad_left` be a fixed magic number instead? it always starts at a fixed column, doesn't it?

skbio/io/format/embl.py
```diff
@@ -1440,26 +1419,22 @@ def _serialize_feature_table(intervals, indent=21):
...
-    label, section = line.split(label_delimiter, 1)
 
-    # add section to data to return
-    data += [section]
+    # take the first line, and derive a label
+    label, section = lines[0].split(label_delimiter, 1)
 
     # read all the others dates and append to data array
-    data.extend(line.split(label_delimiter, 1)[-1] for line in lines[1:])
+    data.extend(line.split(label_delimiter, 1)[-1] for line in lines)
```
RNAer: no need to to do `extend`. you can define `data` here the first time instead of an empty `[]` in line 1431

skbio/io/format/embl.py
bunop: Yes it shuold be. I calculate pad_left it since indent is a parameter, and so it could change if one change indent parameter. But even indent is a magic number for embl: shuld I assign all those value a magic number?

skbio/io/format/embl.py
bunop: I will check with an embl assembly record

bunop: From uniprot manual 
```
... Columns 73-80 of each 
sequence line contain base numbers for easier reading and quick 
location of regions of interest. The numbers are right justified and indicate
the number of the last base on each line.
```
So if we consider `80-73+1=8` columns, the biggest sequence that could be included in embl file will be `pow(10,9)-1` bp. [Here](http://book.bionumbers.org/how-big-are-genomes/) is an article on biggest genomes: *Paris Japonica* and *P.aethiopicus* are not yet sequenced. Then the biggest is Wheat, which assembly is not yet complete and largest chromsome shold be ~700Mb. So, I don't think we could found an example of such type.

bunop:
Damn space in `skbio/io/format/tests/test_embl.py:899`. I will fix it in a new commit

skbio/io/format/tests/test_embl.py
bunop: Even with this modification we can't convert an embl to NCBI and then embl. EMBL has the `class`attribute that can't be translated into genbank, it has the `DOI` cross reference and can have multiple dates for each revision. Moreover is possible start from a genbank, convert to embl and then to genbank.

skbio/io/format/genbank.py
```diff
@@ -562,8 +562,7 @@ def _parse_locus(lines):
                ' +(bp|aa|rc)'
                ' +(.*DNA|.*RNA)?'
                ' +(linear|circular)?'
-               ' +(PRI|ROD|MAM|VRT|INV|PLN|BCT|VRL|PHG|'
-               'SYN|UNA|EST|PAT|STS|GSS|HTG|HTC|ENV|CON)'
+               ' +(?!.*DNA|.*RNA)([A-Z]{3})'
```
bunop: I have to exclude *DNA and * RNA from genbank division, since if molecule type and shape can be empty, and I can't fail the test for missing division (which is expected to fail)

skbio/io/format/embl.py
RNAer: I would use magic numbers for all fixed values.

skbio/io/format/genbank.py
RNAer: don't previous 2 lines make both molecule type and shape optional?

bunop: Yes, if you read comments both `+(.*DNA|.*RNA)?` and `+(linear|circular)?` could be optional. The problem arises with ```LOCUS       SCU49845     5028 bp    DNA                    21-JUN-1999```, which lack division and must raise an exception. Since the two previous line can be empty, DNA is read by division and so no exception is raised. Maybe this is the reason why you specified all the possible divisions in regular expression.

RNAer:
sorry, I don't follow you. We should raise an exception for `LOCUS SCU49845
5028 bp DNA 21-JUN-1999`, right? this is tested in the unit test.

skbio/io/format/genbank.py
bunop: Yes, we should raise an exception. if you put simply `([A-Z]{3})` no exception will be raised, cause DNA or RNA satisfy the regular expression.

RNAer: It is a bit confusing to argue with words now. could you add test cases in the unit test to show why original code needs this update?

bunop: We need this update when you want to write a embl record in genbank format and then read your genbank file: see [embl](https://github.com/bunop/scikit-bio/blob/issue-1483/skbio/io/format/tests/data/embl_single_record#L1) and [genbank](https://github.com/bunop/scikit-bio/blob/issue-1483/skbio/io/format/tests/data/genbank_single_record#L1) for the same object `M14399`: the genbank division is `BCL` while the embl is `PRO`. If I write the embl version, I will write the `PRO` division in genbank locus (header), and then I cannot use the genbank parser since `PRO` is not specified in pattern. I could express all EMBL division in genbank (but I don't like it, what happens if embl change division? or I wrote a new parser for a different nucleotide database?), or read any tree letters as division. If you read any three letter as division, you could read also DNA as division, like in [this](https://github.com/bunop/scikit-bio/blob/issue-1483/skbio/io/format/tests/test_genbank.py#L236) test: this test need to fail (as we lack division) but the greedy re `([A-Z]{3})` catch DNA, while the two previous re could be empty: 
```python
import re
import pprint

# this is the example in test_genbank.test_parse_locus_invalid
lines = [
    # missing unit
    ['LOCUS       NC_005816               9609 '
     '    DNA     circular CON 07-FEB-2015'],
    # missing division
    ['LOCUS       SCU49845     5028 bp'
     '    DNA                    21-JUN-1999'],
    # wrong date format
    ['LOCUS       NP_001832                360 aa'
     '            linear   PRI 2001-12-18']]

# getting the missing division example
line = lines[1][0]

# define a generic 3 letters pattern
pattern = (
   r'LOCUS'
   ' +([^\s]+)'
   ' +([0-9]+)'
   ' +(bp|aa|rc)'
   ' +(.*DNA|.*RNA)?'
   ' +(linear|circular)?'
   ' +([A-Z]{3})'
   ' +([0-9]{2}-[A-Z]{3}-[0-9]{4})')

# search the pattern
matches = re.match(pattern, line[0])

# reading values like _parse_locus does
res = dict(zip(
    ['locus_name', 'size', 'unit', 'mol_type',
     'shape', 'division', 'date'],
    matches.groups()))

# getting values
pprint.pprint(res)
{'date': '21-JUN-1999',
 'division': 'DNA',
 'locus_name': 'SCU49845',
 'mol_type': None,
 'shape': None,
 'size': '5028',
 'unit': 'bp'}
```
This is the reason why I need to exclude DNA and RNA from division, in order to pass the `test_genbank.test_parse_locus_invalid`. Sorry for this confusion, but I think that all tests needed are already coded in `test_genbank` and `test_embl`

RNAer: oh, I see now. Gosh, why genbank and embl have to use different division code! Thanks for explanation. I agree with the changes.

RNAer:
@bunop , I am testing it by runing it on some records and found this problem: when I try to read a following record from ENA, it reads into a empty `DNA` object with only `metadata` attribute. This is less ideal. We should either raise an exception, or return both `metadata` and `interval_metadata`. Also open for other suggestions.

````
ID   BA000049; SV 1; linear; genomic DNA; CON; FUN; 6520266 BP.
XX
AC   BA000049;
XX
PR   Project:PRJNA20809;
XX
DT   09-JUN-2011 (Rel. 109, Created)
DT   23-DEC-2016 (Rel. 131, Last updated, Version 2)
XX
DE   Aspergillus oryzae RIB40 DNA, chromosome 1.
XX
KW   HIGH_QUALITY_DRAFT.
XX
OS   Aspergillus oryzae RIB40
OC   Eukaryota; Fungi; Dikarya; Ascomycota; Pezizomycotina; Eurotiomycetes;
OC   Eurotiomycetidae; Eurotiales; Aspergillaceae; Aspergillus.
XX
RN   [1]
RP   1-6520266
RA   Fujita N., Sawano T., Yamada O., Tanaka T., Machida M.;
RT   ;
RL   Submitted (03-JUN-2011) to the INSDC.
RL   Contact:Masayuki Machida National Institute of Advanced Industrial Science
RL   and Technology, Bioproduction Research Institute; Higashi 1-1-1, Tsukuba,
RL   Ibaraki 305-8566, Japan
XX
RN   [2]
RX   DOI; 10.1038/nature04300.
RX   PUBMED; 16372010.
RA   Machida M., Asai K., Sano M., Tanaka T., Kumagai T., Terai G., Kusumoto K.,
RA   Arima T., Akita O., Kashiwagi Y., Abe K., Gomi K., Horiuchi H.,
RA   Kitamoto K., Kobayashi T., Takeuchi M., Denning D.W., Galagan J.E.,
RA   Nierman W.C., Yu J., Archer D.B., Bennett J.W., Bhatnagar D.,
RA   Cleveland T.E., Fedorova N.D., Gotoh O., Horikawa H., Hosoyama A.,
RA   Ichinomiya M., Igarashi R., Iwashita K., Juvvadi P.R., Kato M., Kato Y.,
RA   Kin T., Kokubun A., Maeda H., Maeyama N., Maruyama J., Nagasaki H.,
RA   Nakajima T., Oda K., Okada K., Paulsen I., Sakamoto K., Sawano T.,
RA   Takahashi M., Takase K., Terabayashi Y., Wortman J., Yamada O.,
RA   Yamagata Y., Anazawa H., Hata Y., Koide Y., Komori T., Koyama Y.,
RA   Minetoki T., Suharnan S., Tanaka A., Isono K., Kuhara S., Ogasawara N.,
RA   Kikuchi H.;
RT   "Genome sequencing and analysis of Aspergillus oryzae";
RL   Nature 438(7071):1157-1161(2005).
XX
DR   MD5; 79c5384fd0ef8fb71ef360426c7b5efa.
DR   BioSample; SAMD00067075.
DR   EnsemblGenomes-Gn; AO090005t00001.
DR   EnsemblGenomes-Gn; AO090005t00002.
DR   EnsemblGenomes-Gn; AO090005t00003.
DR   EnsemblGenomes-Gn; AO090005t00004.
DR   EnsemblGenomes-Gn; AO090005t00005.
DR   EnsemblGenomes-Gn; AO090005t00006.
DR   EnsemblGenomes-Gn; AO090005t00007.
DR   EnsemblGenomes-Gn; AO090005t00008.
DR   EnsemblGenomes-Gn; AO090005t00009.
DR   EnsemblGenomes-Gn; AO090005t00010.
DR   EnsemblGenomes-Gn; AO090005t00011.
DR   EnsemblGenomes-Gn; AO090005t00012.
DR   EnsemblGenomes-Gn; AO090005t00013.
DR   EnsemblGenomes-Gn; AO090005t00014.
DR   EnsemblGenomes-Gn; AO090005t00015.
DR   EnsemblGenomes-Gn; AO090005t00016.
DR   EnsemblGenomes-Gn; AO090005t00017.
DR   EnsemblGenomes-Gn; AO090005t00018.
DR   EnsemblGenomes-Gn; AO090005t00019.
DR   EnsemblGenomes-Gn; AO090005t00020.
DR   EnsemblGenomes-Gn; AO090005t00021.
DR   EnsemblGenomes-Gn; AO090005t00022.
DR   EnsemblGenomes-Gn; AO090005t00023.
DR   EnsemblGenomes-Gn; AO090005t00024.
DR   EnsemblGenomes-Gn; AO090005t00025.
DR   EnsemblGenomes-Gn; AO090005t00026.
DR   EnsemblGenomes-Gn; AO090009t00001.
DR   EnsemblGenomes-Gn; AO090009t00002.
DR   EnsemblGenomes-Gn; AO090009t00003.
DR   EnsemblGenomes-Gn; AO090009t00004.
DR   EnsemblGenomes-Gn; AO090009t00005.
DR   EnsemblGenomes-Gn; AO090009t00006.
DR   EnsemblGenomes-Gn; AO090009t00007.
DR   EnsemblGenomes-Gn; AO090009t00008.
DR   EnsemblGenomes-Gn; AO090009t00009.
DR   EnsemblGenomes-Gn; AO090009t00010.
DR   EnsemblGenomes-Gn; AO090009t00011.
DR   EnsemblGenomes-Gn; AO090009t00012.
DR   EnsemblGenomes-Gn; AO090009t00013.
DR   EnsemblGenomes-Gn; AO090009t00014.
XX
CC   This genome assembly has been updated so
CC   as to reflect the location and orientation of
CC   contigs on the chromosomes. Gap lengths were
CC   estimated using the results from Optical
CC   Mapping (http://www.bio.nite.go.jp/dogan/project/view/AO).
CC   This Aspergillus oryzae RIB40 genome consists of a nuclear genome
CC   assembly represented by eight chromosomes (assembled from 24
CC   entries) and three unplaced contigs (AP007156, AP007160 and
CC   AP007177) and a complete mitochondrial genome (AP007176).
CC   ##Genome-Assembly-Data-START##
CC   Assembly Method :: Phred/Phrap software
CC   Assembly Name   :: Aoryzae_assembly02
CC   ##Genome-Assembly-Data-END##
XX
FH   Key             Location/Qualifiers
FH
FT   source          1..6520266
FT                   /organism="Aspergillus oryzae RIB40"
FT                   /chromosome="1"
FT                   /strain="RIB40"
FT                   /mol_type="genomic DNA"
FT                   /db_xref="taxon:510516"
FT   assembly_gap    1913119..2001628
FT                   /estimated_length=88510
FT                   /gap_type="unknown"
FT   assembly_gap    6430709..6460477
FT                   /estimated_length=29769
FT                   /gap_type="unknown"
FT   assembly_gap    6511837..6517223
FT                   /estimated_length=5387
FT                   /gap_type="unknown"
XX
CO   join(AP007150.1:1..1913118,gap(88510),complement(AP007151.1:1..4429080),
CO   gap(29769),AP007152.1:1..51359,gap(5387),complement(AP007153.1:1..3043))
//
````

**Comment by bunop** [comment]
Yes, I known. For this record and for other assembly record in general we can't have interval metadata using `skbio` philosophy, since `IntervalMetadata` (if I correctly understand) is a `panda` dataframe (or `numpy` array, I don't remember) and somewere in the code interval metadata size is checked with sequence length, and this raise an error when sequence is empty. So I've [disabled](https://github.com/bunop/scikit-bio/blob/issue-1483/skbio/io/format/embl.py#L753) interval metadata for such entries (however, I read it!). Personally, I dislike raising an exception, since someone wants to process assembly data. I think that we could define a *fake* metadata, or something that could be defined when sequence is absent. This could satisfy `skbio` intentions?

**Comment by RNAer** [comment]
`IntervalMetadata` is not a pandas dataframe. You can supply `None` to the
`upper_bound` parameter - this allows you to not specify sequence length
ahead of time.

**Comment by bunop** [comment]
good to know, thank! I will try to model IntervalMetadata for such entries (but this will be on Monday :smile: )

**Comment by RNAer** [comment]
thanks! have a good weekend!

**Comment by bunop** [comment]
Sorry for this late reply, but I was very busy this week. Is not so simple to set a unbounded interval metadata for such records since they have no sequence data. Consider the example:
```python
In [1]: from skbio.metadata import IntervalMetadata

In [2]: from skbio import DNA

In [3]: imd = IntervalMetadata(None)

In [4]: imd.add([(1, 3)], metadata={'gene': 'sagA'})
Out[4]: Interval(interval_metadata=<140225444410592>, bounds=[(1, 3)], fuzzy=[(False, False)], metadata={'gene': 'sagA'})

In [5]: seq = DNA('', interval_metadata=imd)
# omissis
ValueError: Cannot set `bounds` ([(1, 3)]) with coordinate larger than upper bound (0)
```
This seems due to [this](https://github.com/bunop/scikit-bio/blob/issue-1483/skbio/metadata/_mixin.py#L472) line of code, which set upper bound as sequence length when constructing sequence. Passing a `None` value when `interval_metadata.upper_bound` is `None` invalidate 12 tests in `skbio/metadata/_testing.py`, like `test_constructor_empty_interval_metadata_upper_bound_is_none` since the [upper bound is reset to seq/axis length](https://github.com/bunop/scikit-bio/blob/issue-1483/skbio/metadata/_testing.py#L929) when instantiating an interval metadata. A workaround could be to create a fake sequence, for example:
```python
seq = DNA('NNN', interval_metadata=imd)
```
However it doesn't seem useful to create an empty sequence for an assembly record which could define a whole chromosome assembly

wasade:
@bunop no worries at all!

@mortonjt, if you have a second, could you please see @bunop's recent comment regarding interval metadata?

RNAer:
@bunop , thanks for your update. The upper bound is enforced to the sequence length once the interval metadata is attached to a seq object, like positional metadata. This is what we want. In this example of embl record, there is no sequence section, we should not read it into sequence object. The real problem is that we don't have a data model to encapsulate the metadata and interval metadata of a sequence. If we don't care about orphan metadata/interval metadata, we can design the read API of embl to accept a DNA object and attach the parsed metadata/interval metadata from embl file to this DNA object.

bunop:
So, if I have understand, I could do like this: I will open a new issue like "add an object to deal with assembly information", drop support to such embl entries (I think I could do it when parsing the first line) and raise an exception citing such iussue number. At the end, only Sequence object will be returned from embl parser. Do you think that it will be correct?

RNAer:
An object for assembly sounds great, but is not trivial and will need careful thoughts on data structure and APIs. Please feel free create a issue as you like. I think it is fine to leave such embl entries out for now, unless you guys think differently, @jairideout , @ebolyen  and @wasade

ebolyen:
Sorry for the delayed response. That sounds like a good plan to me @RNAer and @bunop.

RNAer:
I am ok with merging this. @wasade , does the PR address all your comments?

Issue #1497, opened 2017-02-27
bug: make test fails in travis ci (python 3.6?)


bunop:
Dear all,

I found that my PR #1495 fail to pass the `travis CI` test for a problem related to the conda environment. Briefly, by inspecting the travis output I noticed that modules were tested on a python `3.6` environment, even if `3.5` version is specified in configuration. The commands to repruduce error locally are the following:
```bash
$ conda create -n travis python=3.6
$ source activate travis
$ conda install --file ci/conda_requirements.txt
$ pip install -r ci/pip_requirements.txt
$ pip install . --no-deps
$ WITH_COVERAGE=TRUE make test
```
This will raise the same errors raised by `Travis CI`. By inspecting code, I found that the problem arises from `skbio.stats.power`, at line [1076](https://github.com/biocore/scikit-bio/blob/master/skbio/stats/power.py#L1076): The order of items of `collections.Counter(set_pos)` is different from python `3.5` to `3.6`. Here is a sample code to reproduce the error:
```python
import collections
set_pos = [2,0,3,1,3,2,2,0,4,3,2,1,5,0,0,3,0,3,3,1,1]
collections.Counter(set_pos).items()
# in python 3.5
# dict_items([(0, 5), (1, 4), (2, 4), (3, 6), (4, 1), (5, 1)])
# in python 3.6
# dict_items([(2, 4), (0, 5), (3, 6), (1, 4), (4, 1), (5, 1)])
```
I think i've fixed the issue in python code, I will open a new PR. Another solution could be fixing `.travis.yml` to force python `3.5` installation, however make test will fail in `3.6` enviroment.
Thank you for your support,
regards,
Paolo

wasade:
Thanks, @bunop! I think this has to do with Python3.6's `OrderedDict`. Good find!

wasade:
And thank you also for the reproducible examples

ElDeveloper:
Just tried installing the latest release in python 3.6 and saw a bunch of failures when I ran the test suite, also note that the [current build is failing](https://travis-ci.org/biocore/scikit-bio).

PR #1498, opened 2017-02-27
BUG: passing tests in python 3.6 - references #1497

bunop:
Fixed a bug in skbio/stats/power.py and added newlines to
pass pyflake tests

### DIFF
--- a/skbio/stats/power.py
+++ b/skbio/stats/power.py
@@ -1076,9 +1076,16 @@ def _draw_paired_samples(meta_pairs, index, num_samps):
     counter = collections.Counter(set_pos)
 
     # counter.items() order isn't guaranteed in python 3.6 and then the random
-    # coiche isn't reproducible between python version, even specifying seed
-    # so we access such values through sets
-    for set_ in set(set_pos):
+    # choice isn't reproducible between python version, even specifying seed;
+    # so we access such values through sets.
+    set_list = set(set_pos)
+
+    # as stated by @RNAer, since we can assure that item in sets are ordered,
+    # we choose to order set_list before accessing values
+    set_list = sorted(set_list)
+
+    # now set_list is ordered and we can iterate over it to get counter obj
+    for set_ in set_list:
         num_ = counter[set_]
         r2 = [np.random.choice(col, num_, replace=False) for col in
               meta_pairs[set_]]
--- a/skbio/stats/power.py
+++ b/skbio/stats/power.py
@@ -1073,7 +1073,13 @@ def _draw_paired_samples(meta_pairs, index, num_samps):
     subs = []
 
     # Draws the other groups
-    for set_, num_ in collections.Counter(set_pos).items():
+    # for set_, num_ in collections.Counter(set_pos).items():
+    counter = collections.Counter(set_pos)
+
+    # collections.Counter(set_pos).items() order isn't guaranteed
+    # in python 3.6, so we access such values through sets
+    for set_ in set(set_pos):
+        num_ = counter[set_]
         r2 = [np.random.choice(col, num_, replace=False) for col in
               meta_pairs[set_]]
         subs.append(r2)
--- a/skbio/stats/power.py
+++ b/skbio/stats/power.py
@@ -1080,8 +1080,8 @@ def _draw_paired_samples(meta_pairs, index, num_samps):
     # so we access such values through sets.
     set_list = set(set_pos)
 
-    # as stated by @RNAer, since we can assure that item in sets are ordered,
-    # we choose to order set_list before accessing values
+    # then, as stated by @RNAer, since we can't assure that items in sets are
+    # ordered, we choose to order set_list before accessing values
     set_list = sorted(set_list)
 
     # now set_list is ordered and we can iterate over it to get counter obj
--- a/skbio/stats/power.py
+++ b/skbio/stats/power.py
@@ -1072,12 +1072,12 @@ def _draw_paired_samples(meta_pairs, index, num_samps):
 
     subs = []
 
-    # Draws the other groups
-    # for set_, num_ in collections.Counter(set_pos).items():
+    # Draws the other groups. Get a collection.Counter object for simplicity
     counter = collections.Counter(set_pos)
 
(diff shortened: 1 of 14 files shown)


wasade:
@jairideout @ebolyen, I think it would be useful for one of you to review as this involves Python versions.

@jwdebelius a review would be great as this involves power.

skbio/stats/power.py
```diff
@@ -1073,7 +1073,13 @@ def _draw_paired_samples(meta_pairs, index, num_samps):
     subs = []
 
     # Draws the other groups
-    for set_, num_ in collections.Counter(set_pos).items():
+    # for set_, num_ in collections.Counter(set_pos).items():
```
wasade: Can this comment be removed please?

skbio/stats/power.py
```diff
@@ -1073,7 +1073,13 @@ def _draw_paired_samples(meta_pairs, index, num_samps):
...
+    counter = collections.Counter(set_pos)
+
+    # collections.Counter(set_pos).items() order isn't guaranteed
```
wasade: Is this due to order, or due to a random seed with `np.random.choice`?

skbio/stats/power.py
bunop: Ok, I can remove comment and update pull request

skbio/stats/power.py
bunop: No. The problem arises when compiling examples. In example, a `np.seed(25)` is instantiated, so you can reproduce always your sample code and `np.random.choice` will be the same. The example reported in #1497 is derived from your examples: the point is that `collections.Counter(set_pos).items()` change order in different python version, `_draw_paired_samples` returns a different array and `_identify_sample_groups` fails example tests

skbio/stats/power.py
wasade: Thanks!

skbio/stats/power.py
wasade: Okay, thanks

jwdebelius:
:+1:

mortonjt:
Awesome.  Thanks @bunop !

skbio/stats/power.py
```diff
@@ -1072,8 +1072,14 @@ def _draw_paired_samples(meta_pairs, index, num_samps):
 
     subs = []
 
-    # Draws the other groups
-    for set_, num_ in collections.Counter(set_pos).items():
+    # Draws the other groups. Get a collection.Counter object for simplicity
+    counter = collections.Counter(set_pos)
+
+    # counter.items() order isn't guaranteed in python 3.6 and then the random
+    # coiche isn't reproducible between python version, even specifying seed
```
mortonjt: `coiche` -> `choice` right?

bunop: Yes, sorry for this typo. Can you correct this or do you need I update my PR?

jairideout:
> @jairideout @ebolyen, I think it would be useful for one of you to review as this involves Python versions.

Looks like the only real change here is to the power code, which @jwdebelius approved, so I think this is GTG! The other styling issues are due to an updated flake8.

It would be helpful if someone could look into why Travis is using Python 3.6 instead of 3.5, because we explicitly pass `python=3.5` in .travis.yml. I wonder if the same problem is happening for Python 3.4 in the build matrix?

wasade:
@jairideout agree

@bunop, would you be able to add in a "python --version" to the .travis.yml? It would be great to just have this readily within the output.

I'm :+1: following addressing the comment by @mortonjt

skbio/stats/power.py
RNAer: @bunop , it'd better for you to correct it, making the commit history cleaner.

skbio/workflow.py
```diff
@@ -220,6 +220,8 @@ def __init__(self):
     def __call__(self, msg):
         self.msg = msg
         return self
+
+
```
RNAer: this is interesting. I thought flake8 would fail without the empty lines here.

skbio/stats/power.py
```diff
@@ -1072,8 +1072,14 @@ def _draw_paired_samples(meta_pairs, index, num_samps):
...
+    # so we access such values through sets
```
RNAer: I don't think `set` solves the problem because `set` is also an unordered data structure. Why the order matters here? If it is only about the doctest/unit test, it is better to just sort the return from this function and then compare to the expect.

bunop:
@jairideout , @wasade the problem in python version seems to be related to conda not travis. I've opened a [issue](https://github.com/conda/conda/issues/4756) on conda repository. In order to fix temporarily this problem and to test against the appopriate python version, I've changed `.travis.yml` by splitting the creation of the environment from package installation. I've also put a `3.6` enviroment to test against new python enviroment. After conda installation, the current python version is shown (before skbio installation: there are other python version displayed by travis before). Now it seems to best tested against `3.4`, `3.5` and `3.6`

skbio/stats/power.py
bunop: Ok, I added an intermediate step to ensure that the list derived from set is ordered

skbio/workflow.py
bunop: Seems to be an update on pyflake. A fresh install will fail without these spaces

jairideout:
Thanks @bunop for investigating and fixing the conda python version issue, *and* adding Python 3.6 testing!

@jwdebelius, can you verify that the latest updates to the power code are acceptable?

And can someone look over the Travis-CI output to make sure things are in order? Travis-CI is basically down due to S3 outage, so the build logs currently aren't accessible.

wasade:
Thank you @bunop!

@jairideout I think we should wait until it comes back up?

jwdebelius:
@jairideout: code looks fine to me.

Thanks

jairideout:
Thanks for verifying @jwdebelius!

@wasade, yeah it'd be good to wait for those logs to show up, just to sanity-check the CI system.

RNAer:
I will merge this since all reviewers thumbed up. We need it for other PRs to pass travis.

Issue #431, opened 2014-06-12
improvements to neighbor joining


gregcaporaso:
@ElBrogrammer and I discussed this while working on #420. 
- [ ] should investigate building TreeNodes as the algorithm proceeds, rather than newick string, this will likely be more efficient (but maybe not if the user just wants a newick string back, since it'd be a lot of TreeNode instantiation) 
- [ ] some functionality should be ported to methods of the `DistanceMatrix`, including:
  - [ ] min, max (see #228): the current `_lowest_index` method
  - [ ] collapsing a pair of values: the current `_compute_collapsed_dm` and `_otu_to_new_node` methods
  - [ ] probably should rename `_otu_to_new_node` since technically it's computing the distance from a node (either internal or a tip) to the new node

Anything else @ElBrogrammer? 

Depends on:
- [ ] TreenNode Refactor (#809)
- [ ] DistanceMatrix Improvement (#824)

PR #1512, opened 2017-04-24
vectorized _compute_q (issue #431)

stevendbrown:
Please complete the following checklist:

* [X] I have read the guidelines in [CONTRIBUTING.md](https://github.com/biocore/scikit-bio/blob/master/CONTRIBUTING.md).

* [X] I have documented all public-facing changes in [CHANGELOG.md](https://github.com/biocore/scikit-bio/blob/master/CHANGELOG.md).

* [ ] **This pull request includes code, documentation, or other content derived from external source(s).** If this is the case, ensure the external source's license is compatible with scikit-bio's [license](https://github.com/biocore/scikit-bio/blob/master/COPYING.txt). Include the license in the `licenses` directory and add a comment in the code giving proper attribution. Ensure any other requirements set forth by the license and/or author are satisfied. **It is your responsibility to disclose code, documentation, or other content derived from external source(s).** If you have questions about whether something can be included in the project or how to give proper attribution, include those questions in your pull request and a reviewer will assist you.

* [X] **This pull request does not include code, documentation, or other content derived from external source(s).**

**Note:** [REVIEWING.md](https://github.com/biocore/scikit-bio/blob/master/REVIEWING.md) may also be helpful to see some of the things code reviewers will be verifying when reviewing your pull request.

----------

Hello,

This pull request references issue #431 by modifying the `_compute_q` function to use CPU vector operations to achieve a significant speedup. I timed the code using `%timeit`, whereby the original function resulted in: `100 loops, best of 3: 181 ms per loop` and the vectorized function resulted in: `100 loops, best of 3: 568 µs per loop`. In addition, for a 10x larger distance matrix, the original code is still running in the time it took me to modify the code and create this pull request. Running `make test` appears to pass without issue and testing the results of the two versions of the function with `np.all(np.isclose(x.to_data_frame(), y.to_data_frame()))` returns `True`.

There are some obvious time sinkholes which remain, but I hope this code can speed up the `nj` function for other users as well.

Best,
Steve

PR #1513, opened 2017-04-25
more speedups to `nj` (issue #431)

stevendbrown:

* [X] I have read the guidelines in [CONTRIBUTING.md](https://github.com/biocore/scikit-bio/blob/master/CONTRIBUTING.md).

* [X] I have documented all public-facing changes in [CHANGELOG.md](https://github.com/biocore/scikit-bio/blob/master/CHANGELOG.md).

* [X] **This pull request does not include code, documentation, or other content derived from external source(s).**

------

Hello,

This pull request references issue #431 by modifying `_lowest_index` and `_compute_collapsed_dm` to avoid item-by-item testing or copying. In the case of `_lowest_index`, the item-by-item checking is replaced with a call to `amin` on the dense form of the distance matrix data followed by point selection to produce the same behavior as point-by-point ordered checking. In the case of `_compute_collapsed_dm`, the existing `DistanceMatrix` data which can be re-used is copied as a single block before point-by-point computation of the distance values for `new_node_id`. On my test 500x500 distance matrix, computation time for `nj` is reduced from 289 seconds to 6 seconds, a 97% reduction in wall-clock time.

Steve

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 ### Backward-incompatible changes [experimental]
 
 ### Performance enhancements
-* `skbio.tree.nj` wall-clock time was decreased about 66% by vectorizing `skbio.tree._nj._compute_q`. ([#1512](https://github.com/biocore/scikit-bio/pull/1512))
+* `skbio.tree.nj` wall-clock time was decreased about 99% by vectorizing `skbio.tree._nj._compute_q` and avoiding unnecessary recalculations. ([#1512](https://github.com/biocore/scikit-bio/pull/1512), [#XXXX](https://github.com/biocore/scikit-bio/pull/XXXX))
 
 ### Bug fixes
 * The `include_self` parameter was not being honored in `skbio.TreeNode.tips`. The scope of this bug was that if `TreeNode.tips` was called on a tip, it would always result in an empty `list` when unrolled.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 ### Backward-incompatible changes [experimental]
 
 ### Performance enhancements
-* `skbio.tree.nj` wall-clock time was decreased about 99% by vectorizing `skbio.tree._nj._compute_q` and avoiding unnecessary recalculations. ([#1512](https://github.com/biocore/scikit-bio/pull/1512), [#XXXX](https://github.com/biocore/scikit-bio/pull/XXXX))
+* `skbio.tree.nj` wall-clock time was decreased about 99% by vectorizing `skbio.tree._nj._compute_q` and avoiding unnecessary recalculations. ([#1512](https://github.com/biocore/scikit-bio/pull/1512), [#1513](https://github.com/biocore/scikit-bio/pull/1513))
 
 ### Bug fixes
 * The `include_self` parameter was not being honored in `skbio.TreeNode.tips`. The scope of this bug was that if `TreeNode.tips` was called on a tip, it would always result in an empty `list` when unrolled.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 ### Backward-incompatible changes [experimental]
 
 ### Performance enhancements
-* `skbio.tree.nj` wall-clock time was decreased about 99% by vectorizing `skbio.tree._nj._compute_q` and avoiding unnecessary recalculations. ([#1512](https://github.com/biocore/scikit-bio/pull/1512), [#1513](https://github.com/biocore/scikit-bio/pull/1513))
+* `skbio.tree.nj` wall-clock runtime was decreased by 99% for a 500x500 distance matrix and 93% for a 100x100 distance matrix. ([#1512](https://github.com/biocore/scikit-bio/pull/1512), [#1513](https://github.com/biocore/scikit-bio/pull/1513))
 
 ### Bug fixes
 * The `include_self` parameter was not being honored in `skbio.TreeNode.tips`. The scope of this bug was that if `TreeNode.tips` was called on a tip, it would always result in an empty `list` when unrolled.
--- a/skbio/tree/_nj.py
+++ b/skbio/tree/_nj.py
@@ -10,6 +10,8 @@ import io
 
 import numpy as np
 
+from scipy.spatial.distance import squareform
+
 from skbio.stats.distance import DistanceMatrix
 from skbio.tree import TreeNode
 from skbio.util._decorator import experimental
@@ -197,12 +199,14 @@ def _compute_collapsed_dm(dm, i, j, disallow_negative_branch_length,
     out_ids = [new_node_id]
     out_ids.extend([e for e in dm.ids if e not in (i, j)])
     result = np.zeros((out_n, out_n))
+    # pre-populate the result array with known distances
+    dmdf = dm.to_data_frame()
+    dmdf = dmdf.drop([i, j]).drop([i, j], axis=1)
+    result[1:, 1:] = np.array(dmdf)
+    # scan through the new node intersections & assign those
     for idx1, out_id1 in enumerate(out_ids[1:]):
         result[0, idx1 + 1] = result[idx1 + 1, 0] = _otu_to_new_node(
             dm, i, j, out_id1, disallow_negative_branch_length)
-        for idx2, out_id2 in enumerate(out_ids[1:idx1+1]):
-            result[idx1+1, idx2+1] = result[idx2+1, idx1+1] = \
-                dm[out_id1, out_id2]
     return DistanceMatrix(result, out_ids)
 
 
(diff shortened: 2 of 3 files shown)


skbio/tree/_nj.py
```diff
@@ -216,15 +220,25 @@ def _lowest_index(dm):
...
+    # convert to an array
+    dmarray = np.array(dm.to_data_frame())
+    # convert to dense, 1-d squareform of distance matrix
+    vec = squareform(dmarray)
+    # get the positions of the lowest value
+    results = np.vstack(np.where(dmarray == np.amin(vec))).T
+    # select results in the bottom-left of the array
+    results = [r for r in results if r[0] > r[1]]
+    # calculate the distances of the results to [0, 0]
+    res_distances = np.sqrt([r[0]**2 + r[1]**2 for r in results])
```
stevendbrown: Is there a danger of an overflow in the squaring of the indices of the `results`?

skbio/tree/_nj.py
```diff
@@ -216,15 +220,25 @@ def _lowest_index(dm):
...
-            curr_index = i, j
-            curr_value = dm[curr_index]
-            if curr_value < lowest_value:
-                lowest_value = curr_value
-                result = curr_index
-    return result
+    # convert to an array
+    dmarray = np.array(dm.to_data_frame())
+    # convert to dense, 1-d squareform of distance matrix
+    vec = squareform(dmarray)
```
wasade: Can be `vec = dm.condensed_form()`

skbio/tree/_nj.py
```diff
@@ -216,15 +220,25 @@ def _lowest_index(dm):
...
-    for i in range(dm.shape[0]):
-        for j in range(i):
-            curr_index = i, j
-            curr_value = dm[curr_index]
-            if curr_value < lowest_value:
-                lowest_value = curr_value
-                result = curr_index
-    return result
+    # convert to an array
+    dmarray = np.array(dm.to_data_frame())
```
wasade: I don't think the cast to a `DataFrame` is necessary as the operations below should be able to leverage `dm.data` directly

skbio/tree/_nj.py
```diff
@@ -216,15 +220,25 @@ def _lowest_index(dm):
...
+    # convert to dense, 1-d squareform of distance matrix
+    vec = squareform(dmarray)
+    # get the positions of the lowest value
+    results = np.vstack(np.where(dmarray == np.amin(vec))).T
```
wasade: For instance, `np.vstack(np.where(dm.data == np.amin(dm.condensed_form()))).T`

skbio/tree/_nj.py
```diff
@@ -216,15 +220,25 @@ def _lowest_index(dm):
...
+    # select results in the bottom-left of the array
+    results = [r for r in results if r[0] > r[1]]
```
wasade: perhaps inconsequential, but I believe this can also be:

```python
np.take(results, np.argwhere(results[:, 0] > results[:, 1]), axis=0).squeeze()
```

skbio/tree/_nj.py
wasade: We have other issues if the DMs are getting large enough for this to be a concern I believe. For instance, I don't think there is an architecture right now which has the memory necessary to support a 2^32 x 2^32 dense matrix. If you don't know about it, you can set the overflow behavior explicitly with numpy though:

```python
In [49]: np.geterr()   # np.seterr to change behavior
Out[49]: {'divide': 'warn', 'invalid': 'warn', 'over': 'warn', 'under': 'ignore'}
```

skbio/tree/_nj.py
stevendbrown: I realized I could have simplified this with:

`results[results[:, 0] > results[:, 1]]`

The results of `%timeit` on that and `np.take` gave 5.71 and 19.1 µs respectively.

stevendbrown:
I vectorized the `_otu_to_new_node` but left it in `_compute_collapsed_dm`. Would it be better to remove `_otu_to_new_node` or move the vectorized code to that function?

wasade:
Thanks @stevendbrown! I'm inclined to say the code should be inlined, and that `_otu_to_new_node` be deleted, but I'd like to defer for comment to either @jairideout or @ebolyen on that point.

wasade:
I'm inclined to say that this PR is :+1: with the exception that it is pending a decision on `_otu_to_new_node`, which will require a new commit (either to delete `_otu_to_new_node` or to shuffle the logic into it)

jairideout:
I'm good with whatever is most efficient since these are all private functions/methods being refactored (users won't be affected by private API changes).

wasade:
Thanks, @jairideout.

@stevendbrown, would it be possible to delete the `_otu_to_new_node` method and its tests, as well as address the `.data` comment that @jairideout raised [here](https://github.com/biocore/scikit-bio/pull/1512#discussion_r113531341) (and where else it may be relevant)?

stevendbrown:
@wasade I found one reference at the end of `nj` to `_otu_to_new_node`. Since the only other reference to it has been removed, I inlined that call as well. Let me know if I should revert those changes.

@jairideout Thanks for the suggestions!

stevendbrown:
Ah, crap, that commit was premature. Sorry about that. :-(

skbio/tree/_nj.py
```diff
@@ -197,12 +199,19 @@ def _compute_collapsed_dm(dm, i, j, disallow_negative_branch_length,
...
+                               ij_indexes, axis=1)
+    # calculate the new distances from the current DistanceMatrix
+    k_to_u = 0.5 * (dm[i] + dm[j] - dm[i, j])
+    # set negative branches to 0 if specified
+    if disallow_negative_branch_length:
+        k_to_u[k_to_u < 0] = 0
+    # drop nodes being joined
+    k_to_u = np.delete(k_to_u, ij_indexes)
+    # assign the distances to the result array
+    result[0] = result[:, 0] = np.array([0] + k_to_u.data.tolist())
```
wasade: I think this can be expressed as the following, and avoid the `tolist` call which is relatively expensive:

```python
result[0] = result[:, 0] = np.hstack([0], k_to_u])
```

CHANGELOG.md
```diff
@@ -16,7 +16,7 @@
 ### Backward-incompatible changes [experimental]
 
 ### Performance enhancements
-* `skbio.tree.nj` wall-clock time was decreased about 66% by vectorizing `skbio.tree._nj._compute_q`. ([#1512](https://github.com/biocore/scikit-bio/pull/1512))
+* `skbio.tree.nj` wall-clock time was decreased about 99% by vectorizing `skbio.tree._nj._compute_q` and avoiding unnecessary recalculations. ([#1512](https://github.com/biocore/scikit-bio/pull/1512), [#1513](https://github.com/biocore/scikit-bio/pull/1513))
```
wasade: is this 99% still accurate given the `.data` change that was made? Would it be possible to note the size distance matrix that this performance improvement is over?

skbio/tree/_nj.py
stevendbrown: It's a surprisingly inexpensive operation; both calls (i.e. `tolist()` and `hstack()`) use about 3.2% of the time spent in this function. However, changing from `hstack` to `concatenate` reduces that particular operation to 1.3% of the time spent in that function.

CHANGELOG.md
stevendbrown: Yes. For a 500x500 matrix, while the switch to `.data` produced a measurable speedup in `nj(dm)` (taking runtimes reported by `%timeit -n 10 tree = nj(dm)` from 1.39 seconds to 949 milliseconds), [I presume that] the majority of the speedup was realized by vectorization (taking the post-#1512 runtime of 115 seconds down to 1.39 seconds).

This said, returns from these changes are non-linear. I have some notes from a now-unknown midpoint of tinkering which suggest returns _increase_ with increasing size of the distance matrix: a 100x100 matrix saw a 77% speedup while a 500x500 matrix saw a 98% speedup.

Unfortunately, for my own data's complete 2831x2831 distance matrix, the current-best runtime (from `%timeit -n 1 tree = nj(dm)`) is still 6m 56s.

I'm happy with the improvements so far for my own wet-bench purposes, so I'm open to any and all suggestions as to how to word the statement in `CHANGELOG.md`; to make it easy, we could just drop the number and say "...was decreased. ([refs to PRs])".

wasade: Thanks! I would expect to see further improvements as the matrix size increases. I think for the purpose of the changelog, just mentioning that this improvement was assessed on a 500x500 matrix is sufficient. I think it is useful for the community to have a ballpark figure on performance improvement. This helps users to evaluate if they want to shift to skbio dev or a future release. For instance, a 5% difference may not matter for a downstream package, but a 100% improvement could represent a very dramatic change and motivate a downstream package to bump priority on adopting a different version of skbio.

wasade: Ah, sorry, I "approved" the changes before the matrix size inclusion :) This is good to merge in my opinion once that's in.

CHANGELOG.md
```diff
@@ -16,7 +16,7 @@
 ### Backward-incompatible changes [experimental]
 
 ### Performance enhancements
-* `skbio.tree.nj` wall-clock time was decreased about 99% by vectorizing `skbio.tree._nj._compute_q` and avoiding unnecessary recalculations. ([#1512](https://github.com/biocore/scikit-bio/pull/1512), [#1513](https://github.com/biocore/scikit-bio/pull/1513))
+* `skbio.tree.nj` wall-clock runtime was decreased by 99% for a 500x500 distance matrix and 93% for a 100x100 distance matrix. ([#1512](https://github.com/biocore/scikit-bio/pull/1512), [#1513](https://github.com/biocore/scikit-bio/pull/1513))
```
stevendbrown: I re-ran the `%timeit` for a 100x100 matrix at the pre #1512 commit and with all commits in this PR to arrive at the 93% number.

wasade: Thanks!

wasade:
Thanks again @stevendbrown!

stevendbrown:
Thanks @wasade & @jairideout for your timely attention & helpful comments! 🍻

Issue #1511, opened 2017-04-19
Build errors with Sphinx and Pandas


wasade:
We're observing build errors with Sphinx. This is causing PRs to indicate failures in CI which are unrelated to the PR itself. #1510 is an exemplar of the problem.

On a local build, using Sphinx 1.5.5, I am able to build the docs.

wasade:
Revised the issue title. We're now also seeing what appear to be [build failures](https://travis-ci.org/biocore/scikit-bio/jobs/232604798#L1173) with the Pandas, suggesting we may need to restrict minor version.

PR #1514, opened 2017-04-27
ENH: Adding helper methods for calculating leaf count, height, and depth

mortonjt:

* [x] **This pull request includes code, documentation, or other content derived from external source(s).** If this is the case, ensure the external source's license is compatible with scikit-bio's [license](https://github.com/biocore/scikit-bio/blob/master/COPYING.txt). Include the license in the `licenses` directory and add a comment in the code giving proper attribution. Ensure any other requirements set forth by the license and/or author are satisfied. **It is your responsibility to disclose code, documentation, or other content derived from external source(s).** If you have questions about whether something can be included in the project or how to give proper attribution, include those questions in your pull request and a reviewer will assist you.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Version 0.5.1-dev (changes since 0.5.1 go here)
 
 ### Features
-* Added ``skbio.tree_tree._cache_leafcount`` ``skbio.tree_tree._cache_height`` and ``skbio.tree_tree._cache_depth`` for calculating leaf count in subtrees, subtree height, and node depth. This is particularly useful for calculating tree coordinates for visualization ([#1514](https://github.com/biocore/scikit-bio/pull/1514))
+* Added ``skbio.tree_tree._cache_leafcount`` ``skbio.tree_tree._cache_height`` and ``skbio.tree_tree._cache_depth`` for calculating leaf count in subtrees, subtree height, and node depth. This is was originally ported from pycogent and particularly useful for calculating tree coordinates for visualization ([#1514](https://github.com/biocore/scikit-bio/pull/1514))
 
 * Added ``skbio.io.format.gff3`` for reading and writing GFF3 files for ``DNA``, ``Sequence``, and ``IntervalMetadata`` classes. ([#1450](https://github.com/biocore/scikit-bio/pull/1450))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## Version 0.5.1-dev (changes since 0.5.1 go here)
 
 ### Features
+* Added ``skbio.tree_tree._cache_leafcount`` ``skbio.tree_tree._cache_height`` and ``skbio.tree_tree._cache_depth`` for calculating leaf count in subtrees, subtree height, and node depth. This is particularly useful for calculating tree coordinates for visualization ([#1514](https://github.com/biocore/scikit-bio/pull/1514))
+
 * Added ``skbio.io.format.gff3`` for reading and writing GFF3 files for ``DNA``, ``Sequence``, and ``IntervalMetadata`` classes. ([#1450](https://github.com/biocore/scikit-bio/pull/1450))
 
 * `skbio.metadata.IntervalMetadata` constructor has a new keyword argument, `copy_from`, for creating an `IntervalMetadata` object from an existing `IntervalMetadata` object with specified `upper_bound`.
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -3043,9 +3043,9 @@ class TreeNode(SkbioObject):
                 if n.length is None:
                     raise ValueError('Missing length.')
                 if not n.is_root():
-                    n._depth = n.parent.length + n.length
+                    n._depth = n.parent.depth + n.length
                 else:
-                    n._depth = n.length
+                    n._depth = 0
         return self._depth
 
     @experimental(as_of="0.4.0")
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -3098,3 +3098,33 @@ class TreeNode(SkbioObject):
 
             yield self
             counter += 1
+
+    def _cache_leafcount(self):
+        """ Counts the number of leaves under each subtree."""
+        for n in self.postorder():
+            if n.is_tip():
+                n.leafcount = 1
+            else:
+                n.leafcount = sum(c.leafcount for c in n.children)
+
+    def _cache_height(self):
+        """ Calculates the height of each subtree."""
+        for n in self.postorder():
+            if n.length is None:
+                raise ValueError('%s in %s has no length.' %
+                                 (n.name, self.__name__))
+            if n.is_tip():
+                n.height = n.length
+
+            else:
+                n.height = max([c.height for c in n.children]) + n.length
+
+    def _cache_depth(self):
+        for n in self.preorder():
+            if n.length is None:
+                raise ValueError('%s in %s has no length.' %
+                                 (n.name, self.__name__))
+            if n.parent:
+                n.depth = n.parent.length + n.length
+            else:
+                n.depth = n.length
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -3025,12 +3025,14 @@ class TreeNode(SkbioObject):
             root = self.root()
             for n in root.postorder(include_self=True):
                 if n.length is None:
-                    raise ValueError('Missing length.')
-                if n.is_tip():
-                    n._height = n.length
-
+                    l = 0
                 else:
-                    n._height = max([c._height for c in n.children]) + n.length
+                    l = n.length
+
+                if n.is_tip():
+                    n._height = l
+                else:
+                    n._height = max([c._height for c in n.children]) + l
         return self._height
 
     @property
@@ -3041,9 +3043,12 @@ class TreeNode(SkbioObject):
             root = self.root()
             for n in root.preorder(include_self=True):
                 if n.length is None:
-                    raise ValueError('Missing length.')
+                    l = 0
+                else:
+                    l = n.length
+
                 if not n.is_root():
-                    n._depth = n.parent.depth + n.length
+                    n._depth = n.parent.depth + l
                 else:
                     n._depth = 0
         return self._depth
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -3004,6 +3004,50 @@ class TreeNode(SkbioObject):
             cached.append(cache_type(func(node)))
             setattr(node, cache_attrname, reduce(reduce_f, cached))
 
+    @property
+    @experimental(as_of="0.5.1")
+    def leafcount(self):
+        """ The number of leaves under the tree."""
+        if not hasattr(self, '_leafcount'):
+            root = self.root()
+            for n in root.postorder(include_self=True):
+                if n.is_tip():
+                    n._leafcount = 1
+                else:
+                    n._leafcount = sum(c.leafcount for c in n.children)
+        return self._leafcount
+
+    @property
+    @experimental(as_of="0.5.1")
+    def height(self):
+        """ The height of the tree."""
+        if not hasattr(self, '_height'):
+            root = self.root()
+            for n in root.postorder(include_self=True):
+                if n.length is None:
+                    raise ValueError('Missing length.')
+                if n.is_tip():
+                    n._height = n.length
+
+                else:
+                    n._height = max([c._height for c in n.children]) + n.length
+        return self._height
+
+    @property
+    @experimental(as_of="0.5.1")
+    def depth(self):
+        """ The depth of the tree."""
+        if not hasattr(self, '_depth'):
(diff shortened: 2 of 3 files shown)


mortonjt:
Note that the original code is from pycogent - but I have had permission from the authors to use it.
This is just providing the helper methods for calculating `leafcount`, `height` and `depth` which are useful for creating tree layouts for visualization.

wasade:
Thanks, @mortonjt. Per our email thread, PyCogent should be fine. 

cc @gregcaporaso because of the check on "external sources" but I am not aware of any issues which require your attention given out-of-band conversation.

skbio/tree/_tree.py
```diff
@@ -3098,3 +3098,33 @@ def shuffle(self, k=None, names=None, shuffle_f=np.random.shuffle, n=1):
 
             yield self
             counter += 1
+
+    def _cache_leafcount(self):
```
wasade: These caches need to be invalidated on topology changes. Please see [`TreeNode.invalidate_caches`](https://github.com/biocore/scikit-bio/blob/master/skbio/tree/_tree.py#L1364).

wasade: It is not clear to me whether these should be executed on the `create_caches`, but it perhaps makes sense.

mortonjt:
Cool.

Not sure what's going on sphinx ... any ideas?

skbio/tree/_tree.py
```diff
@@ -3098,3 +3098,33 @@ def shuffle(self, k=None, names=None, shuffle_f=np.random.shuffle, n=1):
...
+        """ Counts the number of leaves under each subtree."""
+        for n in self.postorder():
```
wasade: To protect against future API changes, and `include_self` be explicitly passed here and in the other two methods?

skbio/tree/_tree.py
```diff
@@ -3098,3 +3098,33 @@ def shuffle(self, k=None, names=None, shuffle_f=np.random.shuffle, n=1):
...
+
+            else:
+                n.height = max([c.height for c in n.children]) + n.length
+
+    def _cache_depth(self):
+        for n in self.preorder():
+            if n.length is None:
+                raise ValueError('%s in %s has no length.' %
+                                 (n.name, self.__name__))
+            if n.parent:
```
wasade: `if n.parent is not None`

skbio/tree/_tree.py
```diff
@@ -3098,3 +3098,33 @@ def shuffle(self, k=None, names=None, shuffle_f=np.random.shuffle, n=1):
...
+                n.depth = n.parent.length + n.length
+            else:
+                n.depth = n.length
```
wasade: Is this correct? If `n == self.root()` then it should have a depth of zero, right?

skbio/tree/tests/test_tree.py
```diff
@@ -1303,6 +1303,42 @@ def test_shuffle_raises(self):
...
+        t = TreeNode.read(io.StringIO("((a:2,b:6)c:4,(d:5,e:1)f:7)r:9;"))
+        t._cache_depth()
+
+        self.assertEquals(t.find('a').depth, 6.0)
+        self.assertEquals(t.find('b').depth, 10.0)
+        self.assertEquals(t.find('c').depth, 13.0)
+        self.assertEquals(t.find('d').depth, 12.0)
+        self.assertEquals(t.find('e').depth, 8.0)
+        self.assertEquals(t.find('f').depth, 16.0)
+        self.assertEquals(t.find('r').depth, 9.0)
```
wasade: This doesn't make sense for the root to have a depth. This length is also not represented in the other depths

skbio/tree/tests/test_tree.py
```diff
@@ -1303,6 +1303,42 @@ def test_shuffle_raises(self):
...
+        t._cache_height()
+        self.assertEquals(t.find('r').height, 21.0)
+        self.assertEquals(t.find('f').height, 12.0)
+        self.assertEquals(t.find('c').height, 10.0)
+        self.assertEquals(t.find('d').height, 5.0)
+        self.assertEquals(t.find('e').height, 1.0)
+        self.assertEquals(t.find('b').height, 6.0)
+        self.assertEquals(t.find('a').height, 2.0)
+
+    def test_cache_depth(self):
```
wasade: Can a test for the `ValueError` be added please?

jairideout:
> cc @gregcaporaso because of the check on "external sources" but I am not aware of any issues which require your attention given out-of-band conversation.

Can you check that the authors of this ported PyCogent code are included in [this list](https://github.com/biocore/scikit-bio#the-pre-history-of-scikit-bio)? They probably are already but good to verify.

Adding a short note in the changelog message giving attribution to PyCogent would also be nice.

Neither of these things are critical since permission has already been obtained.

skbio/tree/_tree.py
wasade: ...actually, this should be `if not n.is_root()`

skbio/tree/tests/test_tree.py
```diff
@@ -1303,6 +1303,42 @@ def test_shuffle_raises(self):
         with self.assertRaises(MissingNodeError):
             next(self.simple_t.shuffle(names=['x', 'y']))
 
+    def test_cache_leafcount(self):
+        t = TreeNode.read([r'((a,b)e,(c,d)f,g)r;'])
+        t._cache_leafcount()
+        self.assertEquals(t.find('r').leafcount, 5)
```
wasade: I'm concerned about this. The developer will need to know to issue a call to a private member method of the tree in order to have access to this public attribute. This present approach will lead to a lot of `AttributeErrors` from users. I think it would be more appropriate for these to be properties so that the caching and construction of the attributes lazy and seamless.

mortonjt:
Yup.  The authors of the pycogent code are @gavin-huttley, Peter Maxwell @gregcaporaso @jairideout and @wasade .  And these authors are on the original list.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## Version 0.5.1-dev (changes since 0.5.1 go here)
 
 ### Features
+* Added ``skbio.tree_tree._cache_leafcount`` ``skbio.tree_tree._cache_height`` and ``skbio.tree_tree._cache_depth`` for calculating leaf count in subtrees, subtree height, and node depth. This is particularly useful for calculating tree coordinates for visualization ([#1514](https://github.com/biocore/scikit-bio/pull/1514))
```
wasade: Adding an inline comment on @jairideout's request of asserting the PyCogent contributors are noted in skbio's pre history, and that PyCogent attribution is also noted

wasade:
@mortonjt, the Sphinx failure is noted in #1511

skbio/tree/tests/test_tree.py
mortonjt: Note that these are helper methods -- I was planning on calling them whenever the coordinates are calculated (for rendering).  

If I'm understanding you - whenever `leafcount` is called, then it should call these helpers?  I'm not sure if I follow, especially for traversal up the the parent nodes.

skbio/tree/tests/test_tree.py
mortonjt: Not sure if I follow this either.  If the root has a `length=5`, shouldn't the depth of the root be `5`?  Or zero?

skbio/tree/tests/test_tree.py
wasade: `TreeNode` is a general object though. We should not merge methods into `TreeNode` which are specific for a downstream application. Regardless, the downstream application should use public API methods, and nothing outside of the `TreeNode` namespace should call a `TreeNode` private member method. 

It is the case that these helpers are generally useful though, and it would be nice if they were easily and appropriately accessible. A `property` is one possible means to achieve this. It will be necessary to ensure that a call does not trigger an unnecessary recursion.

skbio/tree/tests/test_tree.py
wasade: If it is the root, then by definition it does not have a parent -- I'm not sure what the depth of `r` refers to here. It should be zero.

The other confusing thing about these tests is that the depth associated with the root is not reflected in any descendants from the root. However, I think that is a nonissue once the root depth is corrected.

skbio/tree/_tree.py
mortonjt: :+1:

skbio/tree/_tree.py
mortonjt: I don't think that'll work, since there can be un-named nodes that have branch length.

skbio/tree/_tree.py
mortonjt: Done

skbio/tree/_tree.py
mortonjt: Not here, since this won't evaluate root nodes - that's in the if statement.

mortonjt: Spoke too fast - that is fixed!

skbio/tree/tests/test_tree.py
mortonjt: I still don't understand how the property decorator will work.  Would we need to create stale parameters to perform the recursion if these attributes aren't filled?  Also how would be traverse up the tree if say the attributes of a leaf were accessed - but not filled?

mortonjt: Also, would it help if we have push PR specific for the coordinate calculation to motivate this PR?

mortonjt:
The depth, leafcount and height are now property attributes.

skbio/tree/tests/test_tree.py
mortonjt: I'm still not sure about this ...

skbio/tree/_tree.py
wasade: I don't understand how that is relevant to cache invalidation?

skbio/tree/tests/test_tree.py
wasade: A root cannot have a depth as the depth is defined from the root, and a tree cannot have a self-referential edge. The depth is either zero or None here.

skbio/tree/_tree.py
```diff
@@ -3004,6 +3004,50 @@ def reduce_f(a, b):
...
+
+    @property
+    @experimental(as_of="0.5.1")
+    def height(self):
+        """ The height of the tree."""
+        if not hasattr(self, '_height'):
+            root = self.root()
+            for n in root.postorder(include_self=True):
+                if n.length is None:
+                    raise ValueError('Missing length.')
```
wasade: The root does not have a `length` but it does have a `height`.

skbio/tree/tests/test_tree.py
```diff
@@ -1303,6 +1303,51 @@ def test_shuffle_raises(self):
...
+        self.assertEquals(t.find('c').height, 10.0)
+        self.assertEquals(t.find('d').height, 5.0)
+        self.assertEquals(t.find('e').height, 1.0)
+        self.assertEquals(t.find('b').height, 6.0)
+        self.assertEquals(t.find('a').height, 2.0)
+
+    def test_cache_height_error(self):
+        t = TreeNode.read(io.StringIO("((a:2,b:6)c:4,(d:5,e)f:7)r:9;"))
+        with self.assertRaises(ValueError):
+            t.height
```
wasade: This doesn't make sense, the root of a tree has a height

skbio/tree/tests/test_tree.py
wasade: These are generally useful properties. The issues aren't motivation to get them merged, its on ensuring the properties are correct and on smoothing out the API

skbio/tree/tests/test_tree.py
mortonjt: Fixed

skbio/tree/tests/test_tree.py
mortonjt: Right - but node `e` can't have a height, since there isn't a branch length.  So not all of the heights can be cached.  This can get worse if an intermediate node doesn't have a length.

What do you think?  Do you think the absence of a length should be treated as zero?

wasade: This exception will be thrown if the root does not have a length as well given how the code is written

I can see why it would be nice to implicitly treat length as zero, or to except. I'd prefer to get a second opinion on that

mortonjt: @tanaes do you have any thoughts on this?

tanaes: Agree with @wasade on this one. In general trees aren't so well regulated and you get a lot of programs that do things like representing polytomies as random branching clades with length zero (or potentially undefined). And Daniel is totally right that the root is very frequently not given an explicit length, even when most of the tree does. I can see raising an error here in that case being a big headache.

mortonjt: Fixed!  If the lengths are none - they are implicitly assumed to be zero for the sake of the calculations.  But the branch lengths themselves aren't modified.

mortonjt:
I'm going to close this, since I don't know if I'll revisit this.  Anyone who wants this functionality is welcome to fork and create a PR.

antgonza:
Then, what about leaving open so people know that this is a feature request?

Issue #1228, opened 2015-12-09
Add permdisp to distance stats 


johnchase:
I am unable to find a python implementation of permdisp, and it seems like it fit well in scikit-bio and would be useful in conjuction with permanova which is already present. There is a good description of permdisp and it's value [here](http://www.esajournals.org/doi/abs/10.1890/12-2010.1).

Additionally the current version of QIIME [compare_categories.py](https://github.com/biocore/qiime/blob/master/qiime/compare_categories.py#L95) calls permdisp from R. This could be easily replaced in future versions of QIIME if implemented in scikit-bio.

gregcaporaso:
:+1: This will be an important addition as we're planning to use it as a default in QIIME 2.

lkursell:
Any movement on this one?

gregcaporaso:
No, I wish. I'd love to get this in QIIME 2, but it's not highest priority for us at the moment (though if someone did get it in scikit-bio I'd get it hooked up to `qiime diversity beta-group-significance` quickly).

lkursell:
+2

PR #1532, opened 2017-09-12
Permdisp

nreeve17:

A part of this program uses the function hdmedians, which is under a GNU-GPL license. We have reached out to the author of hdmedians and asked if he could use a double license so that it could be compatible with scikit-bio. The author has yet to respond to our request. 

Benchmarking was done to compare the performance of the two test types that PERMDISP performs, median and centroid, against the performance of the analogous betadisper function in R, used by QIIME. Betadisper uses the median test by default and is what is used by QIIME.

Matrix Size | Time (s) Median scikit-bio | Time (s) Centroid scikit-bio | Time (s) QIIME PERMDISP (median)
-- | -- | -- | --
10 | 4.8 | 2.93 | 3.3
50 | 4.94 | 2.99 | 3.56
100 | 5.21 | 3.1 | 3.43
150 | 5.79 | 3.38 | 3.56
200 | 7.73 | 3.66 | 3.67
300 | 8.6 | 4.47 | 4.13
500 | 15.4 | 7.22 | 5.79
1000 | N/A | 26.9 | 12.72

image
image
image


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 
 * `skbio.metadata.IntervalMetadata.drop` has a new boolean parameter `negate` to indicate whether to drop or keep the specified `Interval` objects.
 
+* `skbio.stats.distance` adding function permdisp. This fixes issue-1228
```
gregcaporaso: Change last sentence to: 

This fixes [issue 1228](https://github.com/biocore/scikit-bio/issues/1228).

skbio/stats/distance/_permdisp.py
```diff
@@ -0,0 +1,269 @@
...
+        IDs in `distance_matrix` (i.e., the row labels must be distance matrix
+        IDs), but the order of IDs between `distance_matrix` and the
+        ``DataFrame`` need not be the same. All IDs in the distance matrix must
+        be present in the ``DataFrame``. Extra IDs in the ``DataFrame`` are
+        allowed (they are ignored in the calculations).
+    column : str, optional
+        Column name to use as the grouping vector if `grouping` is a
+        ``DataFrame``. Must be provided if `grouping` is a ``DataFrame``.
+        Cannot be provided if `grouping` is 1-D ``array_like``.
+    test : {'centroid', 'median'}
```
gregcaporaso: Is `median` the default in the R package too? If not, it's worth noting here that we're doing something different than the R implementation, and noting why (with a citation). Otherwise we'll get a lot of support questions asking why the results don't match the R results.

skbio/stats/distance/_permdisp.py
```diff
@@ -0,0 +1,269 @@
...
+        ``p-value``.
+
+    Raises
+    ------
+    TypeError
+        If, when using the spatial median test, the pcoa ordination is not of
+        type np.float32 or np.float64, the spatial median function will fail
+        and the centroid test should be used instead
+    ValueError
+        If the test is not centroid or median. test is set to median by default
```
gregcaporaso: I would drop the note about the default setting here - this statement could easily be overlooked if the default were ever changed.

skbio/stats/distance/_permdisp.py
```diff
@@ -0,0 +1,269 @@
...
+    TypeError
+        If the distance matrix is not an instance of a distance matrix
```
gregcaporaso: *of a distance matrix* should be *of a `skbio.DistanceMatrix`.

skbio/stats/distance/_permdisp.py
```diff
@@ -0,0 +1,269 @@
...
+        statistical significance calculations will be skipped and the p-value
+        will be ``np.nan``.
+
+    Returns
+    -------
+    pandas.Series
+        Results of the statistical test, including ``test statistic`` and
+        ``p-value``.
+
+    Raises
```
gregcaporaso: Should there be a note here for issues with `grouping`? This would include:
 - if a `list` is provided and `column` is provided
 - if a `list` is provided and there is a length mismatch with `distance_matrix`
 - if a `pd.DataFrame` is provided and there are ids in `distance_matrix` that are not in `grouping`
 - there is only one group

skbio/stats/distance/_permdisp.py
```diff
@@ -0,0 +1,269 @@
...
+    original distances to principle coordinates. PERMDISP calculates an
+    F-statistic to assess whether the dispersions between groups is significant
+
+
+    Parameters
+    ----------
+    distance_matrix : DistanceMatrix
+        Distance matrix containing distances between objects (e.g., distances
+        between samples of microbial communities).
+    grouping : 1-D array_like or pandas.DataFrame
```
gregcaporaso: Are there any restrictions on the number of samples that must be in each group? Are groups with one sample in them valid, or should an error be raised?

skbio/stats/distance/_permdisp.py
```diff
@@ -0,0 +1,269 @@
...
+    test statistic name        F-value
+    sample size                      6
+    number of groups                 2
+    test statistic             1.03296
+    p-value                        NaN
+    number of permutations           0
+    Name: PERMDISP results, dtype: object
+
+    PERMDISP computes variances based on two types of tests, using either
+    centroids or spatial medians, also commonly reffered to as a geometric
```
gregcaporaso: reffered -> referred

skbio/stats/distance/_permdisp.py
```diff
@@ -0,0 +1,269 @@
...
+    median or mediancentre. The spatial median is thought to yeild a more
```
gregcaporaso: yeild -> yield

gregcaporaso: Is `mediancentre` a term used in an implementation of this method? Would this more correctly be referred to in text as median center?

skbio/stats/distance/_permdisp.py
```diff
@@ -0,0 +1,269 @@
...
+    ...                       [0.66, 0.77, 0.44, 0.75, 0, 0.77],
+    ...                       [0.33, 0.61, 0.55, 0.88, 0.77, 0]],
+    ...                       ['s1', 's2', 's3', 's4', 's5', 's6'])
+    >>> grouping = ['G1', 'G1', 'G1', 'G2', 'G2', 'G2']
+
+    Run PERMDISP using 99 permutations to caluculate the p-value:
+
+    >>> from skbio.stats.distance import permdisp
+    >>> import numpy as np
+    >>> #make output deterministic; not necessary for normal use
```
gregcaporaso: Please make the following change each time it comes up in this doc string:

*not necessary for normal use* -> *should not be included during normal use*

(I think that statement should be a little stronger, since it really would technically be wrong for the user to run this a bunch of times with the same seed.)

skbio/stats/distance/_permdisp.py
```diff
@@ -0,0 +1,269 @@
...
+    behavior: note that even though the ``DataFrame`` had 7 objects, only 6
+    were used in the test (see the "Sample size" row in the results above to
+    confirm this). Thus, the ``DataFrame`` can be a superset of the distance
+    matrix IDs. Note that the reverse is not true: IDs in the distance matrix
+    *must* be present in the ``DataFrame`` or an error will be raised.
+
+    PERMDISP should be used to determine wether the dispersions between the
+    groups in your distance matrix are significantly separated.
+    A non-significant test result indicates that group dispersions are similar
+    to eachother. PERMANOVA or ANOSIM should then be used in conjunction to
```
gregcaporaso: eachother -> each other

skbio/stats/distance/_permdisp.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+@experimental(as_of="0.5.1")
+def permdisp(distance_matrix, grouping, column=None, test='median',
+             permutations=999):
+    """Test for Homogeneity of Multivariate Groups Disperisons using Marti
+    Anderson's PERMDISP2 procedure.
+
+    PERMDISP is a multivariate analogue of Levene's test for homogeneity of
+    multivariate variances. Non-euclidean distances are handled by reducing the
```
gregcaporaso: It looks like `pcoa` is always the first step. Is this comment about non-euclidean distances relevant?

skbio/stats/distance/_permdisp.py
```diff
@@ -0,0 +1,269 @@
...
+        groups.append(cdist(df.values[:, :-1], [centroids.loc[label].values],
+                      metric='euclidean'))
+
+    stat, _ = f_oneway(*groups)
+    stat = stat[0]
+
+    return stat
+
+
+def _config_med(x):  # not explicitly tested by permdisp test suite
```
gregcaporaso: We should have some basic tests, in part to know that we'll catch a change in the behavior of `hdmedians` if that ever happens.

skbio/stats/distance/_permdisp.py
nreeve17: The most recent documentation for betadisper (the permdisp function in R) says that `median` is the default, some older documentation says only `centroid` is supported.

skbio/stats/distance/_permdisp.py
nreeve17: I am bad at spelling. The term `mediancentre` was one I found reading through literature that supported this method. I don't think it is necessary to include, I can remove it.

ElDeveloper:
> @nreeve17, @ElDeveloper, this looks great!

@gregcaporaso Thanks for reviewing.

> Did you ever hear back on whether we can use hdmedians here? I guess one option, if we can't use it, is to only support method=centroid for now... I realize that's not ideal, but it might be better than nothing.

Yes, we heard back. hdmedians is now Apache licensed (see [this issue](https://github.com/daleroberts/hdmedians/issues/1)).

> When we are implementing a Python version of some existing non-Python method (like we are here), it's standard practice to include tests that ensure that we achieve the same results as that method on complex data. In this case, a good "complex" data set might the QIIME 2 moving pictures tutorial unifrac distance matrix, and the "body site" and "subject" groupings (sample metadata here). Could you run that through betadispr and add tests to confirm that you get the same F-statistics for these groupings with the centroid and median tests? This could all be in one new unit test function (e.g., test_confirm_betadispr_results). It's important that we check this (and our users will).

@gregcaporaso, yes this sounds like a great idea.

gregcaporaso:
Excellent news about the license! It sounds like the `hdmedians` project is in a fairly early stage of development based on the author's comments on that thread, which is a little worrying for including it as a dependency. But, I think as long as we're sure that our unit tests would catch any breaking changes in `hdmedians`, we're ok. For that reason, it might make sense to add a test explicitly checking the results of a call into `hdmedians` (so it's easy for us to differentiate a breaking change in `hdmedians` from an error in skbio).

ElDeveloper:
I agree, that's a really good point. Perhaps we should also pin down the 
exact version we use?

Do we need to include the license? It's now listed as a dependency and 
we use a function from their public API.

gregcaporaso:
> Perhaps we should also pin down the exact version we use?

I thought about that initially, but the developer mentioned concern about making sure that packages which depend on `hdmedians` are able to get updates... I'm worried about pinning the version, and then forgetting about it and never getting updates as a result. What do you think @ElDeveloper? Maybe @ebolyen can weigh in on this too? 

> Do we need to include the license? It's now listed as a dependency and we use a function from their public API.

Oops, you're right, no license is needed.

ebolyen:
The convention thus far in skbio has been to pin a minimum version and then wait for travis to signal an API break via tests. This puts you on the hook for writing tests (against the skbio API, which in turn act as coarse integration tests of the dependency), but those should probably exist anyhow.

I'm sure there's better approaches, like pinning a range and then automatically testing the bounds, but without some fancy CI machinery, that's not the most feasible thing for skbio.

skbio/stats/distance/_permdisp.py
nreeve17: Well, I believe we added that comment because PCoA does handle non-euclidean distances, but it could be changed to simply says `Distances are handled...`

skbio/stats/distance/_permdisp.py
nreeve17: I don't believe we considered this, it is not reflected in the program or the tests. I can write those considerations in. I am not sure how the program will work if theres only one sample.

skbio/stats/distance/_permdisp.py
nreeve17: will do!

skbio/stats/distance/_permdisp.py
nreeve17: I checked and groups with one sample in them are indeed valid, but the F-statistic will be nan if all groups contain only one sample

ElDeveloper:
@gregcaporaso I believe @nreeve17 should have addressed all the comments now. 🎉 

While going through this we encountered that running `permdisp` on randomly generated matrices (as generated by `randdm` in skbio) will result in large discrepancies between this implementation and R. This problem is not observed in **real** datasets, we tested this in at least 5 other published studies.

The problem arises (we believe) from the computation of Principal Coordinates Analyses, we see a warning that indicates that the computation might be misleading i.e. `RuntimeWarning: The result contains negative eigenvalues ... The smallest eigenvalue is -0.6690970550290323 and the largest is 0.8581202056664433.`. As such we think it is safe to ignore this issue, but we wanted to bring this up here for your consideration.

skbio/stats/distance/tests/test_permdisp.py
```diff
@@ -0,0 +1,253 @@
...
+        np.isnan(pval)
+
+    def test_hdmedians(self):
+        exp = np.array([2.01956244, 1.53164546, 2.60571752, 0.91424179,
+                        1.76214416, 1.69943057])
+        obs = np.array(hd.geomedian(self.eq_mat.data))
+        npt.assert_almost_equal(obs, exp, decimal=6)
+
+    def test_confirm_betadispr_results(self):
+        mp_dm = DistanceMatrix.read('data/moving_pictures_dm.tsv')
```
ElDeveloper: Can you replace this and other instances with `get_data_path`, this function allows you to fetch data files found in the test module ... `from skbio.util import get_data_path`. For example see how it is being used [here](https://github.com/biocore/scikit-bio/blob/master/skbio/stats/ordination/tests/test_principal_coordinate_analysis.py#L25)

skbio/stats/distance/tests/test_permdisp.py
```diff
@@ -0,0 +1,253 @@
...
+        mp_mf.set_index('#SampleID', inplace=True)
+
+        exp_med_mp = 10.4105
+        exp_cen_mp = 17.727
+        obs_med_mp = permdisp(mp_dm, mp_mf,
+                              column='BodySite')['test statistic']
+        obs_cen_mp = permdisp(mp_dm, mp_mf, column='BodySite',
+                              test='centroid')['test statistic']
+
+        self.assertAlmostEqual(exp_med_mp, obs_med_mp, places=0)
```
ElDeveloper: Can you test for the whole pd.Series object? While you do that can you validate that the p-value doesn't change in significance i.e. that both are <0.05 or >0.05.

skbio/stats/distance/tests/test_permdisp.py
nreeve17: can do!

skbio/stats/distance/tests/test_permdisp.py
nreeve17: great idea!

skbio/stats/distance/_permdisp.py
nreeve17: did this

skbio/stats/distance/_permdisp.py
nreeve17: should be done now ✅

nreeve17:
All of the comments have been addressed. We compiled a table to summarize the results between permdisp in skbio and betadisper used by R, with links to study ids in Qiita. 

# Spatial Median Comparisons
**Data Set**|**Sample Size**|**metadata grouping**|**Number groups**|**Median F-val skbio**|**Median F-val betadisper**|**M p-val skbio**|**M p-val betadisper**|Qiita Study|PubMed
:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|:----:|:----:|
Moving Pictures|33|BodySite|4|10.1956|10.4105|0.001|0.001|[550](https://qiita.ucsd.edu/study/description/550)|[21624126](https://www.ncbi.nlm.nih.gov/pubmed/21624126)
Moving Pictures|33|Subject|2|0.0228117|0.02202|0.868|0.91|[550](https://qiita.ucsd.edu/study/description/550)|[21624126](https://www.ncbi.nlm.nih.gov/pubmed/21624126)
Keyboard|77|sample\_type|2|7.64607|7.64604|0.015|0.009|[232](https://qiita.ucsd.edu/study/description/232)|[20231444](https://www.ncbi.nlm.nih.gov/pubmed/20231444)
Skin-forensic|17|sample\_type|2|0.989882|0.9901|0.309|0.351|[393](https://qiita.ucsd.edu/study/description/393)|[20231444](https://www.ncbi.nlm.nih.gov/pubmed/20231444)
diet mammals converge|39|host\_clade|6|30.0668|30.07818|0.989|0.001|[626](https://qiita.ucsd.edu/study/description/626)|[25974302](https://www.ncbi.nlm.nih.gov/pubmed/25974302)
diet mammals converge|39|host\_family|22| 21.1988|21.19901| 0.521| 0.001|[626](https://qiita.ucsd.edu/study/description/626)|[25974302](https://www.ncbi.nlm.nih.gov/pubmed/25974302)
diet mammals converge|39|elevation|5|72.6057|72.61026|0.084|0.001|[626](https://qiita.ucsd.edu/study/description/626)|[25974302](https://www.ncbi.nlm.nih.gov/pubmed/25974302)
diet mammals converge|39|diet|3|1.39989|1.39994| 0.439|0.293|[626](https://qiita.ucsd.edu/study/description/626)|[25974302](https://www.ncbi.nlm.nih.gov/pubmed/25974302)
newborns|79|sex|2|2.2111|2.211|0.154|0.144|[395](https://qiita.ucsd.edu/study/description/395)|[20566857](https://www.ncbi.nlm.nih.gov/pubmed/20566857)
newborns|79|delivery\_mode|3|10.8083|10.808|0.001|0.001|[395](https://qiita.ucsd.edu/study/description/395)|[20566857](https://www.ncbi.nlm.nih.gov/pubmed/20566857)

# Centroid Comparisons
**Data Set**|**Sample Size**|**metadata grouping**|**Number groups**|**Centroid F-val skbio**|**Cetroid F-val betadisper**|**M p-val skbio**|**M p-val betadisper**|Qiita Study|PubMed
:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|:----:|:----:|
Moving Pictures|33|BodySite|4|17.4242|17.72748|0.001|0.001|[550](https://qiita.ucsd.edu/study/description/550)|[21624126](https://www.ncbi.nlm.nih.gov/pubmed/21624126)
Moving Pictures|33|Subject|2|0.0467885|0.04549|0.848|0.854|[550](https://qiita.ucsd.edu/study/description/550)|[21624126](https://www.ncbi.nlm.nih.gov/pubmed/21624126)
Keyboard|77|sample\_type|2|9.28355|9.28355|0.006|0.007|[232](https://qiita.ucsd.edu/study/description/232)|[20231444](https://www.ncbi.nlm.nih.gov/pubmed/20231444)
Skin forensic|17|sample\_type|2|1.40085|1.40085|0.329|0.266|[393](https://qiita.ucsd.edu/study/description/393)|[20231444](https://www.ncbi.nlm.nih.gov/pubmed/20231444)
diet mammals converge|39|host\_clade|6|47.556|47.55598  |0.994|0.001|[626](https://qiita.ucsd.edu/study/description/626)|[25974302](https://www.ncbi.nlm.nih.gov/pubmed/25974302)
diet mammals converge|39|host\_family|22|71.9981|71.99812|0.451|0.001|[626](https://qiita.ucsd.edu/study/description/626)|[25974302](https://www.ncbi.nlm.nih.gov/pubmed/25974302)
diet mammals converge|39|elevation|5|90.2223|90.2223|0.103|0.001|[626](https://qiita.ucsd.edu/study/description/626)|[25974302](https://www.ncbi.nlm.nih.gov/pubmed/25974302)
diet mammals converge|39|diet|3|2.08823|2.08823|0.405|0.135|[626](https://qiita.ucsd.edu/study/description/626)|[25974302](https://www.ncbi.nlm.nih.gov/pubmed/25974302)
newborns|79|sex|2|2.88853|2.88853|0.138|0.108|[395](https://qiita.ucsd.edu/study/description/395)|[20566857](https://www.ncbi.nlm.nih.gov/pubmed/20566857)|
newborns|79|delivery\_mode|3|14.0445|14.04454|0.001|0.001|[395](https://qiita.ucsd.edu/study/description/395)|[20566857](https://www.ncbi.nlm.nih.gov/pubmed/20566857)

gregcaporaso:
That table is very helpful, thanks for compiling that! I think it may be worth putting a note in the doc string indicating that the results are slightly different from betadispr, and indicating why - this will save a lot of support questions about that. I'll leave this up to @nreeve17 and @ElDeveloper to wrap up and merge (i.e., I don't need to see it again before merge). Thanks for the work on this @nreeve17!

RNAer:
@nreeve17 , thanks for the tables! very helpful. I am wondering why there are big differences in p-value in the centroid table. Like in "diet mammals converge" data set, the betadisper p-val is 0.001 while skbio is 0.994.

.travis.yml
```diff
@@ -24,6 +24,7 @@ install:
   - python --version
   - pip install -r ci/pip_requirements.txt
   - pip install . --no-deps
+  - pip install hdmedians
```
RNAer: not big deal, but could you put `hdmedians` into `ci/pip_requirement.txt` and remove this line?

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 
 * `skbio.metadata.IntervalMetadata.drop` has a new boolean parameter `negate` to indicate whether to drop or keep the specified `Interval` objects.
 
+* `skbio.stats.distance` adding function permdisp. This fixes issue 1228
```
RNAer: this needs to be updated to 0.5.2-dev as we already released 0.5.2. update after #1565 gets merged.

skbio/stats/distance/_permdisp.py
```diff
@@ -0,0 +1,281 @@
...
+
+import hdmedians as hd
+
+from ._base import (_preprocess_input, _run_monte_carlo_stats, _build_results)
+
+from skbio.stats.ordination import pcoa
+from skbio.util._decorator import experimental
+
+
+@experimental(as_of="0.5.1")
```
RNAer: again needs to update to `0.5.2`

skbio/stats/distance/_permdisp.py
```diff
@@ -0,0 +1,281 @@
...
+    ValueError
+        If a list and a column name are both provided
+    ValueError
+        If a list is provided for `grouping` and it's length does not match
+        the number of ids in distance_matrix
+    ValueError
+        If all of the values in the grouping vector are unique
+    KeyError
+        If there are ids in grouping that are not in distance_matrix
+
```
RNAer: I would collapse the exception types together to be clearer. but it's up to you.

RNAer:
@nreeve17 , sorry if you have discussed this before: it looks like the only `hdmedians.geomedian` is used? I just looked briefly, it does the same computation as `scipy.stats.mstats.gmean`? is it much faster than scipy's version?

skbio/stats/distance/_permdisp.py
```diff
@@ -0,0 +1,281 @@
...
+    were used in the test (see the "Sample size" row in the results above to
+    confirm this). Thus, the ``DataFrame`` can be a superset of the distance
+    matrix IDs. Note that the reverse is not true: IDs in the distance matrix
+    *must* be present in the ``DataFrame`` or an error will be raised.
+
+    PERMDISP should be used to determine wether the dispersions between the
+    groups in your distance matrix are significantly separated.
+    A non-significant test result indicates that group dispersions are similar
+    to each other. PERMANOVA or ANOSIM should then be used in conjunction to
+    determine wether clustering within groups is significant.
```
wasade: whether

skbio/stats/distance/_permdisp.py
```diff
@@ -0,0 +1,281 @@
...
+def _config_med(x):
+    """
+    slice the vector up to the last value to exclude grouping column
+    and transpose the vector to be compatible with hd.geomedian
+    """
+    X = x.values[:, :-1]
+    return np.array(hd.geomedian(X.T))
+
+
+def _compute_median_groups(samples, grouping):
```
wasade: it looks like the only difference between this function and `_compute_centroid_groups` is the method passed to `.aggregate`? Would it make sense to decompose the logic to reduce code replication?

nreeve17: I could do this, we should've considered this before. It would require some of the tests to be refactored as they test those different methods explicitly

RNAer:
@nreeve17 , thanks for explaining and addressing comments! I think it's good to merge

PR #1548, opened 2018-02-04
Optionally install matplotlib/IPython (updated version of #886)

jvkersch:
This is an updated version of #886. I merged in master, resolved the conflicts, and updated the places in the code that had grown matplotlib dependencies since that PR was created. The final version of the code doesn't differ very much from #886, except the following:

* Doctests are skipped when matplotlib is not installed.
* Matplotlib support is bundled as an extra requirement. I am not sure this is the right thing. Given that most users probably want matplotlib installed by default, it would probably be better to install it always, but have a flag to turn off matplotlib when the user requests it.


jvkersch:
Is there someone who could give this a review? Thanks!

jvkersch:
Thanks @antgonza for the quick review! There's one thing that I'm not sure about, and it's whether the code should make matplotlib an opt-out instead of an opt-in (as this PR would change it). It feels to me that many people use skbio as an all-round analysis package (like Pandas and sklearn) and they would be surprised if new installations stopped installing matplotlib. By contrast, someone who is keen on running skbio in an environment with limited resources can probably be pointed to a flag to supply if they don't want matplotlib. What do you think? If you agree that matplotlib should be an opt-out, then I can probably quickly make the changes.

antgonza:
Good point! I like the idea of the opt-out ...

jvkersch:
@antgonza Apologies for the delay, but I can't find a good way to make matplotlib an opt-out... I was looking at how other projects tackle this (Pandas, scikit-learn, ...) and they have a core of required dependencies (Numpy, SciPy, ...) and then a large list of optional dependencies, like Matplotlib, that are isolated in conditional imports. With that in mind, perhaps it isn't so bad to leave the current behavior (Matplotlib opt-in) as-is?

antgonza:
OK, thanks. @wasade or @RNAer, could you take a look?

RNAer:
agreed on keep a core and a optional dependency lists

On Tue, Feb 20, 2018 at 12:54 PM, Antonio Gonzalez <notifications@github.com

RNAer:
per the discussion in the today's sprint, we are not convinced that matplotlib need to switch from a required dep to optional dep because 1) matplotlib is such commonly used pkg; 2) we don't see the concern of having matplotlib on local or remote machine. We will unmerge this PR unless there is any problem emerging in the future

jvkersch:
@RNAer Fair enough. Feel free to close this PR if there is no chance of it getting merged anytime soon.

For the record, my main motivation for avoiding Matplotlib is that it pulls in a lot of extra dependencies, and this can be quite hard to install in a Docker container (my main use case), where you typically don't want all this extra baggage. However, I was able to avoid much of the superfluous bloat by having Matplotlib around, but with a non-GUI backend (`export MPLBACKEND=agg`).

RNAer:
Thanks for your work on this, @jvkersch ! Although it turned out it was not merged, we really appreciate your input on this and let us know the solution of `export MPLBACKEND=agg` solves the problem.

mortonjt:
@jvkersch feel free to reopen later.  We are welcome up reopen discussion on making matplotlib an optional dependency in the future.

PR #1577, opened 2018-06-13
Add PCoA biplot

ElDeveloper:
Implement the biplot projection into a matrix of eigenvectors resulting from PCoA. The interface is pretty straightforward and takes as input a set of ordination results and a matrix of descriptors (abundances, environmental variables, etc). For example see the sketch below; we load a table, calculate a distance matrix, then compute the PCoA and finally get the ordination results but with the `features` attribute updated:

```python
table = load_table('my-table.biom')
dm = beta_diversity('jaccard', table.T)
res = pcoa(dm)

# Notes:
# It could alternatively *only* return the biplot matrix.
# Table can be bacterial abundances or any other environmental variables
res = pcoa_biplot(res, table)
```

The method itself is based on the text from Legendre and Legendre:

screen shot 2018-06-12 at 10 30 37 pm

~~I've checked this against what I think is the only reference implementation of this method, found in R's ape package. Note that the results seem to be slightly off, see notes in the tests.~~

Would appreciate if @Jorge-C, @antgonza, and @mortonjt can have a look at the implementation.

---------------------


skbio/stats/ordination/_principal_coordinate_analysis.py
```diff
@@ -142,3 +142,63 @@ def pcoa(distance_matrix):
...
+    Numerical Ecology 3rd edition.
+
+    Parameters
+    ----------
+    ordination: OrdinationResults
+        The computed principal coordinates analysis of dimensions (n, c) where
+        the matrix ``y`` will be projected onto.
+    y: DataFrame
+        Samples by features table of dimensions (n, m). These can be
+        environmental features or abundance counts. This table should be
```
mortonjt: Would it make much sense to feed in the same table used to calculate the PCoA?  Something seems a bit circular about using abundance counts ...

skbio/stats/ordination/_principal_coordinate_analysis.py
```diff
@@ -142,3 +142,63 @@ def pcoa(distance_matrix):
...
+        normalized in cases of dimensionally heterogenous physical variables.
+
+    Returns
+    -------
+    OrdinationResults
+        The modified input object that includes projected features onto the
+        ordination space in the ``features`` attribute.
```
mortonjt: An example on how to run this would be great, especially given that this is a fairly new method in this ecosystem -- it isn't exactly clear what proper inputs / outputs should look like.

skbio/stats/ordination/__init__.py
```diff
@@ -126,11 +127,11 @@
 from ._redundancy_analysis import rda
 from ._correspondence_analysis import ca
 from ._canonical_correspondence_analysis import cca
-from ._principal_coordinate_analysis import pcoa
+from ._principal_coordinate_analysis import pcoa, pcoa_biplot
 from ._ordination_results import OrdinationResults
 from ._utils import (mean_and_std, scale, svd_rank, corr, e_matrix, f_matrix)
 
-__all__ = ['ca', 'rda', 'cca', 'pcoa', 'OrdinationResults',
+__all__ = ['ca', 'rda', 'cca', 'pcoa', 'pcoa_biplot', 'OrdinationResults',
```
mortonjt: Might be something to bring up to the rest of the group -- would it make more sense to pass in one optional argument into `pcoa` to create biplots vs adding a new method?  Maybe we could have something like
```python
def pcoa(distance_matrix, y=None):
'''
If y is specified, biplots will be generated ...
'''
```

skbio/stats/ordination/_principal_coordinate_analysis.py
ElDeveloper: Although circular, it seems fine. The book makes mentions of this use-case and it is the use-case of the biplots impelmentation in QIIME 1.

From a different book by Legendre:

> [On PCoA biplots] ... For projection of species data onto a PCoA plot, it is important to use the species data with the same transformation (if any) as the one used to compute the dissimilarity matrix.

skbio/stats/ordination/_principal_coordinate_analysis.py
ElDeveloper: Great idea, I'll add something.

skbio/stats/ordination/__init__.py
ElDeveloper: The interface mimics what `ape` does. Which is likely for cases where a *large* PCoA has already been computed. In that case, there's no reason for recomputing the eigendecomposition.

mortonjt:
That quite cringeworthy. But if it's quoted in the book, the odus is on
them, not us. So this should be good as is

ElDeveloper:
😆

skbio/stats/ordination/__init__.py
mortonjt: @antgonza do you have any thoughts on this?  I was thinking that it maybe a good idea to incorporate this functionality directly into the pcoa function.

antgonza: I think going what's implemented is the easiest. Note that I agree that having the pcoa calculation and the pcoa_biplot in different steps makes sense due to large distance matrices. BTW that's how biplots used to work in Qiime1 too.

mortonjt:
Alright then. Can we get this merged in then?

antgonza:
The tests are failing, right? So no until they pass?

skbio/stats/ordination/_principal_coordinate_analysis.py
```diff
@@ -142,3 +142,63 @@ def pcoa(distance_matrix):
...
+    """
+
+    # acknowledge that most saved ordinations lack a name, however if they have
+    # a name, it should be PCoA
+    if (ordination.short_method_name is not '' and
```
RNAer: just curious, is `is not` on purpose (instead of `==`?

RNAer:
The travis failure is due to the pandas DataFrame change. The column (or row) orders between `exp` and `obs` in the tests aren't aligned anymore...

ElDeveloper:
@RNAer, those are unrelated to this PR (as far as I understand). I've opened an issue here: #1578. Any chance you or @mortonjt could have a look?

ElDeveloper:
Tests are passing 🎆

mortonjt:
I think this is ready to go.  Any remaining comments @RNAer ?

ElDeveloper:
Thanks so much for reviewing and merging! 🎉

PR #1599, opened 2018-08-27
Update version string and changelog to 0.5.4-dev

HannesHolste:


ElDeveloper:
Thanks @HannesHolste!! Lets wait until we have pushed the release to PYPI to merge this (just in case we needed to make any additional and unforeseen modifications).

PR #1641, opened 2018-12-02
Added .editorconfig

KOLANICH:

* [x] **This pull request includes code, documentation, or other content derived from external source(s).**  - I'm the author of the copied content.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+root = true
+
+[*]
+charset=utf-8
+indent_style = space
+indent_size = 4
+insert_final_newline = false


wasade:
@KOLANICH, when you have a moment, could you provide a little more context here? This file is not part of, for example, numpy or scikit-learn.

KOLANICH:
> This file is not part of, for example, numpy or scikit-learn.

It is not, they have decided they don't need it. But you should make your own decision.

>could you provide a little more context here?

it changes editor prefs to match project's conventions. Tabs/spaces, tab width (or count of spaces to insert), line ending, stuff like that. It is not a replacement for autopep8 and black. They format source code after it is ready for a PR. This tool changes editor prefs making applying these tools unneeded in order to make the sourcs work (if editor is set up to differrent count of spaces or to using tabs instead of spaces, one will get a syntax error on any modification of pep-8 compliant source. A possible solution (promoted by pep-8 fanatics) is to set an editor to always use pep-8-compliant settings on python files, but it is only acceptable solution for fanatics. I am not a fanatic, so if a project doesn't use .editorconfig, I usually just copy it from an another project rather than changing my editor settings manually).

.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*]
+charset=utf-8
```
thermokarst:
```suggestion
charset = utf-8
```

.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*]
```
thermokarst: Can you make this section specific for Python source files? Also, this project contains a bit of c source - can you add a section for that, too? A Makefile section would be nice, too. My preference is to not apply these settings globally, since we do have different _types_ of source in this repo.

KOLANICH: Done. C sources seem to have the same conventions - both generated with cython and third-party libs.

.editorconfig
KOLANICH: done.

wasade:
@thermokarst do the changes here satisfy your requests?

thermokarst (approved):

PR #1643, opened 2018-12-08
skbio version 0.5.5 changelog

mortonjt:

Note that this should be the final changelog before the scikit-bio 0.5.5 release.  @thermokarst @ebolyen @wasade @ElDeveloper

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 
 
-## Version 0.5.5
+## Version 0.5.5 (2018-12-10)
 
 ### Features
 
@@ -21,14 +21,15 @@
 ### Deprecated functionality [experimental]
 
 ### Miscellaneous
+* Python 3.6 and 3.7 compatibility is now supported
 
 * A pytest runner is shipped with every installation ([#1633](https://github.com/biocore/scikit-bio/pull/1633))
 
-* The nosetest framework has has been replace in favor of pytest ([#1624](https://github.com/biocore/scikit-bio/pull/1624))
+* The nosetest framework has been replaced in favor of pytest ([#1624](https://github.com/biocore/scikit-bio/pull/1624))
 
-* The numpy docs are deprecated in favor of napoleon ([#1629](https://github.com/biocore/scikit-bio/pull/1629))
+* The numpy docs are deprecated in favor of [Napoleon](http://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html) ([#1629](https://github.com/biocore/scikit-bio/pull/1629))
 
-* This version is now compatiable with pandas 0.23 ([#1627](https://github.com/biocore/scikit-bio/pull/1627))
+* This version is now compatible with NumPy >= 1.9.2 and Pandas >= 0.23. ([#1627](https://github.com/biocore/scikit-bio/pull/1627))
 
 
 ## Version 0.5.4 (2018-08-23)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # scikit-bio changelog
 
-## Version 0.5.4-dev
+
+
+## Version 0.5.5
 
 ### Features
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -20,6 +22,15 @@
 
 ### Miscellaneous
 
+* A pytest runner is shipped with every installation ([#1633](https://github.com/biocore/scikit-bio/pull/1633))
+
+* The nosetest framework has has been replace in favor of pytest ([#1624](https://github.com/biocore/scikit-bio/pull/1624))
```
ElDeveloper: has has -> has
replace -> replaced

CHANGELOG.md
```diff
@@ -20,6 +22,15 @@
...
+
+* The numpy docs are deprecated in favor of napoleon ([#1629](https://github.com/biocore/scikit-bio/pull/1629))
```
ElDeveloper: I think I would change this to be `* The documentation is now built using [Napoleon](http://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html) .......`.

CHANGELOG.md
```diff
@@ -20,6 +22,15 @@
...
+
+* This version is now compatiable with pandas 0.23 ([#1627](https://github.com/biocore/scikit-bio/pull/1627))
```
ElDeveloper: Can you also mention NumPy? Maybe mention the ranges themselves, something like:

` This version is now compatible with NumPy >= 1.9.2 and Pandas >= 0.23`.

mortonjt:
Thanks for the feedback @ElDeveloper !  

I now passed into the changelog that there is compatibility with py3.6 and py3.7
However, looking at the travis log, it looks like these versions of python are throwing errors in the docstrings: https://travis-ci.org/biocore/scikit-bio/jobs/465455313

I'm guessing that we should fix these issues in this release.

ElDeveloper:
👍 Looks good to me!

ElDeveloper:
I missed your message. Yeah, it looks like there's a few flake8 errors 😞

mortonjt:
Yes, but the flake8 errors seem to be specific to python 3.6 and 3.7. Not sure why -- I'm double checking the napolean docs to make sure those are rendered properly

mortonjt:
Ok, I'm disabling W605 -- really not sure how you can get working regex tests with this weird limitation.

Otherwise, this should be good to go. @ElDeveloper want to merge this in after the tests pass?

ElDeveloper:
I don't think it is a good idea to disable W605, since this is a rule that affects newer versions of Python. I also wouldn't be in favor of dropping support for python 3.5 based on a flake8 failure. This is the error in the 3.5 build:

```
flake8 skbio setup.py checklist.py --extend-ignore=W605
Usage: flake8 [options] file file ...
flake8: error: no such option: --extend-ignore
```

mortonjt:
Ok that's fine. Could you look at one of the regular expressions then? It
appeared to be a bit of a catch 22

ElDeveloper:
Yes, I'll have a look in a bit. :+1:

mortonjt:
@antgonza or @wasade could either of you double check this?

skbio/io/format/tests/test_stockholm.py
```diff
@@ -853,8 +853,8 @@ def test_gr_wrong_dataframe_item_length_error(self):
                               positional_metadata=pos_metadata_dataframe)])
         with self.assertRaisesRegex(StockholmFormatError,
                                     r'Sequence-specific positional metadata.*'
-                                    'must contain a single character.*Found '
-                                    'value\(s\) in column AC'):
+                                    r'must contain a single character.*Found '
```
antgonza: It seems like the error messages are missing a space. For example:
- here: `character.*Found`
- the test below `character.*Found value`
is that right? If it is, can we fix?

mortonjt: @antgonza note that this is a regex check -- that * is ignoring characters between the `.` and `Found`

antgonza: duh! thanks!

Issue #1654, opened 2019-03-22
Different lengths for aligned query/target sequence (StripedSmithWaterman)


diegozea:
I've found the following problem/error in scikit-bio 0.5.5 (Python 3.6.7): `aligned_query_sequence` and `aligned_target_sequence` has different lengths for some alignments (using StripedSmithWaterman). For example:

#### code 

```python
from skbio.alignment import StripedSmithWaterman                                                                                                                                                                                      
from skbio.alignment._pairwise import blosum50                                                                                                                                                                                        
query = StripedSmithWaterman(
'SFLSFCCVMANIDLSSIFLSCLILDIDQWNKVIEQLGTPCPEFMKKLQPTVRTYVENRPKYAGYSFEKLFPDVLFPADSEHNKLKA',
gap_open_penalty=10,
gap_extend_penalty=1,
substitution_matrix=blosum50)
aln = query('CICFLAASALHYDFSSFYFGLIKTSLQNVIHRIQGSLSHLS')
assert len(aln.aligned_query_sequence) == len(aln.aligned_target_sequence)
```

#### code + output
```python
In [1]: from skbio.alignment import StripedSmithWaterman                                                                                                                                                                                      

In [2]: from skbio.alignment._pairwise import blosum50                                                                                                                                                                                        

In [3]: query = StripedSmithWaterman( 
   ...: 'SFLSFCCVMANIDLSSIFLSCLILDIDQWNKVIEQLGTPCPEFMKKLQPTVRTYVENRPKYAGYSFEKLFPDVLFPADSEHNKLKA', 
   ...: gap_open_penalty=10, 
   ...: gap_extend_penalty=1, 
   ...: substitution_matrix=blosum50)                                                                                                                                                                                                         

In [4]: aln = query('CICFLAASALHYDFSSFYFGLIKTSLQNVIHRIQGSLSHLS')                                                                                                                                                                              

In [5]: aln                                                                                                                                                                                                                                   
Out[5]: 
{
    'optimal_alignment_score': 221,
    'suboptimal_alignment_score': 0,
    'query_begin': 5,
    'query_end': 48,
    'target_begin': 0,
    'target_end_optimal': 40,
    'target_end_suboptimal': 0,
    'cigar': '31M4I9M',
[...]

PR #1660, opened 2019-05-30
provisional bugfix for local_pairwise_align_ssw()

KatyBrown:
Hi,
I think this addresses a small bug in _ssw_wrapper.pyx - in some cases (about 1% in my experience)  running skbio.alignment.local_pairwise_align_ssw() on protein sequences gives the following error:

> Each sequence's length must match the number of positions in the MSA.

I've attached two sequences which will generate this error when the BLOSUM matrix is used (only when s1 is passed as the first argument and s2 as the second).
[example.txt](https://github.com/biocore/scikit-bio/files/3238610/example.txt)

This is also the error described in issue #1654 

As far as I can tell, a couple of extra bases are sometimes present at the beginning of one sequence in the alignment after running the StripedSmithWaterman algorithm, meaning they are not the same length and the initialisation of the TabularMSA object fails.  This fix is a workaround for this, however I wasn't completely clear as to where the error arose so I'm not 100% sure it will work in every scenario

.


ebolyen:
Thanks @KatyBrown! I'll try to take a look at this before the weekend. Please ping me if I forget!

ebolyen:
Looking at the unit-tests, it seems this is breaking another example of cigar parsing (which looks correct to me, but I may be mistaked). This makes me suspect that SSW may be the ultimate culprit here. It does look like both the sequences in your provided example, and in the linked issue, are very different from each other, so the alignment is mostly made of mismatches. It may be that we are running into something strange in SSW. It's probably due for a review as the SSW vendored in skbio is really quite old, so this may have been fixed upstream.

This is also probably one of the most fragile bits of skbio, since you are clearly a user of this API, are there better tools we could use instead?

wasade:
@ebolyen do you want to see this to merge or close it?

ebolyen:
Closing due to #1814

PR #1703, opened 2020-05-14
WIP: Networkx convertors

mortonjt:

This addresses https://github.com/biocore/scikit-bio/issues/809
Here we have both `to_networkx` and `from_networkx` conversion methods implemented, along with bare-bones unittests.

There are a few things that would be need to resolved before merging, namely

- How to handle networkx as an optional dependency
- Should we allow for lengths + other attributes be exchanged between nx.Graph and skbio.TreeNode? For strings / scalars this is totally doable, but could get complicated if we allow lists + dicts to be passed as attributes
- [ ] Missing labels. In theory this `to_networkx` should work with missing node names, but we need additional tests to sanity check this
- [ ] Root node specification.  As the code currently stands, `from_networkx` requires the specification of a root node. If this isn't specified, it just chooses the first node as the root node.

If there are any other outstanding issues, feel free to pile them on the above todo list :)

### DIFF
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -3269,3 +3269,44 @@ class TreeNode(SkbioObject):
                 nodes_to_unpack.append(node)
         for node in nodes_to_unpack:
             node.unpack()
+
+
+    @experimental(as_of="0.5.5-dev")
+    def to_networkx(self):
+        import networkx as nx
+
+        edges = []
+        def get_name(x, i):
+            return x.name if hasattr(x, 'name') else f"unnamed_{i}"
+
+        for i, n in enumerate(self.postorder(include_self=False)):
+            name = get_name(n, i)
+            parent_name = get_name(n.parent, i)
+
+            # TODO add length
+            edges.append((name, parent_name))
+
+        G = nx.Graph()
+        G.add_edges_from(edges)
+        return G
+
+    @classonlymethod
+    @experimental(as_of="0.5.5-dev")
+    def from_networkx(self, G, root=None):
+        import networkx as nx
+        nodes = list(G.nodes())
+        if root is None:
+            root = nodes[0]
+        dfs = nx.dfs_tree(G, root)
+        if len(dfs.edges()) != len(G.edges()):
+            raise ValueError('Graph `G` cannot be represented as a tree.')
+        edges = list(dfs.edges())
+        # each edge element is of the form are (curr, parent)
+        node_dict = {n: TreeNode(n) for n in nodes}
+        node_dict[root].parent = None
+        # perform depth search traversal
+        for e in edges:
+            p, c = e
+            node_dict[p].append(node_dict[c])
+        return node_dict[root]
+
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -3270,15 +3270,14 @@ class TreeNode(SkbioObject):
         for node in nodes_to_unpack:
             node.unpack()
 
-
     @experimental(as_of="0.5.5-dev")
     def to_networkx(self):
         import networkx as nx
 
-        edges = []
         def get_name(x, i):
             return x.name if hasattr(x, 'name') else f"unnamed_{i}"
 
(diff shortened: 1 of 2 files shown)


skbio/tree/_tree.py
```diff
@@ -3269,3 +3269,42 @@ def unpack_by_func(self, func):
                 nodes_to_unpack.append(node)
         for node in nodes_to_unpack:
             node.unpack()
+
+    @experimental(as_of="0.5.5-dev")
+    def to_networkx(self):
+        import networkx as nx
```
wasade, line 3275: Should be up top :)

skbio/tree/_tree.py
```diff
@@ -3269,3 +3269,42 @@ def unpack_by_func(self, func):
...
+
+        def get_name(x, i):
+            return x.name if hasattr(x, 'name') else f"unnamed_{i}"
```
wasade, line 3278: `.name` is on all `TreeNode` objects

skbio/tree/_tree.py
mortonjt: Agreed. The only thing is that if we want optional imports, we'd need to have a try/except at the top - which means that flake8 need to modified to ignore this error.

wasade: I don't think we want to introduce optional imports unless there is already precedence within scikit-bio

wasade: ...rather, I think this would motivate including it as a dependency

mortonjt: The main reason why I brought this up was previous discussions regarding having matplotlib as an optional dependency, since it having more required dependencies can lead to more maintenance burden.  I've seen other packages use optional dependencies to help mitigate these issues (i.e. pandas with the xlrd dependency for reading excel notebook).

That being said, I don't have super strong feelings about this, so I'll include networkx as a required dependency until someone objects.

wasade: @ebolyen @ElDeveloper @thermokarst any comment on preference for networkx being an optional or required dependency? It looks like it's already installed with QIIME 2 2020.2, so I don't think it would increase burden more than any other dependency 

@mortonjt I don't think the dependency question should hold up the rest of the implementation IMO.

skbio/tree/_tree.py
mortonjt: I've addressed this.

mortonjt:
I've addressed the outstanding concerns. Code for lengths and supports have been added and everything should be fully covered by unittests.

I'm still a bit confused about the formatting of supports - I don't see any examples where nodes have names, supports, and length values, so its currently unclear on how to construct unittests. Not sure how critical this is, @qiyunzhu feel free to comment.

Otherwise, this is ready for a solid review.

qiyunzhu:
@mortonjt The `support` attribute of a `TreeNode` object is documented [here](http://scikit-bio.org/docs/latest/generated/skbio.tree.TreeNode.assign_supports.html#skbio.tree.TreeNode.assign_supports). To generate a tree with names, lengths and supports for unit-test, you can do something like this:
```
nwk = "((a:1.2,b:1.4)'95:i':2.3,(c:0.8,d:1.1)'97:j':1.6)r;"
tree = TreeNode.read([nwk])
tree.assign_supports()
```
Then, internal nodes will have both `name` and `support` attributes. For example, `tree.children[0]` will have `name == 'i'` and `support == 95`, meanwhile `length == 2.3`.

mortonjt:
Thank you!

wasade:
@mortonjt is this still a WIP?

mortonjt:
I don’t have time to fix this. So I’d close

wasade:
Okay thanks. @qiyunzhu, are you interested and have bandwidth to finish this off?

wasade:
Closing for now as it's unclear if this will be finished off. Please reopen if this is incorrect

qiyunzhu:
Thanks for pinging me @wasade but I am not clear about the mission of this PR so perhaps can't easily take over...

PR #1717, opened 2020-08-05
Allow `DissimilarityMatrix` to use `np.float32`

wasade:
We observed a Mantel R^2 of 0.9999 between float32 and float64 results with weighted UniFrac when applied to the EMP dataset (~27k samples). This observation arose from a recent GPU adaption of [UniFrac](https://arxiv.org/abs/2005.05826) when testing for a performance difference when computing float32 and float64. We're now producing float32 matrices from UniFrac as it's appreciably faster, however the `DissimilarityMatrix` object at the moment requires double precision. This PR allows `DissimilarityMatrix` to both support float32, and to avoid an implicit copy that would occur if passing in float32 data. 

The implicit copy is important to avoid for large matrices to keep resident memory low.  
 
### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ install:
   - conda info --envs
   - python --version
   - pip install -r ci/pip_requirements.txt
-  - pip install .
+  - pip install . --no-deps
 
 script:
   - WITH_COVERAGE=TRUE make test
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,27 @@
 # scikit-bio changelog
 
+## Version 0.5.6-dev
+
+### Features
+
+* Added support for `np.float32` with `DissimilarityMatrix` objects.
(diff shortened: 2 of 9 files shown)


wasade:
So I think this is good too, cc @gwarmstrong @ElDeveloper

.travis.yml
```diff
@@ -28,7 +28,7 @@ install:
   - conda info --envs
   - python --version
   - pip install -r ci/pip_requirements.txt
-  - pip install .
+  - pip install . --no-deps
```
ElDeveloper, line 31: What's the reason for this change?

ElDeveloper (approved):

wasade:
Without it, we risk having the dependency files falling out of sync with
whats actually needed

ElDeveloper:
Looks good to me!

wasade:
Thanks!!

Issue #1685, opened 2019-12-06
Buffer dtype mismatch in weighted_unifrac


AaronBlare:
I try to run the example of weighted_unifrac from the docs: http://scikit-bio.org/docs/0.5.5/generated/skbio.diversity.beta.weighted_unifrac.html

```
from io import StringIO
from skbio.tree import TreeNode
from skbio.diversity.beta import unweighted_unifrac

tree = TreeNode.read(StringIO(
    '(((((OTU1:0.5,OTU2:0.5):0.5,OTU3:1.0):1.0):0.0,'
    '(OTU4:0.75,(OTU5:0.5,((OTU6:0.33,OTU7:0.62):0.5'
    ',OTU8:0.5):0.5):0.5):1.25):0.0)root;'))

u_counts = [1, 0, 0, 4, 1, 2, 3, 0]
v_counts = [0, 1, 1, 6, 0, 1, 0, 0]
otu_ids = ['OTU1', 'OTU2', 'OTU3', 'OTU4', 'OTU5', 'OTU6', 'OTU7', 'OTU8']

uu = unweighted_unifrac(u_counts, v_counts, otu_ids, tree)
```

I obtain the following error:

```
Traceback (most recent call last):
  File "D:/test.py", line 14, in <module>
    wu = weighted_unifrac(u_counts, v_counts, otu_ids, tree)
  File "C:\ProgramData\Anaconda3\envs\mtDNA\lib\site-packages\skbio\diversity\beta\_unifrac.py", line 281, in weighted_unifrac
    normalized=normalized, unweighted=False)
  File "C:\ProgramData\Anaconda3\envs\mtDNA\lib\site-packages\skbio\diversity\beta\_unifrac.py", line 315, in _setup_pairwise_unifrac
    _vectorize_counts_and_tree(counts, otu_ids, tree)
  File "C:\ProgramData\Anaconda3\envs\mtDNA\lib\site-packages\skbio\diversity\_util.py", line 113, in _vectorize_counts_and_tree
    counts_by_node = _nodes_by_counts(counts, otu_ids, tree_index)
  File "skbio/diversity/_phylogenetic.pyx", line 213, in skbio.diversity._phylogenetic._nodes_by_counts (skbio/diversity/_phylogenetic.c:2641)
  File "skbio/diversity/_phylogenetic.pyx", line 68, in skbio.diversity._phylogenetic._traverse_reduce (skbio/diversity/_phylogenetic.c:1832)
[...]

PR #1729, opened 2020-10-19
BUG: Fix windows and 32bit support of `unweighted_unifrac`

seberg:
This forces the tree `child_index` to be exported as 64bit and
also adds a cast (just in case) into `_nodes_by_counts` in the
`_phylogenetic.pyx` cython code.

I have tested that `unweighted_unifrac` succeeds with the change
to the `child_index`. I have not done a full test run, so while
this closes gh-1685, it is very likely that the similar issues
exist elsewhere.

Ideally, it would be great if you find the time to add an Azure job
for windows, or a 32bit job (which will at least find some of these
issues as well).

On a related note: Since this is an `index`, the correct dtype is
likely `np.intp`, which is 32bit on 32bit systems and 64bit on
64 bit systems. So it might be better if the cython `DTYPE_t` was
changed that (and the `child_index` code here as well), but
of course that relies on `DTYPE_t` not being used for anything
that is not an index operation but should have 64bit accuracy.


wasade:
Thanks @seberg! The build error looks to be associated with an out-of-sync .c [file](https://travis-ci.com/github/biocore/scikit-bio/jobs/402190291#L1301) -- would it be possible to regenerate and add to the PR? Second, could you add a brief note to the ChangeLog?

seberg:
Oh, didn't realize you add simple bug fixes to the change log, added.  Regenerated the cython file, I assume there isn't anything to take care of specifically).  Btw. in NumPy we stopped generating the cython file and checking it in quite long ago, we assume those who build from source can be expected to have cython.  But, I admit that we ship wheels, so few users have to build from source, so I guess you probably have good reasons for shipping the generated version.

wasade:
Thanks, @seberg! It depends on the issue for adding to the changelog, but seeing as there have been a few comments on this, I thought it may be nice to easily summarize for users. W.r.t. regenerating the cython file, I agree with the practice you note. The approach were taking is largely for historical (and possibly outdated) reasons, and path of least resistance at the moment is to stay the course. I'll open an issue about it though as it may be a small hanging issue to knock out for the next release.

Issue #1751, opened 2021-01-26
mantel filters ids even when there is no need


sfiligoi:
The mantel function is performing filtering even when the ids are not changing.
This is expensive, especially for large matrices, and should be avoided.
Note that passing two matrices with identical ids is very common, so this code path should be optimized.

PR #1752, opened 2021-01-26
Avoid actual filtering, if ids have not changed. #1751

sfiligoi:
Check if ids have been changed before filtering, avoiding expensive memory operations.
In the case of mantel, identical ids are the norm, so optimization of this scenario is warranted. #1751


sfiligoi:
CC @wasade

skbio/stats/distance/_base.py
```diff
@@ -369,6 +369,9 @@ def filter(self, ids, strict=True):
         MissingIDError
             If an ID in `ids` is not in the object's list of IDs.
         """
+        if tuple(self._ids) == tuple(ids):
```
wasade, line 372: This would allow for permutations which are otherwise equivalent to be put through expensive operations. For example:

```
  A B C
A 0 1 2
B 1 0 3
C 2 3 0

...is equivalent too:

  A C B
A 0 2 1
C 2 0 3
B 1 3 0
```

The use of `.filter` here is in effect a reorder operation, not a filter. However, changing it so that the conditional is `set(self._ids) == set(ids)` would violate the return assumptions of order w.r.t. `ids` with implications to consumers. 

Would it make sense to add in a new kwarg to resolve whether order matters here?

sfiligoi: It is certainly possible to make it smarter.
However, checking for identity is very cheap and this patch tries to optimize for that only.
Additional tests would either be more expensive or potentially dangerous.

Note: tuple(['A','B','C'])!=tuple(['A','C','B']), so the current patch is semantically correct.
(even if I agree it could potentially be further improved)

wasade:
Looks like there is a conflict and possible real failure on py3.6

wasade:
Looks like a possible real failure on py36

sfiligoi:
> Looks like a possible real failure on py36

@wasade All tests are passing
Did I miss something?

wasade:
looks like its passing now, but there was a prior failure (see https://github.com/biocore/scikit-bio/runs/1841515590#step:6:183) which appears to be the stochastic one on small floats in the treenode doctest. I'll clean up the conflict here and merge in a mmoment

PR #1753, opened 2021-01-29
migrate testing from TravisCI to GitHub Actions

gregcaporaso:
This pull request migrates scikit-bio testing from TravisCI to GitHub Actions. This was pair-programmed with @thermokarst. 

At the moment the Python 3.7 and 3.8 builds are not passing - it looks like there might be some unit test issues that need to be addressed for those Python versions. @wasade, it might make sense to merge this so that the tests are run again, and then the test failures can be addressed in other PRs. I'm happy with however you want to handle that. 

I'm also not certain that Coveralls is working correctly in here. `coverage.py` is running, and the output is presented in the console output, but the `coveralls_finish` job won't run until all `install-and-test` steps complete successfully (so Python 3.7 and 3.8 tests need to pass first). If Coveralls still isn't working after the test failures are fixed, I'm happy to jump back in to try to help.

There are also a couple of differences from what we were running in TravisCI. 
1) I dropped the Gitter stuff (I don't think we're using that anymore) 
2) Before `USE_CYTHON` was only set to `TRUE` in a Python 3.6 run. I now have this as an axis in the build matrix, so that each combination of `os` and `python-version` is tested with both settings of `USE_CYTHON`. 

<hr>


wasade:
I think this makes sense to merge so we can start to close out other PRs w.r.t. py36. Should we comment out py37/38 support right now, and create a separate issue noting that master is at the moment not compatible with 37/38?

sfiligoi:
Permdisp test errors are due to tests that are too tight;
a minimal rounding FP error can move the p-value by more than 1%, since it is internally based on discrete counts.
The FP optimizations between python versions can easily explain the difference in the result (one out of 99 >= comparisons changed).

wasade:
Pairing with @gregcaporaso. We're merging now to restore CI for py36. The failures on py37/38 appear at the moment minor, but nonetheless important to resolve which will occur in a follow up issue.

sfiligoi:
Note: The build uses the system provided clang, instead of the conda provided one. That version has no support for openmp, making it impossible to write optimized code.
@wasade @gregcaporaso Can that be changed?

wasade:
:+1:, issue created, thanks @sfiligoi!

Issue #1779, opened 2021-08-03
Adding TreeNode methods for Evolutionary Distinctiveness measure


FranckLejzerowicz:
Hi,

Please consider this new issue for the PR of a feature addition: a new TreeNode method allowing the computing of the Evolutionary Distinctiveness metric (ED).

This time, the code was re done without looking at the R code (thanks @wasade!) and it not only returns ED for each tip of a tree, but also for each internal node. 

ED is important for conservation and could be used as feature metadata and/or to identify redundant groups of features in a densely populated tree. Many other usage are certainly possible and not envisioned here. Please see these few references on ED:

* Metric comparison to other phylogenetic metrics [A guide to phylogenetic metrics for conservation, community ecology and macroecology](https://onlinelibrary.wiley.com/doi/10.1111/brv.12252)
* Concept on conservation [It Pays to Be Different: Evolutionary Distinctiveness and Conservation Priorities](https://blogs.ucl.ac.uk/gee-research/2014/07/15/it-pays-to-be-differentevolutionary-distinctiveness-and-conservation-priorities/)
* Use by IUCN for conservation (11 May 2020) [Species IUCN threat status level increases with elevation: a phylogenetic approach for Neotropical tree frog conservation](https://link.springer.com/article/10.1007/s10531-020-01986-8)

I have made unittests and compared the results to the [fastest R implementation](https://cran.r-project.org/web/packages/phyloregion/vignettes/ed.html#:~:text=Evolutionary%20distinctiveness%20is%20a%20metric,R%20packages%20picante%20or%20caper%20.). As far as I can tell, there is no python solution to compute this metric, and a user has [requested it from the DendroPy forum](https://groups.google.com/g/dendropy-users/c/hkCQAjjmIG4).

Just like the R function used for testing, this PR for a python-based ED method, allows the users to chose whether to use branch length (or not: only the furcations and node depth), as well as to normalize the tree branch lengths (or not) before computing ED. To do so, two accessory methods are also included that are also not derived from external source (`is_ultrametric` and `scale_length`).

In advance, please consider these items and feel free to express interest for me to proceed with the PR following the guidelines.

* [x] I have read the guidelines in [CONTRIBUTING.md](https://github.com/biocore/scikit-bio/blob/master/CONTRIBUTING.md).

* [x] I have documented all public-facing changes in [CHANGELOG.md](https://github.com/biocore/scikit-bio/blob/master/CHANGELOG.md).

[...]

PR #1780, opened 2021-08-03
Issue 1779

FranckLejzerowicz:
Hi,

This PR is for the [issue #1779 ](https://github.com/biocore/scikit-bio/issues/1779).

It consists of allowing the calculation of Evolutionary Distinctiveness (ED), which is calculated for each node of a TreeNode object.

Here's its components, as some new methods were necessary:
    * main method `evol_distinct`
    * private methods `_calc_fair_proportion` and `_calc_equal_splits` to calculate ED under two different node furcation modes (thanks again @wasade for the more efficient proposal!)
    * method `name_unnamed_internal_nodes` to name unnamed nodes (necessary for ED calculation)
    * method `is_ultrametric` to check is tree is ultrametric or not
    * method `scale_length` to normalize branch length
    * append this added feature description to CHANGELOG.md

Note that all unittests pass, as well as all the tests of `make test` that are associated with this PR. Yet, some tests of `make test` fail, but these failures are associated with others part of skbio which are not relevant to this PR.

Please see the extra info about these checked list in the issue #1779  

Thanks,
Franck


skbio/tree/_tree.py
```diff
@@ -3269,3 +3269,295 @@ def unpack_by_func(self, func):
...
             node.unpack()
+
+    @experimental(as_of="0.5.3")
+    def is_ultrametric(self):
+        """Returns whether the tree is ultrametric or not.
+
+        Returns
+        -------
+        bool
+            True is all root-to-tip distance are equal, or False if at least
```
sjanssen2, line 3280: is -> if?

skbio/tree/_tree.py
```diff
@@ -3269,3 +3269,295 @@ def unpack_by_func(self, func):
...
+            one single distance differs
+
+        Notes
+        -----
+        This function collects all the tip to root branch lengths in a set.
+        If this set has a length of 1, then the tree is ultrametric.
```
sjanssen2, line 3286: does this mechanism mind float rounding issues, i.e. 3.999999999 == 4.000000000 ?!

skbio/tree/_tree.py
```diff
@@ -3269,3 +3269,295 @@ def unpack_by_func(self, func):
...
+        >>> tree.is_ultrametric()
+        True
+        >>> tip_root = tree.root()
+        >>> print([tip_root.distance(tree.find(tip)) for tip in tree.tips()])
+        [10.0, 10.0, 10.0, 10.0]
+        """
+        tip_root = self.root()
+        metrics = set()
+        for tip in self.tips():
+            root_to_tip_dist = tip_root.distance(self.find(tip))
```
sjanssen2, line 3321: why are you using `self.find(tip)` instead of directly `tip`? What if there are multiple tips with the same name? Not sure what `find` returns in these ambiguous situations but you probably don't want to rely on tip names (they might be None as well).

skbio/tree/_tree.py
```diff
@@ -3269,3 +3269,295 @@ def unpack_by_func(self, func):
...
+        >>> # example for an ultrametric tree
+        >>> tree = TreeNode.read(['((a:3,b:3)e:7,(c:2,d:2)f:8);'])
+        >>> tree.is_ultrametric()
+        True
+        >>> tip_root = tree.root()
+        >>> print([tip_root.distance(tree.find(tip)) for tip in tree.tips()])
+        [10.0, 10.0, 10.0, 10.0]
+        """
+        tip_root = self.root()
+        metrics = set()
```
sjanssen2, line 3319: a more speaking name might be `path_lengths`?

skbio/tree/_tree.py
```diff
@@ -3269,3 +3269,295 @@ def unpack_by_func(self, func):
...
+        for tip in self.tips():
+            root_to_tip_dist = tip_root.distance(self.find(tip))
+            metrics.add(root_to_tip_dist)
+        if len(metrics) > 1:
```
sjanssen2, line 3323: maybe more stable `if math.isclose(metrics.min(), metrics.max(), abs_tol=0.00003):`, see https://docs.python.org/3/whatsnew/3.5.html#pep-485-a-function-for-testing-approximate-equality

skbio/tree/_tree.py
```diff
@@ -3269,3 +3269,295 @@ def unpack_by_func(self, func):
...
+            return False
+        return True
+
+    @experimental(as_of="0.5.3")
+    def scale_length(self):
+        """Returns the tree where the branch lengths are
+        scaled, i.e. divided by (i) the root-to-tip distance
```
sjanssen2, line 3330: `scaled`: how? Will values be between 0 and 1?

skbio/tree/_tree.py
```diff
@@ -3269,3 +3269,295 @@ def unpack_by_func(self, func):
...
+        for ultrametric trees and (ii) the total branch
+        length (method `descending_branch_length`) for
+        non-ultrametric trees.
+
+        Returns
+        -------
+        TreeNode
+            The length-scaled tree
```
sjanssen2, line 3338: is that a copy of the original object OR is the input tree modified?

skbio/tree/_tree.py
```diff
@@ -3269,3 +3269,295 @@ def unpack_by_func(self, func):
...
+        >>> [(n.length, t2.find(n.name).length) for n in t.preorder(False)]
+        [(7.0, 0.7), (3.0, 0.3), (3.0, 0.3), (10.0, 1.0)]
+        """
+        tcopy = self.deepcopy()
+        if tcopy.is_ultrametric():
+            # if tree is ultrametric, use longest edge
+            tip_root = tcopy.root()
+            for tip in tcopy.tips():
+                ref_branch_length = tip_root.distance(tcopy.find(tip))
+                break
```
sjanssen2, line 3372: I don't understand this for loop?! Why do you break, after the first iteration? See above about use of `.find`. Are you looking for longest root tip distance?

skbio/tree/_tree.py
```diff
@@ -3269,3 +3269,295 @@ def unpack_by_func(self, func):
...
+        else:
+            # if tree is not ultrametric, use total branch length
+            ref_branch_length = tcopy.descending_branch_length()
```
sjanssen2, line 3375: this gives you ALL the length in the tree. Imagine a tree with 9 tips, their distance from root to them shall be 1. Then, descending_branch_length returns 9. But wouldn't you want it to be 1? I.E. I think you need to "normalize" by number of tips.

skbio/tree/_tree.py
```diff
@@ -3269,3 +3269,295 @@ def unpack_by_func(self, func):
...
+
+    @experimental(as_of="0.5.3")
+    def name_unnamed_internal_nodes(self):
+        """Name the internal nodes that do not have a name.
+
+        Notes
+        -----
+        First, the existing node names are collected in a set, and then the
+        tree is traversed in preorder so that the unnamed nodes are given an
+        incremental `N#` name.
```
sjanssen2, line 3391: do you ensure that there is no given `N#` name in the tree?

sjanssen2:
Hi @FranckLejzerowicz I am curious about the use cases for this distance. I tried to get that from the code and left some comments. I am not trough yet, but get distracted by teaching obligations. Hope to find time in some days.

wasade:
I'm closing as it looks like there are a few important outstanding and unresolved issues here, please reopen if this is planned to be finished

PR #1782, opened 2021-09-03
Adding in correlation metrics (`vlr` and `pairwise_vlr` to `skbio.stats.compositional`)

jolespin:

* [x] **This pull request includes code, documentation, or other content derived from external source(s).** If this is the case, ensure the external source's license is compatible with scikit-bio's [license](https://github.com/biocore/scikit-bio/blob/master/COPYING.txt). Include the license in the `licenses` directory and add a comment in the code giving proper attribution. Ensure any other requirements set forth by the license and/or author are satisfied. **It is your responsibility to disclose code, documentation, or other content derived from external source(s).** If you have questions about whether something can be included in the project or how to give proper attribution, include those questions in your pull request and a reviewer will assist you.

### DIFF
--- a/skbio/stats/composition.py
+++ b/skbio/stats/composition.py
@@ -106,7 +106,7 @@ import pandas as pd
 import scipy.stats
 import skbio.util
 from skbio.util._decorator import experimental
-from skbio.stats.distance import DistanceMatrix
+from skbio.stats.distance import DistanceMatrix, DissimiliarityMatrix
 
 
 @experimental(as_of="0.4.0")
@@ -807,118 +807,120 @@ def centralize(mat):
     return perturb_inv(mat, cen)
 
 
-
 @experimental(as_of="0.5.7")
-def _vlr(x:np.array, y:np.array, ddof:int):
+def _vlr(x: np.array, y: np.array, ddof: int):
     r"""
     Calculates variance log ratio
-    
+
     Parameters
     ----------
     x : array_like, float
        a 1-dimensional vector of proportions
     y : array_like, float
        a 1-dimensional vector of proportions
-        
+
     ddof: int
         degrees of freedom
-        
-       
+
     Returns
     -------
     float
          variance log ratio value
-         
-    
+
     References
     ----------
-    .. [1] V. Lovell D, Pawlowsky-Glahn V, Egozcue JJ, Marguerat S, Bähler J (2015) 
-           Proportionality: A Valid Alternative to Correlation for Relative Data. 
-           PLoS Comput Biol 11(3): e1004075. https://doi.org/10.1371/journal.pcbi.1004075
-    .. [2] Erb, I., Notredame, C. 
-           How should we measure proportionality on relative gene expression data?. 
-           Theory Biosci. 135, 21–36 (2016). https://doi.org/10.1007/s12064-015-0220-8
+    .. [1] V. Lovell D, Pawlowsky-Glahn V, Egozcue JJ, Marguerat S,
+           Bähler J (2015) Proportionality: A Valid Alternative to
+           Correlation for Relative Data. PLoS Comput Biol 11(3): e1004075.
+           https://doi.org/10.1371/journal.pcbi.1004075
+    .. [2] Erb, I., Notredame, C.
+           How should we measure proportionality on relative gene
+           expression data?. Theory Biosci. 135, 21–36 (2016).
+           https://doi.org/10.1007/s12064-015-0220-8
     """
 
     # Log transformation
     x = np.log(x)
     y = np.log(y)
-    
-    #Variance log ratio
+
+    # Variance log ratio
     return np.var(x - y, ddof=ddof)
 
 
 @experimental(as_of="0.5.7")
-def _robust_vlr(x:np.array, y:np.array, ddof:int):
+def _robust_vlr(x: np.ndarray, y: np.ndarray, ddof: int):
     r"""
     Calculates variance log ratio while masking zeros
-    
+
     Parameters
     ----------
     x : array_like, float
        a 1-dimensional vector of proportions
     y : array_like, float
        a 1-dimensional vector of proportions
-        
+
     ddof: int
         degrees of freedom
-        
+
     Returns
     -------
     float
          variance log ratio value
-         
-    
+
+
     References
     ----------
-    .. [1] V. Lovell D, Pawlowsky-Glahn V, Egozcue JJ, Marguerat S, Bähler J (2015) 
-           Proportionality: A Valid Alternative to Correlation for Relative Data. 
-           PLoS Comput Biol 11(3): e1004075. https://doi.org/10.1371/journal.pcbi.1004075
-    .. [2] Erb, I., Notredame, C. 
-           How should we measure proportionality on relative gene expression data?. 
-           Theory Biosci. 135, 21–36 (2016). https://doi.org/10.1007/s12064-015-0220-8
+    .. [1] V. Lovell D, Pawlowsky-Glahn V, Egozcue JJ, Marguerat S,
+           Bähler J (2015) Proportionality: A Valid Alternative to
+           Correlation for Relative Data. PLoS Comput Biol 11(3): e1004075.
+           https://doi.org/10.1371/journal.pcbi.1004075
+    .. [2] Erb, I., Notredame, C.
+           How should we measure proportionality on relative gene
+           expression data?. Theory Biosci. 135, 21–36 (2016).
+           https://doi.org/10.1007/s12064-015-0220-8
     """
 
     # Mask zeros
     x = np.ma.masked_array(x, mask=x == 0)
     y = np.ma.masked_array(y, mask=y == 0)
-    
+
     # Log transformation
     x = np.ma.log(x)
     y = np.ma.log(y)
-    
-    #Variance log ratio
+
+    # Variance log ratio
     return np.ma.var(x - y, ddof=ddof)
 
+
 @experimental(as_of="0.5.7")
-def vlr(x:np.array, y:np.array, ddof:int=1, robust:bool=False):
+def vlr(x: np.ndarray, y: np.ndarray, ddof: int = 1, robust: bool = False):
     r"""
     Calculates variance log ratio
-    
+
     Parameters
     ----------
     x : array_like, float
        a 1-dimensional vector of proportions
     y : array_like, float
        a 1-dimensional vector of proportions
-        
+
     ddof: int
         degrees of freedom
-        
+
     robust: bool
         mask zeros at the cost of performance
-       
+
     Returns
     -------
     float
          variance log ratio value
-         
+
     Examples
     --------
     No zeros
     >>> x = [1,2,3]
     >>> y = [5,8,13]
-    >>> %timeit vlr(x,y)
+    >>> vlr(x,y)
     >>> 0.01277962183258352
 
     Zeros without robust
@@ -932,148 +934,163 @@ def vlr(x:np.array, y:np.array, ddof:int=1, robust:bool=False):
     >>> y = [5,8,13,21]
     >>> vlr(x,y, robust=True)
     >>> 0.01277962183258352
-    
+
     References
     ----------
-    .. [1] V. Lovell D, Pawlowsky-Glahn V, Egozcue JJ, Marguerat S, Bähler J (2015) 
-           Proportionality: A Valid Alternative to Correlation for Relative Data. 
-           PLoS Comput Biol 11(3): e1004075. https://doi.org/10.1371/journal.pcbi.1004075
-    .. [2] Erb, I., Notredame, C. 
-           How should we measure proportionality on relative gene expression data?. 
-           Theory Biosci. 135, 21–36 (2016). https://doi.org/10.1007/s12064-015-0220-8
+    .. [1] V. Lovell D, Pawlowsky-Glahn V, Egozcue JJ, Marguerat S,
+           Bähler J (2015) Proportionality: A Valid Alternative to
+           Correlation for Relative Data. PLoS Comput Biol 11(3): e1004075.
+           https://doi.org/10.1371/journal.pcbi.1004075
+    .. [2] Erb, I., Notredame, C.
+           How should we measure proportionality on relative gene
+           expression data?. Theory Biosci. 135, 21–36 (2016).
+           https://doi.org/10.1007/s12064-015-0220-8
     """
     # Convert array_like to numpy array
     x = np.asarray(x)
     y = np.asarray(y)
 
     # Set up input and parameters
-    kwargs = { 
-        "x":x,
-        "y":y,
-        "ddof":ddof,
+    kwargs = {
+        "x": x,
+        "y": y,
+        "ddof": ddof,
     }
 
-    # Run backend function 
+    # Run backend function
     if robust:
-        return _robust_vlr(**kwargs) 
+        return _robust_vlr(**kwargs)
     else:
         return _vlr(**kwargs)
-    
+
 
 @experimental(as_of="0.5.7")
-def _pairwise_vlr(mat:np.array, ddof:int):
+def _pairwise_vlr(mat: np.ndarray, ddof: int):
     r"""
     Performs pairwise variance log ratio transformation
-    
+
     Parameters
     ----------
     mat : array_like, float
        a matrix of proportions where
        rows = compositions and
        columns = components
-       
+
     ids: array_like, str
         component names
-        
+
     ddof: int
         degrees of freedom
-       
+
     Returns
     -------
     skbio.DistanceMatrix
          distance matrix of variance log ratio values
-         
-    
+
+
     References
     ----------
-    .. [1] V. Lovell D, Pawlowsky-Glahn V, Egozcue JJ, Marguerat S, Bähler J (2015) 
-           Proportionality: A Valid Alternative to Correlation for Relative Data. 
-           PLoS Comput Biol 11(3): e1004075. https://doi.org/10.1371/journal.pcbi.1004075
-    .. [2] Erb, I., Notredame, C. 
-           How should we measure proportionality on relative gene expression data?. 
-           Theory Biosci. 135, 21–36 (2016). https://doi.org/10.1007/s12064-015-0220-8
+    .. [1] V. Lovell D, Pawlowsky-Glahn V, Egozcue JJ, Marguerat S,
+           Bähler J (2015) Proportionality: A Valid Alternative to
+           Correlation for Relative Data. PLoS Comput Biol 11(3): e1004075.
+           https://doi.org/10.1371/journal.pcbi.1004075
+    .. [2] Erb, I., Notredame, C.
+           How should we measure proportionality on relative gene
+           expression data?. Theory Biosci. 135, 21–36 (2016).
+           https://doi.org/10.1007/s12064-015-0220-8
     """
-        
+
     # Log Transform
     X_log = np.log(mat)
-    
+
     # Variance Log Ratio
-    covariance = np.cov(X_log.T, ddof=ddof) 
+    covariance = np.cov(X_log.T, ddof=ddof)
     diagonal = np.diagonal(covariance)
-    vlr_data = -2*covariance + diagonal[:,np.newaxis] + diagonal 
+    vlr_data = -2*covariance + diagonal[:, np.newaxis] + diagonal
     return vlr_data
 
+
 @experimental(as_of="0.5.7")
-def _robust_pairwise_vlr(mat:np.array, ddof:int):
+def _robust_pairwise_vlr(mat: np.ndarray, ddof: int):
     r"""
     Performs pairwise variance log ratio transformation while masking zeros
-    
+
     Parameters
     ----------
     mat : array_like, float
        a matrix of proportions where
        rows = compositions and
        columns = components
-       
+
     ids: array_like, str
         component names
-        
+
     ddof: int
         degrees of freedom
-       
+
     Returns
     -------
     skbio.DistanceMatrix
          distance matrix of variance log ratio values
-         
-    
+
+
     References
     ----------
-    .. [1] V. Lovell D, Pawlowsky-Glahn V, Egozcue JJ, Marguerat S, Bähler J (2015) 
-           Proportionality: A Valid Alternative to Correlation for Relative Data. 
-           PLoS Comput Biol 11(3): e1004075. https://doi.org/10.1371/journal.pcbi.1004075
-    .. [2] Erb, I., Notredame, C. 
-           How should we measure proportionality on relative gene expression data?. 
-           Theory Biosci. 135, 21–36 (2016). https://doi.org/10.1007/s12064-015-0220-8
+    .. [1] V. Lovell D, Pawlowsky-Glahn V, Egozcue JJ, Marguerat S,
+           Bähler J (2015) Proportionality: A Valid Alternative to
+           Correlation for Relative Data. PLoS Comput Biol 11(3): e1004075.
+           https://doi.org/10.1371/journal.pcbi.1004075
+    .. [2] Erb, I., Notredame, C.
+           How should we measure proportionality on relative gene
+           expression data?. Theory Biosci. 135, 21–36 (2016).
+           https://doi.org/10.1007/s12064-015-0220-8
     """
     # Mask zeros
     X = np.ma.masked_array(mat, mask=mat == 0)
-    
+
     # Log Transform
     X_log = np.ma.log(X)
-    
+
     # Variance Log Ratio
-    covariance = np.ma.cov(X_log.T, ddof=ddof) 
+    covariance = np.ma.cov(X_log.T, ddof=ddof)
     diagonal = np.ma.diagonal(covariance)
-    vlr_data = -2*covariance + diagonal[:,np.newaxis] + diagonal 
+    vlr_data = -2 * covariance + diagonal[:, np.newaxis] + diagonal
     return vlr_data
 
+
 @experimental(as_of="0.5.7")
-def pairwise_vlr(mat, ids=None, ddof=1, robust=False) -> DistanceMatrix:
+def pairwise_vlr(mat,
+    ids=None,
+    ddof: int = 1,
+    robust: bool = False,
+    validate: bool = True):
+
     r"""
     Performs pairwise variance log ratio transformation
-    
+
     Parameters
     ----------
     mat : array_like, float
        a matrix of proportions where
        rows = compositions and
        columns = components
-       
+
     ids: array_like, str
         component names
-        
+
     ddof: int
         degrees of freedom
-        
+
     robust: bool
         mask zeros at the cost of performance
-       
+
     Returns
     -------
-    skbio.DistanceMatrix
+    skbio.DistanceMatrix if validate=True
          distance matrix of variance log ratio values
-         
+    skbio.DissimilarityMatrix if validate=False
+         dissimilarity matrix of variance log ratio values
+
     Examples
     --------
     >>> mat = np.asarray([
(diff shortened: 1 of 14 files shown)


skbio/stats/composition.py
```diff
@@ -806,6 +807,163 @@ def centralize(mat):
     cen = scipy.stats.gmean(mat, axis=0)
     return perturb_inv(mat, cen)
 
+@experimental(as_of="0.5.7")
+def vlr(x,y, ddof=1, robust=False):
```
mortonjt, line 811:
```suggestion
def vlr(x, y, ddof=1, robust=False):
```

skbio/stats/composition.py
```diff
@@ -806,6 +807,163 @@ def centralize(mat):
...
+    float
+         variance log ratio value
+         
+    Examples
+    --------
+    # No zeros
+    >>> x = [1,2,3]
+    >>> y = [5,8,13]
+    >>> %timeit vlr(x,y)
+    # 21.7 µs ± 979 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
mortonjt, line 839: Let's remove the `timeit` statements from the docs -- the user doesn't need to know this. But we can keep this sort information under the thread in the pull request to keep track of the runtime of the functions.

skbio/stats/composition.py
```diff
@@ -806,6 +807,163 @@ def centralize(mat):
...
+    # 0.01277962183258352
+
+    # Zeros without robust
+    >>> x = [1,2,3,0]
+    >>> y = [5,8,13,21]
+    >>> %timeit vlr(x,y)
+    # 27.5 µs ± 2.48 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```
mortonjt, line 846: same here

skbio/stats/composition.py
```diff
@@ -806,6 +807,163 @@ def centralize(mat):
...
+    # nan
+
+    # Zeros with robust
+    >>> x = [1,2,3,0]
+    >>> y = [5,8,13,21]
+    >>> %timeit vlr(x,y, robust=True)
+    # 570 µs ± 21 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```
mortonjt, line 853: same here

skbio/stats/composition.py
```diff
@@ -806,6 +807,163 @@ def centralize(mat):
...
+    y = np.asarray(y)
+
+    # Mask zeros
+    if robust:
+        x = np.ma.masked_array(x, mask=x == 0)
+        y = np.ma.masked_array(y, mask=y == 0)
+    
+    # Log transformation
+    x = np.log(x)
+    y = np.log(y)
```
mortonjt, line 876: If we are going to use the robust option, we probably want to use the appropriate functions.  In this particular case, we'd want `np.ma.log`

skbio/stats/composition.py
mortonjt: While we are at it, we probably want to have 2 different private functions, namely `_robust_vlr` and `_vlr`, where

```python
def _vlr(x, y, ddof=1):
    """ Assumes x and y don't have zeros"""
    return np.var(np.log(x), np.log(y))

def _robust_vlr(x, y, ddof=1):
   x = np.ma.masked_array(x, mask=x == 0)
   y = np.ma.masked_array(y, mask=y == 0)
   x = np.ma.log(x)
   y = np.ma.log(y)
   return np.ma.var(x - y)

def vlr(x, y, ddof=1, robust=False):
    x = closure(x) # checks for negative proportions and normalizes to 1
    y = closure(y)
    if robust:
        return _robust_vlr(x, y, ddof)
    return _vlr(x, y)
```

skbio/stats/composition.py
```diff
@@ -806,6 +807,163 @@ def centralize(mat):
...
+    
+    #Variance log ratio
+    return np.var(x - y, ddof=ddof)
+
+@experimental(as_of="0.5.7")
+def pairwise_vlr(mat:Union[np.array, pd.DataFrame], ids:list=None, ddof:int=1, robust=False) -> DistanceMatrix:
```
mortonjt, line 882: I think the same strategy discussed with `vlr` should also apply here.

Also, let's not use the Union typing for now.  None of the other functions in composition don't use pandas yet, so we should be consistent for now -- we can add pandas in another pull request.

skbio/stats/composition.py
```diff
@@ -806,6 +807,163 @@ def centralize(mat):
...
+    .. [1] V. Lovell D, Pawlowsky-Glahn V, Egozcue JJ, Marguerat S, Bähler J (2015) 
+           Proportionality: A Valid Alternative to Correlation for Relative Data. 
+           PLoS Comput Biol 11(3): e1004075. https://doi.org/10.1371/journal.pcbi.1004075
+    .. [2] Erb, I., Notredame, C. 
+           How should we measure proportionality on relative gene expression data?. 
+           Theory Biosci. 135, 21–36 (2016). https://doi.org/10.1007/s12064-015-0220-8
+    """
+    # Identifiers
+    components = None
+    if isinstance(mat, pd.DataFrame):
```
mortonjt, line 945: let's knock this out for now.

skbio/stats/composition.py
```diff
@@ -806,6 +807,163 @@ def centralize(mat):
...
+    # Mask zeros
+    X = mat.astype(np.float64)
+    if robust:
+        X = np.ma.masked_array(X, mask=X == 0)
+        
+    # Log Transform
+    X_log = np.log(X)
+    
+    # Variance Log Ratio
+    covariance = np.cov(X_log.T, ddof=ddof)
```
mortonjt, line 960: there is a `np.ma.cov` we could replace this with in the `_robust_pairwise_vlr_method`

skbio/stats/composition.py
```diff
@@ -1153,6 +1311,7 @@ def ancom(table, grouping,
         return ancom_df, percentile_df
 
 
+
```
mortonjt, line 1314:
```suggestion
```

skbio/stats/composition.py
```diff
@@ -1324,3 +1483,4 @@ def _check_orthogonality(basis):
     if not np.allclose(inner(basis, basis), np.identity(len(basis)),
                        rtol=1e-4, atol=1e-6):
         raise ValueError("Aitchison basis is not orthonormal")
+
```
mortonjt, line 1486:
```suggestion
```

skbio/stats/composition.py
```diff
@@ -806,6 +807,163 @@ def centralize(mat):
...
+    >>> mat = np.asarray([
+        [1,2,3],
+        [5,8,13],
+        [21,34,55],
+    ])
+    >>> dism = pairwise_vlr(mat)
+    >>> dism.redundant_form()
+    # array([[0.        , 0.01576411, 0.00649553],
+    #        [0.01576411, 0.        , 0.00202147],
+    #        [0.00649553, 0.00202147, 0.        ]])
```
mortonjt, line 919: I don't think this is going to pass the doc tests -- the doc tests won't like to have comments in it.
So this should be instead
```python
     array([[0.        , 0.01576411, 0.00649553],
            [0.01576411, 0.        , 0.00202147],
            [0.00649553, 0.00202147, 0.        ]])
```

skbio/stats/composition.py
```diff
@@ -806,6 +807,163 @@ def centralize(mat):
...
+        mat, 
+        index=["sample_1", "sample_2", "sample_3"], 
+        columns=["component_1", "component_2", "component_3"],
+    )
+    >>> dism = pairwise_vlr(df)
+    >>> dism.to_data_frame()
+    # component_1	component_2	component_3
+    # component_1	0.000000	0.015764	0.006496
+    # component_2	0.015764	0.000000	0.002021
+    # component_3	0.006496	0.002021	0.000000
```
mortonjt, line 932: Same here

mortonjt (changes requested):
First pass is looking good!  I've added some comments to restructure the code to make it more modular, particularly around reorganizing the robust arguments.

We do also need to have unittests for all of these functions. 
At a minimum we need 1 test per method, in addition to tests for the failure cases (i.e. multiple zeros for the robust case).

jolespin:
All these are great ideas!  I also really like the private methods as well.  Seems like a really clean way to keep things organized.  I'll make sure to incorporate all of these changes. 

Quite embarrassing to admit but I've never done a unit test (nor a PR for a major package) before.  What do/should I do and where do I do/put for the unit test?

mortonjt:
It's not too dissimilar to the doctests proposed.  You'd be adding more test cases to [this file](https://github.com/biocore/scikit-bio/blob/master/skbio/stats/tests/test_composition.py).

I'd recommend adding something like
```python
class TestVLR(unittest.TestCase):
    def test_normal_vlr(self):
        # stuff happens here
    def test_robust_vlr(self):
        # stuff happens here
    def test_normal_pairwise_vlr(self):
        # stuff happens here
    def test_robust_vlr(self):
        # stuff happens here

which you can test like a regular python file via `python test_composition.py`

**Comment by jolespin** [comment]
Just checking in to see if there are any pending items on my end.

File: skbio/stats/composition.py [hunk 13]
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
...
+    }
+    
+    # Variance Log Ratio
+    if robust:
+        vlr_data = _robust_pairwise_vlr(**kwargs)
+    else:
+        vlr_data = _pairwise_vlr(**kwargs)
+    
+    # Force symmetry
+    vlr_data = (vlr_data + vlr_data.T)/2
```
mortonjt, line 1117: Can you print out an example where this is isn't the case?  VLR should be perfectly symmetrical by default

skbio/stats/composition.py
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
...
+    -------
+    float
+         variance log ratio value
+         
+    Examples
+    --------
+    No zeros
+    >>> x = [1,2,3]
+    >>> y = [5,8,13]
+    >>> %timeit vlr(x,y)
```
mortonjt, line 921: let's cut out the timings out of the docstrings - they probably won't work there anyways

skbio/stats/composition.py
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
...
+           Proportionality: A Valid Alternative to Correlation for Relative Data. 
+           PLoS Comput Biol 11(3): e1004075. https://doi.org/10.1371/journal.pcbi.1004075
+    .. [2] Erb, I., Notredame, C. 
+           How should we measure proportionality on relative gene expression data?. 
+           Theory Biosci. 135, 21–36 (2016). https://doi.org/10.1007/s12064-015-0220-8
+    """
+
+    # Mask zeros
+    mat = mat.astype(np.float64)
+
```
mortonjt, line 1103:
```suggestion
    mat = closure(mat)
```
This will ward against the user inputting negative proportions

skbio/stats/composition.py
```diff
@@ -101,12 +101,12 @@
 #
 # The full license is in the file COPYING.txt, distributed with this software.
 # ----------------------------------------------------------------------------
-
 import numpy as np
 import pandas as pd
 import scipy.stats
 import skbio.util
 from skbio.util._decorator import experimental
+from skbio.stats.distance import DistanceMatrix, DistanceMatrixError
```
mortonjt, line 109:
```suggestion
from skbio.stats.distance import DistanceMatrix
```
`DistanceMatrixError` should be imported in the `test_composition.py`

skbio/stats/composition.py
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
...
+           Proportionality: A Valid Alternative to Correlation for Relative Data. 
+           PLoS Comput Biol 11(3): e1004075. https://doi.org/10.1371/journal.pcbi.1004075
+    .. [2] Erb, I., Notredame, C. 
+           How should we measure proportionality on relative gene expression data?. 
+           Theory Biosci. 135, 21–36 (2016). https://doi.org/10.1007/s12064-015-0220-8
+    """
+    # Convert array_like to numpy array
+    x = np.asarray(x)
+    y = np.asarray(y)
+
```
mortonjt, line 948: We probably should have some checks to make sure that the inputs are sane (i.e. greater or equal to zero).
It'll be great to have a unittest that also accommodates this as well.

mortonjt (commented):
Minor comments to beef up the error handling / unittests.

The underlying logic is fine -- I think the major hurdle here is to make sure that the documentation is styled correctly, in particular pycodestyle, flake8 and sphinx.  See the [here](https://github.com/biocore/scikit-bio/blob/master/CONTRIBUTING.md#setting-up-a-development-environment) on how to build and test the environment on your local machine.

jolespin:
It's one of those situations where you get really small values for differences. 

```python 
def _pairwise_vlr(mat:np.array, ddof:int):
    r"""
    Performs pairwise variance log ratio transformation
    
    Parameters
    ----------
    mat : array_like, float
       a matrix of proportions where
       rows = compositions and
       columns = components
       
    ids: array_like, str
        component names
        
    ddof: int
        degrees of freedom
       
    Returns
    -------
    skbio.DistanceMatrix
         distance matrix of variance log ratio values
         
    
    References
    ----------
    .. [1] V. Lovell D, Pawlowsky-Glahn V, Egozcue JJ, Marguerat S, Bähler J (2015) 
           Proportionality: A Valid Alternative to Correlation for Relative Data. 
           PLoS Comput Biol 11(3): e1004075. https://doi.org/10.1371/journal.pcbi.1004075
    .. [2] Erb, I., Notredame, C. 
           How should we measure proportionality on relative gene expression data?. 
           Theory Biosci. 135, 21–36 (2016). https://doi.org/10.1007/s12064-015-0220-8
    """
        
    # Log Transform
    X_log = np.log(mat)
    
    # Variance Log Ratio
    covariance = np.cov(X_log.T, ddof=ddof) 
    diagonal = np.diagonal(covariance)
    vlr_data = -2*covariance + diagonal[:,np.newaxis] + diagonal 
    return vlr_data

mat = np.array([[1,1,2],[3,5,8],[13,21,55]])
vlr_data = _pairwise_vlr(mat, ddof=1)
vlr_data - vlr_data.T
# array([[ 0.00000000e+00,  2.22044605e-16, -2.22044605e-16],
#        [-2.22044605e-16,  0.00000000e+00,  0.00000000e+00],
#        [ 2.22044605e-16,  0.00000000e+00,  0.00000000e+00]])
```

I think it's similar to why [np.allclose](https://numpy.org/doc/stable/reference/generated/numpy.allclose.html) has the `rtol` and `atol` arguments.  One way around this is to give `DistanceMatrix` a similar `tol=1e-15` argument where it can be symmetric with the exception of minor float errors.

mortonjt:
Got it.  Ok, then the transpose solution you proposed should be ok then.

skbio/stats/composition.py
jolespin: Are you thinking something like this? 

```python
x = np.asarray(x)
y = np.asarray(y)

assert np.all(x >= 0)
assert np.all(y >= 0)
```

skbio/stats/composition.py
jolespin: I thought I removed all of them.  That one must have slipped through the cracks.

mortonjt:
@jolespin on second thought, the transpose solution is going to double the runtime, particularly if the matrix is huge.

Any thoughts on adding a `validate_args` option in the `pairwise_vlr` arguments?  If true, it'll return the transposed solution.  If false, it'll return a `DissimiliarityMatrix` object.

jolespin:
What about something like this? 

```python
    # Return distance matrix
    try:
        return DistanceMatrix(vlr_data, ids=ids)
    except DistanceMatrixError:
        # Force symmetry
        vlr_data = (vlr_data + vlr_data.T)/2
        return DistanceMatrix(vlr_data, ids=ids)
```

mortonjt:
`try / except` statements is an anti-pattern in this case. Plus it doesn't necessarily circumvent the runtime issue.

jolespin:
Good point, how is this? 

```python
@experimental(as_of="0.5.7")
def pairwise_vlr(mat, ids=None, ddof=1, robust=False, validate_args=True) -> DistanceMatrix:
    r"""
    Performs pairwise variance log ratio transformation
    
    Parameters
    ----------
    mat : array_like, float
       a matrix of proportions where
       rows = compositions and
       columns = components
       
    ids: array_like, str
        component names
        
    ddof: int
        degrees of freedom
        
    robust: bool
        mask zeros at the cost of performance
       
    Returns
    -------
    skbio.DistanceMatrix
         distance matrix of variance log ratio values
         
    Examples
    --------
    >>> mat = np.asarray([
        [1,2,3],
        [5,8,13],
        [21,34,55],
    ])
    >>> dism = pairwise_vlr(mat)
    >>> dism.redundant_form()
    array([[0.        , 0.01576411, 0.00649553],
           [0.01576411, 0.        , 0.00202147],
           [0.00649553, 0.00202147, 0.        ]])

    
    References
    ----------
    .. [1] V. Lovell D, Pawlowsky-Glahn V, Egozcue JJ, Marguerat S, Bähler J (2015) 
           Proportionality: A Valid Alternative to Correlation for Relative Data. 
           PLoS Comput Biol 11(3): e1004075. https://doi.org/10.1371/journal.pcbi.1004075
    .. [2] Erb, I., Notredame, C. 
           How should we measure proportionality on relative gene expression data?. 
           Theory Biosci. 135, 21–36 (2016). https://doi.org/10.1007/s12064-015-0220-8
    """

    # Mask zeros
    mat = closure(mat.astype(np.float64))

    # Set up input and parameters
    kwargs = { 
        "mat":mat,
        "ddof":ddof,
    }
    
    # Variance Log Ratio
    if robust:
        vlr_data = _robust_pairwise_vlr(**kwargs)
    else:
        vlr_data = _pairwise_vlr(**kwargs)

    # Return distance matrix
    if validate_args:
        return DistanceMatrix(vlr_data, ids=ids)
    else:
        return DissimiliarityMatrix(vlr_data, ids=ids)
```

mortonjt:
Let's rename `validate_args` -> `validate` to be consistent with [DissimilarityMatrix](https://github.com/biocore/scikit-bio/blob/master/skbio/stats/distance/_base.py#L75)

But yes, I think this should be ok.

jolespin:
> Minor comments to beef up the error handling / unittests.
> 
> The underlying logic is fine -- I think the major hurdle here is to make sure that the documentation is styled correctly, in particular pycodestyle, flake8 and sphinx. See the [here](https://github.com/biocore/scikit-bio/blob/master/CONTRIBUTING.md#setting-up-a-development-environment) on how to build and test the environment on your local machine.

Working on this right now and I have just about all the `skbio/stats/composition.py` adjustments necessary. 

Blanking on this tho...

Here's the 2 lines: 
```python
def pairwise_vlr(mat, ids =None, ddof: int =1,
                robust: bool= False, validate_args: bool= True):
```

Here's the test errors:
```
skbio/stats/composition.py:1062:26: E251 unexpected spaces around keyword / parameter equals
skbio/stats/composition.py:1062:45: E252 missing whitespace around parameter equals
skbio/stats/composition.py:1063:17: E128 continuation line under-indented for visual indent
skbio/stats/composition.py:1063:29: E252 missing whitespace around parameter equals
skbio/stats/composition.py:1063:57: E252 missing whitespace around parameter equals
```

I've tried a bunch of different combinations of spaces around the = or not having them and can't get it to pass.

skbio/stats/composition.py
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
     return perturb_inv(mat, cen)
 
 
+
+@experimental(as_of="0.5.7")
+def _vlr(x:np.array, y:np.array, ddof:int):
```
mortonjt, line 812:
```suggestion
def _vlr(x: np.array, y: np.array, ddof: int):
```

skbio/stats/composition.py
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
...
+    r"""
+    Calculates variance log ratio
+    
+    Parameters
+    ----------
+    x : array_like, float
+       a 1-dimensional vector of proportions
+    y : array_like, float
+       a 1-dimensional vector of proportions
+
```
mortonjt, line 822:
```suggestion
```

skbio/stats/composition.py
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
...
+    # Log transformation
+    x = np.log(x)
+    y = np.log(y)
+    
+    #Variance log ratio
+    return np.var(x - y, ddof=ddof)
+
+
+@experimental(as_of="0.5.7")
+def _robust_vlr(x:np.array, y:np.array, ddof:int):
```
mortonjt, line 852:
```suggestion
def _robust_vlr(x: np.array, y: np.array, ddof: int):
```

skbio/stats/composition.py
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
...
+    
+    # Log transformation
+    x = np.ma.log(x)
+    y = np.ma.log(y)
+    
+    #Variance log ratio
+    return np.ma.var(x - y, ddof=ddof)
+
+@experimental(as_of="0.5.7")
+def vlr(x:np.array, y:np.array, ddof:int=1, robust:bool=False):
```
mortonjt, line 894: same here

skbio/stats/composition.py
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
...
+           Theory Biosci. 135, 21–36 (2016). https://doi.org/10.1007/s12064-015-0220-8
+    """
+    # Convert array_like to numpy array
+    x = np.asarray(x)
+    y = np.asarray(y)
+
+    # Set up input and parameters
+    kwargs = { 
+        "x":x,
+        "y":y,
```
mortonjt, line 952:
```suggestion
        "y": y,
```

skbio/stats/composition.py
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
...
+           How should we measure proportionality on relative gene expression data?. 
+           Theory Biosci. 135, 21–36 (2016). https://doi.org/10.1007/s12064-015-0220-8
+    """
+    # Convert array_like to numpy array
+    x = np.asarray(x)
+    y = np.asarray(y)
+
+    # Set up input and parameters
+    kwargs = { 
+        "x":x,
```
mortonjt, line 951:
```suggestion
        "x": x,
```

skbio/stats/composition.py
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
...
+        "y":y,
+        "ddof":ddof,
```
mortonjt, line 953:
```suggestion
        "ddof": ddof,
```

skbio/stats/composition.py
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
...
+    X_log = np.log(mat)
+    
+    # Variance Log Ratio
+    covariance = np.cov(X_log.T, ddof=ddof) 
+    diagonal = np.diagonal(covariance)
+    vlr_data = -2*covariance + diagonal[:,np.newaxis] + diagonal 
+    return vlr_data
+
+@experimental(as_of="0.5.7")
+def _robust_pairwise_vlr(mat:np.array, ddof:int):
```
mortonjt, line 1007: same here

skbio/stats/composition.py
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
...
+           Theory Biosci. 135, 21–36 (2016). https://doi.org/10.1007/s12064-015-0220-8
+    """
+        
+    # Log Transform
+    X_log = np.log(mat)
+    
+    # Variance Log Ratio
+    covariance = np.cov(X_log.T, ddof=ddof) 
+    diagonal = np.diagonal(covariance)
+    vlr_data = -2*covariance + diagonal[:,np.newaxis] + diagonal
```
mortonjt, line 1003:
```suggestion
    vlr_data = -2 * covariance + diagonal[:, np.newaxis] + diagonal 
```

skbio/stats/composition.py
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
...
+    Parameters
+    ----------
+    mat : array_like, float
+       a matrix of proportions where
+       rows = compositions and
+       columns = components
+       
+    ids: array_like, str
+        component names
+
```
mortonjt, line 1020:
```suggestion
```

skbio/stats/composition.py
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
...
+    r"""
+    Performs pairwise variance log ratio transformation while masking zeros
+    
+    Parameters
+    ----------
+    mat : array_like, float
+       a matrix of proportions where
+       rows = compositions and
+       columns = components
+
```
mortonjt, line 1017:
```suggestion
```

skbio/stats/composition.py
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
...
+    # Mask zeros
+    X = np.ma.masked_array(mat, mask=mat == 0)
+    
+    # Log Transform
+    X_log = np.ma.log(X)
+    
+    # Variance Log Ratio
+    covariance = np.ma.cov(X_log.T, ddof=ddof) 
+    diagonal = np.ma.diagonal(covariance)
+    vlr_data = -2*covariance + diagonal[:,np.newaxis] + diagonal
```
mortonjt, line 1048: same here

mortonjt (commented):
Adding some suggestions on resolving the pycodestyle issues.

skbio/stats/composition.py
```diff
@@ -101,12 +101,12 @@
 #
```
pirovc, line 101: Also add vlr and pairwise_vlr to the function list?

skbio/stats/composition.py
```diff
@@ -807,6 +807,319 @@ def centralize(mat):
...
+    X = np.ma.masked_array(mat, mask=mat == 0)
+    
+    # Log Transform
+    X_log = np.ma.log(X)
+    
+    # Variance Log Ratio
+    covariance = np.ma.cov(X_log.T, ddof=ddof) 
+    diagonal = np.ma.diagonal(covariance)
+    vlr_data = -2*covariance + diagonal[:,np.newaxis] + diagonal 
+    return vlr_data
```
mortonjt, line 1049: Finally getting around to playing with this.  Turns out that vlr_data is a masked array, and the pairs that don't have any common samples will be filled with some random number (i.e. -2).  This won't play nicely with the distance matrix object since you can't have negative distances.

There are a couple of ways we can deal with this.
1. Have this as a separate method (i.e. `masked_pairwise_vlr`) where we just return the masked array and let the user deal with it.  We won't have DistanceMatrix compatibility anymore, but what gives ...
2. Have a parameter in `_robust_vlr` called `fill_value` set to an arbitrary high constant (i.e. 1e6) that fills these missing values to make the DistanceMatrix methods happy.

mortonjt (commented):
I was playing around with this module on some real data, and already I am noticing some neat edge cases.

So turns out if you do a robust VLR, you can still end up with zero problems, since you need the nonzero entries between your pair of taxa to line up.  Even if you select for taxa that are highly abundant, you will have some pairs of taxa that are *never* observed in the same samples.  The `robust_vlr` as it stands will mask these values -- but this will break the existing DistanceMatrix construction.

skbio/stats/composition.py
jolespin: @mortonjt this is interesting.  Can you show an example so I can figure out how we can implement this? I just submitted my PR.  I've almost fixed all of the formatting requests to pass tests.  I could use a little help with these ones: 
```
skbio/stats/composition.py:1063:5: E128 continuation line under-indented for visual indent
skbio/stats/composition.py:1064:5: E128 continuation line under-indented for visual indent
skbio/stats/composition.py:1065:5: E128 continuation line under-indented for visual indent
skbio/stats/composition.py:1066:5: E128 continuation line under-indented for visual indent
skbio/stats/composition.py:1066:5: E125 continuation line with same indent as next logical line
```

jolespin:
I just submitted my PR.  I've almost fixed all of the formatting requests to pass tests.  I could use a little help with these ones: 
```
skbio/stats/composition.py:1063:5: E128 continuation line under-indented for visual indent
skbio/stats/composition.py:1064:5: E128 continuation line under-indented for visual indent
skbio/stats/composition.py:1065:5: E128 continuation line under-indented for visual indent
skbio/stats/composition.py:1066:5: E128 continuation line under-indented for visual indent
skbio/stats/composition.py:1066:5: E125 continuation line with same indent as next logical line
```

skbio/stats/composition.py
mortonjt: You can pull out any microbiome dataset and probably see this issue.  I used the [Global Gut dataset](https://qiita.ucsd.edu/study/description/850) and only looked the top 100 microbes.

I was also noticing that there were some negative VLR values, which was weird.

skbio/stats/composition.py
mortonjt: yes, `ValueError` would be preferable since it provides an informative message.

mortonjt (commented):
Regarding the indentation errors, you can try [autopep8](https://pypi.org/project/autopep8/).

wasade:
@jolespin @mortonjt, it looks like this PR still has pending items, is that accurate?

mortonjt:
Correct. There’s some pep8ing left. Although it’s pro minor

wasade:
Great thanks, any chance you could finish that off this week?

jolespin:
Ah man, I must have missed that. Do you know what could be causing the miniscule error? Maybe something regarding ddof? 

I won't be able to handle this until I finish the paper I'm writing which is time sensitive.

wasade:
Sounds good. Let's leave this open

mortonjt:
Thanks for your contribution @jolespin , this code has now been merged into skbio

jolespin:
@mortonjt That's great news!  Apologies I left this stale.  Had to finish up my dissertation and it took up all of my focus for a while.

I'm trying to find out which changes were added to fix the errors I had before.

Issue #1796, opened 2022-02-25
SIMD Everywhere commit (b470a55) breaks build on Windows


theAeon:
Running setup.py on windows 10 x64 halts on line 91 of setup.py with a KeyError. As far as I can tell, the setup script is trying to look up a config variable with `sysconfig.get_config_vars()['CC']` that does not exist on windows, causing an unhandled exception. Should be a quick patch.

wasade:
@theAeon with #1797 and 1798 merged, can you check if this issue is now resolved?

theAeon:
Well, they were my PRs, I know for a fact they resolve it haha

wasade:
:) great thanks

Issue #1791, opened 2021-12-17
In Python 3.10, Py_REFCNT output is not an lvalue and therefore cannot be an operand for ++ or --


ScottNortonPhD:
Occurs when using either the pypa release (`pip install -U scikit-bio`) or the master branch (`pip install -U "git+https://github.com/biocore/scikit-bio"`).

Relevant portion of log from pip:
```
  gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -I/ysm-gpfs/apps/software/libffi/3.3-GCCcore-10.2.0/include -I/ysm-gpfs/apps/software/SQLite/3.33.0-GCCcore-10.2.0/include -fPIC -I/tmp/pip-build-env-4zx9x2cl/overlay/lib/python3.10/site-packages/numpy/core/include -I/tmp/pip-build-env-4zx9x2cl/overlay/lib/python3.10/site-packages/numpy/core/include -I/home/ssn33/project/local/include/python3.10 -c skbio/alignment/_ssw_wrapper.c -o build/temp.linux-x86_64-3.10/skbio/alignment/_ssw_wrapper.o -I. -fopenmp-simd -DSIMDE_ENABLE_OPENMP
  In file included from /tmp/pip-build-env-4zx9x2cl/overlay/lib/python3.10/site-packages/numpy/core/include/numpy/ndarraytypes.h:1969,
                   from /tmp/pip-build-env-4zx9x2cl/overlay/lib/python3.10/site-packages/numpy/core/include/numpy/ndarrayobject.h:12,
                   from /tmp/pip-build-env-4zx9x2cl/overlay/lib/python3.10/site-packages/numpy/core/include/numpy/arrayobject.h:4,
                   from skbio/alignment/_ssw_wrapper.c:624:
  /tmp/pip-build-env-4zx9x2cl/overlay/lib/python3.10/site-packages/numpy/core/include/numpy/npy_1_7_deprecated_api.h:17:2: warning: #warning "Using deprecated NumPy API, disable it with " "#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION" [-Wcpp]
     17 | #warning "Using deprecated NumPy API, disable it with " \
        |  ^~~~~~~
  skbio/alignment/_ssw_wrapper.c: In function ‘__pyx_tp_dealloc_5skbio_9alignment_12_ssw_wrapper_AlignmentStructure’:
[...]

PR #1798, opened 2022-02-26
Regenerate Cython wrappers on build by default 

theAeon:

* [X] I have read the guidelines in [CONTRIBUTING.md](https://github.com/biocore/scikit-bio/blob/master/CONTRIBUTING.md).

* [X] I have documented all public-facing changes in [CHANGELOG.md](https://github.com/biocore/scikit-bio/blob/master/CHANGELOG.md).


theAeon:
Closes #1791

ebolyen:
Hi @theAeon! I really appreciate this patch (and thank you for the careful attention to copyright), the bad news is this C file is generated with Cython, so your fix will be overwritten by re-compiles of the Cython.

I wonder if updating Cython and rebuilding the C wrappers would be sufficient to fix the issue?

wasade:
@ebolyen, should the `.c` files be removed actually seeing as they're generated?

ebolyen:
They are left in so that source builds will work without Cython as a dependency.

wasade:
Right, but why not make Cython a dependency? It’s mature, and other biocore projects depend on it as far as I know without issue

theAeon:
I'm with @wasade on this: if Cython was *supposed* to be generating the c files, then we need to make sure its a dependency. Judging by the issue I linked, everyone was using the pregenerated files. This is, of course, assuming that updating cython fixes this issue.

theAeon:
Interesting-regenerating also seems to fix #1796. Probably worth pulling in the MSVC tag regardless in case of a non-cython build.

ebolyen:
Relevant references:
https://www.python.org/dev/peps/pep-0518/
https://www.python.org/dev/peps/pep-0621/

I don't have any first-hand experience working with the new pyproject.toml format, I just know it exists to solve these issues.

ebolyen:
As to why there was an env-var, we treated Cython as a "development dependency". In conda-land, this would probably be called a build dependency. 

It's not needed as something a user would use, but developers should remember to re-render the Cython occasionally to the distributed C code which can then be compiled on the user's side.

It's a bit like how nodejs has just an unending number of transpilers which aren't a runtime dependency of the package, but are definitely needed to modify the source in any meaningful way.

theAeon:
I've been using the pyproject.toml format in my own projects by way of https://github.com/pdm-project/pdm, its honestly ridiculous how easy it is compared to setuptools. Not sure if a full conversion would be necessary to get this running again, though.

theAeon:
Seems we're blocking other projects from supporting 3.10-is there anything that needs to be changed here?

tillea:
Am Sat, Mar 26, 2022 at 01:15:00AM -0700 schrieb Andrew Robbins:
> Seems we're blocking other projects from supporting 3.10-is there anything that needs to be changed here?
Well, you can possibly inspect all patches we apply in the current Debian package which you can find here

    https://salsa.debian.org/med-team/python-skbio/-/tree/master/debian/patches

Kind regards, Andreas.

theAeon:
Oh-I meant from a PR perspective. Forcing cython compilation on works.

wasade:
@theAeon I'm going to merge this now, even though we're breaking on a few builds as the failures appear unrelated. I'll address them in a separate PR. Thanks!

PR #1821, opened 2022-09-08
Fix #1818

SilasK:
Please complete the following checklist:

* [ ] I have read the guidelines in [CONTRIBUTING.md](https://github.com/biocore/scikit-bio/blob/master/CONTRIBUTING.md).

* [x] I have documented all public-facing changes in [CHANGELOG.md](https://github.com/biocore/scikit-bio/blob/master/CHANGELOG.md).

* [ ] **This pull request includes code, documentation, or other content derived from external source(s).** If this is the case, ensure the external source's license is compatible with scikit-bio's [license](https://github.com/biocore/scikit-bio/blob/master/COPYING.txt). Include the license in the `licenses` directory and add a comment in the code giving proper attribution. Ensure any other requirements set forth by the license and/or author are satisfied. **It is your responsibility to disclose code, documentation, or other content derived from external source(s).** If you have questions about whether something can be included in the project or how to give proper attribution, include those questions in your pull request and a reviewer will assist you.

* [x] **This pull request does not include code, documentation, or other content derived from external source(s).**

**Note:** [REVIEWING.md](https://github.com/biocore/scikit-bio/blob/master/REVIEWING.md) may also be helpful to see some of the things code reviewers will be verifying when reviewing your pull request.

PR #1826, opened 2022-11-22
updates to support most recent dependency versions

gregcaporaso:
Some backward incompatible changes were introduced in pandas 1.5.0 and scipy 1.9.0 that required some updates here (thanks to @SilasK for sharing a fix for this in #1821).


gregcaporaso:
The current h5py failures seem related to [this issue](https://github.com/conda-forge/h5py-feedstock/issues/92), but the version pinning done there conflicts with our ci version requirement so need to experiment with other versions.

ebolyen:
Ok, USE_CYTHON=TRUE on 3.11 should succeed while USE_CYTHON=FALSE should fail (as a re-render is needed).

ebolyen:
TODO: 
 - update docs to a modern sphinx (`'U'` mode was removed in py3.11 so docutils is raising an error).
 - fix docker build

ebolyen:
Presuming I didn't mess anything up with my rebase, #1832 has cleaned up history.

Issue #1827, opened 2022-11-24
Suggestion: Remove sklearn as a dependency


qiyunzhu:
I noticed that scikit-learn is a dependency of scikit-bio. However, the only place it is used is in the calculation of [beta diversity](https://github.com/biocore/scikit-bio/blob/master/skbio/diversity/_driver.py) metrics. After examining the code, I believe that this heavy dependency can be totally removed, thus making scikit-bio slimmer, quicker to load, and more coherent to the original idea of `scikit` (meaning extensions of SciPy, if I understand correctly; thus I guess scikits probably shouldn't depend on each other).

Specifically, according to the code and [documentation](http://scikit-bio.org/docs/latest/generated/skbio.diversity.beta_diversity.html), it wraps any metric provided by scikit-learn's [`pairwise.distance_metrics`](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.distance_metrics.html). If not found, it will go for SciPy's [`pdist`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.pdist.html). Actually, scikit-learn only provides a handful of distance metrics, most of which are covered by SciPy. Some non-overlapping metrics are synonyms, such as: l1 = manhattan, l2 = euclidean. The only metric that scikit-learn has but SciPy does not is the [Haversine distance](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.haversine_distances.html), but it measures the distance between two points on a spheral surface, and I doubt if anyone would use it for calculating beta diversity between two communities.

Therefore I think the wrapper for scikit-learn metrics can be removed from scikit-bio, letting SciPy be the only source of (non-phylogenetic) beta diversity metrics. If needed (for sake of ecology; not for geometry) one can manually implement additional beta diversity metrics.

ebolyen:
I think this is a good idea, the only catch is we will need to make sure that some of our qualitative metrics cast numeric vectors into boolean before running. This is something sklearn does behind the scenes, but it is really quite trivial and we should be handling it on our end anyhow.

gregcaporaso:
We addressed this in #1831 - thanks for the suggestion @qiyunzhu!

qiyunzhu:
@gregcaporaso @ebolyen Thanks! Looks great!

PR #1831, opened 2022-12-07
remove scikit-learn dependency 

gregcaporaso:
scikit-learn has been removed as a dependency. This was a fairly heavy-weight dependency that was providing minor functionality to scikit-bio. The critical components have been implemented in scikit-bio directly. 

The other change that comes along with this is that three beta diversity metrics that were provided by sklearn are no longer available. These are `wminkowski`, `nan_euclidean`, and `haversine`. On testing, `wminkowski` and `haversine` did not work through `skbio.diversity.beta_diversity` (or `sklearn.metrics.pairwise_distances`). The former was deprecated in favor of calling `minkowski` with a vector of weights provided as kwarg `w` , and the latter does not work with data of this shape. `nan_euclidean` can still be accessed from scikit-learn directly if a user installs scikit-learn in their environment. For example:

```python
import skbio
from sklearn.metrics import pairwise_distances
counts = [[23, 64, 14, 0, 0, 3, 1],
        [0, 3, 35, 42, 0, 12, 1],
        [0, 5, 5, 0, 40, 40, 0],
        [44, 35, 9, 0, 1, 0, 0],
        [0, 2, 8, 0, 35, 45, 1],
        [0, 0, 25, 35, 0, 19, 0],
        [88, 31, 0, 5, 5, 5, 5],
        [44, 39, 0, 0, 0, 0, 0]]

sklearn_dm = pairwise_distances(counts, metric="nan_euclidean")
skbio_dm = skbio.DistanceMatrix(sklearn_dm)

from skbio.diversity import beta_diversity
beta_diversity("minkowski", counts, w=[1,1,1,1,1,1,2])
```

addresses #1827

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,10 @@
 * scikit-learn has been removed as a dependency. This was a fairly heavy-weight dependency that was providing minor functionality to scikit-bio. The critical components have been implemented in scikit-bio directly, and the non-criticial components are listed under "Backward-incompatible changes [experimental]".
 
 ### Backward-incompatible changes [experimental]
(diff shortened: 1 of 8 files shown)


gregcaporaso:
**Note: we should revert the last commit before merging - just trying to get the tests to run.**

ebolyen:
@gregcaporaso looks like there's an actual lint failure this time

ebolyen (approved):
Thanks @gregcaporaso!

PR #1836, opened 2022-12-08
MISC: loose ends from root cleanup

ebolyen:
Forgot to update the relative links in some of the dev readmes, and the `manifest.in` file no longer includes the dropped elements (or things that just don't need to be shipped in an sdist).

Also, it turns out renaming `COPYING.txt` to `LICENSE.txt` means the headers were wrong, fortunately it was quick to use sed to make the update.


ebolyen:
Coveragerc has been moved into pyproject.toml

ebolyen:
Coveragerc has been restored

pyproject.toml
```diff
@@ -5,3 +5,14 @@ requires = ["Cython>=0.29.32", "numpy>=1.9.2", "setuptools", "wheel"]
...
+
+[tool.check-manifest]
+ignore = [
+    ".coveragerc",
+    ".dockerignore",
+    ".editorconfig",
+    "RELEASE.md",
+    "REVIEWING.md",
+    "aarch64.Dockerfile",
+    "checklist.py",
```
gregcaporaso, line 17: I think `PULL_REQUEST_TEMPLATE.md` should also be in this list.

gregcaporaso (approved):
Just one comment - everything else looks good to me!

pyproject.toml
ebolyen: That's already ignored by being in .github (which is a default exclusion for check-manifest)

gregcaporaso: Got it, all good then.

PR #1837, opened 2022-12-08
Github Actions CI Refactor

ebolyen:
Makes the CI configurable at the top and runs the Cython tests on the latest version.

The matrix has been adjusted to test conda and pypi (via setup.py) instead of toggling Cython, since as long as the Cython renders correctly on each platform, the vendored C code is what we actually need to test.
 
Linting and Doc building happen on their own. And everything is chained to limit the number of GH runners needed without compromising initial info from failures.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,112 @@
...
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+env:
+  latest_python: "3.11"
+  supported_pythons: "3.8|3.9|3.10|3.11"
+  miniforge_version: "22.9.0-2"
+  miniforge_variant: "Mambaforge"
```
wasade, line 13: Did not know this was a thing. Very excited that this is a thing.

ebolyen: Right? It sets up everything in a sane way as well.

PR #1852, opened 2023-05-05
Issue 1817

### LINKED ISSUE #1817: Modifying IntervalMetadata.query() to yield all intervals by default (or adding another function to do this)?
Hi! This is not an urgent issue, but it's a situation where I suspect other people might be confused.

The situation: given an [`IntervalMetadata`](http://scikit-bio.org/docs/latest/generated/skbio.metadata.IntervalMetadata.html#skbio.metadata.IntervalMetadata) object (representing the coordinates of predicted genes in a contig), I want to iterate through all intervals included in this object. I do not necessarily know the length of the contig in advance (or, at least, I'd like to avoid reading through a large FASTA file just to figure out the contig length), and I do not necessarily know if any of the intervals share some common piece of metadata.

## Current behavior of `IntervalMetadata.query()`

I assumed that calling [`IntervalMetadata.query()`](http://scikit-bio.org/docs/latest/generated/skbio.metadata.IntervalMetadata.query.html#skbio.metadata.IntervalMetadata.query) with both the `bounds` and `metadata` parameters set to the default of `None` would work, since the documentation says for each of these parameters that "Default (`None`) means all `Intervals` meet this requirement." However, calling `.query()` with both parameters set to `None` does not yield anything.

The reason for this, I think, is that `bounds` being `None` makes `.query()` [assume](https://github.com/biocore/scikit-bio/blob/ecc78ece42ee2fb0658cb0b6586cd5558568d30c/skbio/metadata/_interval.py#L831-L833) that I am trying to filter through intervals that match the `metadata`. This results in a call to the `._query_attribute()` function, where [`metadata` being `None` instantly causes `None` to be returned](https://github.com/biocore/scikit-bio/blob/ecc78ece42ee2fb0658cb0b6586cd5558568d30c/skbio/metadata/_interval.py#L775-L787). This, in turn, causes no intervals to be yielded.

I thought that the `IntervalMetadata.lower_bound` and `IntervalMetadata.upper_bound` attributes might help with this (giving me something I could pass as a tuple to the `bounds` parameter for `.query()`), but the `upper_bound` for my `IntervalMetadata` object (which I read in using `skbio.io.read()` as shown [here](http://scikit-bio.org/docs/latest/generated/skbio.io.format.gff3.html#module-skbio.io.format.gff3)) is `None`. And when I pass `lower_bound = 0` and `upper_bound = None` to `.query()`, I get a `ValueError` (thrown from [here](https://github.com/biocore/scikit-bio/blob/ecc78ece42ee2fb0658cb0b6586cd5558568d30c/skbio/metadata/_interval.py#L1005-L1008)) that "`` `start` (0) cannot be a larger int than `end` (None).``"

## Expected behavior of `IntervalMetadata.query()`

Based on the documentation, I had assumed (and I suspect other people would also assume) that `.query()` called without parameters would select all intervals. The current behavior (selecting no intervals) is counterintuitive, I think; more importantly, there isn't a clear non-hacky way to select all intervals at once.

## Current workarounds

- **Workaround 1:** Directly iterate through `IntervalMetadata._intervals`. However, this seems like a brittle solution (since I assume `._intervals` is an internal attribute, and therefore subject to change in future skbio versions).

- **Workaround 2:** Call `IntervalMetadata.query(metadata={})`. This bypasses the `metadata is None` check in `._query_attribute()`, and causes `.query()` to yield all intervals. (This is a bit better than Workaround 1 IMO, although it's still kind of a hack.)

## Proposed solution

If this is actually an issue (and not just me imagining things :), I can think of two simple ways to address this issue:

- **Option 1:** Modify `.query()` so that both parameters being `None` selects all intervals.

- **Option 2:** Add a new function (named something like `get_all()`? Not sure what would be best) that selects all intervals. I guess changing the default behavior of `.query()` would technically be backwards incompatible, so this solution would bypass that while still allowing for an easy supported method to go through all intervals.

Thanks!

wasade:
Fixes #1817 

@fedarko, can you review please?


skbio/metadata/_interval.py
```diff
@@ -828,6 +828,9 @@ def query(self, bounds=None, metadata=None):
         Interval
             ``Interval`` object satisfying the search criteria.
         """
+        if bounds is None and metadata is None:
+            metadata = {}
```
fedarko, line 832: Thanks -- I've tested this locally, and confirm that this works (now `i.query() == i.query(metadata={})`. However, I believe that the code can be simplified by removing the `bounds is None` check on this line:

```python
if metadata is None:
    metadata = {}
```

This will also let us delete the identical check of `if metadata is None` on lines 840 and 841. Since that check is run through every iteration of the `for loc in bounds` loop, this may add a tiny performance boost (if nothing else, it'll make the code easier to read).

fedarko (changes requested):
Thank you, @wasade! This fixes the issue. I have one small suggestion that will simplify the rest of the function, but aside from that this looks good.

skbio/metadata/_interval.py
wasade: Thanks! Deferring this as at the time the focus is correctness of function

PR #1858, opened 2023-06-08
Fixing dependency conflict: filelock vs lockfile for cachecontrol

qiyunzhu:
[cachecontrol](https://github.com/psf/cachecontrol) replaced the deprecated [lockfile](https://pypi.org/project/lockfile/) with [filelock](https://pypi.org/project/filelock/). This change took place since [version 0.13.0](https://pypi.org/project/CacheControl/0.13.0/). Specifically, it [requires](https://github.com/psf/cachecontrol/blob/master/pyproject.toml) filelock >= 3.8.0. This change will hopefully fix the build problem.


ci/aarch64.conda_requirements.txt
```diff
@@ -1,5 +1,5 @@
-lockfile >= 0.10.2
-cachecontrol >= 0.11.5
+filelock >= 3.8.0
+cachecontrol >= 0.13.0
```
wasade, line 2: @ebolyen, do you anticipate these new minimal requirements presenting issues with QIIME 2 builds?

qiyunzhu:
@wasade @ebolyen A workaround may be having scikit-bio install both `filelock` and `lockfile`, such that whichever version of `cachecontrol` can pick what it likes.

qiyunzhu:
Or, we can completely ditch `cachecontrol`. In the scikit-bio codebase, only one place uses this library:

skbio/io/_iosources.py:

```python
from cachecontrol import CacheControl
from cachecontrol.caches import FileCache

class HTTPSource(IOSource):

    def get_reader(self):
        sess = CacheControl(requests.Session(),
                            cache=FileCache(tempfile.gettempdir()))
        req = sess.get(self.file)

        # if the response is not 200, an exception will be raised
        req.raise_for_status()

        return io.BufferedReader(io.BytesIO(req.content))
```

I suspect that this code can be replaced by Python's built-in code?

qiyunzhu:
Having a mechanism for retrieving remote resources is probably not difficult. GPT suggested the following with just Python's built-in:

```python
    def get_reader(self):
        req = urllib.request.Request(self.file)
        with urllib.request.urlopen(req) as response:
            if response.status != 200:
                raise http.client.HTTPException(f'HTTP request failed with status {response.status}')
            data = response.read()
        return io.BufferedReader(io.BytesIO(data))
```

Or the following with the `requests` library, which is widely adopted, and is already a dependency of QIIME 2.

```python
    def get_reader(self):
        req = requests.get(self.file)
        req.raise_for_status()
        return io.BufferedReader(io.BytesIO(req.content))
```

Both codes don't have a caching mechanism, but it is probably not necessary. As a reference, Pytorch uses the following [code](https://github.com/pytorch/vision/blob/657027f3548f20aea7b1c34e26511823b1607145/torchvision/datasets/utils.py#L46) for downloading bulkly datasets (likely larger than scikit-bio's). As shown, it merely uses Python's built-in.

```python
def _urlretrieve(url: str, filename: str, chunk_size: int = 1024 * 32) -> None:
    with urllib.request.urlopen(urllib.request.Request(url, headers={"User-Agent": USER_AGENT})) as response:
        _save_response_content(iter(lambda: response.read(chunk_size), b""), filename, length=response.length)
```

Just for reference... I don't have a preference here.

wasade:
Agree, however I don't know how this IO source is used so I'm not well positioned to guess the consequences of making a change. If you think this is in scope of maintenance though, then I would not discourage an exploration. If you do, I endorse the use of `requests`.

qiyunzhu:
I don't know either how `HTTPSource` is used. It is not mentioned (or unit-tested) anywhere in scikit-bio. I guess that it is a mechanism of downloading data from a remote location, like pandas, scikit-learn and pytorch does. If so, that would be useful. Probably need input from @ebolyen before taking any action.

qiyunzhu:
Revisited in #1863 .

PR #1866, opened 2023-07-24
Fix installation process by disallowing Cython version 3.0 

yairm210:
Cython released [a major update (3.0)](https://pypi.org/project/Cython/#history) last week, which among other things changed `/-division uses the true (float) division operator, unless cdivision is enabled.` 

This new version - and given the error, probably related to this specific change - leads to the following compile error when attempting to install scikit-bio via Poetry:

```
        [1/6] Cythonizing skbio/alignment/_ssw_wrapper.pyx
        /tmp/pip-build-env-xsru6wu_/overlay/lib/python3.8/site-packages/Cython/Compiler/Main.py:381: FutureWarning: Cython directive 'language_level' not set, using '3str' for now (Py3). This has changed from earlier releases! File: /tmp/pip-req-build-7o17y4ab/skbio/alignment/_ssw_wrapper.pyx
          tree = Parsing.p_module(s, pxd, full_module_name)
        
        Error compiling Cython file:
        ------------------------------------------------------------
        ...
                else:
                    matrix = self._convert_dict2d_to_matrix(substitution_matrix)
                # Set up our mask_length
                # Mask is recommended to be max(query_sequence/2, 15)
                if mask_auto:
                    self.mask_length = len(query_sequence) / 2
                                                           ^
        ------------------------------------------------------------
        
        skbio/alignment/_ssw_wrapper.pyx:584:51: Cannot assign type 'double' to 'int32_t'
```

The solution to this is to limit the Cython version used in the build process to 0.29 versions only - this is the quickest fix possible to allow users to continue installing scikit-bio.

A more knowledgeable person than I would be needed to enable compatibility with Cython 3.0 :)

---

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["Cython>=0.29.32", "numpy>=1.9.2", "setuptools", "wheel"]
+requires = ["Cython>=0.29.32, <3.0.0", "numpy>=1.9.2", "setuptools", "wheel"]
 
 [tool.pytest.ini_options]
 filterwarnings = [


wasade:
Thanks, @yairm210. It would be ideal to address the issues rather than pin unless there is an explicit and impossible incompatibility between 0.29 and 3.0. What happens if the result of that division is cast to satisfy the type constraint?

yairm210:
The compatible solution is - per their documentation - to enable `cdivision`, which is a compiler directive, as mentioned here: https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html

But I'm not sure where in your build process that should go :)

wasade:
What happens if you add a cast to satisfy the error?

yairm210:
No idea :) sounds like it should work?

wasade:
Are you able to test this out? We're extremely limited on developer effort for maintenance on this project, and any assistance is greatly appreciated

yairm210:
I can try tomorrow

yairm210:
After playing around with the installation of this on empty python:3.8 dockers, I discovered that this error *does not* occur when installing 0.5.8, *only* for 0.5.7 which we were using due to dependency hell, and for which I've therefore been testing this fix in my fork 0_0

So... this change definitely need not go into the master branch. The question is, if I do find a fix for 0.5.7 installation, would you even care? What would you do with it?

yairm210:
No problem
I recommend that you also run your CI steps *on the latest commit you have* as well because this is a major version change and stuff could be broken without you knowing :)

PR #1867, opened 2023-07-31
readme update QIIME -> QIIME2

antgonza:

* [ ] I have read the guidelines in [CONTRIBUTING.md](https://github.com/biocore/scikit-bio/blob/master/.github/CONTRIBUTING.md).


skbio/alignment/_tabular_msa.py
```diff
@@ -2416,10 +2416,8 @@ def sort(self, level=None, ascending=True):
         modified (a new object is *not* returned).
 
         """
-        series = self._seqs.sort_index(ascending=ascending, level=level)
-        self._seqs = series
-        positional_metadata = self.positional_metadata.sort_index(axis=1)
-        self.positional_metadata = positional_metadata
+        self._seqs.sort_index(ascending=ascending, level=level, inplace=True)
```
antgonza, line 2419: Hi @ebolyen, adding the note message here so you can see the changes I made to this method. 

Anyway, this PR is almost ready for review, except for one issue: multiple builds are failing one tests, like this:
https://github.com/biocore/scikit-bio/actions/runs/5719059744/job/15496443692?pr=1867#step:6:309

Looking closer, I think the errors are due to sorting by index but then having no specified order of those with the same index; for example the input for the test is: 
```
10    (((T)), ((C)), ((C)), ((G)))
10    (((T)), ((A)), ((G)), ((G)))
8     (((G)), ((G)), ((G)), ((G)))
10    (((T)), ((G)), ((G)), ((G)))
0     (((A)), ((C)), ((G)), ((T)))
10    (((T)), ((A)), ((G)), ((A)))
```
and the output is:
```
0     (((A)), ((C)), ((G)), ((T)))
8     (((G)), ((G)), ((G)), ((G)))
10    (((T)), ((C)), ((C)), ((G)))
10    (((T)), ((A)), ((G)), ((G)))
10    (((T)), ((G)), ((G)), ((G)))
10    (((T)), ((A)), ((G)), ((A)))
```
which means that the last 4 can be in any order. To further test this, I submitted some more changes to see which builds failed and, as expected, some other builds failed (that passed before), like: https://github.com/biocore/scikit-bio/actions/runs/5720081961/job/15499338048?pr=1867.

Do you agree with this assessment? If yes, do you have any suggestions on how to best deal with sorting by value?

antgonza: @ebolyen; another option is to change the test to just check the order of the index vs. the full Series - would this work?

wasade: I think the change in test is a plausible fix here, thanks @antgonza!

wasade:
@antgonza, could you add a note to the change log citing this PR mentioning pinning scipy for the maintenance release being worked on?

@ebolyen, please note that with this merge, scikit-bio will pin scipy to < 1.10.1. Are you aware of any potential issues with Q2 builds? If so then I'll open a separate PR to focus on addressing 1.10.1 support near term.

Issue #1886, opened 2023-11-04
Adopting new NumPy random generator


qiyunzhu:
NumPy adopted a new random number generator policy ([NEP 19](https://numpy.org/neps/nep-0019-rng-policy.html)) (see a [tutorial](https://albertcthomas.github.io/good-practices-random-number-generators/)) since version 1.17:

```python
rng = np.random.default_rng(42)
res = rng.integers(10000, size=10)
```

It is recommended over the convensional way:

```python
np.random.seed(42)
res = np.random.randint(10000, size=10)
```

**Consideration**: In addition to the rationales detailed in [NEP 19](https://numpy.org/neps/nep-0019-rng-policy.html), the new generator also has significant performance advantage in applications we care. Specifically, the rarefaction (i.e., subsampling) process, a very slow step in previous workflows, could gain 10x performance (see #1885 for details).

**Proposal**

Refactor scikit-bio codebase to replace the convensional `np.random` usages with the new random generator.

For example, a function previously written as:

```python
def subsample(counts, n):
    probs = counts / counts.sum()
    return np.random.multinomial(n, probs)
```

Should become:

```python
def subsample(counts, n, seed=None):
    rng = get_rand_gen(seed)
    probs = counts / counts.sum()
    return rng.multinomial(n, probs)
```

The global utility function `get_rand_gen` should be like (Here is a [similar one](https://github.com/scikit-learn/scikit-learn/blob/093e0cf14aff026cca6097e8c42f83b735d26358/sklearn/utils/validation.py#L1249) in scikit-learn, although it hasn't adopted the new random generator):

```python
def get_rand_gen(seed):
    if seed is None or isinstance(seed, int):
        return np.random.default_rng(seed)
    if isinstance(seed, np.random.Generator):
        return seed
    raise ValueError('Invalid seed.')
```

**Workload**: There are a total of 21 instances of `np.random` in the scikit-bio codebase (excluding unit and doc tests).

**Version control**: Scikit-bio currently requires NumPy >= 1.9.2. This needs to change to 1.17 in order to adopt the new random generator. Note that NumPy 1.17 was released in July 2019, which was before the release of Python 3.8 (Oct 2019). Since scikit-bio's Python version requirement is >= 3.8, it should be reasonable to require NumPy >= 1.17.

[...]

Issue #1885, opened 2023-11-04
Potential improvement for `subsample_counts`


qiyunzhu:
Update: Using NumPy's new random generator ([NEP 19](https://numpy.org/neps/nep-0019-rng-policy.html)) improves performance by 10 fold. This is because the new [`numpy.random.Generator.choice`](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.choice.html) uses the [Floyd's algorithm](https://dl.acm.org/doi/10.1145/30401.315746) to accelerate sampling speed, especially when sample size _k_ << total size _n_.

```python
def subsample_without_replacement(counts, n):
    rng = np.random.default_rng()
    unpacked = np.repeat(np.arange(counts.size), counts)
    permuted = rng.choice(unpacked, size=n, replace=False)
    idx, freqs = np.unique(permuted, return_counts=True)
    result = np.zeros_like(counts)
    result[idx] = freqs
    return result
```

However, the new random generator requires NumPy >= 1.19. Currently, skbio's [requirement](https://github.com/scikit-bio/scikit-bio/blob/master/ci/conda_requirements.txt) says NumPy >= 1.9.2. This may need to be updated.

---

The current [`subsample_counts`](https://github.com/scikit-bio/scikit-bio/blob/8c4825757c780ead5a41af8997b699021af0322b/skbio/stats/_subsample.py#L151) function uses a homebrew [Cython function](https://github.com/scikit-bio/scikit-bio/blob/8c4825757c780ead5a41af8997b699021af0322b/skbio/stats/__subsample.pyx#L13) to achieve subsampling, which I think is probably not necessary. An easy rework would be:

```python
def subsample_without_replacement(counts, n):
    unpacked = np.repeat(np.arange(counts.size), counts)
    permuted = np.random.choice(unpacked, size=n, replace=False)
    idx, freqs = np.unique(permuted, return_counts=True)
    result = np.zeros_like(counts)
    result[idx] = freqs
    return result
```

QIIME 2's [`rarefy` function](https://github.com/qiime2/q2-feature-table/blob/a1b415d9bdc06d1ade9ece09ad9c6226dc63f09d/q2_feature_table/_normalize.py#L12) does not use skbio's `subsample_counts`, but instead uses biom-format's table.subsample. Therefore, modifying this function in skbio will not risk breaking QIIME 2.

[...]

PR #1888, opened 2023-11-05
Refactored subsample_counts

qiyunzhu:
This PR rewrote part of `skbio.stat.subsample_counts`. The rationale is provided in #1885 .

Note that it is not yet the ultimate form. It will be further improve after the new NumPy random generator is adopted (see #1886 ).

Note that this change will not affect QIIME 2, which uses `biom.Table.subsample` instead.

***
**Update 2**: for sparse data, the following modification is more efficient. It extracts the indices of non-zero elements, and rans the core algorithm on them only, then convert the result back to the original shape. Inspired again by @wasade .

```python
def subsample_sparse(counts, n, seed=None):
    rng = np.random.default_rng(seed)
    nonzero = counts[nzidx := np.where(counts != 0)[0]]
    permuted = rng.choice(nonzero.sum(), size=n, replace=False, shuffle=False)
    digits = np.digitize(permuted, np.cumsum(nonzero))
    idx, freqs = np.unique(digits, return_counts=True)
    result = np.zeros_like(counts)
    result[nzidx[idx]] = freqs
    return result
```

As I found during the test, most memory consumption was involved in the overhead of getting non-zero indices and writing dense arrays. The core algorithm is quite efficient in both speed and memory usage.

Benchmarks:

1. The HMP2 dataset (10614 features by 1569 samples, density = 0.215), ran on an AMD Ryzen 9 6900HX CPU:

```
           wall time    peak memory
unpacking:    10.5  s   424,044,953
binning:       1.34 s   134,956,259
sparse:        1.24 s   134,967,990
prefilter:     1.18 s    29,144,285
biom *:        1.26 s   101,249,698
```

2. The EMP dataset (317314 features by 22811 samples, density = 0.00317), ran on two Intel Xeon Platinum 8358 CPUs:

```
             wall time      peak memory
unpacking:  didn't run
binning:      2min 24s   57,913,710,757
sparse:       1min 48s   57,911,498,675
prefilter:      27.9 s      187,892,256
biom *:       1min  1s    1,091,993,551
```

* biom is the latest development version (commit d3f82fe) which has the optimized algorithm. It serves as a reference but not a fair comparison because it operates on a different data structure.

***
**Update**: benchmark results based on the HMP2 WGS dataset classified by Woltka. The feature table has 10614 features by 1569 samples. The data and benchmark script are attached.
[data.tsv.zip](https://github.com/scikit-bio/scikit-bio/files/13273062/data.tsv.zip)
[bench.ipynb.zip](https://github.com/scikit-bio/scikit-bio/files/13273063/bench.ipynb.zip)

The original version discussed in #1885 :

```python
def subsample_unpack(counts, n, seed=None):
    rng = np.random.default_rng(seed)
    unpacked = np.repeat(np.arange(counts.size), counts)
    permuted = rng.choice(unpacked, size=n, replace=False, shuffle=False)
    idx, freqs = np.unique(permuted, return_counts=True)
    result = np.zeros_like(counts)
    result[idx] = freqs
    return result
```

The new version inspired by @wasade . It avoids unpacking which is memory intensive.

```python
def subsample_bin(counts, n, seed=None):
    rng = np.random.default_rng(seed)
    choices = rng.choice(counts.sum(), size=n, replace=False, shuffle=False)
    permuted = np.digitize(choices, np.cumsum(counts))
    idx, freqs = np.unique(permuted, return_counts=True)
    result = np.zeros_like(counts)
    result[idx] = freqs
    return result
```

Results:
`subsample_unpack`: Wall time: 24.9 s. Peak memory: 406,433,133.
`subsample_bin`: Wall time: 3.36 s. Peak memory: 266,910,395.
`skbio.stats.subsample_counts` as of v0.5.8: Wall time: 8min 39s. Peak memory: 696,843,704.
`biom.Table.subsample` as of v2.1.15: Wall time: 7min 36s. Peak memory: 444,404,878.

(Note: it may not be completely fair to include biom, which deals with a different data structure.)

Therefore, the new version is advantageous in both wall time and memory usage.

Comparison of distributions of subsampled counts of one sample, showing high overlap.

image

***


mortonjt:
This is *great* -- especially since it removes all of that extra cython code

qiyunzhu:
@mortonjt Thanks! @wasade What do you think? Especially because this is relevant to the same function in biom-format.

mortonjt:
Hi @qiyunzhu see my comment here : https://github.com/scikit-bio/scikit-bio/pull/1889

I think we need another test to ensure that the seeds are disjoint

qiyunzhu:
@mortonjt This PR and #1889 do not affect each other. Once both are merged, I intend to create another or a couple of PRs to connect every function that uses random number with the new random number generator.

wasade:
Instead of removing the Cython code, I recommend adopting the revised and optimized approach now taken in BIOM (see https://github.com/biocore/biom-format/pull/935)

qiyunzhu:
@wasade Thanks for the input! The subsampling without unpacking is certainly a big advantage. I read your code but couldn't understand well because I currectly do not have much knowledge of sparse matrix. Specifically, I don't know what is `indptr`. Do you want to suggest a way to move forward?

wasade:
Thanks! The wikipedia page on CSR has historically been pretty good (https://en.wikipedia.org/wiki/Sparse_matrix#Compressed_sparse_row_(CSR,_CRS_or_Yale_format)). But briefly, intptr represents the start/stop indices in data that correspond to a given row of data. The biom implementation is designed for many samples not a single sample, so that can be worked around.

qiyunzhu:
@wasade Thank you! Based on your inspiration, I came up with a new, simple NumPy solution that doesn't require unpacking:

```python
def subsample_without_replacement(counts, n):
        permuted = np.random.choice(counts_sum, size=n, replace=False)
        digitized = np.digitize(permuted, np.cumsum(counts))
        idx, freqs = np.unique(digitized, return_counts=True)
        result = np.zeros_like(counts)
        result[idx] = freqs
```

Note that the `np.random.choice` (including the new `np.random.Generator.choice`) step still requires generating `np.arange(counts_sum)` under the hood. Therefore, memory-wise it is not that efficient either. In my test, this entire function uses 50% memory compared with my the previous version.

I noticed that [your code](https://github.com/biocore/biom-format/blob/d3f82fe666c18bb30cbf571a7e114042308a326c/biom/_subsample.pyx#L99) also has this step. Therefore, I suspect that the memory issue also persists.

I think that a better solution may be using `np.random.randint` (or the new `np.random.Generator.integers`), which does not require unpacking (tested and working). However, unfortunately, it does not support `replace=False`. Therefore, perhaps a way is to let it return more than `n` numbers and keep only unique values. If the number of unique values falls below `n`, then do another round of `randint` until there are sufficient. This apparently is complicated.

qiyunzhu:
@wasade Not really. Using the new `np.random.Generator.choice` with `shuffle=False` significantly reduces memory usage. I think that it doesn't require unpacking. Therefore, problem solved! I would suggesting merging this PR as well as #1889 . Then we can add the new random generator to further improve this function (and others).

sfiligoi:
@qiyunzhu Can you provide some benchmark numbers for some sizable inputs?
I.e. EMP or larger?
Also, some benchmark numbers for a high-count (>4G) input would also be very valuable.

Both CPU-wise and memory wise, and comparing to the current master branch?

qiyunzhu:
@sfiligoi Sure! Benchmark results provided in the main thread. The input dataset is pretty large in the real world, but not as large as the metrics you suggested. If you have them, you are welcome to test them too!

wasade:
@qiyunzhu @sfiligoi and I had an out-of-band discussion on this. We have merged but unreleased code in BIOM, which was not part of the benchmark. When incorporated, it is quite a bit faster and less memory. This suggests the revised implementation in BIOM `master` should be considered for implementation here.

Screen Shot 2023-11-06 at 2 55 13 PM
Screen Shot 2023-11-06 at 2 55 36 PM

qiyunzhu:
This is being addressed in #2016 by @wasade . So closing the current PR.

Issue #946, opened 2015-06-11
Request: Adding in numpy.random.RandomState parameter into all randomized methods


mortonjt:
One problem with using straight up `np.random.seed` is there is a single global variable that handle all of the randomization calls.

If we wanted to parallel random sampling, this would quickly become a bottleneck.
Right now, if I wanted to use `skbio.stats.subsample_counts` to perform multiple rarefaction on a boat load of samples, I cannot fork off multiple processes to get this job done in parallel and get different random numbers - many of the processes will end up with the same random seed.  Take a look at this [gist](https://gist.github.com/brantfaircloth/73bbad82a89ccfe98dd7)

Using `np.random.RandomState` resolves this problem since it modularizes this functionality into separate objects.  This will allow for multiple independent streams to be created - making it possible to parallelize random sampling. 

Here are a few more discussions about numpy.random vs. numpy.random.RandomState on stackoverflow
http://stackoverflow.com/questions/5836335/consistenly-create-same-random-numpy-array
http://stackoverflow.com/questions/22994423/difference-between-np-random-seed-and-np-random-
http://stackoverflow.com/questions/21494489/what-does-numpy-random-seed0-do
http://stackoverflow.com/questions/14504866/python-multiprocessing-numpy-random
http://stackoverflow.com/questions/24345637/why-doesnt-numpy-random-and-multiprocessing-play-nice

Jorge-C:
Big +1, this is the way to go and has all the advantages that Jamie
mentioned!

mortonjt:
This is going to critical for reproducible random numbers, particularly in the context of parallelization. It would be great if this was bumped up the priority list

RNAer:
second that. this problem dawned to me as well.

PR #1889, opened 2023-11-05
Adopted NumPy's new random generator

qiyunzhu:
Addressing #1886 . This PR adopts NumPy's random generator to enable reproducible results, as well as better performance in some applications. See #1886 for a detailed explanation of this issue. See also #946 for a previous discussion among users and developers.

In order to adopt this new feature, support for NumPy <1.17 is being dropped.

Theoretically, it is possible to cross-support both old and new NumPy, as follows:

```python
def get_rng(seed=None):

    # new random generator (numpy >= 1.17)
    if hasattr(np.random, 'default_rng'):
        if seed is None or isinstance(seed, int):
            return np.random.default_rng(seed)
        if isinstance(seed, np.random.Generator):
            return seed
        raise ValueError('Invalid seed.')

    # legacy random generator (numpy < 1.17)
    else:
        if seed is None or isinstance(seed, int):
            return np.random.RandomState(seed)
        if isinstance(seed, np.random.RandomState):
            return seed
        raise ValueError('Invalid seed.')
```

However, this workaround has limitations:

1) Some methods are not cross-supported, such as `integers` (new) vs `randint` (old).

2) Old and new generators produce different results given the same seed. For example,

```python
>>> rs = np.random.RandomState(42)
>>> rs.randint(100)
51
>>> rng = np.random.default_rng(42)
>>> rng.integers(100)
8
```

Therefore, I think keeping backward compatibility is disadvantagous.


skbio/util/tests/test_misc.py
```diff
@@ -231,5 +233,41 @@ def gen():
         self.assertEqual(find_duplicates(gen()), set(['a', 2]))
 
 
+class TestGetRng(unittest.TestCase):
```
mortonjt, line 236: We should probably add a test to make sure that the seeds are disjoint.

Maybe for one of the use cases (i.e. subsample), we should have a test generating two random generators with 2 different seeds and assert that the subsampled counts aren't identical.

mortonjt (approved):
I think this pull request is good.  But we should get the suggested unittest implemented in https://github.com/scikit-bio/scikit-bio/pull/1888 before merging this pull request in.

Just to confirm, does #1888 depend on this PR?

qiyunzhu:
@mortonjt I added a test to see if random integers generated using 10 different seeds on my machine are the same as in the GitHub CI. Btw, this PR is not related to #1888 .

PR #1887, opened 2023-11-04
Removing beta diversity metric kulsinski to unlock SciPy 1.11+

qiyunzhu:
Addressing #1875 . Also reverting #1867 . My rationales for this change are:

1. SciPy replaced `kulsinski` with `kulczynski1` because the former was problematic (see SciPy issue [#2009](https://github.com/scipy/scipy/issues/2009)).
2. As a consequence, scikit-bio was not compatible with SciPy 1.11+.
3. When testing a vector against itself, `kulczynski1` will return inf. `kulsinski` will return a positive number, instead of 0. Therefore, neither function is a valid beta diversity metric.
4. Literature search didn't find anyone using it as a beta diversity metric.
5. `kulsinski` was not exposed to the user anywhere (documentation, function call, etc.). The user wouldn't know it is there unless they explicitly type "kulsinski". Removing it has minimum effect on backward compatibility.

Before merging this PR, we need to coordinate with the QIIME 2 team, because QIIME 2 has it as a beta diversity metric. @ebolyen What do you think?

Please complete the following checklist:

* [x] I have read the guidelines in [CONTRIBUTING.md](https://github.com/biocore/scikit-bio/blob/master/.github/CONTRIBUTING.md).

* [x] I have documented all public-facing changes in [CHANGELOG.md](https://github.com/biocore/scikit-bio/blob/master/CHANGELOG.md).

* [ ] **This pull request includes code, documentation, or other content derived from external source(s).** If this is the case, ensure the external source's license is compatible with scikit-bio's [license](https://github.com/biocore/scikit-bio/blob/master/LICENSE.txt). Include the license in the `licenses` directory and add a comment in the code giving proper attribution. Ensure any other requirements set forth by the license and/or author are satisfied. **It is your responsibility to disclose code, documentation, or other content derived from external source(s).** If you have questions about whether something can be included in the project or how to give proper attribution, include those questions in your pull request and a reviewer will assist you.

* [ ] **This pull request does not include code, documentation, or other content derived from external source(s).**

**Note:** [REVIEWING.md](https://github.com/biocore/scikit-bio/blob/master/REVIEWING.md) may also be helpful to see some of the things code reviewers will be verifying when reviewing your pull request.

PR #1890, opened 2023-11-06
Improved Fisher's alpha calculation (which unblocks SciPy 1.11+)

qiyunzhu:
The original motivation of this PR is to address the CI failure in #1887 , which attempted to introduce support for SciPy 1.11+, but got errors in `fisher_alpha`. Specifically, SciPy's `minimize_scalar` function failed to solve the equation and returned `nan`.

I looked into SciPy's changelog. There are several notes about changes in `minimize` and `minimize_scalar`, but it is not obvious how the calculation was affected.

However, I read the current [implementation](https://github.com/scikit-bio/scikit-bio/blob/ed3332a9b674823b5b52e29e44f3f2adb4e355ad/skbio/diversity/alpha/_base.py#L259) of `fisher_alpha`, the literature (especially [Fisher et al. (1943)](https://www.jstor.org/stable/1411), the original paper introducing this metric), and alternative implementations (such as vegan's [`fisherfit`](https://github.com/vegandevs/vegan/blob/master/R/fisherfit.R). I think that our implementation needs to be improved.

The new implementation is presented, tested and working with SciPy versions 1.8.1, 1.10.0 and 1.11.3.

***

Most importantly, the objective function that needs to be minimized, changed from:

```python
(alpha * np.log(1 + (n / alpha)) - s) ** 2
```

To:

```python
(alpha * np.log(1 + (n / alpha)) - s) ** 2 if alpha > 0 else np.inf
```

In the old version, there is risk of division by zero, because the function does not bound the value of $\alpha$. The old version uses a hack that modifies NumPy's global setting and reset it to mute the warnings. This solution was not ideal and, in SciPy 1.11, this causes problems in the optimization process. In the new version, when $\alpha$ reaches zero or below (detailed below), the objective function becomes infinite, therefore not optimal. This approach effectively prevented the problem.

The new version added handling of two special cases: $\alpha$ = infinite when all species are singletons (i.e., $S = N$) (detailed below). $\alpha$ = 0 when all species are zero.

The old version doesn't exploit SciPy's `OptimizeResult` class, which has an attribute `success`. The new version does. Meanwhile, the old version uses error = 1.0 as a threshold to determine whether the solution is converged. This is probably not right, as $\alpha$ can be very small (like 0.1234) in a normal community (i.e., a community that fits Fisher's log-series) or be very large (like 1,234,000) in a community full of singletons (detailed below). Therefore, setting a convergence threshold = 1.0 is not useful. The new version removes it.

Note: The new version uses the new NumPy decorator: `@np.errstate(invalid='ignore')` to mute warnings. There is only one warning, that occurs only once, during the first use of `fisher_alpha` in a matrix multiplication operation. It doesn't cause any problem. Replacing `np.inf` with some big number such as `1e32` can remove this warning. But I don't feel comfortable introducing an arbitrary upper bound (detailed below).

Note: SciPy's `minimize_scalar` allows specifying lower and upper bounds for optimization. However, $\alpha$ doesn't have an upper bound (see below), therefore one cannot utilize this feature.

Note: vegan uses a [different way](https://github.com/vegandevs/vegan/blob/master/R/fisherfit.R) to confine the search space: It sets bracket = (0.1, 50). The optimizer will look into this range, then performs a downhill search beyond the range. `minimize_scalar` also has this feature and I tested. With the same bracket, and replacing `np.inf` in the new objective function with `s ** 2` (which is the correct value at $\alpha = 0$), the function passed all unit tests. However, I am not too sure that it can handle all edge cases, nor I am sure the necessity of introducing an arbitrary bracket.

***

For better implementation of `fisher_alpha` one needs to understand the mathematical properties of $\alpha$. In particular, in the Fisher's alpha equation where $N$ is the number of individuals and $S$ is the number of species:

$$ S = \alpha \ln (1 + \frac{N}{\alpha}) $$

What is the range of $\alpha$? My understanding is:

$$ \alpha \in (0, \infty) $$

Here is Fig. 8 of [Fisher et al. (1943)](https://www.jstor.org/stable/1411):

Screenshot 2023-11-05 at 7 54 41 PM

It (as well as other figures) denotes that $\alpha$ = infinity when $N = S$ (i.e., all species are singletons).

This can also be understood as follows. When $N = S$, the equation becomes:

$$ x = \ln (1 + x) $$

Where $x = N / \alpha$. This function has one solution that is $x = 0$. Therefore $\alpha = \infty$.

Note: vegan does not make this a special case. Instead, it will generate a large number (e.g., 2147483650 for c(1,1,1,1,1)), and warn the user: "all species singletons: alpha arbitrarily high". In Python, if we remove the special case, the outcome will also be a large number. I think that it is due to arithmetic imprecision in the handling of floating-point numbers. The analytical solution should be infinity.

On the other side, $\alpha$ as an alpha diversity metric must be positive. This can also be deduced from equation (4) in Fisher et al. (1943), where $x$ is a finite positive number:

$$ S_{n} = \frac{\alpha}{n} x^{n} $$

Therefore, the coclusion is that $\alpha \in (0, \infty)$. This establishes the basis of the new implementation.

***


skbio/diversity/alpha/tests/test_base.py
```diff
@@ -107,25 +107,41 @@ def _diversity(indices, f):
...
         # case). Verified against R's vegan::fisher.alpha.
         obs = fisher_alpha([999, 0, 10])
-        self.assertAlmostEqual(obs, 0.2396492)
+        self.assertAlmostEqual(obs, 0.2396492, places=6)
+
+        # Should be infinite when all species are singletons
+        obs = fisher_alpha([1, 1, 1, 1, 1])
+        self.assertEqual(obs, np.inf)
+
+        # Should be infinite when there is no individual
```
wasade, line 134:
```suggestion
        # Should be zero when there is no individual
```

wasade (commented):
Minor comment

qiyunzhu:
@mortonjt Cool! It passed the test.

Issue #161, opened 2014-03-14
Define a SubstitutionMatrix, and define the standard substitution matrices


wasade:
This object should should contain a numpy `array` that holds the substitution values and support `__getitem__`. Ideally, it leverages numpy as much as possible through `array` lookups instead of `dict` lookups. E.g.:

``` python
class SubstitutionMatrix(object):
    _invalid_index = -1

    def __init__(self, mat, index):
        self._mat = mat  # the substitution values 
        self._index = index  # maps a character to the index in mat

    def __getitem__(self, args):
        from, to = args

        from_idx = self._index[ord(from)]
        to_idx = self._index[ord(to)]

        if from_idx == self._invalid_index:
            raise IndexError("Unknown element %s!" % from)
        if to_idx == self._invalid_index:
            raise IndexError("Unknown element %s!" % to)

        return self._mat[from_idx, to_idx]

    def to_values(self, seq1, seq2):
        # bulk perform the __getitem__, can take advantage of
        # array(seq1).view(int8) instead of mapping ord

        # should accept skbio.core.sequence objects as well

        # return an array(len(seq1), dtype=int8)
        pass

    @classmethod
    def from_dict(self, d):
        pass

    @classmethod
    def from_dict2d(self, d):
        pass
```

The `skbio.align.matrix` module should then predefine the standard matrices to allow for:

``` python
from skbio.align.matrix import BLOSUM62
His_to_Leu = BLOSUM62['H', 'L']

print BLOSUM62.to_values('AMP', 'AXP')
[4 -1 7]
```

PR #1912, opened 2023-12-22
Allowed whitespace of arbitrary length as delimiter in reading lsmat

qiyunzhu:
The intention is to enable support for fixed-width text matrices, such as substitution matrices (e.g., BLOSUM and PAM).

```
>>> !wget -q https://ftp.ncbi.nlm.nih.gov/blast/matrices/BLOSUM62
>>> !cat BLOSUM62
#  Matrix made by matblas from blosum62.iij
#  * column uses minimum score
#  BLOSUM Clustered Scoring Matrix in 1/2 Bit Units
#  Blocks Database = /data/blocks_5.0/blocks.dat
#  Cluster Percentage: >= 62
#  Entropy =   0.6979, Expected =  -0.5209
   A  R  N  D  C  Q  E  G  H  I  L  K  M  F  P  S  T  W  Y  V  B  Z  X  *
A  4 -1 -2 -2  0 -1 -1  0 -2 -1 -1 -1 -1 -2 -1  1  0 -3 -2  0 -2 -1  0 -4 
R -1  5  0 -2 -3  1  0 -2  0 -3 -2  2 -1 -3 -2 -1 -1 -3 -2 -3 -1  0 -1 -4 
N -2  0  6  1 -3  0  0  0  1 -3 -3  0 -2 -3 -2  1  0 -4 -2 -3  3  0 -1 -4 
D -2 -2  1  6 -3  0  2 -1 -1 -3 -4 -1 -3 -3 -1  0 -1 -4 -3 -3  4  1 -1 -4 
C  0 -3 -3 -3  9 -3 -4 -3 -3 -1 -1 -3 -1 -2 -3 -1 -1 -2 -2 -1 -3 -3 -2 -4 
Q -1  1  0  0 -3  5  2 -2  0 -3 -2  1  0 -3 -1  0 -1 -2 -1 -2  0  3 -1 -4 
E -1  0  0  2 -4  2  5 -2  0 -3 -3  1 -2 -3 -1  0 -1 -3 -2 -2  1  4 -1 -4 
G  0 -2  0 -1 -3 -2 -2  6 -2 -4 -4 -2 -3 -3 -2  0 -2 -2 -3 -3 -1 -2 -1 -4 
H -2  0  1 -1 -3  0  0 -2  8 -3 -3 -1 -2 -1 -2 -1 -2 -2  2 -3  0  0 -1 -4 
[...]

PR #1913, opened 2023-12-23
Added class `SubstitutionMatrix`

qiyunzhu:
This PR provides an explicit substitution matrix class for sequence alignment. It is an follow-up of the previous issue #161 , but uses a new design. It is a subclass of `DissimilarityMatrix`. I found that `DissimilarityMatrix` is a perfect existing structure for this purpose. Specifically, it provides a 2D NumPy array and a character-to-index dictionary that are essential for efficient sequence alignment (which necessitates converting sequences to vectors of indices followed by vectorized calculation). In contrast, the previous design -- a nested dictionary, is not suitable. Meanwhile, the new class `SubstitutionMatrix` provides a few extra methods and attributes over `DissimilarityMatrix` to make it more explict for the purpose.

Several most common matrices are bundled in the code, including NUC.4.4, PAM series and BLOSUM series. This choice follows the options in the NCBI blastn and blastp web servers. The matrix data were derived from the NCBI FTP server (https://ftp.ncbi.nlm.nih.gov/blast/matrices/). To my understanding, these data files are in the public domain (see [here](https://www.ncbi.nlm.nih.gov/books/NBK62345/) and [here](https://www.ncbi.nlm.nih.gov/books/about/copyright/)).

The user can read additional matrices retrieved from the URL above, or any compatible ones (including ones bundled in other popular programs, such as [BioPython](https://github.com/biopython/biopython/tree/master/Bio/Align/substitution_matrices/data)).

The reason it is under `sequence` instead of `alignment` is because substitution matrices are also useful for phylogenetic reconstruction.

The method `identity` replaces `make_identity_substitution_matrix` (deprecated) for creating identity matrices.

To-do 1: Add method `from_msa(TabularMSA, [TreeNode])`, which creates a substitution matrix by calculating the substitution frequencies observed in a multiple sequence alignment, given a phylogenetic tree.

To-do 2: Bundle more matrices, including new ones like VTML, and ones for tree-building, such as JTT, WAG and LG.

***

* [x] I have read the guidelines in [CONTRIBUTING.md](https://github.com/scikit-bio/scikit-bio/blob/master/.github/CONTRIBUTING.md).

* [x] I have documented all public-facing changes in [CHANGELOG.md](https://github.com/scikit-bio/scikit-bio/blob/master/CHANGELOG.md).

* [x] **This pull request includes code, documentation, or other content derived from external source(s).** If this is the case, ensure the external source's license is compatible with scikit-bio's [license](https://github.com/scikit-bio/scikit-bio/blob/master/LICENSE.txt). Include the license in the `licenses` directory and add a comment in the code giving proper attribution. Ensure any other requirements set forth by the license and/or author are satisfied. **It is your responsibility to disclose code, documentation, or other content derived from external source(s).** If you have questions about whether something can be included in the project or how to give proper attribution, include those questions in your pull request and a reviewer will assist you.

**Note:** [REVIEWING.md](https://github.com/scikit-bio/scikit-bio/blob/master/REVIEWING.md) may also be helpful to see some of the things code reviewers will be verifying when reviewing your pull request.


qiyunzhu:
Hi @mortonjt @wasade Any of you interested in taking a look? Thanks!

mortonjt:
Thanks Qiyun.

Any interest in adding in a pytest.fixture to iterate all of the substitution matrices to make sure that they won't break the alignments? If that is in, I'd be convinced that this pull request is good to go.

skbio/alignment/_pairwise.py
```diff
@@ -189,8 +88,8 @@ def local_pairwise_align_nucleotide(seq1, seq2, gap_open_penalty=5,
     # use the substitution matrix provided by the user, or compute from
     # match_score and mismatch_score if a substitution matrix was not provided
     if substitution_matrix is None:
-        substitution_matrix = \
-            make_identity_substitution_matrix(match_score, mismatch_score)
+        substitution_matrix = SubstitutionMatrix.identity(
+            'ACGTU', match_score, mismatch_score).to_dict()
```
wasade, line 92: It doesn't need to happen in this PR, but it would be nice if string literals were represented in stable variables either at the object scope or module scope (e.g., this would change to `'ACGTU'` -> `self.NUCLEOTIDES` or something to that effect. The benefit is often these types of strings are reused, and centralizing them facilitates management and reduces chance of bugs from replication/maintenance errors

skbio/alignment/_pairwise.py
```diff
@@ -258,7 +157,8 @@ def local_pairwise_align_protein(seq1, seq2, gap_open_penalty=11,
                 % type(seq).__name__)
 
     if substitution_matrix is None:
-        substitution_matrix = blosum50
+        substitution_matrix = SubstitutionMatrix.by_name(
+            'BLOSUM50').to_dict()
```
wasade, line 161: Same here re: the use of variables. But again, doesn't need to be in this PR, and may make sense as a separate refactor PR

skbio/alignment/_pairwise.py
```diff
@@ -522,7 +422,8 @@ def global_pairwise_align_protein(seq1, seq2, gap_open_penalty=11,
                 "not dtype %r" % seq.dtype.__name__)
 
     if substitution_matrix is None:
-        substitution_matrix = blosum50
+        substitution_matrix = SubstitutionMatrix.by_name(
+            'BLOSUM50').to_dict()
```
wasade, line 426: Here too

skbio/alignment/tests/test_ssw.py
```diff
@@ -33,6 +32,8 @@ class TestSSW(TestCase):
         "target_sequence"
     ]
 
+    blosum50 = SubstitutionMatrix.by_name('BLOSUM50').to_dict()
```
wasade, line 35: Here too

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+# The full license is in the file LICENSE.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import numpy as np
+
+from skbio.util._decorator import experimental, classonlymethod
+from skbio.stats.distance import DissimilarityMatrix
+
+
+@experimental(as_of='0.5.10')
```
wasade, line 15: Separate from this PR, I wonder about the long term value of these tags. Does anything use them?

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+class SubstitutionMatrix(DissimilarityMatrix):
+    """Matrix for scoring substitutions between characters in biological
+    sequences.
```
wasade, line 18: I'm surprised this made it through lint. PEP8 extends PEP257, which implies use of a single line summary, see [here](https://peps.python.org/pep-0257/#multi-line-docstrings). That said, I think we should definitely switch to `black` or `ruff` and just never talk about style again :)

```suggestion
    """Scoring matrix between characters in biological sequences
```

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+    pre-defined and can be referred to by name. Examples include NUC.4.4 for
+    nucleotides, and variants of BLOSUM and PAM matrices for amino acids.
+
+    ``SubstitutionMatrix`` is a subclass of ``DissimilarityMatrix``. Therefore,
+    all attributes and methods of the latter also apply to the former.
+
+    Examples
+    --------
+    >>> from skbio import SubstitutionMatrix
+    >>> mat = SubstitutionMatrix('ACGT', np.array([
```
wasade, line 65: This string literal, and the one on line 81, might make sense to not be within a variable as it is an example use as opposed to an internal library call. But at least making it possible for a user to rely on a variable would be good as it helps to future proof their code in the event skbio reconsiders e.g. the datatype used

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+# ----------------------------------------------------------------------------
+
+import numpy as np
+
+from skbio.util._decorator import experimental, classonlymethod
+from skbio.stats.distance import DissimilarityMatrix
+
+
+@experimental(as_of='0.5.10')
+class SubstitutionMatrix(DissimilarityMatrix):
```
wasade, line 16: Are there examples of substitution matrices which are not symmetric? If symmetry is expected, then I believe this should use `DistanceMatrix`

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+        -----
+        This is an alias of ``data``.
+
+        """
+        return self._data
+
+    @experimental(as_of='0.5.10')
+    def __init__(self, alphabet, scores, **kwargs):
+        """Initialize a substitution matrix object with an alphabet and a score
+        matrix.
```
wasade, line 127:
```suggestion
        """Initialize a substitution matrix object
```

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+        """
+        super().__init__(scores, alphabet, **kwargs)
+
+    @experimental(as_of='0.5.10')
+    def to_dict(self):
```
wasade, line 132: I somewhat expected this to return `{(row, col): value}` rather than 2D? No strong opinion though

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+        >>> mat = SubstitutionMatrix.from_dict(d)
+        >>> mat.alphabet
+        ('a', 'b', 'c')
+        >>> mat.scores
+        array([[ 1.,  0.,  0.],
+               [ 0.,  1.,  0.],
+               [ 0.,  0.,  1.]])
+
+        """
+        alphabet = tuple(dictionary.keys())
```
wasade, line 182:
```suggestion
        alphabet = tuple(dictionary)
```

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+        alphabet_set = set(alphabet)
+        idmap = {x: i for i, x in enumerate(alphabet)}
+        scores = np.zeros((n := len(alphabet), n))
+        for i, row in enumerate(dictionary.values()):
+            if set(row.keys()) != alphabet_set:
```
wasade, line 187:
```suggestion
            if set(row) != alphabet_set:
```

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+        array([[ 1.,  0.,  0.],
+               [ 0.,  1.,  0.],
+               [ 0.,  0.,  1.]])
+
+        """
+        alphabet = tuple(dictionary.keys())
+        alphabet_set = set(alphabet)
+        idmap = {x: i for i, x in enumerate(alphabet)}
+        scores = np.zeros((n := len(alphabet), n))
+        for i, row in enumerate(dictionary.values()):
```
wasade, line 186: This implicitly assumes Python will ensure the order of evaluation of `.values` and `.keys` is stable and consistent. While it's true, this may be unexpected for developers unfamiliar with Python (or for those familiar only with old versions of Python). It may be pleasant to be explicit, use `.items`, and pull index directly from the `idmap` on line 196

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+            if set(row.keys()) != alphabet_set:
+                raise ValueError('The outer and inner layers of the dictionary'
+                                 ' must have the same set of keys.')
+            for key, value in row.items():
+                if isinstance(value, int):
+                    value = float(value)
+                elif not isinstance(value, float):
+                    raise ValueError('Scores must be integers or floating-'
```
wasade, line 194: I don't think this check or the preceding is necessary:

```python
In [5]: skbio.DistanceMatrix(np.array([['A','B'],['B','A']]), ['foo', 'bar'])
Cell In[5], line 1
----> 1 skbio.DistanceMatrix(np.array([['A','B'],['B','A']]), ['foo', 'bar'])

File ~/miniconda3/envs/qiime2-amplicon-2023.9/lib/python3.8/site-packages/skbio/stats/distance/_base.py:136, in DissimilarityMatrix.__init__(self, data, ids, validate)
    133         _issue_copy = False
    135 if _issue_copy:
--> 136     data = np.asarray(data, dtype='float')
    138 if data.ndim == 1:
    139     # We can assume squareform will return a symmetric square matrix
    140     # so no need for full validation.
    141     # Still do basic checks (e.g. zero length)
    142     # and id validation
    143     data = squareform(data, force='tomatrix', checks=False)

ValueError: could not convert string to float: 'A'
```

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+        >>> from skbio import SubstitutionMatrix
+        >>> mat = SubstitutionMatrix.by_name('BLOSUM62')
+        >>> len(mat.alphabet)
+        24
+        >>> mat['M', 'K']
+        -1.0
+
+        """
+        try:
+            return named_substitution_matrices[name]
```
wasade, line 296: Normalizing case would probably dampen some user frustration, e.g. `name.lower()` and ensuring `named_substitution_matrices` supports both lower, upper and native casing

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+        Returns
+        -------
+        list of str
+            Names of pre-defined substitution matrices.
+
+        See Also
+        --------
+        by_name
+        """
+        return list(named_substitution_matrices.keys())
```
wasade, line 314:
```suggestion
        return sorted(named_substitution_matrices)
```

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+
+
+def _matrix_to_vector(mat):
+    """Convert a square matrix into a flattened vector consisting of the upper
+    triangle and the diagonal.
```
wasade, line 319:
```suggestion
    """Flatten a matrix to a vector of the upper triangle and diagonal.
```

This is very similar to condensed form availalbe in scipy?

Note the code does not assert the matrix is square

wasade (changes requested):
Thanks, @qiyunzhu!!! A few minor comments and suggestions, and a couple of notes for items that I think would be valuable to consider as a separate PR or two

qiyunzhu:
@wasade (and @mortonjt ) thanks for reviewing my code! I am sorry for the late response and I had to deal with quite a few things in the beginning of the semester. Now I am back to skbio development.

skbio/alignment/_pairwise.py
qiyunzhu: This is a good suggestion. Actually, the original skbio codebase already has it. It is the `definite_chars` attribute of each sequence type. For example see [here](https://github.com/scikit-bio/scikit-bio/blob/8a72c20c6ad50f9cb6d44f5fcf87be5f9582de58/skbio/sequence/_dna.py#L166C29-L166C29). Note: I made the edit here only temporarily, as the entire Python implementation of alignment algorithms will be removed and replaced soon (maybe in the next 2-3 PRs).

skbio/alignment/_pairwise.py
qiyunzhu: The same as above.

skbio/sequence/_substitution.py
qiyunzhu: The documentation displays these notes. Back in days the team established a rigorous life cycle mechanism. See below a screenshot from this [video](https://www.youtube.com/watch?v=ZpgkRQooGqo). (I wasn't there so this is just my understanding.) I tend to this that this mechanism is beneficial. Other packages also have similar things.

image

skbio/sequence/_substitution.py
qiyunzhu: Great to know! I wasn't aware of it. I will try to make single-line summaries in the future.

skbio/sequence/_substitution.py
qiyunzhu: The common matrices provided by NCBI are all symmetric (I checked). However, asymmetric substitution matrices do exist, although rare (see [here](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC7586916/) for examples). There could be a reason why NCBI matrices document both upper and lower triangles, and why other packages like biotite support asymmetric matrices (see [here](https://www.biotite-python.org/apidoc/biotite.sequence.align.SubstitutionMatrix.html?highlight=substitutionmatrix#biotite.sequence.align.SubstitutionMatrix.is_symmetric)). I think that one advantage of allowing asymmetric matrices is that the researchers can test specific hypotheses of evolution and mutation in a highly controlled setting (e.g., positively charged amino acids have a selective advantage at a catalytic site, therefore one a site become positively charged, it won't go back).

 Some other packages

skbio/sequence/_substitution.py
qiyunzhu: The current behavior of `to_dict` is consistent with the original approach in skbio (although I think it is bit redundant as one can simply do `mat['M', 'K']` to access a value). Besides, I tend to think that using tuples as dictionary keys, despite valid (as tuples are hashables), may cause confusions to Python beginners.

skbio/sequence/_substitution.py
qiyunzhu: This is a good point. Let me modify.

skbio/sequence/_substitution.py
qiyunzhu: Good point. Let me modify.

skbio/sequence/_substitution.py
qiyunzhu: Re: Normalizing case: Good point! Let me edit.

Re: Global scope. I noted this. At present other global variables are not upper case (for example see [genetic codes](https://github.com/scikit-bio/scikit-bio/blob/8a72c20c6ad50f9cb6d44f5fcf87be5f9582de58/skbio/sequence/_genetic_code.py#L761)). Therefore I followed the convention. But I agree that using upper case for global variables is the way to go. How about we create a separate PR to change all such variables into upper case?

skbio/sequence/_substitution.py
qiyunzhu: Hmmm... I am not sure here. The original order has some meaning. After sorting it could be less intuitive. For example, `PAM250` will be on top of `PAM45`.

skbio/sequence/_substitution.py
wasade: What other packages are doing something like this? My concern is that while this did get setup, I don't recall seeing any PRs over the history of the project which actually changed things associated with these decorators nor am I aware of consumers of the library using this information (though I may be wrong there?)

skbio/sequence/_substitution.py
qiyunzhu: Yes. However SciPy's `squareform` only supports matrices with zero diagonal. Therefore I had to write the code by hand.

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+    """
+    return mat[np.triu_indices(len(mat))]
+
+
+def _vector_to_matrix(vec):
+    """Convert a vector consisting of the upper triangle and the diagonal of a
```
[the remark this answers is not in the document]
qiyunzhu: See my comment above. SciPy can't handle this... Maybe we should consider making ours a public API for users with this need.

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+    matrix to square form.
+    """
+    n = int((np.sqrt(1 + 8 * len(vec)) - 1) / 2)
```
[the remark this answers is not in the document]
qiyunzhu: Sure! Let me do.

skbio/sequence/_substitution.py
wasade: Perhaps though nested `dict`s are also quite complex. As noted though, no strong opinion

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+    mat = np.zeros((n, n))
+    mat[np.triu_indices(n)] = vec
```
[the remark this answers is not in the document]
qiyunzhu: Good point. Let me do.

skbio/sequence/_substitution.py
wasade: The genetic code example links to a variable that is described as module private as it is prefixed with '_'. If the use of `named_substitution_matrices` is similarly intended to be module private then perhaps following that form is best. (Or we create an issue to refactor genetic codes)

skbio/sequence/_substitution.py
wasade: A natural sort could be used. Stability in order is pleasant for testing and developer exploration

skbio/sequence/_substitution.py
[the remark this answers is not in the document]
wasade: :+1: and it may be beneficial to open an issue with SciPy, linking to the methods here, seeing if their maintainers would accept this variant in a PR

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+    return mat + np.triu(mat, k=1).T
+
+
+# Defined according to the matrices hosted at the NCBI FTP server:
+# https://ftp.ncbi.nlm.nih.gov/blast/matrices/
+named_substitution_matrices = {
```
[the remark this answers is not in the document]
qiyunzhu: I like the idea of lazy loading. I am not familiar with the best approaches to achieve this. Some other packages simply store the matrices as external files. I wonder if there is a better approach? It may be the greatest if you can show an example.

skbio/sequence/_substitution.py
```diff
@@ -0,0 +1,518 @@
...
+
+    # NUC.4.4, a.k.a. DNAfull
+    'NUC.4.4': SubstitutionMatrix(
+        'ATGCSWRYKMBVHDN', _vector_to_matrix(np.array([
```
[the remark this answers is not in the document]
qiyunzhu: I though a bit but am not sure what's the best strategy to store the matrix data. One way is to keep them as external files (like [this one](https://ftp.ncbi.nlm.nih.gov/blast/matrices/BLOSUM62)). If we keep the data in the Python script, we cannot make the values look nice anyway because each row will exceed 80 characters and we have to break them into multiple lines.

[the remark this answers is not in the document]
wasade: `# noqa` can always be added to the end of the lines. But I really like the idea of just storing or downloading the files, as we can assert they are identical to what's provided by ncbi

skbio/sequence/_substitution.py
qiyunzhu: I don't quite understand this comment... Could you explain a bit?

skbio/sequence/_substitution.py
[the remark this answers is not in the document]
wasade: Something like the following, and would preserve dict-like interaction?

```python
class _LazySubstitutionMatrix:
    def __init__(self, name, path):
        self.name = name
        self.path = path
        self._data = None
    
    @property
    def data(self):
        if self._data is None:
            self._data = _some_mechanism_to_load_data(self.path)
        return self._data

class _SubstitutionMatrices:
    _matrices = (_LazySubstitutionMatrix('NUC.4.4', 'file://foo/bar'),
                 _LazySubstitutionMatrix('PAM30', http://foo/bar'), 
                 ...)
    def __init__(self):
        self._lazy_data = {m.name: m for m in self._matrices}
     
    def __get__(self, k):
        return self._lazy_data[k].data

    def keys(self):
        return sorted(self._lazy_data)

    def values(self):
        return [self[k] for k in self.keys()]

    def items(self):
        return [(k, self[k]) for k in self.keys()]

named_substitution_matrices = _SubstitutionMatrices()
```

skbio/sequence/_substitution.py
wasade: doing `>>> mat = SubstitutionMatrix(DNA_ALPHABET, ...)` or something to that effect

skbio/sequence/_substitution.py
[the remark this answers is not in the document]
qiyunzhu: At this moment, skbio cannot directly parse the matrix files correctly. I added this functionality (see #1912 ) but it is not perfect yet. How about we keep the current solution (in Python code) for now, and have a separate discussion on how to store external data in skbio? I envision that such a mechanism will be highly useful (like `load_iris` in some other Python/R packages).

skbio/sequence/_substitution.py
qiyunzhu: I echo your questions. Here are some examples in [Pandas](https://github.com/pandas-dev/pandas/blob/a671b5a8bf5dd13fb19f0e88edc679bc9e15c673/pandas/io/parsers/readers.py#L171) (and [here](https://github.com/pandas-dev/pandas/blob/a671b5a8bf5dd13fb19f0e88edc679bc9e15c673/pandas/core/frame.py#L12212)) and [scikit-learn](https://github.com/scikit-learn/scikit-learn/blob/3f89022fa/sklearn/decomposition/_pca.py#L199). Instead of adding decorators, they achieve this in the documentation only. We can have a discussion on how to deal with the decorators.

wasade: Deprecation warnings, and noting versions added aren't a bad idea but qualititively different from these decorators

skbio/sequence/_substitution.py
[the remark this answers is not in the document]
wasade: Makes sense. And also like `biom.example_table`... having easy to get data is awesome

qiyunzhu:
@mortonjt Thanks for your feedback! I added unit tests with nucleotide and protein matrices in the last commit. I didn't do it for all matrices though, as the alignment functions will be removed and replaced, and it will make more sense if we test on the new functions.

skbio/sequence/_substitution.py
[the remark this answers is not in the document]
qiyunzhu: Relevant to the other comment. Will put this on hold pending the discussion on external data.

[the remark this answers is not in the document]
qiyunzhu: PS @wasade : If we directly type data in `_LazySubstitutionMatrix` (instead of external files), will it still suffice lazy loading?

skbio/sequence/_substitution.py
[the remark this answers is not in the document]
qiyunzhu: Done.

skbio/sequence/_substitution.py
qiyunzhu: Added normalization. Now `blosum62` and `BLOSUM62` both work. Let me also add `_` before `named_substitution_matrices`.

skbio/sequence/_substitution.py
qiyunzhu: Got it. I will keep this in my note. A tricky thing is that established substitution matrices have different alphabets. They diverge on whether to include degenerate codes and which degenerate codes to include. Some mechanism needs to be implemented in skbio to handle characters not found in the alphabet. For example, replacing `R` in a nucleotide sequence with `N`.

skbio/sequence/_substitution.py
qiyunzhu: I agree with you on this.

skbio/sequence/_substitution.py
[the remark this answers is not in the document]
wasade: You could defer the calls to `_vector_to_matrix` for example but bigger bang may be until these can be pulled from an authoritative source

skbio/sequence/_substitution.py
wasade: Ah ya, that makes sense

skbio/sequence/_substitution.py
qiyunzhu: Natural sort should solve part of the problem. However in some instances, matrix names form a logical order (e.g., from simple to complex), which cannot be handled by natural sort. An example is JC69 -> K80 -> HKY -> GTR (ordered by number of parameters from few to many).

wasade: Okay. Rephrasing :) having a stable rational order would be pleasant, however that is achieved

skbio/sequence/_substitution.py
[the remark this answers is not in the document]
qiyunzhu: @wasade I submitted an issue in SciPy: https://github.com/scipy/scipy/issues/19875

qiyunzhu:
@wasade Thanks again for such careful review! I think I have addressed all the comments you made. Would you like to take a look?

skbio/sequence/_substitution.py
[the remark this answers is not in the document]
qiyunzhu: I envision that skbio should have a local data repository that is distributed with the package, plus a remote server that stores larger datasets. Analogs are scikit-learn's small, local datasets and pyTorch's big, remote datasets.

For substitution matrices specifically, the [NCBI repo](https://ftp.ncbi.nlm.nih.gov/blast/matrices/) has many common matrices, but not all. These are only for alignments. Meanwhile, matrices for tree building are not there. Examples are JTT, WAG, LG, etc. In comparison, BioPython has some matrices that are not found in NCBI. For skbio, I think that having a local repo of most common matrices (like the ones I already typed) and a remote repo of extended ones may be the way to go.

skbio/sequence/_substitution.py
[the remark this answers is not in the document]
qiyunzhu: Done.

skbio/sequence/_substitution.py
wasade: Looking at these changes now. The solution works (ie no concern if this is merged as is), though this would be more direct rather than a search. Alternatively, the keys in _named_substitition_matrices could be forced to a specific case and then the check can be readily normalized. 

```python
if name in _named_substitition_matrices:
    return _named_substitition_matrices[name]
elif name.upper() in _named_substitition_matrices:
    return _named_substitition_matrices[name.upper()]
elif name.lower() in _named_substitition_matrices:
    return _named_substitition_matrices[name.lower()]
else:
    raise KeyError(f"Unrecognized substititution matrix {name}")
```

skbio/sequence/_substitution.py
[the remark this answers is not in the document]
wasade: Sounds good! Could an issue be opened to capture this? It could refer back to the discussion here

wasade (approved):

qiyunzhu:
@wasade Thanks and happy weekend!

PR #1949, opened 2024-02-23
Changed Matplotlib setting to unblock use on systems without graphics

qiyunzhu:

* [ ] I have documented all public-facing changes in the [changelog](https://github.com/scikit-bio/scikit-bio/blob/master/CHANGELOG.md).

* [ ] **This pull request includes code, documentation, or other content derived from external source(s).** If this is the case, ensure the external source's license is compatible with scikit-bio's [license](https://github.com/scikit-bio/scikit-bio/blob/master/LICENSE.txt). Include the license in the `licenses` directory and add a comment in the code giving proper attribution. Ensure any other requirements set forth by the license and/or author are satisfied.


qiyunzhu:
@mortonjt @wasade We noticed that `make test` doesn't work in one student's computer (@christapo). That computer has WSL, which looks fine to us, but it somehow doesn't have graphics enabled. Therefore, the backend of matplotlib has to be changed into `agg` prior to running `make test`. While this can be done via `export MPLBACKEND=Agg`, as was done in `ci.yml`, I think that a more general solution is to let `skbio.test` do this within Python.

mataton:
@qiyunzhu Merging now per your request. This looks good to me.

Issue #1920, opened 2024-01-24
Tokenizer compatibility with existing aligners / language models


mortonjt:
As discussed in #1917 it maybe worthwhile added additional conversion functionality in order to add support to interface with existing alignment and protein language models.

See below for a more complete discussion
https://github.com/scikit-bio/scikit-bio/pull/1917#issuecomment-1908382484

PR #1952, opened 2024-02-23
WIP: Sequence character conversion

mataton:
This function will enable conversion of non-canonical or degenerate characters to characters of the user's choosing. Options for replacement include:

- wildcard:  replace the degenerate/non-canonical characters with 'N' or 'X'
- gap:  replace with gap character
- char:  replace with a single letter string
- trim:  remove all degenerate/non-canonical characters
- raise:  raise an error

This function is a step in the direction of tokenization support for language models, as outlined in issue #1920. The rationale is that skbio needs to be able to convert sequences into alphabets which match those used by the language models. For instance, [ProTrans](https://huggingface.co/Rostlab/prot_t5_xl_uniref50) converted the amino acids `['B', 'U', 'Z', 'O']` to `'X'` prior to tokenization.

@qiyunzhu The docstring needs work, and I'll need to write a unittest, update the `CHANGELOG.md`, and finish the 'raise' option, but what are your thoughts as it is?

* [ ] I have documented all public-facing changes in the [changelog](https://github.com/scikit-bio/scikit-bio/blob/master/CHANGELOG.md).

* [ ] **This pull request includes code, documentation, or other content derived from external source(s).** If this is the case, ensure the external source's license is compatible with scikit-bio's [license](https://github.com/scikit-bio/scikit-bio/blob/master/LICENSE.txt). Include the license in the `licenses` directory and add a comment in the code giving proper attribution. Ensure any other requirements set forth by the license and/or author are satisfied.

### DIFF
--- a/skbio/sequence/_grammared_sequence.py
+++ b/skbio/sequence/_grammared_sequence.py
@@ -749,6 +749,34 @@ class GrammaredSequence(Sequence, metaclass=GrammaredSequenceMeta):
 
         return re.compile(regex_string)
 
+    def to_definites(self, degenerate="wild", char=None):
+        """Convert degenerate or non-canonical characters to definite characters."""
+        sequence = str(self)
+        # print(f"original sequence: {sequence}")
+        # print(len(sequence))
+
+        if degenerate == "wild":
+            for degenerate_char in self.degenerate_chars:
+                sequence = sequence.replace(
+                    degenerate_char, "".join(self.wildcard_char)
+                )
+        elif degenerate == "gap":
+            for degenerate_char in self.degenerate_chars:
+                sequence = sequence.replace(
+                    degenerate_char, "".join(self.default_gap_char)
+                )
+        elif degenerate == "char":
+            for degenerate_char in self.degenerate_chars:
+                sequence = sequence.replace(degenerate_char, "".join(char))
+        # print(f"modified sequence: {sequence}")
+        # print(len(sequence))
+
+        return self._constructor(
+            sequence=sequence,
+            metadata=self.metadata,
+            positional_metadata=self.positional_metadata,
+        )
+
     @stable(as_of="0.4.0")
     def find_motifs(self, motif_type, min_length=1, ignore=None):
         """Search the biological sequence for motifs.
--- a/skbio/sequence/_grammared_sequence.py
+++ b/skbio/sequence/_grammared_sequence.py
@@ -750,32 +750,48 @@ class GrammaredSequence(Sequence, metaclass=GrammaredSequenceMeta):
         return re.compile(regex_string)
 
     def to_definites(self, degenerate="wild", char=None):
-        """Convert degenerate or non-canonical characters to definite characters."""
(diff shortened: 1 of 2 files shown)


mataton:
I think it makes sense to change the function name to something like `to_character`, instead of `to_definite` because the function returns sequences with non-definite characters. What do you think @qiyunzhu?

qiyunzhu:
@mataton Thank you! This is wonderful. I will read and comment. Let me also call @mortonjt who is knowledgeable in this.

skbio/sequence/_grammared_sequence.py
```diff
@@ -749,6 +761,57 @@ def to_regex(self, within_capture=False):
 
         return re.compile(regex_string)
 
+    def to_definites(self, degenerate="wild", char=None, canonical=True):
```
qiyunzhu, line 764: @mataton This is impressive effort! Thanks! I wanted to make a comment on the algorithm. The current code iterates over all degenerate characters, and runs `str.replace` for each of them. There may be room to optimize this. For example, you can use regular expression:

```python
import re
pattern = "[{}]".format("".join(map(re.escape, degenerates)))
sequence = pattern.sub(char, sequence)
```

Or you can use list comprehension:

```python
sequence = "".join(char if x in degenerates else x for x in sequence)
```

I don't know which one is faster. I guess regex should win. But one should do some tests to make sure.

***

Also, you wrote the code block for replacement multiple times. This can be saved with:

```python
if degenerate == "wild":
    sub_char = self.wildcard_char
elif degenerate == "gap":
    sub_char = self.default_gap_char
elif ...
for degenerate_char in degenerates:
    sequence = sequence.replace(
        degenerate_char, sub_char
    )
```

mataton: @qiyunzhu Thank you for the feedback! Surprisingly it actually looks like the list comprehension is faster. Because it is faster and also easier to read, I think we should go with the list comprehension option. 

Screenshot 2024-02-23 170047

Also, I found an error with your regex solution. It needed `re.compile` at the beginning of it to avoid trying to run `sub()` on a `string` object. The version I implemented looked like this:

```python
import re
pattern = re.compile("[{}]".format("".join(map(re.escape, degenerates))))
sequence = pattern.sub(char, sequence)
```

qiyunzhu: Good catch! And thanks for benchmarking! I agree with you that list comprehension should be the solution. Actually, I think there are faster methods. For example, once can pre-generate the regex pattern and store it as a class property.  For another example, one can directly operate on `sequence._bytes` using some vectorized operation like `np.where(np.isin(seq._bytes, seq.definites), seq._bytes, sub_char)`. This could be faster. However, I am not sure if this step can be a bottleneck. List comprehension is the concise and easy-to-understand solution.

qiyunzhu (approved):
@mataton This is a well-structured and carefully considered functionality. Thanks! Let me merge. There are a few minor things left, such as you didn't mention this in the documentation header or in the changelog. But let's tackle that later.

PR #1955, opened 2024-02-24
Skipping Matplotlib during test if not importable

qiyunzhu:
Matplotlib was made an optional dependency. However, it is still required during unittest and doctest. This is not a problem if a developer installed the dependencies in `requirements.test.txt`. However, it could be a problem for regular users, who installed scikit-bio with only mandatory dependencies, and execute `python -m skbio.test` to check if the package is working properly.

This PR makes the unittests that involves Matplotlib conditional: they are only executed when Matplotlib is importable.

Additionally, it skips the doctests that involves Matplotlib. Those lines were not tested anyway, therefore skipping them won't cost anything. (Note: in the Sphinx-rendered documentation, the plots are still there.) However, it is not easy to make these tests conditional. Python's `doctest` module does not have conditional skipping.

There may be hacks though. [This answer](https://stackoverflow.com/questions/42708809/python-doctest-skip-a-test-conditionally) provides some inspiration. In the future, it will be good to implement such a mechanism, such that the examples in the documentation can include other optional dependencies. (For example, we can show that skbio is compatible with both pandas and polars, matplotlib and plotly, numpy and cupy, etc.)


mataton (approved):
@qiyunzhu I built the docs and ran tests locally. This all looks good to me.

PR #1956, opened 2024-02-26
Dirichilet Multinomial differential abundance method

mortonjt:
Addresses https://github.com/scikit-bio/scikit-bio/issues/1948

This includes the simplest sane differential abundance method. The purpose of this methodology is to create an initial differential abundance methodology to serve as a placeholder for future differential abundance methods.  ANCOM is great, but falls short since it doesn't provide directionality (i.e. doesn't tell you if a feature increased or decreased in abundance).  

Underneath the hood, the Dirichilet Multinomial approach will create posterior distributions for the taxa abundances in each biospecimen and perform a statistical test on the resulting posterior draws. This reflects the [ALDEx2 methodology](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4030730/) plus and minus some features (i.e. we aren't computing IQR reference frames or effect sizes as done in ALDEx2, but we are reporting confidence intervals, which isn't currently reported by ALDEx2 by default). The one thing that is worth highlighting is the why which we compute aggregate statistics across all of the posterior draws. For most of the statistics (i.e. log-fold change), we report the average of all of the statistics computed for each posterior draw.  But for  the confidence intervals, we compute the minimum / maximum.  This will make the confidence intervals much more conservative, but (I'd argue) the more desirable quantities to work with, since it will be more robust to low-sequencing depth noise.

Once this is in, we can discuss adding in [generalized linear models](https://www.statsmodels.org/stable/glm.html) and [linear mixed effects](https://www.statsmodels.org/stable/mixed_linear.html) to provide barebones, but a feature complete differential abundance framework that would provide the API infrastructure and formats to onboard other differential abundance methodologies.


qiyunzhu:
@mortonjt Maybe pull upstream main then push? We recently added coverage test.

skbio/stats/composition.py
```diff
@@ -1701,3 +1710,203 @@ def _check_orthogonality(basis):
...
+    res = [_welch_ttest(np.array(table.loc[trt_group.index, x].values),
+                        np.array(table.loc[ref_group.index, x].values))
+           for x in table.columns]
+    res = pd.concat(res)
+
+    for i in range(1, draws):
+        prior = np.random.dirichlet(np.ones(table.shape[1]) + pseudocount,
+                                    size=table.shape[0])
+        dir_table = pd.DataFrame(clr(table.values + prior),
+                                 index=table.index, columns=table.columns)
```
mortonjt, line 1878: I just realized that I got the order of operations backwards.

It should be 
```python
        dir_table = pd.DataFrame(clr([np.random.dirichlet(table.values[j]  + pseudocount) for j in range(table.shape[0]),
                                 index=table.index, columns=table.columns)
```
That way, the uncertainty would factor in the sequencing depth (which will shrink the confidence interval with larger sequencing depths).

Probably should get another unittest to confirm this.

skbio/stats/composition.py
```diff
@@ -1701,3 +1710,203 @@ def _check_orthogonality(basis):
...
+                             index=table.index, columns=table.columns)
+
+    res = [_welch_ttest(np.array(table.loc[trt_group.index, x].values),
+                        np.array(table.loc[ref_group.index, x].values))
+           for x in table.columns]
+    res = pd.concat(res)
+
+    for i in range(1, draws):
+        prior = np.random.dirichlet(np.ones(table.shape[1]) + pseudocount,
+                                    size=table.shape[0])
```
mortonjt, line 1876: we'll ditch this prior for the refactor below

mortonjt:
Hi @qiyunzhu I found one issue that needs to be resolved first. But yes I need to figure out the merge and recommit hooks

qiyunzhu:
@mortonjt I am not sure if we should introduce statsmodels as a dependency only for doing FDR correction. The Benjamini-Hochberg method can be implemented simply with the following code, as suggested by Vladimir in this [answer](https://stackoverflow.com/questions/7450957/how-to-implement-rs-p-adjust-in-python). In my opinion it is more efficient than statsmodels'.

```
def adjust_pvals(pvals):
    n = pvals.size
    order = pvals.argsort()[::-1]
    qvals = pvals[order] * n / np.arange(n, 0, -1)
    qvals = np.minimum(1, np.minimum.accumulate(qvals))
    return qvals[order.argsort()]
```

skbio/stats/composition.py
```diff
@@ -1701,3 +1710,203 @@ def _check_orthogonality(basis):
     basis = np.atleast_2d(basis)
     if not np.allclose(basis @ basis.T, np.identity(len(basis)), rtol=1e-4, atol=1e-6):
         raise ValueError("Basis is not orthonormal")
+
+
+def _welch_ttest(x1, x2):
```
qiyunzhu, line 1715: I thought Welch's _t_-test is already implemented in SciPy, as: `ttest_ind(x, y, equal_var=False)`. But I could be wrong.

mortonjt:
Yes, but it doesnt return confidence intervals, which is critical.

mortonjt:
I’m OK deferring the statsmodels port for now

skbio/stats/composition.py
```diff
@@ -1701,3 +1710,203 @@ def _check_orthogonality(basis):
...
+
+    table_index_len = len(table.index)
+    grouping_index_len = len(grouping.index)
+    mat, cats = table.align(grouping, axis=0, join="inner")
+    if len(mat) != table_index_len or len(cats) != grouping_index_len:
+        raise ValueError("`table` index and `grouping` " "index must be consistent.")
+
+    trt_group = grouping.loc[grouping == treatment]
+    ref_group = grouping.loc[grouping == reference]
+    prior = np.random.dirichlet(np.ones(table.shape[1]) + pseudocount,
```
qiyunzhu, line 1864: Now we can use the new NumPy [random generator](https://numpy.org/doc/stable/reference/random/generator.html) for these. I implemented this in scikit-bio recently (#1888). An example usage is here (#1889). Sorry I forgot to add that to the documentation (I wasn't familiar with RST at that time). You didn't either. To add a function to the documentation you will need to manually add its name to the header of a script file, namely [here](https://github.com/scikit-bio/scikit-bio/blob/72104932830ace439ed64c298ce354264e23c75c/skbio/stats/composition.py#L43)

mortonjt: done

skbio/stats/composition.py
```diff
@@ -1701,3 +1710,203 @@ def _check_orthogonality(basis):
...
+    FDR-corrected pvalues use Benjamini-Hochberg for pvalue adjustment for
+    multiple corrections. The confidence intervals are computed using the
+    mininum 2.5% and maximum 95% bounds computed across all of the posterior draws.
+
+
+    The reference frame here is the geometric mean. Extracting absolute log
+    fold changes from this test assumes that the average microbial abundance
+    between the `treatment` and the `reference` groups are the same. If this
+    assumption is violated, then the log-fold changes will be biased, and the
+    pvalues will not be reliable. However, the bias is the same across each
```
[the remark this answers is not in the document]
mortonjt: no, this was incomplete.  It is the same across the features.

skbio/stats/composition.py
```diff
@@ -1701,3 +1711,283 @@ def _check_orthogonality(basis):
...
+
+@experimental(as_of="0.5.9")
+def dirmult_ttest(
+    table: pd.DataFrame,
+    grouping: str,
+    treatment: str,
+    reference: str,
+    pseudocount: float = 0.5,
+    draws: int = 128,
+    seed=0,
```
qiyunzhu, line 1754: The default value for `seed` can be `None`. See [numpy doc](https://numpy.org/doc/stable/reference/random/generator.html). See also skbio function [`get_rng`](https://github.com/scikit-bio/scikit-bio/blob/72104932830ace439ed64c298ce354264e23c75c/skbio/util/_misc.py#L217).

setup.py
```diff
@@ -241,6 +241,7 @@ def check_bin(ccbin, source, allow_dash):
         "scipy >= 1.9.0",
         "h5py >= 3.6.0",
         "hdmedians >= 0.14.1",
+        "statsmodels >= 0.10.0"
```
qiyunzhu, line 244: Given that we had consensus in the previous discussion, how about dropping this line and adding a `_benjamini_hochberg` helper function [along with this](https://github.com/scikit-bio/scikit-bio/blob/72104932830ace439ed64c298ce354264e23c75c/skbio/stats/composition.py#L1530)?

skbio/stats/tests/test_composition.py
```diff
@@ -1305,5 +1306,144 @@ def test_pairwise_vlr(self):
...
+
+        d = 50
+        n = 200
+        self.depth = depth = 1000
+        p1 = np.random.lognormal(0, 1, size=d) * 10
+        p2 = np.random.lognormal(0.01, 1, size=d) * 10
+        self.p1, self.p2 = p1 / p1.sum(), p2 / p2.sum()
+        self.data2 = np.vstack(
+            (
+                [np.random.multinomial(depth, self.p1) for _ in range(n)],
```
qiyunzhu, line 1331: May consider swapping those with `rng` too. Not important though, if you are not testing the generated values.

skbio/stats/composition.py
```diff
@@ -1701,3 +1711,283 @@ def _check_orthogonality(basis):
...
+            "`grouping` must be a `pd.Series`," " not %r." % type(grouping).__name__
+        )
+
+    if np.any(table < 0):
+        raise ValueError("Cannot handle negative values in `table`. ")
+
+    if (grouping.isnull()).any():
+        raise ValueError("Cannot handle missing values in `grouping`.")
+
+    if (table.isnull()).any().any():
```
qiyunzhu, line 1916: I thought `ndarray.any` without parameter will operate on all dimensions. See [here](https://numpy.org/doc/stable/reference/generated/numpy.any.html). But better double check,

setup.py
mortonjt: I've removed the statsmodels dependency. I've decided to use the `_holm_boniferroni` correction instead, since it is already implemented in skbio

skbio/stats/composition.py
mortonjt: this is copied from the `ancom` method. I suggest keeping this as is for consistency - we can always refactor later.

mortonjt:
It looks like not all of the tests were passing (possibly due to random numerical issues). I'm adding back the fudge factor

qiyunzhu:
> It looks like not all of the tests were passing (possibly due to random numerical issues). I'm adding back the fudge factor

Also there is a documentation building error, likely because something in the `Reference` section is not cited in previous sections.

qiyunzhu:
@mortonjt Looks good now! Shall we merge once all tests pass?

mortonjt:
Hi yes I think this is ready to merge

PR #1902, opened 2023-11-20
Replacing observed OTUs with richness

qiyunzhu:
Please complete the following checklist:

The metric `observed_otus` may be outdated, and was replaced by `observed_features` in QIIME 2. However it is still `observed_otus` in scikit-bio. I think that the most accurate term in community ecology is "observed species richness". It is often denoted as $S_{obs}$ or $S$ in literature.

Therefore I am suggesting a new metric `sobs` to indicate "observed species richness". The word "sobs", although brief, is consistent with the naming convention of current alpha diversity metrics, which usually consists of human names, mathematical symbols and abbreviations (such as `simpson_d`, `fisher_alpha`, `ace`). Whereas a name that describes the nature of the metric may be too long. For example, `observed_richness` is a good descriptive name, but may not be ideal since the metric is the simplest one of all.

For reference: In vegan, this metric can be calculated using the function `specnumber` (meaning species number), and it is denoted as `S.obs` in the output of `estimateR`.

Additionally, I added `observed_features` to be consistent with QIIME 2, and deprecated `observed_otus`. In this PR, the three metrics: `observed_otus`, `observed_features` and `sobs` are aliases.

***

* [x] I have read the guidelines in [CONTRIBUTING.md](https://github.com/biocore/scikit-bio/blob/master/.github/CONTRIBUTING.md).

* [x] I have documented all public-facing changes in [CHANGELOG.md](https://github.com/biocore/scikit-bio/blob/master/CHANGELOG.md).

* [ ] **This pull request includes code, documentation, or other content derived from external source(s).** If this is the case, ensure the external source's license is compatible with scikit-bio's [license](https://github.com/biocore/scikit-bio/blob/master/LICENSE.txt). Include the license in the `licenses` directory and add a comment in the code giving proper attribution. Ensure any other requirements set forth by the license and/or author are satisfied. **It is your responsibility to disclose code, documentation, or other content derived from external source(s).** If you have questions about whether something can be included in the project or how to give proper attribution, include those questions in your pull request and a reviewer will assist you.

* [x] **This pull request does not include code, documentation, or other content derived from external source(s).**

**Note:** [REVIEWING.md](https://github.com/biocore/scikit-bio/blob/master/REVIEWING.md) may also be helpful to see some of the things code reviewers will be verifying when reviewing your pull request.

PR #1965, opened 2024-03-09
Deprecated "OTU" in favor of "taxon"

qiyunzhu:
This PR removes the outdated term "OTU" from both documentation and code, and replaces it with the more generic term "taxon". As a consequence, the parameter `otu_ids` now becomes `taxa`. This affects `faith_pd`, `phydiv`, `unweighted_unifrac` and `weighted_unifrac`. The parameter `otu_ids` is kept as an alias of `taxa`, such that existing code that uses these functions will not break. A deprecation warning is displayed in the documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@
 
 ### Miscellaneous
 
+* Replaced the historical term "OTU" with the more generic term "taxon" (plural: "taxa"). As a consequence, the parameter "otu_ids" in phylogenetic alpha and beta diversity metrics was replaced by "taxa". Meanwhile,
+the old parameter "otu_ids" is still kept as an alias of "taxa" for backward compatibility. However it will be removed in a future release.
 * Revised contributor's guidelines.
 * Enabled code coverage reporting via Codecov. See [#1954](https://github.com/scikit-bio/scikit-bio/pull/1954).
 * Renamed the default branch from "master" to "main". See [#1953](https://github.com/scikit-bio/scikit-bio/pull/1953).
--- a/skbio/diversity/__init__.py
+++ b/skbio/diversity/__init__.py
@@ -42,7 +42,7 @@ individual organism's genome.
    unit (OTU), a historically important term in microbiome research. However,
    as the field advances and the research targets diverge (e.g., amplicon
    sequence variant, or ASV), a more generic term such as "taxon" becomes more
-   appropriate. Therefore, the term OTU was replaced by taxon since scikit-bio
+   appropriate. Therefore, the term OTU was replaced by taxon in scikit-bio
    0.6.0.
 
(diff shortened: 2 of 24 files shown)


wasade:
Hi @qiyunzhu, out of curiosity, what about `feature` instead as some of the calculations are not specific to taxa?

qiyunzhu:
@wasade Thanks for the suggestion. I thought about "feature", but eventually thought that "taxon" is a more familiar term to people who study evolution and ecology. For example, we can call each tip in a phylogenetic tree as a "feature" but that doesn't sound intuitive. Some of my rationales can be found in #1902.

wasade:
True, but in that case, we use the `TreeNode` for more than phylogenies. And, many of the diversity calculations are applicable to mass spectrometry data

qiyunzhu:
I tend to think that "tree" may not be unintuitive to phylogeneticists. I agree that "taxon" does not have a broad meaning as "feature" does. To clarify that, the documentation has an explanation:

Previous verison:

```
We use the term "OTU" here very loosely, as these can in practice represent
diverse feature types including bacterial species, genes, and metabolites. The
term "sample" is also loosely defined for these purposes. These are intended to
represent a single unit of sampling, and as such what a single sample
represents can vary widely. For example, in a microbiome survey, these could
represent all 16S rRNA gene sequences from a single oral swab. In a comparative
genomics study on the other hand, a sample could represent an individual
organism's genome.
```

In this PR:
```
The term "taxon" (plural: "taxa") describes a group of biologically related
organisms that constitute a unit in the community. Taxa are usually defined at
a uniform taxonomic rank, such as species, genus or family. In community
ecology, taxon is usually referred to as "species" (singular = plural), but its
definition is not limited to species as a taxonomic rank. The term "taxonomic
group" is a synonym of taxon in many situations.

In scikit-bio, the term "taxon/taxa" is used very loosely, as these can in
practice represent diverse feature types including organisms, genes, and
metabolites. The term "sample" is also loosely defined for these purposes.
These are intended to represent a single unit of sampling, and as such what a
single sample represents can vary widely. For example, in a microbiome survey,
these could represent all 16S rRNA gene sequences from a single oral swab. In
a comparative genomics study on the other hand, a sample could represent an
individual organism's genome.
```

Regarding diversity calculation on non-community data: Currently, all alpha diversity metrics require input data to be counts (non-negative integers). This might be relaxed to accept floating-point numbers. For example, transcriptomic data in the unit of RPKM or TPM can use these functions too. Not all diversity metrics can accept floating-point numbers (for example, Chao1), but some do (for example, Shannon). This potential may be explored in scikit-bio.

qiyunzhu:
@mataton Thanks for reviewing! I agree with your assessment of the two places in the documentation. I have fixed them and updated this pull request.

PR #1971, opened 2024-03-14
Enhanced to_definites and hash implementations for gaps, degenerates, and definites

mataton:
This PR implements a new version of the `to_definites` function first introduced in #1952. The function is 10x faster than the previous implementation on large sequences (~15k in length). 

It now also implements hash table operations for the `gaps`, `degenerates`, and `definites` functions. `np.in1d` is deprecated as of numpy 2.0 and was what these functions were using previously. Even when tested against `np.isin`, which is the replacement for `np.in1d`, the hash table implementations here are twice as fast. 


skbio/sequence/_grammared_sequence.py
```diff
@@ -809,12 +808,21 @@ def to_definites(self, degenerate="wild", noncanonical=True):
                     f"Invalid character '{degenerate}' in sequence. Character must "
                     f"be within sequence alphabet: {self.alphabet}"
                 )
+        elif isinstance(degenerate, str) and len(degenerate) == 0:
+            sub_char = 0
         else:
             raise ValueError('Invalid value for parameter "degenerate".')
 
-        sequence = "".join(sub_char if x in degenerates else x for x in sequence)
+        htab = np.zeros((256,), dtype=np.uint8)
```
qiyunzhu, line 816: The hash table should be stored as a hidden attribute of the class (like `_degenerate_codes`), rather than being created every time one calls the `to_definites` method. This could save lots of compute. Because there are two versions: 1) degenerate codes, 2) degenerate and noncanonical codes, you will need to create two attributes (e.g., `_degenerate_hash` and `_degen_nonca_hash`).

skbio/sequence/_grammared_sequence.py
```diff
@@ -809,12 +808,21 @@ def to_definites(self, degenerate="wild", noncanonical=True):
                     f"Invalid character '{degenerate}' in sequence. Character must "
                     f"be within sequence alphabet: {self.alphabet}"
                 )
+        elif isinstance(degenerate, str) and len(degenerate) == 0:
```
qiyunzhu, line 811: This can be `elif degenerate == "":`, although I think there is a better method (see below).

skbio/sequence/_grammared_sequence.py
```diff
@@ -809,12 +808,21 @@ def to_definites(self, degenerate="wild", noncanonical=True):
...
         else:
             raise ValueError('Invalid value for parameter "degenerate".')
 
-        sequence = "".join(sub_char if x in degenerates else x for x in sequence)
+        htab = np.zeros((256,), dtype=np.uint8)
+        htab[self._degenerate_codes] = 1
+        if noncanonical:
+            htab[self._noncanonical_codes] = 1
+        pos = htab[self._bytes]
+        if sub_char:
```
qiyunzhu, line 821: Here, I would suggest separating the situations of 1) trimming off degenerate characters and 2) replacing them with a new character. For 2), you don't have to do `np.trim_zeros`, which can save some compute.

For 1), there are a few problems: a) `np.trim_zeros` only removing leading and trailing zeros, but not middle ones. b) 0 can be a valid character code (although it is not usually used). We can't just make 0 a symbol for removal. I would suggest this instead: `self._bytes[np.where(1 - pos)[0]]`. (Also, I would suggest changing the parameter back to `trim` because it is more explicit than an empty string.)

qiyunzhu (changes requested):
@mataton I have a few comments. Thanks!

skbio/sequence/_grammared_sequence.py
qiyunzhu: PS: I am thinking that "del" may be better than "trim" as an option for the parameter, because "trim" could implicate that only leading and trailing degenerates are trimmed.

qiyunzhu:
@mataton Additionally, the hash table idea can be adopted to replace the existing `degenerates`, `gaps` and `definites` methods. Currently, they do:

```python
def gaps(self):
    return np.in1d(self._bytes, self._gap_codes)
```

As we noted, `np.in1d` is deprecated and should be replaced with `np.isin`. However, both functions involve array search, which is expensive. If we do a hash table solution instead, it would be:

```python
def gaps(self):
    return self._gap_hash[self._bytes]
```

In which `self._gap_hash` is a boolean vector of 256 elements. `True` means gap, `False` means all others. This NumPy fancy indexing trick should be much more performant than array search (my guess; subject to test).

mataton:
@qiyunzhu Thanks, this all sounds good. Do you think I should make the changes to `degenerates`, `gaps`, and `definites` in this PR or submit another PR focused on those? I'm working on the proposed changes to `to_definites` now.

qiyunzhu:
@mataton I think that you can change all of them in this PR. These are not public-facing APIs (because they start with `_`). So we don't have to worry that they may break something.

skbio/sequence/_grammared_sequence.py
```diff
@@ -209,6 +211,21 @@ def _noncanonical_codes(cls):
             cls.__noncanonical_codes = np.asarray([ord(c) for c in noncanonical_chars])
         return cls.__noncanonical_codes
 
+    @classproperty
+    def _degenerate_hash(cls):
+        if cls.__degenerate_hash is None:
+            htab = np.zeros((256,), dtype=np.uint8)
```
qiyunzhu, line 210: This can be the following to avoid creating an extra variable (`htab`):

```python
cls.__degenerate_hash = np.zeros((256,), dtype=np.uint8)
cls.__degenerate_hash[cls._degenerate_codes] = 1
```

skbio/sequence/_grammared_sequence.py
```diff
@@ -209,6 +211,21 @@ def _noncanonical_codes(cls):
...
+            htab[cls._degenerate_codes] = 1
+            cls.__degenerate_hash = htab
+        return cls.__degenerate_hash
+
+    @classproperty
+    def _degen_nonca_hash(cls):
```
qiyunzhu, line 216: This place is very tricky: When you do `arr2 = arr1`, you are not copying all elements of `arr1`, but simply creating a reference to `arr1`. As a consequence, when you modify the content of `arr2`, you will find that `arr1` is also modified. This can be a problem. The solution is to do `arr2 = arr1.copy()`. [Here](https://www.educative.io/answers/pass-by-value-vs-pass-by-reference) is some reading materials on this.

mataton: Thank you!

qiyunzhu (approved):

qiyunzhu:
Will merge. The drop of code coverage is just because some code is removed, not that some code is not covered.

PR #1982, opened 2024-03-19
WIP: Registry support for HDF5

wasade:
This pull request modifies the IO registry to support container types, which currently only includes HDF5.

The changes have not yet been documented, and there are further tests that need to be included. The reason to open a WIP PR right now is the changes hit a milestone in that a "sniffer" was successful, and to trigger the full test suite across architectures we support. 

The changes were modeled approximately after the compression support. However, a container may or may not be compressed, and a container may have higher level constructs, so lumping them under those models was not correct. 

The `h5py.File` object does not quack like a regular IO object in Python. To make this work with the current framework, it was necessary to mix with `BytesIO` (n.b. other `io` types like `Buffered*` and `RawIOBase` or `IOBase` did not work). I did not obviously see an approach in h5py to test e.g. `readable`. What's important here is that the `h5py.File` model is not of a file but instead a `h5py.Group` so there are some differences in expectation. 

At the moment, though I don't fully understand the inner workings, I believe format interaction will need to be via `fh.raw` (see the sniffer test). 

@ebolyen if you're able, a quick pass through this would be greatly appreciated. 


skbio/io/_iosources.py
```diff
@@ -53,6 +58,12 @@ def get_compression_handler(name):
     return compressors.get(name, False)
 
 
+def get_container_handler(name):
+    containers = {c.name: c for c in _containers()}
+    containers["auto"] = AutoContainer
```
wasade, line 63: I don't _think_ it makes sense to have an automatic container

skbio/io/_iosources.py
```diff
@@ -254,3 +273,54 @@ def get_reader(self):
...
+
+class H5PYFile(h5py.File, io.BytesIO):
+    # h5py.File does not provide typical attributes and methods for a file
+    # like closed, readable, seekable, writable, etc. A mixin with BytesIO
+    # solves this
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+
+class AutoContainer(Container):
```
wasade, line 313: Drop

skbio/io/registry.py
```diff
@@ -911,13 +917,15 @@ def wrapped_sniffer(
...
                     errors=errors,
+                    container=container,
                     **kwargs,
                 ) as fh:
                     try:
                         # Some formats may have headers which indicate their
                         # format sniffers should be able to rely on the
                         # filehandle to point at the beginning of the file.
-                        fh.seek(0)
+                        if fh.seekable():
```
wasade, line 927: Add a comment about why this was needed

skbio/io/tests/test_registry.py
```diff
@@ -790,6 +791,23 @@ def formatc_sniffer(fh):
         self.assertTrue(self._check_f2)
         self.assertTrue(self._check_f3)
 
+    def test_sniff_hdf5(self):
+        expected = [1, 2, 3]
+
+        formata = self.registry.create_format('formata', encoding='binary',
```
wasade, line 797: Add a format construction which is not the hdf5 container

skbio/io/util.py
```diff
@@ -151,6 +160,9 @@ def open(
     compresslevel : int (0-9 inclusive), optional
         The level of compression to use, will be passed to the appropriate
         compression handler. This is only used when writing.
+    container : {'hdf5', None}, optional
+        TODO
+        "auto" while currently in the code may not actually make sense
```
wasade, line 165: Correct this

skbio/io/util.py
```diff
@@ -183,17 +195,31 @@ def _munge_file(file, is_binary_file, arguments):
...
         raise ValueError("Cannot use `newline` with binary encoding.")
 
     if compression is not None and not compression_handler:
         raise ValueError("Unsupported compression: %r" % compression)
 
+    if container is not None and not container_handler:
+        raise ValueError("Unsupported container: %r" % container)
+
+    if container is not None:
+        if compression == "auto":
```
wasade, line 215: Drop this

skbio/io/util.py
```diff
@@ -183,17 +195,31 @@ def _munge_file(file, is_binary_file, arguments):
...
+            compression = None
+        elif compression is not None:
+            raise ValueError("Compression with a container is not supported.")
+
+    if container is not None and not is_binary_file:
+        raise ValueError("Containers are binary.")
```
wasade, line 221: I think this is effectively true for container types we may care about. It is necessary to define in the documentation what we consider a container though as JSON _contains_ data. Perhaps the binary constraint can be relaxed in the future.

skbio/io/util.py
```diff
@@ -204,6 +230,15 @@ def _munge_file(file, is_binary_file, arguments):
             else:
                 newfile = CompressedBufferedReader(file, c.get_reader())
 
+        if container:
+            c = container_handler(newfile, arguments)
+            if mode == "w":
+                newfile = ContainerWriter(
+                    file, c.get_writer(), streamable=c.streamable
```
wasade, line 237: HDF5 I don't think can be `streamable` but other future types might be

skbio/io/_fileobject.py
```diff
@@ -64,6 +64,34 @@ class CompressedBufferedWriter(CompressedMixin, io.BufferedWriter):
...
+    """Act as a bridge between worlds."""
+
+    def __init__(self, before_file, *args, **kwargs):
+        self.streamable = kwargs.pop("streamable", True)
+        self._before_file = before_file
+        super(ContainerMixin, self).__init__(*args, **kwargs)
+
+    @property
+    def closed(self):
+        return self.raw.closed or self._before_file.closed
```
wasade, line 77: Add comment on where `raw` comes from

Add comment on differentiating `before_file` and `raw`

skbio/io/registry.py
wasade: ...not actually needed

skbio/io/tests/test_registry.py
wasade: On closer look, I dont think that makes sense here

wasade:
I don't think this is actually necessary anymore. It was a good dive into the registry, but we can handle the IO directly without adding complexity here.

PR #1988, opened 2024-03-21
House keeping works for the composition module

qiyunzhu:
This PR touches up the `skbio.stats.composition` module to improve its usability.

Main changes (following discussion with @mortonjt):

- Abbreviated `multiplicative_replacement` as `multi_replace` for simplicity. The old function is kept as an alias for backward compatibility.
- Implemented the Benjamini-Hochberg method for FDR correction, in addition to the existing Holm-Bonferroni method.
- Added a parameter `p_adjust` for both `ancom` and `dirmult_ttest`.
  - The word `p_adjust` is consistent with R, statsmodels, and QIIME2's ANCOM-BC plugin.
  - The value of `p_adjust` can be None, "holm", "holm-bonferroni", "bh", "fdr_bh", "benjamini-hochberg" (case insensitive). This is to ensure compatibility with statsmodels' `multipletests`, which we may adopt in the future.
  - The original parameter `multiple_comparisons_correction` of `ancom` is too long, so I renamed it as `p_adjust`. The old name is kept for backward compatibility.

Notes:
- I left the default FDR correction method as Holm-Bonferroni (Holm) for both `ancom` and `dirmult_ttest`. However, I acknowledge that ALDEx2 defaults to Benjamini-Hochberg (BH). ANCOM also defaults to BH ([source code](https://github.com/FrederickHuangLin/ANCOM-Code-Archive/blob/a1c47296b5e8e518387b4885699587274c2ab6b1/programs/ancom.R#L129)), whereas ANCOM-BC defaults to Holm. I left them Holm because I fear that my edits could break the code. @mortonjt 

Minor changes:
- Standardized docstrings throughout the module.
- For `ancom`, the user now can specify a statistical test by name (like "ttest_ind") in addition to a callable.

Issues noticed (@mortonjt please check):
- In the test code, the [`multiplicative_replacement_warning`](https://github.com/scikit-bio/scikit-bio/blob/3fb72e437841e01690a02d44a35c290953ff12d1/skbio/stats/tests/test_composition.py#L242) function was skipped because it doesn't starts with `test_`. I manually ran it and found that it failed. That is, a large delta value doesn't cause multiplicative replacement to generate negative proportions ([source code](https://github.com/scikit-bio/scikit-bio/blob/3fb72e437841e01690a02d44a35c290953ff12d1/skbio/stats/composition.py#L228)). I tried very large delta values but it consistently failed. Is there something overlooked in the function?
- The function `vlr`'s docstring doesn't have formula and description as other transformation functions do. Also, `vlr` does not seem to support 2D matrices as input, while other transformations do.

Thoughts:
- The functions [`sbp_basis`](https://github.com/scikit-bio/scikit-bio/blob/3fb72e437841e01690a02d44a35c290953ff12d1/skbio/stats/composition.py#L1603C5-L1603C14) and [`_gram_schmidt_basis`](https://github.com/scikit-bio/scikit-bio/blob/3fb72e437841e01690a02d44a35c290953ff12d1/skbio/stats/composition.py#L1581) use a slow `for` loop and can be optimized using NumPy's vectorization. Can be a practice project for a junior developer.
- The `draws` parameter of `dirmult_ttest` may be too informal. This parameter defines the number of Monte Carlo samplings. In ALDEx2, it is called `mc.samples`. In SciPy it is called `n_resamples` ([link](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.monte_carlo_test.html)).
- "significance_test" is bit long. But I don't have a better idea.
- "Reject null hypothesis" is bit long. How about "Reject null"?
- "Treatment" vs "Reference" sounds too specific. But I don't have a better idea.

### DIFF
--- a/skbio/stats/composition.py
+++ b/skbio/stats/composition.py
@@ -1292,6 +1292,11 @@ def ancom(
         accept at least two vectors of floats and returns a test statistic and
         a *p*-value. Functions under ``scipy.stats`` can be directly specified
         by name. The default is one-way ANOVA ("f_oneway").
+
+        .. versionchanged:: 0.6.0
+
+            Accepts test names in addition to functions.
+
     percentiles : iterable of floats, optional
         Percentile abundances to return for each feature in each group. By
         default, will return the minimum, 25th percentile, median, 75th
@@ -1330,6 +1335,9 @@ def ancom(
     --------
     ``multiple_comparisons_correction`` is deprecated as of ``0.6.0``. It has
     been renamed to ``p_adjust``.
+
+    ``significance_test=None` is deprecated as of ``0.6.0``. The default value
+    is now "f_oneway".
 
     Notes
     -----
@@ -1490,7 +1498,7 @@ def ancom(
     if not 0 < theta < 1:
         raise ValueError("`theta`=%f is not within 0 and 1." % theta)
 
-    # deprecated parameter
+    # @deprecated
     if multiple_comparisons_correction != "holm-bonferroni":
         _warn_deprecated(ancom, "0.6.0")
         p_adjust = multiple_comparisons_correction
@@ -1538,6 +1546,7 @@ def ancom(
             "single group)."
         )
 
+    # @deprecated
     if significance_test is None:
         significance_test = "f_oneway"
 
--- a/skbio/stats/composition.py
+++ b/skbio/stats/composition.py
@@ -92,6 +92,7 @@ from scipy.stats import t, gmean
 from skbio.stats.distance import DistanceMatrix
 from skbio.util import find_duplicates
 from skbio.util._misc import get_rng
+from skbio.util._warning import _warn_deprecated
 
 
 def closure(mat):
@@ -238,14 +239,7 @@ def multiplicative_replacement(mat, delta=None):
     multi_replace
 
     """
-    # @deprecated
-    if not hasattr(multi_replace, "warned"):
-        simplefilter("once", DeprecationWarning)
-        warn(
-            "multiplicative_replacement is deprecated as of 0.6.0.", DeprecationWarning
-        )
-        multi_replace.warned = True
-
+    _warn_deprecated(multiplicative_replacement, "0.6.0")
     return multi_replace(mat, delta)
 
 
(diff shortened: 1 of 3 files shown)


skbio/stats/composition.py
```diff
@@ -511,13 +516,14 @@ def clr_inv(mat):
 def ilr(mat, basis=None, check=True):
     r"""Perform isometric log ratio transformation.
 
-    This function transforms compositions from Aitchison simplex to
-    the real space. The :math: ilr` transform is both an isometry,
-    and an isomorphism defined on the following spaces
+    This function transforms compositions from Aitchison simplex to the real
+    space. The :math: ilr` transform is both an isometry, and an isomorphism
```
mataton, line 520: Missing tick mark before ilr. This currently doesn't render properly.

skbio/stats/composition.py
```diff
@@ -1235,55 +1325,65 @@ def ancom(
...
     scipy.stats.f_oneway
     scipy.stats.wilcoxon
     scipy.stats.kruskal
 
+    Warnings
+    --------
+    ``multiple_comparisons_correction`` is deprecated as of ``0.6.0``. It has
+    been renamed to ``p_adjust``.
+
+    ``significance_test=None` is deprecated as of ``0.6.0``. The default value
```
mataton, line 1339: Missing second tick mark after "significance_test=None". This causes rendering issues.

skbio/stats/composition.py
mataton, line 520: Missing tick mark before ilr. This currently doesn't render properly.

skbio/stats/composition.py
mataton, line 1339: Missing second tick mark after "significance_test=None". This causes rendering issues.

mataton (changes requested):
@qiyunzhu Besides these minor changes everything looks good to me.

qiyunzhu:
@mataton Thanks for noting the typos. I have addressed them.

mataton (approved):
@qiyunzhu Looks good! Once the checks pass I'll merge.

Issue #1990, opened 2024-03-21
Intermediate BIOM file left after test


qiyunzhu:
Every time one executes `make test` to run all tests (including doctests), there will be an additional file `test.biom` created under `ci`. This file stays after the test finishes. I guess this is because of the doctest [here](https://github.com/scikit-bio/scikit-bio/blob/4891a54569f5a2031b0951094513696f124bf1b9/skbio/io/format/biom.py#L35). @wasade Can you address it? Thanks!

PR #1994, opened 2024-03-25
Use os.remove to remove generated .biom file

mataton:
This PR implements a simple solution to the problem illustrated in issue #1990. It also includes some formatting changes to the file made by ruff.

### DIFF
--- a/skbio/io/format/biom.py
+++ b/skbio/io/format/biom.py
@@ -30,17 +30,15 @@ The official format specification for BIOM-Format can be found at [1]_.
 
 Examples
 --------
-Here we're demonstrating how to convert a BIOM table into HDF5 format and then
-read it back.
+Here we will write an existing BIOM table, and re-read it. Note that the Table
+from ``biom`` implicitly gets the ``.write`` method from the IO registry. This
+``ByteIO`` object can be a file path in a regular use case.
 
->>> import io
->>> import biom
->>> import skbio
->>> import h5py
+>>> import io, skbio
 >>> f = io.BytesIO()
->>> with h5py.File(f, 'w') as fp:
-...     biom.example_table.to_hdf5(fp, 'asd')
->>> roundtrip = skbio.read(io.BytesIO(f.getvalue()), into=skbio.Table)
+>>> skbio.feature_table.example_table.write(f)  # doctest: +ELLIPSIS
+<_io.BytesIO object at ...>
+>>> roundtrip = skbio.read(f, into=skbio.Table)
 >>> roundtrip
 2 x 3 <class 'biom.table.Table'> with 5 nonzero entries (83% dense)
 
--- a/skbio/io/format/biom.py
+++ b/skbio/io/format/biom.py
@@ -9,8 +9,10 @@ Internally, it stores the data in both compressed sparse row, and compressed
 sparse column representation. It additionally has support for representing sample
 and feature metadata.
 
-.. note:: Internally, BIOM describes features as observations, whereas scikit-bio
-uses the term features.
+.. note::
+
+   Internally, BIOM describes features as observations, whereas scikit-bio uses the
+   term features.
 
 Format Support
 --------------
@@ -31,10 +33,17 @@ Examples
 Here we will write an existing BIOM table, and re-read it. Note that the Table
 from ``biom`` implicitly gets the ``.write`` method from the IO registry.
 
->>> import biom, skbio, os
+>>> import biom, skbio
 >>> biom.example_table.write('test.biom')
 'test.biom'
->>> roundtrip = skbio.read('test.biom', into=skbio.Table)
+
+>>> table = skbio.read('test.biom', into=skbio.Table)
+>>> table
+2 x 3 <class 'biom.table.Table'> with 5 nonzero entries (83% dense)
+
+Clean up:
+
+>>> import os
 >>> os.remove('test.biom')
 
 References
--- a/skbio/io/format/biom.py
+++ b/skbio/io/format/biom.py
@@ -30,21 +30,19 @@ The official format specification for BIOM-Format can be found at [1]_.
 
 Examples
 --------
-Here we will write an existing BIOM table, and re-read it. Note that the Table
-from ``biom`` implicitly gets the ``.write`` method from the IO registry.
+Here we're demonstrating how to convert a BIOM table into HDF5 format and then
+read it back.
 
->>> import biom, skbio
->>> biom.example_table.write('test.biom')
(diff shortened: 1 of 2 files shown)


mataton:
@wasade Do you see any issue in handling things with os.remove?

qiyunzhu:
> @wasade Do you see any issue in handling things with os.remove?

Specifically, we were a bit concerned whether writing a file in the current directory might cause an issue if skbio is installed into a non-writable directory (e.g., through `sudo apt install skbio`). We attempted to do this in a writtable temporary directory by Sphinx decoratives `.. testsetup::` and `.. testcleanup::`. However, they didn't work despite our trying. Therefore, we had to give up but simply use `os.remove` to delete the file in the current directory.

wasade:
Does writing to an instance of `io.BytesIO` work here? see eg:

```
In [7]: f = io.BytesIO()

In [8]: fp = h5py.File(f, 'w')

In [9]: biom.example_table.to_hdf5(fp, 'asd')

```

mataton:
@qiyunzhu @wasade I made a new commit with a `io.BytesIO` solution, but I'm not sure if it's exactly what we want.

skbio/io/format/biom.py
```diff
@@ -28,13 +30,19 @@
...
 --------
-Here we will write an existing BIOM table, and re-read it. Note that the Table
-from ``biom`` implicitly gets the ``.write`` method from the IO registry.
-
->>> import biom, skbio
->>> biom.example_table.write('test.biom')
-'test.biom'
->>> roundtrip = skbio.read('test.biom', into=skbio.Table)
+Here we're demonstrating how to convert a BIOM table into HDF5 format and then
+read it back.
```
qiyunzhu, line 34: Can add a sentence: "This `ByteIO` object can be a file path in a regular use case."

skbio/io/format/biom.py
```diff
@@ -28,13 +30,19 @@
...
+
+>>> import io
+>>> import biom
+>>> import skbio
+>>> import h5py
+>>> f = io.BytesIO()
+>>> with h5py.File(f, 'w') as fp:
+...     biom.example_table.to_hdf5(fp, 'asd')
+>>> roundtrip = skbio.read(io.BytesIO(f.getvalue()), into=skbio.Table)
```
qiyunzhu, line 43: What if `io.BytesIO(f.getvalue())` is replaced with `f`? If not working, what if adding `f.seek(0)` beforehand?

qiyunzhu (approved):
This looks about right to me! Thanks @mataton @wasade

skbio/io/format/biom.py
mataton: I have this written now:

````
>>> import io
>>> import skbio
>>> f = io.BytesIO()
>>> skbio.feature_table.example_table.write(f)
<_io.BytesIO object at 0x7f969d056390>
>>> f.seek(0)
>>> roundtrip = skbio.read(f, into=skbio.Table)
>>> roundtrip
2 x 3 <class 'biom.table.Table'> with 5 nonzero entries (83% dense)
````

But the memory address of the `BytesIO` object is different when I run `make test` compared to when I just run it in python, which makes the test fail. I'm not very experienced with this so I'm not sure how to proceed.

Comment (wasade) [reply skbio/io/format/biom.py:43]: what about [`skip`](https://docs.pytest.org/en/7.1.x/how-to/doctest.html#skipping-tests)?

Comment (mataton) [reply skbio/io/format/biom.py:43]: Importantly, all of the following only occurs when I run `make test`. When I perform the same commands in the Python shell the errors don't occur and `f` correctly contains the content of `skbio.feature_table.example_table`.

````
>>> skbio.feature_table.example_table.write(f)  # doctest: +SKIP
<_io.BytesIO object at 0x7f969d056390>
>>> print(f.getvalue())
Expected nothing
Got:
    b''
````

It appears that `f` is an `<emptyfile>` while it should contain the content of `skbio.feature_table.example_table`, if my understanding is correct. This leads to the following error.

````
>>> roundtrip = skbio.read(f, into=skbio.Table)
UNEXPECTED EXCEPTION: UnrecognizedFormatError("Cannot read '<emptyfile>' from <WrappedBufferedRandom>, no Table reader found.  ")
````

This error occurs whether I have `f.seek(0)` preceding the `roundtrip = skbio.read(f, into=skbio.Table)` line or not. 

Thank you for the help so far @wasade.

Comment (wasade) [reply skbio/io/format/biom.py:43]: Doesn't look empty? I'm not sure what `getvalue` does off hand, but a `seek(0)` followed by `read` confirms there are quite a few bytes of data, and I'm able to load from a `seek(0)`:

```
>>> import skbio
>>> import io
>>> f = io.BytesIO()
>>> skbio.feature_table.example_table.write(f)
<_io.BytesIO object at 0x7fa6a014e890>
>>> f.seek(0)
0
>>> what = f.read()
>>> len(what)
37914
>>> f.seek(0)
0
>>> tab = skbio.read(f, into=skbio.feature_table.Table)
>>>
```

mataton: I agree, when running it interactively it works and I get the same result as you. However, I am still encountering the above errors when I run the test by the command `make test`:

````
========================================== FAILURES ==========================================
_______________________________ [doctest] skbio.io.format.biom _______________________________
035
036 >>> import io
037 >>> import skbio
038 >>> f = io.BytesIO()
039 >>> skbio.feature_table.example_table.write(f)  # doctest: +SKIP
040 <_io.BytesIO object at 0x7ff6309e0f40>
041 >>> f.seek(0)
042 0
043 >>> what = f.read()
044 >>> len(what)
Expected:
    37914
Got:
    0

/home/maton/qlab/skbio/github/scikit-bio/skbio/io/format/biom.py:44: DocTestFailure
````

qiyunzhu:
@wasade @mataton I did some research and found that `.. testsetup::` could not be triggered because we are using pytest instead of Sphinx's `make doctest` to run the doctests. The latter would make the doctests more flexible. There are additional limitations associated with the pytest method (e.g., you can't run a single doctest). This can potentially be addressed in the future.

skbio/io/format/biom.py
mataton: The most recent commit works. The issue was when running `make test` I was skipping the line which wrote the file by using `doctest: +SKIP`. Instead, I used `doctest: +ELLIPSIS` and this works. Thank you again @wasade!

wasade: Thanks for all the hard work!! I know very much so how tedious doctest can get :/

Issue #1996, opened 2024-03-26
hdmedians is not available from conda-forge for windows


mataton:
[hdmedians](https://github.com/daleroberts/hdmedians) is unavailable from conda-forge on windows. Adding windows support is a goal of scikit-bio over the next year. Currently hdmedians is only used in a single function in skbio:  `_permdisp._config_med()`. 

We may consider implementing hdmedians within skbio, as hdmedians has the Apache 2.0 license which may be compatible with our own license. Alternatively we may consider replacing hdmedians with another package which is available on conda-forge for windows and more actively maintained.

@qiyunzhu had this idea initially. @wasade and @mortonjt what are your thoughts?

mortonjt:
Looking at the source code, it actually looks pretty minimal. The only
possible concern is porting the extra cython code from hdmedians - we’d
need to make sure that the cython code also ports to windows

qiyunzhu:
@mataton This is great to know! Thanks! Yes let's port hdmedians.

The relevant hdmedians function is [`geomedian`](https://github.com/daleroberts/hdmedians/blob/19113b0a3496fed45bdf75740ff47f35bd420271/hdmedians/geomedian.pyx#L393C1-L393C16), which in turn calls [`geomedian_axis_one`](https://github.com/daleroberts/hdmedians/blob/19113b0a3496fed45bdf75740ff47f35bd420271/hdmedians/geomedian.pyx#L150). I think that `geomedian_axis_one` is the only function we need to port (along with any utility functions it calls, like `dist_euclidean` and `norm_euclidean`).

mataton:
Looks good, would `stats/distance/_utils.py` be a good place to put these?

mataton:
Closed with PR #2003.

PR #2003, opened 2024-04-03
WIP:  Porting hdmedians

mataton:
This PR will port the `geomedian` function from hdmedians. See issue #1996 for rationale. 

### DIFF
--- a/skbio/stats/distance/_cutils.pyx
+++ b/skbio/stats/distance/_cutils.pyx
@@ -293,3 +293,112 @@ def permanova_f_stat_sW_cy(TReal[:, ::1] distance_matrix,
             s_W += local_s_W/group_sizes[group_idx]
 
     return s_W
+
+from libc.math cimport isnan, sqrt, acos, fabs
+
+cimport numpy as cnp
+
+ctypedef fused floating:
+    cnp.float32_t
+    cnp.float64_t
+
+ctypedef cnp.float32_t float32_t
+ctypedef cnp.float64_t float64_t
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+cdef geomedian_axis_one(floating[:, :] X, floating eps=1e-7,
+                           size_t maxiters=500):
+    """Compute high dimensional median."""
+    cdef size_t p = X.shape[0]
+    cdef size_t n = X.shape[1]
+
+    cdef floating[:] y = np.mean(X, axis=1)
+
+    if n == 1:
+        return y
+
+    if floating is cnp.float32_t:
+        dtype = np.float32
+    else:
+        dtype = np.float64
+
+    cdef floating[:] D = np.empty(n, dtype=dtype)
+    cdef floating[:] Dinv = np.empty(n, dtype=dtype)
+    cdef floating[:] W = np.empty(n, dtype=dtype)
+    cdef floating[:] T = np.empty(p, dtype=dtype)
+    cdef floating[:] y1 = np.empty(p, dtype=dtype)
+    cdef floating[:] R = np.empty(p, dtype=dtype)
+
+    cdef floating dist, Dinvs, total, r, rinv, tmp, Di
+    cdef size_t nzeros = n
+    cdef size_t iteration
+
+    with nogil:
+        iteration = 0
+        while iteration < maxiters:
+
+            for i in range(n):
+                Di = dist_euclidean(X[:, i], y)
+                D[i] = Di
+                if fabs(Di) > eps:
+                    Dinv[i] = 1. / Di
+                else:
+                    Dinv[i] = 0.
+
+            Dinvs = sum(Dinv)
+
+            for i in range(n):
+                W[i] = Dinv[i] / Dinvs
+
+            for j in range(p):
+                total = 0.
+                for i in range(n):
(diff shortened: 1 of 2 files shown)


skbio/stats/distance/_cutils.pyx
```diff
@@ -293,3 +293,119 @@ def permanova_f_stat_sW_cy(TReal[:, ::1] distance_matrix,
             s_W += local_s_W/group_sizes[group_idx]
 
     return s_W
+
+from libc.math cimport isnan, sqrt, acos, fabs
+
+cimport numpy as cnp
```
qiyunzhu, line 299: imports should be uniformly placed at the beginning of the script.

skbio/stats/distance/_cutils.pyx
```diff
@@ -293,3 +293,119 @@ def permanova_f_stat_sW_cy(TReal[:, ::1] distance_matrix,
...
+
+ctypedef fused floating:
```
qiyunzhu, line 301: Should add a note saying that these types and functions are ported from hdmedians. Should indicate if you made any change to the code. Then copy the hdmedians license files to the `licenses` folder of skbio.

qiyunzhu (changes requested):
@mataton Good job! To use code from an Apache 2.0 licensed project, you will need to include their license in our `licenses` folder, and add a note before the ported code. You can find a few examples in the codebase.

skbio/stats/distance/_cutils.pyx
```diff
@@ -293,3 +293,119 @@ def permanova_f_stat_sW_cy(TReal[:, ::1] distance_matrix,
...
+    cnp.float32_t
+    cnp.float64_t
+
+ctypedef cnp.float32_t float32_t
+ctypedef cnp.float64_t float64_t
```
qiyunzhu, line 306: hdmedians supports float32 and float64 (according to its [readme](https://github.com/daleroberts/hdmedians)). But in our case only float64 will be provided as input. Therefore we can remove float32 support.

skbio/stats/distance/_cutils.pyx
```diff
@@ -293,3 +293,119 @@ def permanova_f_stat_sW_cy(TReal[:, ::1] distance_matrix,
...
+    return <floating>sqrt(d)
+
+cdef floating norm_euclidean(floating[:] x) nogil:
+    cdef size_t n = x.shape[0]
+    cdef float64_t d = 0.
+    for i in range(n):
+        d += x[i] * x[i]
+    return <floating>sqrt(d)
+
+cdef floating sum(floating[:] x) nogil:
```
qiyunzhu, line 406: I don't know the purpose of this function, but I guess that it is for backward compatibility with float32. In our case we can simply do `x.sum()` (my guess; pending test).

skbio/stats/distance/_cutils.pyx
```diff
@@ -293,3 +293,119 @@ def permanova_f_stat_sW_cy(TReal[:, ::1] distance_matrix,
...
+cdef floating dist_euclidean(floating[:] x, floating[:] y) nogil:
+    cdef size_t n = x.shape[0]
+    cdef float64_t d = 0.
+    cdef float64_t tmp
+    for i in range(n):
+        tmp = x[i] - y[i]
+        d += tmp * tmp
+    return <floating>sqrt(d)
+
+cdef floating norm_euclidean(floating[:] x) nogil:
```
qiyunzhu, line 404: This can be replaced with `np.linalg.norm(x)`, I guess, but I am not sure about efficiency or correctness.

skbio/stats/distance/_cutils.pyx
mataton: I'm not sure exactly what is going on as today is my first time working with Cython, but when I try using Python's built-in `sum()` function I get the following errors when I try to compile the Cython using `python setup.py build_ext --inplace`, all in reference to the `Dinvs = sum(Dinv)` line:
- `Converting to Python object not allowed without gil`
- `Constructing Python tuple not allowed without gil`
- `Calling gil-requiring function not allowed without gil`

[Here](https://cython.readthedocs.io/en/latest/src/userguide/nogil.html) is the documentation for the GIL. I don't think the Cython will compile correctly without this version of `sum()`.

qiyunzhu: On a second thought, the originally design (together with the use of `nogil`) might be for parallelization.

skbio/stats/distance/_cutils.pyx
mataton: Similarly to the `sum()` function, when I tested replacing `norm_euclidean` and `dist_euclidean` with `np.linalg.norm()`, I ran into errors regarding the use of Python objects without gil.

skbio/stats/distance/_cutils.pyx
mataton: I believe that to be true, should we keep the original design then?

qiyunzhu (commented):
@mataton For float32 removal, you may delete the entire `ctypedef fused floating` statement, which creates a new type by fusing float32 and float64. Then, you can replace every instance of `floating` with `float64_t`.

PR #2008, opened 2024-04-12
WIP : Embedding format

mortonjt:
This pull request includes a new format and wrapper for encoding protein embeddings from protein language models. Many finetuning methods (i.e. classifiers) operate on the protein embeddings, so exposing this interface will lower barriers for downstream applications. 

The underlying `ProteinEmbedding` object is simple, it only contains an embedding matrix and a string. The residues in the string correspond to the row vectors in the embedding matrix.  We are currently only implementing a light-weight wrapper to aid the standardization of a corresponding file format, while enabling the embeddings to be identified / indexed by the original sequence used to generate the embedding.

The implemented file format also simple, it takes in one big matrix `embedding` and an `id` string. In order to enable storing multiple embedding objects, we also store a `idptr` dataset that keeps track which portions of the `embedding` and `id` correspond to different sequences. This design will allow this file format to take in multiple `ProteinEmbedding` objects, which will help with mitigating the memory issues that are commonly encountered with encoding many embedding objects. 

- [x] Implement `ProteinEmbedding` class
- [x] Implement ABC classes `Embedding` and `SequenceEmbedding` 
- [x] Add hdf5 format to the IO registry
- [x] Include tests for `pad_start` and `pad_end`
- [x] Add tests for assertion errors
- [x] Add format error
- [x] Ensure that doc tests are complete and passing


mortonjt:
Not ready for review, having a really hard time adding documentation due to strict ruff requirements

mortonjt:
This pull request is about 90% done, there are a few house keeping tasks that need to be done. But it is becoming sufficiently large where I think it warrants an intermediate review. @wasade @qiyunzhu

skbio/embedding/_protein.py
```diff
@@ -0,0 +1,90 @@
...
+                f"The embedding ({embedding.shape[0]}) must have the "
+                f"same length as the sequence ({len(sequence)})."
+            )
+
+        super(ProteinEmbedding, self).__init__(
+            embedding=embedding, sequence=sequence, **kwargs
+        )
+
+    def __str__(self):
+        return str(self._ids)
```
mataton, line 67: This looks good, but when I do functional tests and I do `test_object.__str__` the output is identical to `test_object.__repr__`. Is this correct? It looks like `test_object.__str__` should just give me a string of the sequence(s). If I do `str(test_object._ids` I get the expected output of a string.

mataton (commented):
Besides the one comment, everything looks good to me.

skbio/embedding/_protein.py
```diff
@@ -0,0 +1,97 @@
...
+# ----------------------------------------------------------------------------
+from skbio.sequence import Protein
+from skbio.embedding._embedding import SequenceEmbedding
+from skbio.util import get_data_path
+from pathlib import Path
+import pandas as pd
+import numpy as np
+
+
+class ProteinEmbedding(SequenceEmbedding):
```
mataton, line 16: May also be nice to have a simple boolean like `clip_head` but to check if the input sequence has characters separated by spaces, and if it does, do `sequence_example = sequence_example.replace(" ", "")` or something similar. Not a huge deal for the user to do this themselves, but some models like `prot_bert` need the sequences to have spaces in between characters, where this class currently assumes that there are no spaces between characters.

skbio/embedding/_protein.py
```diff
@@ -0,0 +1,97 @@
...
+    """Stores the embeddings of the protein sequence.
+
+    Parameters
+    ----------
+    sequence : str, Sequence, or 1D np.ndarray
```
qiyunzhu, line 21: For `ProteinEmbedding` specifically, this `Sequence` can be `Protein`.

skbio/embedding/_embedding.py
```diff
@@ -0,0 +1,39 @@
...
+# The full license is in the file LICENSE.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+import numpy as np
+
+from skbio.sequence import Sequence
+from skbio._base import SkbioObject
+from skbio.metadata._mixin import MetadataMixin
+
+
+class Embedding(SkbioObject):
```
qiyunzhu, line 15: I think the current structure (embedding is a 1st-level module) is great. Just wanted to ask if there are non-sequence embeddings envisioned in our plan.

skbio/io/format/embedding.py
```diff
@@ -0,0 +1,212 @@
...
+    # 137
+    # >>> ord('\x89')
+    # 137
+    # >>> ord('\032')
+    # 26
+    # >>> ord('\x1a')
+    # 26
+    if magic == b"\x89HDF\r\n\x1a\n":
+        with h5py.File(fh, "r") as h5file:
+            if "embedding" in h5file and "id" in h5file and "idptr" in h5file:
```
qiyunzhu, line 94: Interesting that h5py can do this!

skbio/embedding/_protein.py
```diff
@@ -0,0 +1,100 @@
...
+        indent = " " * n_indent
+        rstr = rstr.replace(
+            "has gaps",
+            f"embedding dimension: {self.embedding.shape[1]}\n{indent}has gaps",
+        )
+        return rstr
+
+
+example_protein_embedding = ProteinEmbedding(
+    np.load(np.random.randn(62, 1024)),
```
mortonjt, line 99:
```suggestion
    np.random.randn(62, 1024),
```

skbio/embedding/_protein.py
```diff
@@ -96,5 +96,5 @@ def __repr__(self):
 
 
 example_protein_embedding = ProteinEmbedding(
-    np.load(np.random.randn(62, 1024)),
+    np.random.randn(62, 1024),
```
qiyunzhu, line 99: `randn` might be replaced by [`standard_normal`](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.standard_normal.html) in future NumPy versions.

skbio/embedding/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+This module provides support for storing embeddings for biological objects, such
+as protein embeddings outputted from protein language models (pLMs).
+
+Embedding types
+---------------
+
+.. autosummary::
+    :toctree: generated/
+
+    ProteinEmbedding
```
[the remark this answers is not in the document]
mortonjt: Its a thought, and it does raise questions about the difference between a `ProteinEmbedding` and a `SequenceEmbedding` -- it suggests that there should be protein-specific checks in place. This would be one of the benefits of having the `ProteinEmbedding` object inherent from `Protein`, it would enable validation of the input string (i.e. confirm that they are valid amino acids via IUPAC).

skbio/embedding/_embedding.py
mortonjt: Yes, once we have pLMs in place, we will want to think about molecular and mass-spectrometry embeddings / vectors

skbio/embedding/_embedding.py
```diff
@@ -0,0 +1,39 @@
...
+    @property
+    def ids(self):
+        # each embedding row corresponds to an id
+        return self._ids
+
+    def __init__(self, embedding, ids, **kwargs):
+        self._embedding = embedding
+        self._ids = ids
+
+        if not isinstance(embedding, np.ndarray):
```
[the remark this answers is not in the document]
mortonjt: I'm OK with having it be array-like, so as long as it is can be converted to an array

skbio/embedding/_protein.py
mortonjt: I added this in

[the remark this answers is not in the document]
mortonjt: Its a thought, but there is merit in checking if the protein is a valid sequence per the comment above.

skbio/embedding/_protein.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __init__(
+        self, embedding, sequence: str, clip_head=False, clip_tail=False, **kwargs
+    ):
+        if clip_head:
+            embedding = embedding[1:]
+        if clip_tail:
+            embedding = embedding[:-1]
+
+        if isinstance(sequence, Protein):
```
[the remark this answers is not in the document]
mortonjt: I'm not clear on this, how exactly would references in python be stored? Python doesn't have explicit support for pointers AFAIK

In this case, this is a minimal memory footprint, since the vast majority of proteins have less than 1K residues (and consequentially the vast majority of pLMs can only handle proteins with less than 1K residues). Plus the embeddings here are by far takes up most the memory, for each residue, there is a vector of length 1024 for ProTrans

I'm OK with having the internal sequence attribute be a protein.  Now the question is, does it make more sense for `ProteinEmbedding` to have a protein attribute, or for it to subclass `Protein`?

skbio/embedding/_protein.py
mortonjt: Could you clarify this with a minimal example? There is a unittest for the repr that seems to be passing on my end.

skbio/io/_exception.py
```diff
@@ -130,3 +130,9 @@ class BIOMFormatError(Exception):
     """Raised when a ``BIOM`` formatted file cannot be parsed."""
 
     pass
+
+
+class EmbeddingFormatError(Exception):
+    """Raised when an ``Embedding`` formatted file cannot be parsed."""
```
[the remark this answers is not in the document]
mortonjt: I'm following the formatting of the other errors -- if this will be changed, we will need to change the rest of the error docs (which I think we should defer to another PR).

mortonjt (commented):
First round of comments addressed. Some of the comments surrounding the `SequenceEmbedding` vs `ProteinEmbedding` remain outstanding.

skbio/embedding/_protein.py
mataton: Sure, the repr looks good to me. I am just confused whether the `__str__` is outputting what it should. 

````python
>>> test = ProteinEmbedding(sequence=sequence, embedding=embeddings, clip_head=True, clip_tail=True)
>>> test.__repr__
... <bound method ProteinEmbedding.__repr__ of ProteinEmbedding
... -------------------------
... Stats:
...     length: 7
...     embedding dimension: 1024
...     has gaps: False
...     has degenerates: True
...     has definites: True
...     has stops: False
... -------------------------
... 0 AETCXAX>
>>> test.__str__
... <bound method ProteinEmbedding.__repr__ of ProteinEmbedding
... -------------------------
... Stats:
...     length: 7
...     embedding dimension: 1024
...     has gaps: False
...     has degenerates: True
...     has definites: True
...     has stops: False
... -------------------------

I'm not sure if there's anything wrong with this. I'm just curious if `__str__` and `__repr__` are supposed to output the same thing?

File: skbio/io/format/embedding.py [hunk 12]
```diff
@@ -0,0 +1,212 @@
...
+    constructor=ProteinEmbedding,
+    obj_constructor=Protein,
+    kwargs: dict = {},
+):
+    h5grp = h5py.File(fh, "r")
+    embed_fh = h5grp["embedding"]
+    id_fh = h5grp["id"]
+
+    # assumes that there is only a single object in the file
+    emb = embed_fh[()].squeeze()
```
wasade, line 133:
```suggestion
    emb = np.array(embed_fh[:])
```

?

wasade: May not even need the `:`

skbio/io/format/embedding.py
```diff
@@ -0,0 +1,212 @@
...
+    id_ = id_fh[()]
+    string = str(id_.tobytes().decode("ascii")).replace("\x00", "")
```
wasade, line 135: Same comment as above

skbio/io/format/embedding.py
```diff
@@ -0,0 +1,212 @@
...
+            arr = np.frombuffer(str(obj).encode("ascii"), dtype=np.uint8)
+            # Store the embedding itself. We are assuming that the
+            # embbedding is a 2D numpy array
+            emb = obj.embedding
+            # store the pointers that keep track of the start and
+            # end of the embedding for each object, as well as well as
+            # the corresponding string representation
+            if "idptr" in h5grp:
+                idptr_fh = h5grp["idptr"]
+                idptr_fh.resize((i + 1,))
```
wasade, line 172: This likely will be expensive if done many times

skbio/io/format/embedding.py
```diff
@@ -0,0 +1,212 @@
...
+            else:
+                idptr_fh = h5grp.create_dataset(
+                    "idptr", data=[len(arr)], maxshape=(None,), dtype=np.int32
+                )
+            if "id" in h5grp:
+                id_fh = h5grp["id"]
+                id_fh.resize((idptr_fh[i],))
+                id_fh[idptr_fh[i - 1] : idptr_fh[i]] = arr
+            else:
+                id_fh = h5grp.create_dataset(
```
wasade, line 183: Should any of these have compression?

skbio/io/format/embedding.py
```diff
@@ -0,0 +1,212 @@
...
+                    "id", data=arr, maxshape=(None,), dtype=np.int32
```
wasade, line 184: IDs above are converted to ascii then utf-8, why use np.int32?

skbio/io/format/tests/test_embedding.py
```diff
@@ -0,0 +1,121 @@
...
+                 ),
+                (
+                    np.load(get_data_path('embed2.txt.npy')),
+                    Protein(('QQNKELNFKLREKQNEIFELKKIAETLRSKL'
+                             'EKYVDITKKLEDQNLNLQIKISDLEKKLSDA'),
+                            metadata={"id": "seq2"})
+                )
+            ]
+        )
+        self.tempdir = tempfile.TemporaryDirectory()
```
wasade, line 50: Could a `tearDown` be added which calls `self.tempdir.cleanup()`? I _think_ it's necessary to pass `ignore_cleanup_errors=True` to `TemporaryDirectory` as well

skbio/io/format/tests/test_embedding.py
```diff
@@ -0,0 +1,121 @@
...
+        self.valid_embed_path = get_data_path('prot.emb')
+        self.invalid_embed_path = str(tempdir / Path('invalid'))
+        self.nonembed_hdf5_path = str(tempdir / Path('other.hdf5'))
+
+        with open(self.invalid_embed_path, 'wb') as fp:
+            fp.write(b'this is not a embed file')
+
+        with h5py.File(self.nonembed_hdf5_path, 'w') as fp:
+            fp['stuff'] = [1, 2, 3]
+
```
wasade, line 64:
```suggestion
    
    def tearDown(self):
        self.tempdir.cleanup()
```

skbio/io/format/tests/test_embedding.py
wasade, line 50:
```suggestion
        self.tempdir = tempfile.TemporaryDirectory(ignore_cleanup_errors=True)
```

skbio/io/format/tests/test_embedding.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def test_sniffer(self):
+        self.assertEqual(_embed_sniffer(self.valid_embed_path), (True, {}))
+        self.assertEqual(_embed_sniffer(self.invalid_embed_path), (False, {}))
+        self.assertEqual(_embed_sniffer(self.nonembed_hdf5_path), (False, {}))
+
```
wasade, line 70:
```suggestion
```

skbio/io/format/embedding.py
```diff
@@ -0,0 +1,212 @@
...
+
++------+------+---------------------------------------------------------------+
+|Reader|Writer|                          Object Class                         |
++======+======+===============================================================+
+|Yes   |Yes   |generator of :mod:`skbio.embedding.ProteinEmbedding` objects   |
++------+------+---------------------------------------------------------------+
+|Yes   |Yes   |:mod:`skbio.embedding.ProteinEmbedding` objects                |
++------+------+---------------------------------------------------------------+
+
+Format Specification
```
wasade, line 22: Could the format spec include dimension and dtype expectations? 

It may be nice to include a `.attr['format']` and `.attr['format-version']` in the file

wasade: `.attr` is formally part of HDF5

skbio/embedding/_protein.py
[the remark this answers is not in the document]
qiyunzhu: I think that in Python, all variables are [passed by reference](https://realpython.com/python-pass-by-reference/), therefore you don't need to explicitly define a pointer. It is done already. If you do need to pass by value, you will have to manually trigger `.copy` or `.deepcopy` or something.

I would suggest that a `SequenceEmbedding` class has a `sequence` attribute. For each subclass (`ProteinEmbedding`, `DNAEmbedding`, `RNAEmbedding`), we can have explicit constraints (e.g., `sequence` must be `Protein`) and checks.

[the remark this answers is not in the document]
qiyunzhu: PS: Even if you do:
```python
prot = Protein('MKRVGI')
emb = ProteinEmbedding(..., prot)
prot = None
```
The embedding object still has the protein information. I think.

skbio/embedding/_protein.py
mataton:
````python
>>> str(test)
... 'AETCXAX'
>>> print(repr(test))
... ProteinEmbedding
... -------------------------
... Stats:
...     length: 7
...     embedding dimension: 1024
...     has gaps: False
...     has degenerates: True
...     has definites: True
...     has stops: False
... -------------------------
... 0 AETCXAX
````
So if done with `repr(test)` and `str(test)` it works as intended.

Comment (wasade) [reply skbio/embedding/_protein.py:70]: Awesome awesome :) so in case this wasn't known, many (not all) of the dunders have functions that access like these, but also see `hash`, `len`, etc. And `==` for `__eq__`, etc

File: skbio/embedding/__init__.py [hunk 21]
```diff
@@ -0,0 +1,29 @@
...
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+from skbio.embedding._protein import ProteinEmbedding, example_protein_embedding
+
+
+__all__ = ["ProteinEmbedding", "example_protein_embedding"]
```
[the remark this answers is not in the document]
mortonjt: I don't understand the merits of this - could you elaborate? It seems a little wasteful on memory to package an example embedding on every instance on `ProteinEmbedding`

skbio/io/format/embedding.py
```diff
@@ -0,0 +1,212 @@
...
+    id_fh = h5grp["id"]
+    idptr_fh = h5grp["idptr"]
+    j = 0
+    n = idptr_fh.shape[0]
+    for i in range(n):
+        idptr = idptr_fh[i]
+        id_ = id_fh[j:idptr]
+        emb = embed_fh[j:idptr]
+        string = str(id_.tobytes().decode("ascii")).replace("\x00", "")
+        j = idptr
```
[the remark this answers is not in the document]
mortonjt: It should just be a number via `idptr = idptr_fh[i]`
But yes it is a pointer to a position within the jagged array

skbio/io/format/embedding.py
mortonjt: That has crossed my mind (I don't have intuition on large-scale benchmarks).
The only caveat here is that in order to allocate the entire array at once, we'd need to do a sweep over the entire generator to tally up the sum of the sequence lengths. It should be doable, I'd just need to figure out the schematics of iterating over a generator multiple times without triggering a `StopIteration`

skbio/embedding/_protein.py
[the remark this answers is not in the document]
qiyunzhu: I totally agree that protein embedding should check whether it is a protein.

skbio/io/format/embedding.py
mortonjt: The sequences, probably not worth it.  The embedding arrays possibly yes -- since that is where most of the memory footprint is going to be coming from.

skbio/io/format/embedding.py
```diff
@@ -0,0 +1,212 @@
...
+    embed_fh = h5grp["embedding"]
+    id_fh = h5grp["id"]
+    idptr_fh = h5grp["idptr"]
+    j = 0
+    n = idptr_fh.shape[0]
+    for i in range(n):
+        idptr = idptr_fh[i]
+        id_ = id_fh[j:idptr]
+        emb = embed_fh[j:idptr]
+        string = str(id_.tobytes().decode("ascii")).replace("\x00", "")
```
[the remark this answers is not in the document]
mortonjt: Could you clarify?  I was under the impression that this was storing the utf-8 and pulling it out ...

skbio/embedding/_embedding.py
qiyunzhu: Cool! Two more questions: 1) Will ordinations be considered as embeddings? 2) We need to think about whether we do: sequence -> sequence embedding, or embedding -> sequence embedding. Since embedding can host multiple heterogerous biological objects, I wonder if it makes more sense to separate them under sequence, structure, spectrum, etc. But I also see the point of other way round. Hard question...

skbio/embedding/__init__.py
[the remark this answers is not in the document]
wasade: If there is a way to make instantiation lazy then that would reduce import time

mortonjt (commented):
See questions below

skbio/io/format/embedding.py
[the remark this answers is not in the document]
wasade: No need for `tobytes` or juggling the null character. You get back bytes, which is confusing me at the moment, but I can't tinker more on it right now -- we coerce utf-8 in biom though.

```python

In [41]: f = h5py.File('foo.h5', 'w')

In [42]: f['example'] = ['foo', 'bar', 'baz']

In [43]: f['example'][0]
Out[43]: b'foo'

In [44]: f['example'][0].decode('utf-8')
Out[44]: 'foo'

In [45]: f.create_dataset('example-2', data=['foo', 'bar', 'baz'])
Out[45]: <HDF5 dataset "example-2": shape (3,), type "|O">

In [46]: f['example-2'][0]
Out[46]: b'foo'

In [47]: f['example-2'][0].decode('utf-8')
Out[47]: 'foo'
```

skbio/io/format/embedding.py
wasade: In `demux` if I remember, what we did was overallocate, trigger resize only when needed, and resize once more when done. This reduces the number of reallocs with a minor amount of book keeping. 

If these data are already in memory, then exhausting the generator won't impact memory, and would allow writing in bulk

skbio/embedding/_embedding.py
mortonjt: As discussed previously, let's do embedding -> sequence embedding
The reason being is that there are some attributes in sequence that we don't want to expose in sequence embedding (i.e. positional indexing for sequence objects doesn't make much sense for sequence embeddings, since the embeddings are generated using the entire sequence, the context is important).

skbio/embedding/__init__.py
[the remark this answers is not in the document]
mortonjt: Oh .. I see.  Yea we can definitely define `ProteinEmbedding.example`.  Maybe we should do this for other objects as well?

skbio/io/format/embedding.py
mortonjt: If the data are in memory, its not a problem. The issue is if we are doing with >100M proteins, that'll blow past 1TB of memory, so we would want to dynamically allocate these objects to disk in a streaming fashion.

wasade: But this is on write, so at least the current item being yielded is in memory. Is it the case that we cannot assume other items in the generator are not in memory? Is the intended workflow something like the following?

```python
def transform(something):
    for stuff in skbio.read(something):
        cooler_stuff = do_the_thing(stuff)
        yield cooler_stuff

skbio.write(transform(a_thing), format='embedding')
```

?

skbio/embedding/_embedding.py
```diff
@@ -0,0 +1,43 @@
...
+
+class Embedding(SkbioObject):
+    """Store embeddings for a biological object."""
+
+    @property
+    def embedding(self):
+        return self._embedding
+
+    @property
+    def ids(self):
```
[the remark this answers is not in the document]
wasade: Oh I see, this is an `id` in the sense that an ASV is an `id`. The naming is a bit confusing, as often IDs do not have meaning but are for establishing linking relationships. Should this be renamed to `peptides` or `sequences`?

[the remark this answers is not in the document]
mortonjt: Yes we can add a `residues` method to make this more transparent on a per object basis. But maybe we should have an `id` method in the `Embedding` ABC so that all subclasses can still be queried with the same attribute?

skbio/io/format/embedding.py
mortonjt: Basically yes, let me rewrite it into the typical workflow that I'm envisioning
```python
def encode_protein_from_llm(fasta_filepath):
    for protein in Protein.read(fasta_filepath):
        embedding = LLM(protein)   # outputs a ProteinEmbedding object
        yield embedding

embeddings = encode_protein_from_llm(fasta_file)
skbio.io.write(embeddings, format='embed', into=writable_emb_path)
```

skbio/io/format/tests/test_embedding.py
mortonjt: Is that the case? `ignore_cleanup_errors=True` isn't being passed in the biom tests https://github.com/scikit-bio/scikit-bio/blob/main/skbio/io/format/tests/test_biom.py#L23

skbio/io/format/embedding.py
mortonjt: So it turns out that it'll be very complicated to clone generators (in order to figure out how much space to allocate).
https://stackoverflow.com/a/29835957

So let me proceed with the original suggestion and only resize only as necessary (this will amortize the runtime).

skbio/io/format/embedding.py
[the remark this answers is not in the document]
mortonjt: Hi @wasade I'm circling back to this and am still a little confused about the issue. 
The amino acid sequences are all ascii representations, so utf-8 is not needed for this specific use-case.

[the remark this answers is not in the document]
qiyunzhu: A side note: scikit-bio sequence characters are currently limited to ASCII (0-127). I had a thought to relax it to any length of alphabet. But this is of low priority.

skbio/io/format/tests/test_embedding.py
wasade: It was necessary on another project, or possibly for windows support somewhere, but I can't remember exactly why right now

skbio/io/format/embedding.py
wasade: I do recommend testing overallocate then reduce at the end as realloc is often expensive

skbio/io/format/embedding.py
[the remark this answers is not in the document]
wasade: Just store bytes then. It's super weird to see the null character, and remember that `str()` is `utf-8`. The example code I put above shows writing bytes without having to deal with encoding

skbio/embedding/__init__.py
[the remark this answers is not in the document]
valentynbez: For now I think it's okay, other sequence embeddings are not that popular yet. It can always be generalised and abstracted if users show interest to such functionality/

skbio/io/format/embedding.py
mortonjt: Yes adding dtype information could be extremely useful for dealing with 16/8/4-bit precision tensors.  I'm not as sure about the dimensionality, since that can be explicitly inferred from the embeddings.

wasade: This may be a challenge point. `bfloat16` for example is I believe still in [RFC](https://forum.hdfgroup.org/t/hdf5-rfc-adding-support-for-16-bit-floating-point-and-complex-number-datatypes-to-hdf5/11975) stage for HDF5

skbio/io/format/embedding.py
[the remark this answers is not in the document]
mortonjt: It turned out the null bytes were arising due to a type mismatch. The `create_dataset` function was allocating `np.int32`, when it should have been `np.uint8`. Replacing with the appropriate type resolves the need to replace null bytes

skbio/embedding/tests/test_protein.py
```diff
@@ -0,0 +1,75 @@
...
+        with self.assertRaises(ValueError):
+            ProteinEmbedding(self.emb, self.seq + "A")
+
+    def test_repr(self):
+        emb, s = self.emb, self.seq
+        p_emb = ProteinEmbedding(emb, s)
+        self.assertEqual(p_emb.embedding.shape, (62, 1024))
+        res_rstr = repr(p_emb)
+        exp_rstr = (
+            ("ProteinEmbedding\n"
```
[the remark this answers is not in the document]
mortonjt: I'm going to go head and delete this entire unittest. It is redundant anyways, since we have a doctest for this

skbio/embedding/_protein.py
```diff
@@ -0,0 +1,100 @@
...
+            sequence = str(sequence)
+
+        if " " in sequence:
+            sequence = sequence.replace(" ", "")
+
+        # make sure that the embedding has the same length as the sequence
+        sequence_len = len(sequence)
+        if embedding.shape[0] != sequence_len:
+            raise ValueError(
+                f"The embedding ({embedding.shape[0]}) must have the "
```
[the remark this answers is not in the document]
mortonjt: Yes, because there are often extra tokens (i.e. <start>, <end> tokens). The lengths are only comparable after clipping

skbio/io/format/tests/test_embedding.py
mortonjt: I'm not able to pass this in.  I'm getting a `TypeError: __init__() got an unexpected keyword argument 'ignore_cleanup_errors' error with my version of python. Given that this isn't in the biom tests perhaps we can offload this to a future PR if necessarily

skbio/io/format/embedding.py
```diff
@@ -0,0 +1,212 @@
...
+    # Note that Wikipedia specifies: "\211HDF\r\n\032\n" which is an ordinal form:
+    # >>> ord('\211')
+    # 137
+    # >>> ord('\x89')
+    # 137
+    # >>> ord('\032')
+    # 26
+    # >>> ord('\x1a')
+    # 26
+    if magic == b"\x89HDF\r\n\x1a\n":
```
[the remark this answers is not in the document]
mortonjt: Raised an issue

qiyunzhu:
@mortonjt Great job! As we discussed, I will merge.

Issue #1974, opened 2024-03-17
Efficient sequence alignment path data structure


qiyunzhu:
### Discussed in https://github.com/scikit-bio/scikit-bio/discussions/1973

@mortonjt @wasade Will appreciate your thoughts!

<div type='discussions-op-text'>

<sup>Originally posted by **qiyunzhu** March 17, 2024</sup>
Here I am describing the plan to implement a multiple and pairwise sequence **alignment path** data structure in scikit-bio. In summary:

- Central idea: [run-length encoding](https://en.wikipedia.org/wiki/Run-length_encoding) (RLE) and [bit array](https://en.wikipedia.org/wiki/Bit_array).
- It is detached from the original sequences.
- It is highly memory efficient, compared with tabular alignments.
- It permits fully vectorized operations, therefore time efficient.
- It can be efficiently converted to/from various formats, such as CIGAR (supported by many alignment libraries), tabular (original format in scikit-bio), indices (Biotite), coordinates (Biopython), etc.

I guess there might be similiar implementations in the past, although I haven't come across them through limited literature search. Will appreciate it if anyone can inform me of the relevant literature.

A Jupyter notebook implementing a prototype of this data structure and relevant functionality is attached.
[align_path.zip](https://github.com/scikit-bio/scikit-bio/files/14628752/align_path.zip)

Will appreciate any feedback from the team and community!

---

For example, we have a multiple sequence alignment (MSA) consisting of three aligned DNA sequences.

```
seq1: CGGTCGTAACGCGTA---CA
seq2: CAG--GTAAG-CATACCTCA
seq3: CGGTCGTCAC-TGTACACTA
```

### Suggested solution

I suggest the following solution, which is essentially a combination of [**run-length encoding**](https://en.wikipedia.org/wiki/Run-length_encoding) (RLE) and [**bit array**](https://en.wikipedia.org/wiki/Bit_array). Through reasoning, comparison, and benchmarking, I concluded that it is the most suitable for the variety of target applications of scikit-bio.

We divide the alignment position-wise into **segments** based on gap status. Each segment represents a consecutive block that has a consistent gap status in all sequences. In this example, the alignment can be divided into seven segments.

```
seq1: CGG TC GTAAC G CGTA --- CA
seq2: CAG -- GTAAG - CATA CCT CA
seq3: CGG TC GTCAC - TGTA CAC TA
```

Calculate the **lengths** of segments:

```
lens: 3 2 5 1 4 3 2
```

[...]

PR #2011, opened 2024-04-18
Efficient sequence alignment path data structure

mataton:
This PR will implement the sequence alignment path data structure as outlined in issue #1974. 

### DIFF
--- a/skbio/alignment/_path.py
+++ b/skbio/alignment/_path.py
@@ -317,9 +317,9 @@ class AlignPath(SkbioObject):
         Examples
         --------
         >>> from skbio.alignment import AlignPath
-        >>> path = Align.Path(lengths=[1, 2, 2, 1],
-        ...                   states=[0, 5, 2, 6],
-        ...                   starts=[0, 0, 0])
+        >>> path = AlignPath(lengths=[1, 2, 2, 1],
+        ...                  states=[0, 5, 2, 6],
+        ...                  starts=[0, 0, 0])
         >>> path.to_indices()
         array([[ 0, -1, -1,  1,  2,  3],
                [ 0,  1,  2, -1, -1, -1],
@@ -379,10 +379,10 @@ class AlignPath(SkbioObject):
         --------
         >>> import numpy as np
         >>> from skbio.alignment import AlignPath
-        >>> indices = np.array([[ 0, -1, -1,  1,  2,  3],
-        ...                     [ 0,  1,  2, -1, -1, -1],
-        ...                     [ 0, -1, -1,  1,  2, -1]])
-        >>> path = AlignPath.from_indices()
+        >>> indices = np.array([[0, -1, -1,  1,  2,  3],
+        ...                     [0,  1,  2, -1, -1, -1],
+        ...                     [0, -1, -1,  1,  2, -1]])
+        >>> path = AlignPath.from_indices(indices)
         >>> path
         AlignPath
         Shape(sequence=3, position=6)


mataton:
@qiyunzhu @mortonjt This PR is more than 90% completed. A preliminary review would be appreciated.

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+
+import numpy as np
+
+from skbio._base import SkbioObject
+from skbio.util._decorator import overrides, classonlymethod
+
+
+class AlignPath(SkbioObject):
+    def __init__(self, lengths, states, starts):
+        """Create an alignment path from segment lengths and states.
```
mortonjt, line 17:
```suggestion
        r"""Create an alignment path from segment lengths and states.
```

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+
+        See Also
+        --------
+        skbio.sequence.Sequence
+        skbio.alignment.TabularMSA
+
+        """
+        # Number of sequences needs to be explicitly provided, because the packed bits
+        # does not contain this information. (It is merely in multiples of 8.)
+        self.lengths = np.asarray(lengths, dtype=int)
```
mortonjt, line 37:
```suggestion
        self.lengths = np.asarray(lengths, dtype=dtype)
```
I wonder if we should have a `dtype` flag to be passed in, and have `np.uint32` be the default. 
But again, I'm not sure if this data structure is sufficient for genome-scale alignments (it can't recombination right?) Otherwise, these alignments could get *huge* (i.e. trillion bp genomes seen in some plants)

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+
+        # TODO: An additional parameter `starts` should record the starting positions
+        # of the alignment in each sequence (shape: (n_seqs,)). This is important
+        # especially for local alignments (e.g., search for a short read in a genome
+        # sequence). If it is provided, `n_seqs` is no longer necessary.
+        # self.starts = starts
+
+        # TODO: Needs to think about whether reverse complemented (Boolean array of
+        # (n_seqs,)) should be included as a parameter. It is only relevant for
+        # nucleotide sequences.
```
mortonjt, line 61: I think we can scrap this requirement -- users should in theory reverse complement their own sequences, right?

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+
+        # Shape is n_seqs (rows) x n_positions (columns), which is consistent with
+        # TabularMSA
+        self.shape = (n_seqs, n_positions)
+
+        # TODO: An additional parameter `starts` should record the starting positions
+        # of the alignment in each sequence (shape: (n_seqs,)). This is important
+        # especially for local alignments (e.g., search for a short read in a genome
+        # sequence). If it is provided, `n_seqs` is no longer necessary.
+        # self.starts = starts
```
mortonjt, line 57: Has this been addressed?

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+
+        # TODO: Needs to think about whether reverse complemented (Boolean array of
+        # (n_seqs,)) should be included as a parameter. It is only relevant for
+        # nucleotide sequences.
+
+    def __str__(self):
+        """Return string representation of this AlignPath."""
+        # Not sure if this makes sense for this class, but it is needed for all
+        # SkbioObjects.
+        return self.__repr__()
```
mortonjt, line 67: It may not be a bad idea to return the full CIGAR string for a pairwise alignment. Not sure what the analogy is for an MSA

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+            )
+            lens = np.append(idx[1:] - idx[:-1], ints.shape[1] - idx[-1])
+            ints = ints[:, idx]
+
+        # return per-segment lengths and bits
+        return cls(lens, ints, np.zeros(bits.shape[0], dtype=int))
+
+    @classonlymethod
+    def from_tabular(cls, msa):
+        """Create an alignment path from a `TabularMSA` object."""
```
mortonjt, line 124: Is it worthwhile to have a `to_tabular` version method? It maybe ideal for easier manipulation / visualization

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+            f"<{self.__class__.__name__}, shape: {self.shape}, lengths: "
+            f"{self.lengths}, states: {self.states}"
+        )
+
+    def subset(self):
+        """Select subset of sequences from AlignPath.
+
+        Better to have ability to index an AlignPath for particular sequences,
+        then compress these sequences using something like _fix_arrays"""
+        pass
```
mortonjt, line 81: Should we delete this, since it isn't implemented? We can raise a `NotImplementedError` otherwise

skbio/alignment/_util.py
```diff
@@ -0,0 +1,51 @@
+import numpy as np
```
mataton, line 1: @mortonjt @qiyunzhu This file is incomplete and untested, feel free to ignore for now.

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+# The full license is in the file LICENSE.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import numpy as np
+
+from skbio._base import SkbioObject
+from skbio.util._decorator import overrides, classonlymethod
+
+
+class AlignPath(SkbioObject):
```
mortonjt, line 15: Is it worthwhile to flesh out a practical use-case of this object? Pseudocode in the issue tracker may be useful to help us understand how this code is going to be used (i.e. pseudocount for progressive alignment)

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+
+    def to_bits(self):
+        """Unpack states into an array of bits."""
+        return np.unpackbits(
+            np.atleast_2d(self.states), axis=0, count=self.shape[0], bitorder="little"
+        )
+
+    @classonlymethod
+    def from_bits(cls, bits):
+        """Create an alignment path from a bit array (0 - char, 1 - gap)."""
```
mortonjt, line 91:
```suggestion
        r"""Create an alignment path from a bit array (0 - char, 1 - gap)."""
```

skbio/alignment/_path.py
mortonjt, line 124:
```suggestion
        r"""Create an alignment path from a `TabularMSA` object."""
```

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+
+        # Get gap character code.
+        gap_chars = [ord(x) for x in msa.dtype.gap_chars]
+
+        # Identify gap positions, and convert them into a bit array, then create an
+        # alignment path based on it.
+        return cls.from_bits(np.isin(byte_arr, gap_chars))
+
+    def to_indices(self, gap=-1):
+        """Generate an array of indices of characters in the original sequences.
```
mortonjt, line 138:
```suggestion
        r"""Generate an array of indices of characters in the original sequences.
```

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+            keep = np.repeat(self.states == 0, self.lengths)
+            return idx[:, keep]
+        elif gap == "mask":
+            return np.ma.array(idx, mask=(1 - pos))
+        else:
+            return np.where(pos, idx, gap)
+
+    @classonlymethod
+    def from_indices(cls, indices, gap=-1):
+        """Create an alignment path from character indices in the original sequences.
```
mortonjt, line 163:
```suggestion
        r"""Create an alignment path from character indices in the original sequences.
```

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+
+        gap can be a value or "mask", but cannot be "del".
+        """
+        if gap == "mask":
+            return cls.from_bits(np.ma.getmask(indices))
+        else:
+            return cls.from_bits(indices == gap)
+
+    def to_coordinates(self):
+        """Generate an array of segment coordinates in the original sequences."""
```
mortonjt, line 173:
```suggestion
        r"""Generate an array of segment coordinates in the original sequences."""
```

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+        lens = self.lengths * (1 - self.to_bits())
+        col0 = np.zeros((self.shape[0], 1), dtype=int)
+        lens = np.append(col0, lens, axis=1)
+        return lens.cumsum(axis=1)
+
+    @classonlymethod
+    def from_coordinates(cls, coords):
```
mortonjt, line 180:
```suggestion
    def from_coordinates(cls, coords):
        r"""Generate the an alignment path from an array of segment coordinates."""
```
Don't forget the docstring, this won't be documented otherwise

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+        if ints.shape[0] == 1:
+            ints = np.squeeze(ints, axis=0)
+        return cls(lens, ints, np.zeros(diff.shape[0], dtype=int))
+
+
+class PairAlignPath(AlignPath):
+    @overrides(AlignPath)
+    @classonlymethod
+    def from_bits(cls, bits):
+        """Create an alignment path from a bit array."""
```
mortonjt, line 194:
```suggestion
        r"""Create an alignment path from a bit array."""
```

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+        # This should be faster than the generic solution I guess.
+        # TODO: Pending benchmark and optimization.
```
mortonjt, line 196: Should we perform a benchmark before merging?

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+            raise ValueError(
+                "For pairwise alignment, `states` must only contain "
+                "zeros, ones, or twos."
+            )
+        return np.stack([self.states & 1, self.states >> 1])
+
+    def to_cigar(self, seqs=None):
+        """Get a CIGAR string representing the pairwise alignment path.
+
+        seqs: If provided, will distinguish match (=) and mismatch (X). Otherwise,
```
mortonjt, line 224:
```suggestion
    Parameters
    ----------
        seqs: If provided, will distinguish match (=) and mismatch (X). Otherwise,
```

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+        # TODO: Pending benchmark and optimization.
+        if not np.all(np.isin(self.states, [0, 1, 2])):
+            raise ValueError(
+                "For pairwise alignment, `states` must only contain "
+                "zeros, ones, or twos."
+            )
+        return np.stack([self.states & 1, self.states >> 1])
+
+    def to_cigar(self, seqs=None):
+        """Get a CIGAR string representing the pairwise alignment path.
```
mortonjt, line 222:
```suggestion
        r"""Get a CIGAR string representing the pairwise alignment path.
```

skbio/alignment/_path.py
mortonjt: Also need to document the parameteres

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+                    cigar += "X"
+            return self._run_length_encode(cigar)
+        else:
+            for i, length in enumerate(lengths):
+                cigar += str(length) + codes[gaps[i]]
+            return cigar
+
+    @classonlymethod
+    def from_cigar(cls, cigar):
+        """Create a pairwise alignment path from a CIGAR string."""
```
mortonjt, line 251:
```suggestion
        r"""Create a pairwise alignment path from a CIGAR string."""
```

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+            else:
+                raise ValueError("Invalid characters in CIGAR string.")
+        if fix_arrs:
+            lengths, gaps = cls._fix_arrays(
+                lengths=np.array(lengths), gaps=np.array(gaps)
+            )
+        return cls(np.asarray(lengths), np.asarray(gaps), [0, 0])
+
+    def _run_length_encode(self, input_string):
+        """Perform run length encoding on a string."""
```
mortonjt, line 289:
```suggestion
        r"""Perform run length encoding on a string."""
```
Last one, make sure all of the docstrings are formatted appropriately

skbio/alignment/_tabular_msa.py
```diff
@@ -2477,3 +2477,16 @@ def _is_sequence_axis(self, axis):
     @overrides(PositionalMetadataMixin)
     def _positional_metadata_axis_len_(self):
         return self.shape.position
+
+
+class TabularMSA2(TabularMSA):
```
mortonjt, line 2482: Can you clarify where this is coming from? Do we need this? Or can we just include this in `TabularMSA`?

skbio/alignment/_util.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def align_score(path, seqs, match=2, mismatch=-3, gap_open=-5, gap_extend=-2):
```
mortonjt, line 4: There are already quite a few alignment methods in skbio, so we need to differientiate. Maybe `pairwise_align_score`?

skbio/alignment/_util.py
```diff
@@ -0,0 +1,51 @@
...
+    # get total score
+    return (
+        match * n_matches
+        + mismatch * n_mismatches
+        + gap_open * n_gap_opens
+        + gap_extend * n_gap_extends
+    )
+
+
+def align_scores(path, seqs, match, mismatch, gap_open, gap_extend):
```
mortonjt, line 36: Same, maybe `multiple_align_score`?

skbio/alignment/_util.py
```diff
@@ -0,0 +1,51 @@
...
+    """Calculate pairwise alignment scores of a multiple alignment and return a
+    matrix."""
+    pass
+
+
+def align_sp_scores(path, seqs, match, mismatch, gap_open, gap_extend):
```
mortonjt, line 42: Maybe `multiple_align_sp_score`

mortonjt: What do you think of having a `multiple_dm_score` that instead passes in distance matrix (which could be sum-of-pairs, or any distance metric). This could look like

`def align_sp_scores(path, seqs, match_dm, mismatch, gap_open, gap_extend):`

Smith-waterman can work with just gaps + a matrix distance matrix (either blosum, or an index-specific distance matrix).

skbio/alignment/_util.py
```diff
@@ -0,0 +1,51 @@
...
+    """Calculate sum-of-pairs (SP) score of a multiple alignment."""
+    pass
+
+
+def separate_rle(msa):
```
mortonjt, line 47: doc?

skbio/alignment/_path.py
mortonjt: Actually, the `to_cigar` method would round this out.

skbio/alignment/__init__.py
```diff
@@ -218,11 +218,14 @@
     local_pairwise_align_ssw,
 )
 from skbio.alignment._ssw_wrapper import StripedSmithWaterman, AlignmentStructure
+from skbio.alignment._path import AlignPath, PairAlignPath
```
mortonjt, line 221: Make sure that the `toctree` above is documented. Otherwise these objects won't show up in the docs

skbio/alignment/_path.py
```diff
@@ -0,0 +1,308 @@
...
+        lens = diff[bits.argmin(axis=0), np.arange(diff.shape[1])]
+        ints = np.packbits(bits, axis=0, bitorder="little")
+        if ints.shape[0] == 1:
+            ints = np.squeeze(ints, axis=0)
+        return cls(lens, ints, np.zeros(diff.shape[0], dtype=int))
+
+
+class PairAlignPath(AlignPath):
+    @overrides(AlignPath)
+    @classonlymethod
```
mortonjt, line 192: It looks like some of the unittests still throws an error for python=3.8
```
AttributeError: 'classonlymethod' object has no attribute '__name__'
```
Not sure why this is being thrown, but we'll want to investigate this. Does this need the `@overrides(AlignPath)` decorator?

mortonjt (changes requested):
Addressed some comments. Its a relatively incomplete review, so we'll want at least one more pass.

Also, don't forget to add this to the `index.rst` -- this won't show up in the documentation otherwise.

skbio/alignment/_path.py
qiyunzhu: They usually should. Although multiple programs like BLAST and Bowtie2 have the option to search both forward and reverse strands.

skbio/alignment/_path.py
mataton: @qiyunzhu You may have a clearer or more well defined use case in mind than I do?

skbio/alignment/_path.py
mataton: Yes, I can talk with @qiyunzhu in person and flesh this out.

mataton:
@qiyunzhu One question I have regarding `align_scores` or calculating a matrix of pairwise alignment scores from an MSA: If at a single position in the pairwise alignment there are gaps in each sequence, how is this scored? From the example in the original [issue](https://github.com/scikit-bio/scikit-bio/issues/1974):

````python
seq2 = 'CAG--GTAAG-CATACCTCA'
seq3 = 'CGGTCGTCAC-TGTACACTA'
````

At the position where both sequences contain a gap, how is this interpreted or scored?

File: skbio/alignment/_tabular_msa.py (hunk 21 above)
Comment (mataton) [reply skbio/alignment/_tabular_msa.py:2482]: This will be included in `TabularMSA`, will address this in future commit.

File: skbio/alignment/_util.py [hunk 28]
```diff
@@ -0,0 +1,70 @@
...
+    # agnostic of the sequences
+    gap_segs = path.states != 0
+    n_gap_opens = gap_segs.sum()
+    n_gap_extends = (gap_segs * path.lengths).sum() - n_gap_opens
+
+    # step 2: calculate match & mismatch scores
+    # based on the characters within the non-gap segments
+    # This will work for a strictly pairwise alignment, where it is certain that
+    # gaps will not occur in the same position in both sequences.
+    seq1, seq2 = (x._bytes for x in seqs)
```
mataton, line 32: @qiyunzhu This code is much simpler than the original, which has been commented out below. I think the original intended to account for situations where there are gaps in the same position in the two sequences. My understanding is that in a purely pairwise alignment, this will never occur. Is that correct? I feel this function should just be about a straightforward pairwise alignment score. Handling other cases could be put into the other functions in this file. What do you think?

skbio/alignment/__init__.py
```diff
@@ -218,11 +220,14 @@
...
 
 __all__ = [
     "TabularMSA",
     "StripedSmithWaterman",
     "AlignmentStructure",
```
qiyunzhu, line 228: Just realized that there is an `AlignmentStructure` class. Outdated. Should be removed when we remove the SSW wrapper code. The new `AlignPath` class can replace it.

skbio/alignment/_path.py
qiyunzhu: I think we should do it but it is outside the scope of the current PR. Specifically, we can:
1. Generate a pairwise alignment using external Python libraries like BioPython or Biotite and convert the result into `AlignPath`.
2. Generate a multiple alignment using external programs like ClustalO or MAFFT and read the result (multi-fasta format) into `AlignPath`.
3. Perform pairwise or multiple alignment using scikit-bio iteself (to be re-modeled) and return an `AlignPath` object.
4. Manually create a plausible alignment and calculate its alignment score.

There may be variants of each case. For example, for 3, we can return all optimal alignments instead of just one arbitrary alignment (pseudocode):

```python
for aln in pairwise_align(seq1, seq2, return_all=True):
    scores.append(aln.align_score())
```

skbio/alignment/_path.py
qiyunzhu: This might be a good point. It may not be a concern in Linux or MacOS, but in Windows (even 64-bit), `int` defaults to `np.int32`, the maximum of which is 2,147,483,647 (i.e., 2 Gbp), which is shorter than many genomes. Perhaps we can default it to `np.uint64`. I am not sure if we need to create a dtype flag. This may confuse non-technical users.

skbio/alignment/_path.py
qiyunzhu: I don't know if there is a CIGAR equivalent for multiple sequence alignments. I guess our `AlignPath` is the closest data structure.

skbio/alignment/_path.py
qiyunzhu: I benchmarked the conversion part on real-world data and it is pretty fast. Matt added the validation.

skbio/alignment/_path.py
```diff
@@ -0,0 +1,305 @@
...
+    def from_bits(cls, bits):
+        r"""Create an alignment path from a bit array."""
+        # This should be faster than the generic solution I guess.
+        # TODO: Pending benchmark and optimization.
+
+        # Ensure bits is a 2D numpy array of ones and zeros.
+        if not isinstance(bits, np.ndarray) or bits.ndim != 2 or bits.shape[1] == 0:
```
qiyunzhu, line 194: `bits` may be array-like (e.g., a NumPy array or a Python list), and one can use `np.atleast_2d` to convert it into a 2-D array.

skbio/alignment/_path.py
```diff
@@ -0,0 +1,305 @@
...
+            raise ValueError("Input 'bits' must be a non-empty 2D numpy array.")
+        if not np.all(np.logical_or(bits == 0, bits == 1)):
```
qiyunzhu, line 196: This seems to be a brilliant solution. It outperforms the best GPT-4o could do (it suggests `np.isin(arr, [0, 1]).all()`).

Minor suggestion: `np.logical_or(bits == 0, bits == 1).all()` should be slightly faster than `np.all(...)`.

skbio/alignment/_path.py
```diff
@@ -0,0 +1,305 @@
...
+
+        """
+        # Number of sequences needs to be explicitly provided, because the packed bits
+        # does not contain this information. (It is merely in multiples of 8.)
+        self.lengths = np.asarray(lengths, dtype=int)
+        n_positions = self.lengths.sum()
+        self.states = np.atleast_2d(np.asarray(states, dtype=np.uint8))
+
+        # start positions
+        self.starts = np.asarray(starts, dtype=int)
```
mataton, line 42: Is this okay as a regular `int`, or should we change this to `np.uint64` as well? As mentioned previously ~2Gbp is shorter than many genomes, and it is possible that the start position of a sequence will be beyond that point in the genome.

skbio/alignment/_path.py
mataton: I can scrap it for now. If we want to implement in the future we always can.

skbio/alignment/_path.py
qiyunzhu: Good point. I would suggest changing into `np.uint64` as well. Meanwhile, I don't think supporting negative start positions is necessary (I can't think of a use case).

skbio/alignment/_path.py
mataton: Thank you!

skbio/alignment/_util.py
qiyunzhu: This `separate_rle` is a completely different data structure. Before we sufficiently explore it, I would suggest that we don't include it in this PR.

skbio/alignment/_path.py
```diff
@@ -0,0 +1,341 @@
...
+
+        See Also
+        --------
+        skbio.sequence.Sequence
+        skbio.alignment.TabularMSA
+
+        """
+        # Number of sequences needs to be explicitly provided, because the packed bits
+        # does not contain this information. (It is merely in multiples of 8.)
+        self.lengths = np.asarray(lengths, dtype=np.int64)
```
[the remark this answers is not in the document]
qiyunzhu: Good to know. Thanks for exploring!

skbio/alignment/_path.py
mataton: Removing the `overrides` decorator solved this issue without affecting functionality.

skbio/alignment/_path.py
```diff
@@ -0,0 +1,410 @@
...
+            if len(seqs[0]) != self.shape[1] or len(seqs[1]) != self.shape[1]:
+                raise ValueError(
+                    f"At least one of sequence lengths "
+                    f"{len(seqs[0]), len(seqs[1])} does not match "
+                    f"alignment length ({self.shape[1]})."
+                )
+            # Also need to ensure that the sequences provided actually match the
+            # lengths and states of the instantiated object.
+            byte_arr = np.stack([x._bytes for x in seqs])
+            gap_chars = [ord(x) for x in seqs[0].gap_chars]
```
qiyunzhu, line 290: I think that there is a pre-calculated class property `_gap_codes`.

skbio/alignment/_path.py
```diff
@@ -0,0 +1,410 @@
...
+                    "Provided sequences do not match existing states. Consider the "
+                    "order of the provided sequences."
+                )
+
+        cigar = ""
+        lengths = self.lengths
+        gaps = np.squeeze(self.states)
+        codes = ["M", "I", "D", "P"]
+        if seqs is not None:
+            query = str(seqs[0])
```
qiyunzhu, line 310: Instead of comparing strings, we can directly compare the bytes underlying the strings.

skbio/alignment/_path.py
```diff
@@ -0,0 +1,410 @@
...
+        ----------
+        seqs : list of skbio.sequence
+            Pair of aligned sequences to generate cigar string. If provided, will
+            distinguish match (=) and mismatch (X). Otherwise, will uniformly note
+            them as (mis)match (M).
+        """
+        # If `seqs` is provided, ensure that the sequence lengths match the alignment
+        # length.
+        if seqs:
+            if len(seqs[0]) != self.shape[1] or len(seqs[1]) != self.shape[1]:
```
qiyunzhu, line 278: I am under impression that this code assumes that the two sequences are of the same length. But in fact they don't have to be. The logic of this method may be revised.

skbio/alignment/_path.py
```diff
@@ -0,0 +1,410 @@
...
+                )
+
+        cigar = ""
+        lengths = self.lengths
+        gaps = np.squeeze(self.states)
+        codes = ["M", "I", "D", "P"]
+        if seqs is not None:
+            query = str(seqs[0])
+            ref = str(seqs[1])
+            for qchar, rchar in zip(query, ref):
```
qiyunzhu, line 312: See above. This code assumes that the two sequences are of the same length, which may not be the case.

skbio/alignment/_path.py
```diff
@@ -0,0 +1,410 @@
...
+                    lengths.append(current_length)
+                gaps.append(mapping[char])
+                current_length = 0
+                no_ones = True
+            else:
+                raise ValueError("Invalid characters in CIGAR string.")
+        lengths, gaps = cls._fix_arrays(lengths=np.array(lengths), gaps=np.array(gaps))
+        return cls(lengths, gaps, [0, 0])
+
+    def _run_length_encode(self, string_in):
```
qiyunzhu, line 374: This method does not use `self`. Therefore, it does not have to be a method of the class. Instead, it can be a standalone function.

skbio/alignment/_path.py
```diff
@@ -0,0 +1,410 @@
...
+        """
+        input_arr = np.array(list(string_in))
+        idx = np.append(0, np.where(input_arr[:-1] != input_arr[1:])[0] + 1)
+        count = np.diff(np.concatenate((idx, [len(string_in)])))
+        unique = input_arr[idx]
+        encoded_str = "".join(str(c) + u for c, u in zip(count, unique))
+
+        return encoded_str
+
+    def _fix_arrays(lengths, gaps):
```
qiyunzhu, line 390: Likewise, this can be a standalone function.

skbio/alignment/_path.py
```diff
@@ -0,0 +1,407 @@
...
+
+    def to_cigar(self, seqs=None):
+        r"""Get a CIGAR string representing the pairwise alignment path.
+
+        Parameters
+        ----------
+        seqs : list of skbio.sequence
+            Pair of aligned sequences to generate cigar string. If provided, will
+            distinguish match (=) and mismatch (X). Otherwise, will uniformly note
+            them as (mis)match (M).
```
qiyunzhu, line 273: Should note that seq1 is "query", seq2 is "reference".

skbio/alignment/_path.py
```diff
@@ -146,13 +146,20 @@ def to_indices(self, gap=-1):
 
         Parameters
         ----------
-        gap : -1, "del", or "mask", optional
-            If -1, replace gaps with this value. If "del", delete columns that have any
-            gap. If "mask", mask gaps. Default is -1.
+        gap : int, float, "del", or "mask", optional
```
qiyunzhu, line 149: I think an index cannot be a float. It is possible to cast floats into integers, but it shouldn't be a normal use case.

skbio/alignment/_path.py
```diff
@@ -0,0 +1,446 @@
...
+                seq2 = np.array([ord(char) for char in str(seqs[1])])
+            elif isinstance(seqs[0], GrammaredSequence) and isinstance(
+                seqs[1], GrammaredSequence
+            ):
+                seq1 = seqs[0]._bytes
+                seq2 = seqs[1]._bytes
+            else:
+                raise ValueError("`seqs` must be of type string or Sequence object.")
+
+            idx1, idx2 = [int(x) for x in self.starts]
```
qiyunzhu, line 331: I thought that `starts` is always an integer array, therefore `int` is not necessary?

skbio/alignment/_path.py
```diff
@@ -0,0 +1,446 @@
...
+            if isinstance(seqs[0], str) and isinstance(seqs[1], str):
+                seq1 = np.array([ord(char) for char in str(seqs[0])])
+                seq2 = np.array([ord(char) for char in str(seqs[1])])
+            elif isinstance(seqs[0], GrammaredSequence) and isinstance(
+                seqs[1], GrammaredSequence
+            ):
+                seq1 = seqs[0]._bytes
+                seq2 = seqs[1]._bytes
+            else:
+                raise ValueError("`seqs` must be of type string or Sequence object.")
```
qiyunzhu, line 329: Can be "`seqs` must be strings or Sequence objects."

mataton:
@qiyunzhu @mortonjt I believe this PR is ready for the final round of review. As an update from the initial round, there have been a few major changes. Most notably, all of the scoring functions such as `pairwise_align_scores` have been removed from this PR and will be addressed in a future separate PR. Additionally, the `to_cigar` method has been entirely rewritten. Other than those, the changes have addressed all comments from the previous rounds of review. Thank you for your help on this!

skbio/alignment/_path.py
mataton: `starts` is defined initially as `self.starts = np.asarray(starts, dtype=np.uint64)`, but when I unpacked `starts` with `idx1, idx2 = self.starts`, `idx1` and `idx2` became `float`s when we add them to `length` in line 335. Using `int` here solves this problem.

For example:
````python
>>> starts = np.array([0, 1], dtype=np.uint64)
>>> lengths = np.array([3, 4], dtype=np.int64)
>>> a, b = starts
>>> c, d = lengths
>>> type(a)
<class 'numpy.uint64'>
>>> type(c)
<class 'numpy.int64'>
>>> a + c
3.0
````

The other solution is to use `np.int64` instead of `np.uint64` when we define `starts` at the beginning of the class. `np.uint64` has strange behavior as I have mentioned elsewhere in this PR.

Comment (qiyunzhu) [reply skbio/alignment/_path.py:329]: Thanks for checking this! I think that using `np.int64` instead of `np.uint64` would be the best solution.

File: skbio/alignment/_path.py [hunk 44]
```diff
@@ -0,0 +1,447 @@
...
+                    idx1 += length
+                    idx2 += length
+                elif state == 1:
+                    cigar.append(str(length) + "I")
+                    idx2 += length
+                elif state == 2:
+                    cigar.append(str(length) + "D")
+                    idx1 += length
+                elif state == 3:
+                    cigar.append(str(length) + "P")
```
qiyunzhu, line 355: There is a shorter may that can replace the code between lines 332 to 354:

```python
if state == 0:
    match_arr = ...
else:
    cigar.append(str(length) + codes[state])
flipped = state ^ 3
idx1 += flipped & 1 and length
idx2 += flipped & 2 and length
```

This method uses indexing and bitwise operations to avoid rewriting `cigar.append` and `idx +=` for each condition. It is not necessarily faster than the current solution, nor is it more elegant (bitwise operation is hard to understand!). Therefore, it is only for reference.

skbio/alignment/_path.py
```diff
@@ -0,0 +1,447 @@
...
+            The alignment path created from the given CIGAR string.
+        """
+        # Make sure cigar is not empty.
+        if not cigar:
+            raise ValueError("CIGAR string must not be empty.")
+
+        lengths = []
+        gaps = []
+        current_length = 0
+        mapping = {
```
qiyunzhu, line 381: Likewise, this `mapping` can be a constant outside the method or the class.

mortonjt (approved):
My comments have been addressed

skbio/alignment/_path.py
```diff
@@ -237,29 +265,54 @@ def from_indices(cls, indices, gap=-1):
...
-            return cls.from_bits(np.ma.getmask(indices))
+            return cls.from_bits(
+                np.ma.getmask(indices),
+                [x[np.argmax(x != indices.fill_value)] for x in indices],
+            )
         else:
-            return cls.from_bits(indices == gap)
+            indices = np.asarray(indices)
+            return cls.from_bits(
+                indices == gap, [x[np.argmax(x != gap)] for x in indices]
```
qiyunzhu, line 290: @mataton After some exploration, the best non-Cython solution I learned is:

Replace:

```python
[x[np.argmax(x != gap)] for x in indices]
```

With:

```python
indices[np.arange(indices.shape[0]), np.argmax(indices != gap, axis=1)]
```

This [thread](https://stackoverflow.com/questions/16243955/numpy-first-occurrence-of-value-greater-than-existing-value) has some nice discussions and benchmarks. The most interesting thing is Mr Fooz's reply.

PR #2021, opened 2024-05-06
Protein Vector object

mortonjt:

There are a few linting issues that need to be addressed -- the major modification here was updating the embedding format to handle both protein embeddings and protein vectors (vectors here will allow for fast queries), in addition to conversion utilities discussed in https://github.com/scikit-bio/scikit-bio/issues/2006

This is the last pull request that would need top be merged into skbio before we start putting together the tutorials.


mortonjt:
I wonder if codecov doesn't like it when there are multiple files with the same name (i.e. `skbio.io.format._embedding.py` and `skbio.embedding._embedding`). Renaming these files to confirm.

If so, then the same logic would apply to `skbio.embedding._protein` since there is also `skbio.sequence._protein`.

skbio/io/__init__.py
```diff
@@ -246,6 +246,7 @@ class of Python's standard library. The goal of a ``sniffer`` is two-fold: to
     GFF3FormatError,
     EMBLFormatError,
     BIOMFormatError,
+    EmbedFormatError,
```
mortonjt, line 249: It turned out that this wasn't even being imported -- hinting that we don't have tests that verify any of these errors. 
Raising an issue, since this could be a blind spot when creating new file formats.

mortonjt:
Waiting on feedback regarding codecov. Otherwise, this pull request is ready for review.

Major changes since https://github.com/scikit-bio/scikit-bio/pull/2008
1. Adding in support for `ProteinVector` objects. This will allow for consuming the output of pLMs like TMvec for rapid remote homology search and light-weight protein function annotation. Furthermore, there are a number of conversion mechanisms in place to allow for easy manipulation of these objects.
2. Renamed `skbio.io.format.embedding` -> `skbio.io.format.embed` for consistency. It also helps distinguish between the file format `embed` and the object `embedding`
3. We defined a `bytes()` ABC method to further simplify hdf5 conversion. That way, every class inherienting from `Embedding` already has `bytes()` defined and can write to HDF5
4. We added additional flexibility to the HDF5 format to allow for more relaxes string length requirements. In the previous PR, the string length was required to be equal to the embedding length (which is the case with the `ProteinEmbedding` object). But this is not the case with the `ProteinVector` object and other potential downstream objects (i.e. `MolecularEmbedding` objects).  That being said, we have enabled backwards compatibility, since the increased flexibility slightly increases the HDF5 memory footprint (by adding extra pointers).
5. Fixed exception import that was missed in the previous pull request

skbio/embedding/_embedding.py
```diff
@@ -98,3 +108,158 @@ def __repr__(self):
...
+        if len(vector.shape) == 2:
+            assert vector.shape[0] == 1, (
+                "Only 1 vector per sequence is allowed."
+            )
+
+        seq = np.array([sequence], dtype='O')
+        super(SequenceVector, self).__init__(vector, seq, **kwargs)
+
+    def __str__(self):
+        return str(self._ids[0].decode('ascii'))
```
wasade, line 133: The `decode` here, with the `encode on line 60, is essentially:

`b'asdasd'.decode('ascii').encode('ascii')`

...right? Is there room to reduce or remove some of the decode/encode operations?

skbio/embedding/_embedding.py
```diff
@@ -98,3 +108,158 @@ def __repr__(self):
...
+        )
+        return rstr
+
+    @property
+    def embedding(self):
+        r""" The embedding tensor. """
+        return self._embedding.reshape(1, -1)
+
+    @staticmethod
+    def to_numpy(sequence_vectors : List["SequenceVector"]):
```
wasade, line 176: I don't think we use type annotations elsewhere. If this is kept, shouldn't it be `List[SequenceVector]`?

skbio/embedding/_embedding.py
```diff
@@ -98,3 +108,158 @@ def __repr__(self):
...
+        lens = [len(pv.vector) for pv in sequence_vectors]
+        if not all(ln == lens[0] for ln in lens):
+            raise ValueError("All vectors must have the same length.")
+        data = np.vstack([pv.vector for pv in sequence_vectors])
+        return data
+
+    @staticmethod
+    def to_distance_matrix(sequence_vectors : List["SequenceVector"],
```
wasade, line 184: Same question here

skbio/embedding/_embedding.py
```diff
@@ -98,3 +108,158 @@ def __repr__(self):
...
+        DistanceMatrix
+            A DistanceMatrix object.
+
+        See Also
+        --------
+        DistanceMatrix
+        """
+        data = SequenceVector.to_numpy(sequence_vectors)
+        ids = [str(pv) for pv in sequence_vectors]
+        dm = squareform(pdist(data, metric))
```
wasade, line 208: Should this use the `beta_diversity` driver?

skbio/embedding/_embedding.py
```diff
@@ -98,3 +108,158 @@ def __repr__(self):
...
+        return DistanceMatrix(dm, ids=ids)
+
+    @staticmethod
+    def to_ordination(sequence_vectors : List["SequenceVector"]):
```
wasade, line 212: I'll stop flagging these

skbio/embedding/_embedding.py
```diff
@@ -98,3 +108,158 @@ def __repr__(self):
...
+
+        Returns
+        -------
+        OrdinationResults
+            An Ordination object.
+
+        See Also
+        --------
+        OrdinationResults
+        """
```
wasade, line 229: It may be nice, in a separate PR, to refactor and move this logic to `skbio.stats.ordination` -- it seems generally useful to perform SVD -> `OrdinationResults`

skbio/embedding/_protein.py
```diff
@@ -96,3 +123,80 @@ def __repr__(self):
...
+        -------
+        str
+            A string representation of the ProteinEmbedding object.
+
+        See Also
+        --------
+        Protein
+        """
+        seq = Protein(str(self))
+
```
wasade, line 193: There is a lot in common here with `SequenceVector.__repr__` and may be an opportunity to decompose

skbio/embedding/tests/test_embedding.py
```diff
@@ -69,6 +82,63 @@ def test_assert_length(self):
...
+        for i, vector in enumerate([self.vector1, self.vector2, self.vector3]):
+            self.assertTrue(np.array_equal(self.sequence_vectors[i].vector, vector))
+
+    def test_sequence(self):
+        # Test if the sequence attribute is set correctly
+        for i, sequence in enumerate(["ACGT", "GCTA", "TTAG"]):
+            self.assertEqual(self.sequence_vectors[i].sequence, sequence)
+
+    def test_repr(self):
+        # Test if the __repr__ method returns the correct string
```
wasade, line 107: I dont think that's what this test does. It may be nice if it checked some bare minimum content about the object rather than just the object name, but I don't have strong feelings about that

skbio/embedding/tests/test_embedding.py
```diff
@@ -69,6 +82,63 @@ def test_assert_length(self):
...
+    def test_to_numpy(self):
+        # Test if to_numpy returns the correct numpy array
+        expected_result = np.array([self.vector1, self.vector2, self.vector3])
+        result = SequenceVector.to_numpy(self.sequence_vectors)
+        self.assertTrue(np.array_equal(result, expected_result))
+
+    def test_to_distance_matrix(self):
+        # Test if to_distance_matrix returns a DistanceMatrix object
+        distance_matrix = SequenceVector.to_distance_matrix(self.sequence_vectors)
+        self.assertEqual(distance_matrix.shape, (3, 3))
```
wasade, line 127: Testing the exact values present in the distance matrix should be easy and more precise of a test

wasade: Same re: `to_dataframe` and `to_ordination`

skbio/io/format/embed.py
```diff
@@ -165,12 +186,25 @@ def _embed_to_protein(
...
+    )
 
+
+def _objects_to_embed(objs, fh, include_embedding_pointer=True):
+    with h5py.File(fh, "w") as h5grp:
         h5grp.attrs["format"] = "embedding"
         h5grp.attrs["format-version"] = "1.0"
-        maxsize = 1
+        max_idsize = 1
+        max_embsize = 1
```
wasade, line 207:
```suggestion
        max_embsize = 1
        resize_by = 1.38
```

skbio/io/format/embed.py
```diff
@@ -191,16 +225,24 @@ def _objects_to_embed(objs, fh):
...
+            if i > 0:
+                if include_embedding_pointer:
+                    if ((len(arr) + idptr_fh[i - 1]) > max_idsize or
+                        (emb.shape[0] + embptr_fh[i - 1]) > max_embsize ):
+                        max_idsize = ceil(len(arr) + idptr_fh[i - 1]) * 1.38
+                        max_embsize = ceil(emb.shape[0] + embptr_fh[i - 1]) * 1.38
+                        resize = True
+                else:
+                    if len(arr) + idptr_fh[i - 1] > max_idsize:
+                        max_idsize = ceil(len(arr) + idptr_fh[i - 1] * 1.38)
```
wasade, line 238:
```suggestion
                        max_idsize = ceil(len(arr) + idptr_fh[i - 1]) * resize_by
                        max_embsize = ceil(emb.shape[0] + embptr_fh[i - 1]) * resize_by
                        resize = True
                else:
                    if len(arr) + idptr_fh[i - 1] > max_idsize:
                        max_idsize = ceil(len(arr) + idptr_fh[i - 1] * resize_by)
```

skbio/io/format/embed.py
```diff
@@ -191,16 +225,24 @@ def _objects_to_embed(objs, fh):
...
-            if i > 0 and "idptr" in h5grp:
+            if "idptr" in h5grp:
                 idptr_fh = h5grp["idptr"]
-
-                if len(arr) + idptr_fh[i - 1] > maxsize:
-                    maxsize = ceil(len(arr) + idptr_fh[i - 1] * 1.38)
-                    resize = True
-
                 if resize:
                     idptr_fh.resize((ceil(i * 1.38),))
```
wasade, line 247:
```suggestion
                    idptr_fh.resize((ceil(i * resize_by),))
```

skbio/io/format/embed.py
```diff
@@ -212,23 +254,43 @@ def _objects_to_embed(objs, fh):
...
                     "id", data=arr, maxshape=(None,), dtype=np.uint8,
                     compression='gzip'
                 )
 
+            if include_embedding_pointer:
+
+                if "embedding_ptr" in h5grp:
+                    embptr_fh = h5grp["embedding_ptr"]
+                    if resize:
+                        embptr_fh.resize((ceil(i * 1.38),))
```
wasade, line 270:
```suggestion
                        embptr_fh.resize((ceil(i * resize_by),))
```

skbio/embedding/_embedding.py
mortonjt: I'm OK with removing type annotations and moving it to the docs

Regarding `List[SequenceVector]`, that won't here, since the `SequenceVector` object needs to be first created before evaluating the `SequenceVector` type

wasade: Thanks! 

🐔 🥚

skbio/embedding/_embedding.py
mortonjt: Yes! I find myself often recycling the SVD / biplot code

skbio/embedding/_embedding.py
mortonjt: erm ... yes I think we can rid of the outer `str` call, since the input is already a `str`

mortonjt, line 133:
```suggestion
        return self._ids[0].decode('ascii')
```

skbio/embedding/_protein.py
```diff
@@ -7,10 +7,27 @@
...
+def _validate_protein(sequence):
+    if isinstance(sequence, Protein):
+        sequence = str(sequence)
+    elif isinstance(sequence, str):
+        if " " in sequence:
+            sequence = sequence.replace(" ", "")
+
+        # perform a check to make sure the sequence is a valid
+        # protein sequence
+        Protein(sequence)
```
qiyunzhu, line 29: Will this print something to the screen? To avoid this behavior, maybe `_ = Protein(sequence)`?

skbio/embedding/_embedding.py
qiyunzhu: I wonder if we may pass sequences as byte arrays under the hood among `Sequence`, `Embedding` and `Vector`, since this is the native format for `Sequence`, and it is efficient. On the other hand, `str` is less native or efficient.

skbio/embedding/_embedding.py
```diff
@@ -98,3 +108,158 @@ def __repr__(self):
...
+            f"embedding dimension: {dim}\n{indent}has gaps",
+        )
+        return rstr
+
+    @property
+    def embedding(self):
+        r""" The embedding tensor. """
+        return self._embedding.reshape(1, -1)
+
+    @staticmethod
```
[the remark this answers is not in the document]
mortonjt: That is probably OK to separate it out. The only thing that I may want to add is type checking to make sure that all of the vectors have the same type signature.

skbio/embedding/tests/test_embedding.py
mortonjt: tests have been added

skbio/embedding/_embedding.py
mortonjt: We're currently passing around a numpy of str. This is what is done in the `DistanceMatrix` object, so I'm in favor of currently keeping this to avoid a major refactor here

mortonjt:
Hi all, thanks for the comments.

After carefully scrutinizing the type signatures, I realized that there is an edge case that could complicate the incorporating of non-sequence vector objects (i.e. molecular vectors). 

I've created an intermediate class called `EmbeddingVector` that primarily serves to (1) address this future edge case and (2) enabling better type checking of vector objects (i.e. checking to make sure that all conversion methods accept iterables of objects that subclass `EmbeddingVector`).

I realize that this is complicating the hierarchy : `Embedding` -> `EmbeddingVector` -> `SequenceVector` -> `ProteinVector`, but this will be the last change to the hierarchy, and will avoid short-circuiting ourselves if we wanted to add in additional types in the future.

skbio/embedding/_protein.py
```diff
@@ -36,6 +54,25 @@ class ProteinEmbedding(SequenceEmbedding):
     --------
     Protein
 
+
```
qiyunzhu, line 57: This empty line is not needed.

skbio/embedding/_protein.py
```diff
@@ -36,6 +54,25 @@ class ProteinEmbedding(SequenceEmbedding):
...
+    --------------------------
+    Stats:
+        length: 10
+        embedding dimension: 3
+        has gaps: False
+        has degenerates: False
+        has definites: True
+        has stops: False
+    --------------------------
+    0 ACDEFGHIKL
```
qiyunzhu, line 75: Should have an empty line before the end of docstring.

skbio/embedding/_protein.py
```diff
@@ -48,16 +85,7 @@ def __init__(
         if clip_tail:
```
qiyunzhu, line 85: In the docstring of `ProteinEmbedding`, `clip_end` should be `clip_tail`.

skbio/embedding/_protein.py
```diff
@@ -81,18 +109,85 @@ def __repr__(self):
...
-            f"embedding dimension: {self.embedding.shape[1]}\n{indent}has gaps",
+        rstr = _repr_helper(
+            repr(seq), "Protein", "ProteinEmbedding", "embedding",
+            regex_match="has gaps", shape=self.embedding.shape[1]
         )
         return rstr
 
 
 example_protein_embedding = ProteinEmbedding(
     np.random.randn(62, 1024),
```
qiyunzhu, line 120: Should use [`np.random.default_rng`](https://numpy.org/doc/stable/reference/random/generator.html) to ensure reproducibility.

skbio/embedding/_embedding.py
```diff
@@ -7,10 +7,28 @@
...
 import numpy as np
-
+import pandas as pd
 from skbio.sequence import Sequence
 from skbio._base import SkbioObject
+from skbio.stats.distance import DistanceMatrix
+from skbio.stats.ordination import OrdinationResults
 from skbio.metadata._mixin import MetadataMixin
+from skbio.diversity import beta_diversity
+from scipy.spatial.distance import pdist, squareform
```
qiyunzhu, line 17: They are not used in the code.

skbio/embedding/_embedding.py
```diff
@@ -86,15 +110,212 @@ def __repr__(self):
...
+        An iterable of EmbeddingVector objects, or objects that
+        subclass EmbeddingVector.
+    validate : bool, optional
+        If True, validate that all vectors have the same length
+        and are valid types
+
+    Returns
+    -------
+    np.ndarray
+        A numpy array of shape (n_sequences, n_features) where
```
qiyunzhu, line 196: This sentence is incomplete.

skbio/embedding/_protein.py
```diff
@@ -7,10 +7,28 @@
...
+        if " " in sequence:
+            sequence = sequence.replace(" ", "")
+
+        # perform a check to make sure the sequence is a valid
+        # protein sequence
+        _ = Protein(sequence)
+    return sequence
 
 
 class ProteinEmbedding(SequenceEmbedding):
```
qiyunzhu, line 34: The parameter `sequence` may be "array_like" instead of just "ndarray", such that plain Python lists can be supported. The code can add `sequence = np.asarray(sequence)` to convert lists into ndarray, or leave as-is if already array.

skbio/embedding/_embedding.py
```diff
@@ -7,10 +7,28 @@
 # ----------------------------------------------------------------------------
```
qiyunzhu, line 7: Following the naming convention in scikit-bio, the name of this script file can be changed from `_embedding.py` into `_base.py`.

skbio/embedding/_embedding.py
```diff
@@ -86,15 +110,212 @@ def __repr__(self):
...
+            raise ValueError("All objects must be of the same type.")
+
+        lens = [len(ev.vector) for ev in embedding_vectors]
+        if not all(ln == lens[0] for ln in lens):
+            raise ValueError("All vectors must have the same length.")
+    data = np.vstack([ev.vector for ev in embedding_vectors])
+    return data
+
+
+def embedding_vectors_to_distance_matrix(embedding_vectors, metric='euclidean',
```
qiyunzhu, line 224: This function name is bit too long. How about abbreviate `distance_matrix` as `distmat`? This should be a commonly used abbreviation. See [here](https://www.bioinformatics.nl/cgi-bin/emboss/distmat) and [here](https://www.rdocumentation.org/packages/pracma/versions/1.9.9/topics/distmat).

skbio/embedding/_embedding.py
```diff
@@ -86,15 +110,212 @@ def __repr__(self):
...
+                                         validate=True):
+    r""" Convert an iterable of EmbeddingVector objects to a
+    DistanceMatrix object.
+
+    Parameters
+    ----------
+    embedding_vectors : iterable of EmbeddingVector objects
+        An iterable of EmbeddingVector objects, or objects that
+        subclass EmbeddingVector.
+    metric : str, optional
```
qiyunzhu, line 234: `metric` can be upgraded to support a user function. But it can be postponed to a future PR.

skbio/embedding/_embedding.py
```diff
@@ -86,15 +110,212 @@ def __repr__(self):
...
+    def __str__(self):
+        return self._ids[0].decode('ascii')
+
+    @property
+    def vector(self):
+        return self.embedding.squeeze()
+
+    @property
+    def embedding(self):
+        r""" The embedding tensor. """
```
qiyunzhu, line 135: Is "tensor" the right term here? I thought that a tensor is basically an array that can be stored in GPU.

skbio/embedding/_embedding.py
```diff
@@ -86,15 +110,212 @@ def __repr__(self):
...
+        lens = [len(ev.vector) for ev in embedding_vectors]
+        if not all(ln == lens[0] for ln in lens):
+            raise ValueError("All vectors must have the same length.")
+    data = np.vstack([ev.vector for ev in embedding_vectors])
+    return data
+
+
+def embedding_vectors_to_distance_matrix(embedding_vectors, metric='euclidean',
+                                         validate=True):
+    r""" Convert an iterable of EmbeddingVector objects to a
```
qiyunzhu, line 226: First sentence of a docstring should not exceed one line.

qiyunzhu:
@mortonjt I see the logic of `EmbeddingVector` and I like that. I wonder if this term should be called "vector embedding" instead?

One thing to note, which is somehow relevant but shouldn't affect the design here: scikit-bio's `Sequence` only support ASCII characters, which is a big limitation (e.g., you can't easily represent methylated nucleotides), but also grants performance since all underlying data are np.uint8. In comparison, Biotite's sequence can contain any custom values (any immutable and hashable)

skbio/embedding/_embedding.py
mortonjt: Not necessarily. There is no `_base.py` in the stats folder for instance.

Issue #2014, opened 2024-04-30
Alpha Diversity indices produce zeros where previously undefined (NaN)


ebolyen:
https://github.com/scikit-bio/scikit-bio/pull/1894 added short-circuits to produce `0` where previously the formula would produce `NaN`.

This is currently impacting QIIME 2's diversity measures and represents a breaking change from our perspective. For the moment, we're going to copy the effected indices prior to these updates into q2-diversity-lib, but we don't want this to be a long-term solution. 

What was the rationale for this particular change? 

Reviewing Pielou's `J'` for instance, there doesn't appear to be a definition for zero-feature samples, so the formula producing `NaN` seems appropriate as it isn't defined (this also lets the user know that they shouldn't interpret that sample). Shannon has a similar reasoning.

cc @gregcaporaso

ebolyen:
(had the wrong PR linked, updated)

qiyunzhu:
Hello all, I recently carefully reviewed the alpha diversity functionality in scikit-bio, and improved some of them. There are several things we need to be cautious about and try to improve:

1. Empty community: It should return 0. The code should force that. Some metrics already have this implemented (I did them). Though not all of them. This is a dilemma that many original authors have also explored (e.g., as @mortonjt noticed, a sample with one species may have the same alpha diversity as a sample with zero species).

2. It has been planned that we let all alpha diversity metrics to operate on sparse matrices provided by BIOM. Once this is implemented, some of the challenges discussed here will be gone. That is, instead of a vector with some zeros, you will start with a vector with only positive numbers. A community without any species will be an empty numpy array.

3. Pielou's _J_ (a.k.a., Shannon's equitability, _E<sub>H</sub>_): There is a logical problem in the current implementation: By definition, _J_ is the ratio of observed Shannon Index vs. maximum Shannon Index allowed by the observed number of species. However, in scikit-bio, Shannon index is calculated by a base of 2, whereas Pielou's _J_ is calculated by a base on e. This is problematic. I investigated literature and other software packages, as well as those of derivatives of Shannon (e.g., Pielou's J, and complex metrics like Hill number), and am under impression that most (if not all) calculates Shannon index using a base of e. This even includes SciPy's `entropy`. Therefore I would suggest that we replace 2 with e in scikit-bio.

I noticed this a long time ago. I planned to bring this up after more thorough literature review. However, given the current discussion, I think it is time for me to mention it.

mortonjt:
[...]

PR #2024, opened 2024-05-12
Improving alpha diversity calculation

qiyunzhu:
@ebolyen This addresses the relevant issue #2014 . Now the alpha diversity metrics return `np.nan` instead of zero when the community is empty.

Exceptions are metrics that are solely based on observed counts, including `sobs`, `singles`, `doubles` and `osd`. They will return zero.

One relevant question is how to deal with one-taxon communities when calculating species evenness metrics. At present, `pielou_e` and `heip_e` return `np.nan`, whereas `simpson_e` returns 1. I tend to think 1 makes sense. Evenness should range between 0 and 1, in which 1 means all species have the same abundance (i.e., completely even). When there is only one species, it is guaranteed that evenness is maximized. On the other hand, it will not be ideal if a non-empty community returns a NaN value. What do you think?

@wasade This also improves how counts are validated. Specifically, it will remove zeros from the counts before calculating alpha diversity. This will save troubles in the downstream implementation of alpha diversity metrics. This will also permit efficiency improvement by directly feeding sparse matrices from BIOM into the calculation. However, the latter is not done yet. This needs further discussion. That being said, the current PR should be safe to use before the final solution.

Also cc'ing @mortonjt 


qiyunzhu:
Hello @ebolyen would you like to comment on whether the consideration of edge cases (empty community, one-species community) is sane in this PR (see my comments above)? Also cc'ing @gregcaporaso

qiyunzhu:
Now there is a significant enrichment of the parameters and documentation of alpha diversity metrics.

qiyunzhu:
@wasade @mortonjt Here is a very large PR. I did some substantial amount of literature review in order to improve the calculation and documentation of all alpha diversity metrics. I wonder if either of you can do a quick sanity check of the technical side of my code (especially counts vector validation and transformation), while trusting my professionality in community ecology?

skbio/diversity/tests/test_driver.py
```diff
@@ -168,20 +168,17 @@ def test_invalid_input_phylogenetic(self):
 
     def test_empty(self):
         # empty vector
-        actual = alpha_diversity('sobs',
-                                 np.array([], dtype=np.int64))
+        actual = alpha_diversity('sobs', np.array([], dtype=np.int64))
         expected = pd.Series([0], dtype=int)
```
mataton, line 172: `expected` should have `dtype=np.int64` to pass on Windows.

This is also true for lines 208 in `test_empty`, line 213 in `test_single_count_vector`, lines 255 and 262 in `test_sobs`, and line 315 in `test_no_ids` in this class.

skbio/diversity/tests/test_driver.py
```diff
@@ -168,20 +168,17 @@ def test_invalid_input_phylogenetic(self):
...
         assert_series_almost_equal(actual, expected)
 
         # array of empty vector
-        actual = alpha_diversity('sobs',
-                                 np.array([[]], dtype=np.int64))
+        actual = alpha_diversity('sobs', np.array([[]], dtype=np.int64))
         expected = pd.Series([0], dtype=int)
```
mataton, line 177: `dtype=np.int64`

skbio/diversity/tests/test_driver.py
```diff
@@ -168,20 +168,17 @@ def test_invalid_input_phylogenetic(self):
...
         assert_series_almost_equal(actual, expected)
 
         # array of empty vectors
-        actual = alpha_diversity('sobs',
-                                 np.array([[], []], dtype=np.int64))
+        actual = alpha_diversity('sobs', np.array([[], []], dtype=np.int64))
         expected = pd.Series([0, 0], dtype=int)
```
mataton, line 182: `dtype=np.int64`.

mataton (changes requested):
Other than the windows tests failures which I commented on, everything looks good.

qiyunzhu:
@mataton Thanks! I have made the edits you suggested.

mataton (approved):

qiyunzhu:
Let me merge this PR and move on.

Issue #1959, opened 2024-02-29
Rename DistanceMatrix and OrdinationResults IDs


wasade:
Currently, it is not easy to rename `DistanceMatrix` or `OrdinationResults` IDs. Work around are possible, but exposing a stable `.rename(rename_dict, strict=True | False)` member method would be pleasant, where `strict` requires all IDs have a mapping.

constellation99:
@Bharath-Sathappan and myself are currently working on this issue. We have implemented rename functions for DistanceMatrix and OrdinationResults and are finishing writing test cases before submitting our pull request.

PR #2025, opened 2024-05-12
Added Rename and Test Cases in DistanceMatrix Issue: #1959

code-wolf-byte:
Added a rename function to the class DistanceMatrix as well as apt test cases for the function.

[* ] I have read the [contribution guidelines](https://scikit.bio/contribute.html).

[* ] I have documented all public-facing changes in the [changelog](https://github.com/scikit-bio/scikit-bio/blob/main/CHANGELOG.md).

[* ] This pull request includes code, documentation, or other content derived from external source(s). If this is the case, ensure the external source's license is compatible with scikit-bio's [license](https://github.com/scikit-bio/scikit-bio/blob/main/LICENSE.txt). Include the license in the licenses directory and add a comment in the code giving proper attribution. Ensure any other requirements set forth by the license and/or author are satisfied.

It is your responsibility to disclose code, documentation, or other content derived from external source(s). If you have questions about whether something can be included in the project or how to give proper attribution, include those questions in your pull request and a reviewer will assist you.
[* ] This pull request does not include code, documentation, or other content derived from external source(s).

### DIFF
--- a/skbio/stats/distance/_base.py
+++ b/skbio/stats/distance/_base.py
@@ -213,7 +213,46 @@ class DissimilarityMatrix(SkbioObject, PlottableMixin):
                 dm[i, j] = metric(a, b)
 
         return cls(dm, keys_)
+    def rename(self, rename_dict, strict=True):
+        """
+        Rename IDs in the DistanceMatrix.
 
+        Parameters
+        ----------
+        rename_dict : dict
+            A dictionary mapping current IDs to new IDs.
+        strict : bool, optional
+            If True, requires that all existing IDs in the DistanceMatrix must be
+            included in the rename_dict. If False, IDs not in rename_dict will be
+            kept as is.
+
+        Returns
+        -------
+        DistanceMatrix
+            A new instance of DistanceMatrix with IDs renamed.
+
+        Raises
+        ------
+        ValueError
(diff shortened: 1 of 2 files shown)


qiyunzhu (commented):
@code-wolf-byte Good job!

The error in the test appears to be because `rename` returns a new instance of the distance matrix, but the example code doesn't indicate that.

Ideally, there should be two optional behaviors, controlled by an optional parameter `inplace`. When it is true, the method will rename IDs _in place_ and doesn't return a value. When it is false, the method will return a new distance matrix with renamed IDs.

An example of such a method is [`pandas.DataFrame.rename`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.rename.html). You can see how the API is like (also, instead of `rename_dict` they name it as `mapper`, which can take either a dictionary or a function).

constellation99:
Hi! Myself and @Bharath-Sathappan are actually also working on this issue and would be happy to share what we've implemented with you for our DistanceMatrix rename function as well as the test cases we've created. Additionally, we are working on the renaming function for OrdinationResults and are currently finishing up writing test cases.

qiyunzhu:
@constellation99 @Bharath-Sathappan Thanks for the offering! @code-wolf-byte do you have a plan to complete this PR in the near future? Let us know!

Issue #2017, opened 2024-05-01
Support for stdin for at least FASTA/FASTQ


wasade:
Many libraries support streaming FASTA/FASTQ data through stdin. It would be pleasant for that mode of operation to be supported as it is handy when writing small stream processors

```bash
$ cat test.py
import skbio
import sys
for r in skbio.read(sys.stdin, format='fasta'):
    print(r.metadata['id'])
$ cat q.fna 
>qx1
TTTTTTTTT
>qx2
TTTTTTTTNNNNNN
$ cat q.fna | python test.py 
Traceback (most recent call last):
  File "test.py", line 3, in <module>
    for r in skbio.read(sys.stdin, format='fasta'):
  File "/home/mcdonadt/miniconda3/envs/test/lib/python3.8/site-packages/skbio/io/registry.py", line 1160, in read
    return io_registry.read(file, format=format, into=into, verify=verify,
  File "/home/mcdonadt/miniconda3/envs/test/lib/python3.8/site-packages/skbio/io/registry.py", line 506, in read
    return (x for x in itertools.chain([next(gen)], gen))
  File "/home/mcdonadt/miniconda3/envs/test/lib/python3.8/site-packages/skbio/io/registry.py", line 529, in _read_gen
    reader, kwargs = self._init_reader(file, fmt, into, verify, kwargs,
  File "/home/mcdonadt/miniconda3/envs/test/lib/python3.8/site-packages/skbio/io/registry.py", line 543, in _init_reader
    backup = file.tell()
io.UnsupportedOperation: underlying stream is not seekable
```

RaeedA:
Hey! This issue seems to stem from the fact that stdin isn't a seekable stream, which messes with the file sniffers. The problem specifically arises when [saving the pointer position](https://github.com/scikit-bio/scikit-bio/blob/e3a27f56970fded68ba7561767cec41e47ba3c78/skbio/io/registry.py#L532) before sniffing, and [restoring it afterwards](https://github.com/scikit-bio/scikit-bio/blob/e3a27f56970fded68ba7561767cec41e47ba3c78/skbio/io/registry.py#L534C1-L534C17).

I found that commenting them out caused the piping to work as expected, but with residual warnings.
```bash
(skbio-dev) azomss-air:scikit-bio raeedazom$ cat skbio/io/q.fna | python skbio/io/temp.py 
[...]

PR #2029, opened 2024-05-24
WIP: stdin support for FASTA/FASTQ files

RaeedA:
Issue #2017 
Adding support for FASTA/FASTQ files to make them accept stdin streaming, as well as adding compatibility for other file formats in the future.
Modified the FASTQ file parser to match the FASTA structure, specifically moving the parser functionality out of the reader and into a dedicated helper function.
Making most sniffer/parsers pass out consumed lines.
Coverage still in progress.


skbio/io/_iosources.py
```diff
@@ -190,6 +201,9 @@ def can_read(self):
             if isinstance(head, str):
```
qiyunzhu, line 201: These two checks can be merged:

```python
if isinstance(head, (str, bytes)):
    self.repaired = itertools.chain([head], iterator)
    return True
```

skbio/io/format/_base.py
```diff
@@ -227,13 +229,18 @@ def _line_generator(fh, skip_blanks=False, strip=True):
 def _too_many_blanks(fh, max_blanks):
     count = 0
     too_many = False
+    consumed = []
     for line in _line_generator(fh, skip_blanks=False):
+        consumed += [line]
```
qiyunzhu, line 234: This can be replaced with:

```python
consumed.append(line)
```

skbio/io/format/fasta.py
```diff
@@ -640,22 +640,26 @@ def _fasta_sniffer(fh):
     #   not be validated but it probably isn't what the user wanted). Also, if
     #   we add QUAL as its own file format in the future, we wouldn't want the
     #   FASTA and QUAL sniffers to both positively identify a QUAL file.
-    if _too_many_blanks(fh, 5):
-        return False, {}
+    blanks, consumed = _too_many_blanks(fh, 5)
+    if blanks:
+        return False, {}, consumed
+    if not fh.seekable():
+        fh = itertools.chain(consumed, fh)
```
qiyunzhu, line 647: Very good. It looks like the right solution.

skbio/io/format/_base.py
RaeedA: Is this more efficient than just doing +=? I'm using += in other places as sometimes I need to add multiple lines, and wanted to keep it consistent.

skbio/io/registry.py
```diff
@@ -514,24 +519,37 @@ def _read_gen(self, file, fmt, into, verify, kwargs):
...
+                try:
+                    # Attempt to reset and sniff file
+                    backup = file.tell()
+                    res = sniffer(file, **io_kwargs)
+                    is_format = res[0]
+                    skwargs = res[1]
+                    file.seek(backup)
+                except io.UnsupportedOperation:
+                    # File is not seekable, so get consumed lines to reset it
+                    is_format, skwargs, consumed = sniffer(file, **io_kwargs)
```
wasade, line 551: `peek` for `stdin` (see `sys.stdin.buffer.peek`) and many other IO types (not `io.StringIO` AFAIK) would work without advancing the pointer too. But, it's read request is not assured to be what you want in terms of bytes

skbio/io/registry.py
```diff
@@ -914,7 +939,8 @@ def wrapped_sniffer(
                         # Some formats may have headers which indicate their
                         # format sniffers should be able to rely on the
                         # filehandle to point at the beginning of the file.
-                        fh.seek(0)
+                        if self.header:
+                            fh.seek(0)
```
wasade, line 943: Is this still safe for `stdin` which is not seekable?

skbio/io/tests/test_util.py
```diff
@@ -634,6 +638,27 @@ def test_open_write_mode(self):
         fh.close()
         self.assertTrue(fh.closed)
 
+class TestReadStandardInput(unittest.TestCase):
+    num_text = ["ONE", "TWO", "THREE", "FOUR", "FIVE", "SIX", "SEVEN"]
+
+    def test_stdin_read_fasta(self):
```
wasade, line 644: These tests should stay, but I recommend additionally setting up an independent integration test run through the `Makefile`. Something like the below. The inline python though could also be placed in a minor script. 

```bash
#!/bin/bash

cat skbio/io/tests/test_data/some_fasta.fna | \
    python -c "import skbio, sys; print([r.metadata['id'] for r in skbio.read(sys.stdin, format='fasta')]) > /tmp/obs_stdin.ids

grep "^>" skbio/io/tests/test_data/some_fasta.fna | tr -d ">" > /tmp/exp_stdin.ids

obs_md5=$(md5sum /tmp/obs_stdin.ids)
exp_md5=$(md5sum /tmp/exp_stdin.ids)

if [[ ${obs_md5} != ${exp_md5} ]]; then
    exit 1
fi
```

skbio/io/format/_base.py
qiyunzhu: `append` is just a "pythonic" way :) I don't know if it's more efficient or not. PS: if  you want to add multiple elements, you can do `consumed.extend([line1, line2, line3...])`.

skbio/io/registry.py
qiyunzhu: I think that @RaeedA added a `seekable` flag to different input types to control this behavior.

RaeedA: That was the original plan, I moved some things around in the current changes to make it work more intuitively. It's now in a try/catch block.

wasade:
Does this work?How to check if file object is random accessstackoverflow.comOn Jun 3, 2024, at 15:05, Raeed Azom ***@***.***> wrote:﻿
@RaeedA commented on this pull request.



In skbio/io/registry.py:
> @@ -914,7 +939,8 @@ def wrapped_sniffer(
                         # Some formats may have headers which indicate their
                         # format sniffers should be able to rely on the
                         # filehandle to point at the beginning of the file.
-                        fh.seek(0)
+                        if self.header:
+                            fh.seek(0)

That was the original plan, I moved some things around in the current changes to make it work more intuitively. It's now in a try/catch block.

qiyunzhu:
@wasade According to the Python [documentation](https://docs.python.org/3/library/io.html#io.IOBase.seekable), `seekable` is a valid check. It says: "Return True if the stream supports random access".

wasade:
Right, and it would be preferable to perform the explicit check rather than a try/except

wasade:
Oh I see, the SO post didn't actually come through in email -- this [one](https://stackoverflow.com/a/49376762/19741)

qiyunzhu:
> Right, and it would be preferable to perform the explicit check rather than a try/except

I agree with that. In the current PR there are both. Redundancy may not be bad. But @RaeedA you may want to check to make sure we don't solely rely on try/except. (I am okay with solely relying on seekable, if that works.)

RaeedA:
Hey! I'm actually not using the seekable call here, as I've found that it behaves irregularly with some data formats. I've instead been using tell() wrapped inside of a try/catch, as tell will consistently produce an UnsupportedOperation error that I can use to detect seekable versus unseekable.

skbio/io/tests/test_util.py
RaeedA: I'm not super familiar with these, but I've been looking for a way to do just this! I'll take a look at implementing this for testing.

skbio/io/registry.py
RaeedA: Yeah, I've tested a couple other methods of reading nonseekables including this but it just seems that reconstructing it as a chain is the best.

qiyunzhu:
@RaeedA I tend to think that using try/except is fine, given that you found its result consistent, whereas that of seekable isn't.

PR #2032, opened 2024-05-30
Windows support for ProteinEmbedding

mataton:
This PR implements Windows support for the new `ProteinEmbedding` class. There was a problem with how temporary files are handled on Windows vs Mac/Linux. See details of the issue from my Windows support [PR](https://github.com/biocore/biom-format/pull/951) for the BIOM format. The current solution in this PR relies on implicit cleanup of the `TemporaryDirectory`s, which works and was already happening with some of the other tests in `test_embed.py`, so I think this is an okay solution. We may want to revisit this and adopt a more elegant solution, but one of our goals is to have skbio have Windows support by the ISMB conference in July, so I feel this quick solution works for now.


qiyunzhu (approved):

mortonjt:
looks good to me! merge?

qiyunzhu:
@mataton According to the [documentation](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory), the `delete` parameter was added to Python 3.12. Therefore we probably cannot use it for the sake of backward compatibility.

mataton:
I didn't notice that it was introduced in 3.12. Tests pass for all python 3.12 builds. I'll have to look for an alternative solution.

PR #2034, opened 2024-06-05
Fix default int types for Pandas DataFrame and Series

mataton:
This PR fixes underlying issues with the way Pandas handles `int`s on Windows.

### DIFF
--- a/skbio/diversity/tests/test_driver.py
+++ b/skbio/diversity/tests/test_driver.py
@@ -170,19 +170,19 @@ class AlphaDiversityTests(TestCase):
         # empty vector
         actual = alpha_diversity('sobs',
                                  np.array([], dtype=np.int64))
-        expected = pd.Series([0])
+        expected = pd.Series([0]).astype(int)
         assert_series_almost_equal(actual, expected)
 
         # array of empty vector
         actual = alpha_diversity('sobs',
                                  np.array([[]], dtype=np.int64))
-        expected = pd.Series([0])
+        expected = pd.Series([0]).astype(int)
         assert_series_almost_equal(actual, expected)
 
         # array of empty vectors
         actual = alpha_diversity('sobs',
                                  np.array([[], []], dtype=np.int64))
-        expected = pd.Series([0, 0])
+        expected = pd.Series([0, 0]).astype(int)
         assert_series_almost_equal(actual, expected)
 
         # empty vector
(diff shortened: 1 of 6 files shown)


skbio/util/_testing.py
```diff
@@ -349,6 +349,23 @@ def assert_data_frame_almost_equal(left, right, rtol=1e-5):
...
+    in comparing observed and expected data frames in Windows. This function repairs
+    the issue by converting int64 columns of a data frame into int32 in Windows.
+
+    See: https://github.com/unionai-oss/pandera/issues/726
+
+    """
+    for col in df.select_dtypes("int").columns:
+        df[col] = df[col].astype(int)
+
+    return df
```
qiyunzhu, line 366: This function doesn't have to return `df` -- it manipulates `df` in place. Returning `df` can actually confuse the user, who might assume that the function creates a copy while not touching the original input.

skbio/stats/distance/tests/test_mantel.py
```diff
@@ -549,7 +552,7 @@ def test_minimal_compatible_input(self):
         # input as array_like
         obs = pwmantel((self.minx, self.miny, self.minz),
                        alternative='greater')
-        assert_data_frame_almost_equal(obs, self.exp_results_minimal)
+        assert_data_frame_almost_equal(obs, self.exp_results_minimal) #
```
qiyunzhu, line 555: What does the trailing `#` do?

skbio/diversity/tests/test_driver.py
```diff
@@ -170,19 +170,19 @@ def test_empty(self):
         # empty vector
         actual = alpha_diversity('sobs',
                                  np.array([], dtype=np.int64))
-        expected = pd.Series([0])
+        expected = pd.Series([0]).astype(int)
```
qiyunzhu, line 173: You can do `pd.Series([0], dtype=int)`, which is simpler. (Assuming it works...)

qiyunzhu (commented):
@mataton Good job! I have two minor comments.

skbio/stats/distance/tests/test_mantel.py
mataton: Accidentally left it there, it's been removed with the most recent commits.

skbio/util/_testing.py
mataton: Without the `return` statement, the tests fail with the following error:

`AssertionError: DataFrame Expected type <class 'pandas.core.frame.DataFrame'>, found <class 'NoneType'> instead`

qiyunzhu (approved):
Looks good! Let's merge.

Issue #2009, opened 2024-04-18
TreeNode.from_taxonomy should take a pandas dataframe as input


mortonjt:
Dataframes are easier to obtain than list of tuple of lists (which is the current input of [`TreeNode.from_taxonomy`](https://github.com/scikit-bio/scikit-bio/blob/main/skbio/tree/_tree.py#L1822) )

The current workaround (using the metaphlan4 database as a reference) is as follows
```python

from skbio import TreeNode
taxonomy = pd.read_table('MetaPhlAn4/mpa_vJan21_CHOCOPhlAnSGB_202103_species.txt', header=None, index_col=0)
taxonomy = pd.DataFrame(list(taxonomy[1].apply(lambda x: x.split('|')).values), index=taxonomy.index)
taxonomy.columns = ['k', 'p', 'c', 'o', 'f', 'g', 's']
lineages = [(tup.Index, [tup.k, tup.p, tup.c, tup.o, tup.f, tup.g, tup.s]) for tup in taxonomy.itertuples(index=True)]
tree = TreeNode.from_taxonomy(lineages)
```

wasade:
It would be pleasant for `from_taxonomy` to detect the separator used, and to split accordingly as well.

PR #2042, opened 2024-06-10
Added support for dict and dataframe for `TreeNode.from_taxonomy`

qiyunzhu:
Addressing #2009 

### DIFF
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -1824,14 +1824,14 @@ class TreeNode(SkbioObject):
 
         Parameters
         ----------
-        lineage_map : iterable of tuple
-            A id to lineage mapping where the first index is an ID and the
-            second index is an iterable of the lineage.
+        lineage_map : dict, iterable of tuples, or pd.DataFrame
+            Mapping of taxon IDs to lineages (iterables of taxonomic units
+            from high to low in ranking).
 
         Returns
         -------
         TreeNode
-            The constructed taxonomy
+            The constructed taxonomy.
 
         See Also
         --------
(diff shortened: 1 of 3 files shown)


skbio/tree/_tree.py
```diff
@@ -1874,6 +1874,11 @@ def from_taxonomy(cls, lineage_map):
         root = cls(name=None)
         root._lookup = {}
 
+        if isinstance(lineage_map, dict):
+            lineage_map = lineage_map.items()
+        elif isinstance(lineage_map, pd.DataFrame):
+            lineage_map = ((idx, row.tolist()) for idx, row in lineage_map.iterrows())
```
wasade, line 1880: I'm hesitant about this as it assumes the frame is limited to lineage information, and that the columns are ordered properly. Enforcing column order though may require we limit users to specific column names though, which is viable also not great. If instead we require a column `taxonomy` or similar, and that it contains a delimited lineage then we could better control this e.g., `[(r.Index, delimiter.split(r.Taxonomy)) for r in lineage_map.itertuples()]`

Note that, minor right now, but `itertuples` is much faster than `iterrows`.

qiyunzhu: @wasade Thanks for the review. I know that lineage strings are common, but taxonomic ranks separated by column is probably also common, and more friendly to downstream analysis. Not sure what use case @mortonjt encountered, pending his comment. It may be useful to have a function in skbio to convert lineage strings from/into columns.

mortonjt: Hi yes I like the proposed approach of anticipating both taxonomic strings and data frame. Regarding the functionality I proposed earlier, I had taxonomic strings from metaphlan that I manually parsed into a data frame.  In my mind, having a dataframe as input could also be valuable, since it is easier to obtain summaries for each taxonomic level, as well as manipulate the taxonomic names. As @wasade mentioned earlier, the delimiters could widely vary -- I've seen both `;` and `|` used in practice.

qiyunzhu: Will be something like:

```python
df['Taxonomy'].str.split('; ', expand=True).apply(lambda x: x.str.split('__').str[1])
```

qiyunzhu:
@wasade @mortonjt Maybe let's merge this PR, and have another PR for lineage-to-columns conversion? Which skbio module do you think it will fit in? Currently there is no `taxonomy` module in scikit-bio. A work around might be to have a `lineage` file format.

mortonjt:
+1 for lineage format. Then we can have TreeNode.read take in files as input

PR #2050, opened 2024-06-15
Adding Nearest Neighbor Interchange (NNI) algorithm for phylogenetic reconstruction

christapo:
This PR adds the Nearest Neighbor Interchange (NNI) algorithm as an additional method for phylogenetic reconstruction.

This method can be used by itself or as a post-processing option for the Neighbor Joining (NJ) algorithm.


skbio/tree/_nj.py
```diff
@@ -286,3 +287,332 @@ def _pair_members_to_new_node(dm, i, j, disallow_negative_branch_length):
...
+                sub_tips = []
+                for subtree in [b, c, d]:
+                    sub_tips.append(1 if subtree.is_tip() else subtree.count(tips=True))
+                b_, c_, d_ = sub_tips
+                a_ = n_taxa - b_ - c_ - d_
+                # calculate the swap length for the two possible swaps given the edge
+                swap_1 = _swap_length(a_, b_, c_, d_, i1, i2, i3, i4, adm)
+                swap_2 = _swap_length(a_, b_, d_, c_, i1, i2, i4, i3, adm)
+                # store the best possible swap into a maxheap
+                if swap_1 > swap_2 and swap_1 > 0:
```
qiyunzhu, line 471: These two conditions are not mutually exclusive. Could there be other conditions?

skbio/tree/_nj.py
```diff
@@ -286,3 +287,332 @@ def _pair_members_to_new_node(dm, i, j, disallow_negative_branch_length):
...
+        parent = node
+        a = parent.parent
+        # identify the index of each neighboring node
+        for index, node in enumerate(ordered):
+            if node == a:
+                i1 = index
+        for child in parent.children:
+            if child.is_tip():
+                continue
+            else:
```
qiyunzhu, line 449: Since you already did `continue` in the previous line, the current `else` can be omitted.

skbio/tree/_nj.py
```diff
@@ -286,3 +287,332 @@ def _pair_members_to_new_node(dm, i, j, disallow_negative_branch_length):
...
+    for node in ordered:
+        # ignore tips of the tree
+        if node.is_tip():
+            continue
+        # identify the parent and grandparent nodes
+        parent = node
+        a = parent.parent
+        # identify the index of each neighboring node
+        for index, node in enumerate(ordered):
+            if node == a:
```
qiyunzhu, line 444: It may be better to do `if node is a`.

skbio/tree/_nj.py
```diff
@@ -286,3 +287,332 @@ def _pair_members_to_new_node(dm, i, j, disallow_negative_branch_length):
...
+    ordered = list(tree.postorder(include_self=False))
+    root = tree.root()
+    n_taxa = root.count(tips=True) + 1
+    # begin by finding nodes which are the child node of an internal edge
+    for node in ordered:
+        # ignore tips of the tree
+        if node.is_tip():
+            continue
+        # identify the parent and grandparent nodes
+        parent = node
```
qiyunzhu, line 440: Why is it called "parent" since it is the node itself?

skbio/tree/_nj.py
```diff
@@ -286,3 +287,332 @@ def _pair_members_to_new_node(dm, i, j, disallow_negative_branch_length):
...
+        return
+    for node in tree.non_tips():
+        if len(node.children) != 2:
+            print("Could not perform NNI. Tree needs to be a binary tree")
+            return
+    adm = _average_distance_matrix(tree, dm)
+    while True:
+        # create heap of possible swaps and then swapping subtrees
+        # until no more swaps are possible.
+        adm = _average_distance_matrix(tree, dm)
```
qiyunzhu, line 340: I noticed that both `_average_distance_matrix` and `_swap_heap` involve counting tips in the tree (`tree.root().count(tips=True)`). Counting involves a postorder traversal of the entire tree, which is expensive. Since the tree size doesn't change during the swapping process (if I understand correctly), it may be more efficient to pre-calcuate tip count just once and pass it to the helper function.

skbio/tree/_nj.py
```diff
@@ -286,3 +287,332 @@ def _pair_members_to_new_node(dm, i, j, disallow_negative_branch_length):
...
+        # which is above the first node in the postorder as well as an ancestor
+        for j in range(i + 1, n - 1):  # part (a)
+            b = ordered[j]
+            # skipping over ancestors
+            if b in a.ancestors():
+                continue
+            # both nodes are tips
+            if a.is_tip() and b.is_tip():
+                adm[i, j] = adm[j, i] = dm[a.name, b.name]
+            # second node is a tip, but not the first node
```
qiyunzhu, line 515: Could there be a condition where first node is a tip, but not the second node?

qiyunzhu (commented):
@christapo This is awesome work! Since our last in-person discussion, your code quality has significantly improved, and considerably professional. I added a few trivial comments. Looking forward to your presentation tomorrow!

skbio/tree/_nj.py
```diff
@@ -286,3 +287,383 @@ def _pair_members_to_new_node(dm, i, j, disallow_negative_branch_length):
...
+    >>> rat = tree.find('rat')
+    >>> print(rat.length)
+    0.21
+
+    """
+    # Initialize and populate the average distance matrix
+    if not inplace:
+        tree = tree.copy()
+    if len(tree.root().children) != 1:
+        print("Could not perform NNI. Tree needs to be rooted at a leaf node")
```
mataton, line 382: As mentioned in this morning's meeting, this line (the entire print statement) can be replaced by `raise TypeError("Could not perform NNI. Tree needs to be rooted at a leaf node.")`

skbio/tree/_nj.py
```diff
@@ -286,3 +287,383 @@ def _pair_members_to_new_node(dm, i, j, disallow_negative_branch_length):
...
+        return
+    for node in tree.non_tips():
+        if len(node.children) != 2:
+            print("Could not perform NNI. Tree needs to be a binary tree")
```
mataton, line 386: This can also be replaced with `raise TypeError("Could not perform NNI. Tree needs to be a binary tree.")`

skbio/tree/tests/test_nj.py
```diff
@@ -191,6 +221,201 @@ def test_pair_members_to_new_node_zero_branch_length(self):
...
+        pre_str = "((c,b))a;"
+        pre_TreeNode = TreeNode.read(
+                io.StringIO(pre_str))
+        expected_str = "((c:1.0,b:2.0):1.0)a;"
+        expected_TreeNode = TreeNode.read(
+                io.StringIO(expected_str))
+        self.assertEqual(str(nni(pre_TreeNode, dm, inplace=False)),
+                         str(expected_TreeNode))
+
+    def test_nni_binary_flag(self):
```
mataton, line 258: The way to test errors in Python is typically to use a [context manager](https://book.pythontips.com/en/latest/context_managers.html). So here it would look something like:

````python
with self.assertRaises(TypeError, msg="Could not perform NNI. Tree needs to be a binary tree."):
        data = [[0, 3],
                [3, 0]]
        dm = DistanceMatrix(data, list('ab'))
        pre_str = "((b))a;"
        pre_TreeNode = TreeNode.read(
                io.StringIO(pre_str))
        nni(pre_TreeNode, dm)
````

File: skbio/tree/tests/test_nj.py [hunk 10]
```diff
@@ -191,6 +221,201 @@ def test_pair_members_to_new_node_zero_branch_length(self):
...
+        # returns a None object.
+        data = [[0, 3],
+                [3, 0]]
+        dm = DistanceMatrix(data, list('ab'))
+        pre_str = "((b))a;"
+        pre_TreeNode = TreeNode.read(
+                io.StringIO(pre_str))
+        self.assertEqual(nni(pre_TreeNode, dm), None)
+
+    def test_nni_leaf_root_flag(self):
```
mataton, line 269: The context manager approach for testing errors is more appropriate here as well.

skbio/tree/_nj.py
christapo: Thank you for the reminder!

skbio/tree/tests/test_nj.py
christapo: Thank you!

qiyunzhu: @mataton @christapo Actually, this is probably not the very best example. Instead you may want to do:

```python
data = [[0, 3],
        [3, 0]]
dm = DistanceMatrix(data, list('ab'))
pre_str = "((b))a;"
pre_TreeNode = TreeNode.read(io.StringIO(pre_str))
msg = "Could not perform NNI. Tree needs to be a binary tree."
with self.assertRaisesRegex(TypeError, msg):
    nni(pre_TreeNode, dm)
```

Rationales: 1) The error testing context manager should contain only the code that will raise an error, that is, the `nni` call. The preparations don't have to be included. This way can precisely narrow down the test to the `nni` call.

2). The `msg` parameter of [`assertRaises`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) is very confusing. It doesn't actually check the error message. Therefore the original code won't work. Instead, [`assertRaisesRegex`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex) will. [Here](https://stackoverflow.com/questions/58687475/) are some discussions.

If you do intend to use `assertRaises`, the correct code is:

```python
with self.assertRaises(TypeError) as cm:
    nni(pre_TreeNode, dm)
self.assertEqual(str(cm.exception), msg)
```

(I preferred this way because it is more precise than `assertRaisesRegex`.)

christapo: Thank you for your insight! I've updated the error testing for both the binary tree test here and the leaf root test below.

qiyunzhu (approved):

PR #2061, opened 2024-06-21
Fix cython zero division errors on Windows

mataton:
This PR fixes zero division errors which were happening in `skbio/metadata/_intersection.pyx`. The fix is fairly straightforward, just ensure that the random numbers used can never lead to division by zero.

The value chosen for `RAND_MAX` is the lowest possible value from its standard definition. See [here](https://en.cppreference.com/w/cpp/numeric/random/RAND_MAX) for more details.


qiyunzhu (commented):
This is effective but sounds bit scary. Doesn't Python have binomial distribution under `scipy.stats.binom`? @mortonjt

qiyunzhu:
I think @mataton 's solution is right in logic, but just needs a small tweak. How about:

```
self.priority = ceil(nlog * log(-1.0/(1.0 * rand()/(RAND_MAX + 1) - 1)))
```

This will ensure that `rand()/(RAND_MAX + 1)` is always < 1.

Actually, I think the original code doesn't enforce that and is probably not correct. The test passed in Linux simply because `RAND_MAX` is large enough such that zero division _happened_ to not happen. But in Windows it is smaller. @mortonjt

qiyunzhu:
Here is some discussion: https://stackoverflow.com/questions/6218399/how-to-generate-a-random-number-between-0-and-1

mataton:
@qiyunzhu Your solution works! I've pushed the commits so we should be able to merge if everyone thinks this is a good solution.

qiyunzhu (approved):
Great job @mataton !

Issue #2062, opened 2024-06-21
FDR correction using statsmodels


qiyunzhu:
Given that we have decided to include statsmodels as a dependency of scikit-bio (#2049), we can now ditch our homebrew code for Benjamini-Hochberg and Holm-Bonferroni corrections, and adopt `statsmodels.stats.multitest.multipletests`. This will require modifying `_dispatch_p_adjust`, and removing `_benjamini_hochberg` and `_holm_bonferroni`, all under `skbio.stats.composition`.

qiyunzhu:
@aryankeluskar Would you like to give it a quick fix?

aryankeluskar:
@qiyunzhu Thank you for considering me for this task! I'm definitely willing to take it up. I will proceed with making the required changed and submit a PR shortly after testing the new code.

PR #2063, opened 2024-06-22
FDR correction using statsmodels (addressing #2062)

aryankeluskar:
In response to issue #2062, this PR updates the native implementation for Benjamini-Hochberg correction and Holm-Bonferroni correction by utilizing the respective methods from the `statsmodels` library. This update would replace the previous implementation which was derived from [this stackoverflow post](https://stackoverflow.com/questions/7450957/). After calling the necessary functions, the data is stored in a variable with the same name as before and returned. 

The issue also mentioned modifying the `_dispatch_p_adjust` function since `_benjamini_hochberg` and `_holm_bonferroni` functions now seem redundant. However, this will require multiple changes to the unit tests since it currently calls both `_benjamini_hochberg` and `_holm_bonferroni` functions to perform the computations. 

I also noticed that changes other than those made to `composition.py` are included in this PR. This is most likely due to  `ruff format` on my system, which made changes to `embed.py`


qiyunzhu:
Hello @aryankeluskar This looks great as always. I have a suggestion: With `multipletests`, we can have more options than Benjamini-Hochberg and Holm-Bonferroni. The original options like "bh" need to be kept for backward compatibility, but they will be converted into `multipletests`-supported method, and call `multipletests` directly. Therefore, the `p_adjust` method can become:

```
    p_adjust : str or None, optional
        Method to correct *p*-values for multiple comparisons. Options are Holm-
        Boniferroni ("holm" or "holm-bonferroni") (default), Benjamini-
        Hochberg ("bh", "fdr_bh" or "benjamini-hochberg"), or any method supported
        by statsmodels' ``multipletests`` function. Case-insensitive. If None, no
        correction will be performed.

        .. versionchanged:: 0.6.2

            Uses statsmodels' ``multipletests`` function to perform FDR correction. All
            methods supported by this function can be used.

    See Also
    --------
    statsmodels.stats.multitest.multipletests
```

Then `_dispatch_p_adjust` (can be renamed as e.g., `_calc_p_adjust`) will perform name conversion for BH and Holm, do lowercase, then directly call multipletests and return results. There is no need to have the two separate function.

aryankeluskar:
Greetings of the day @qiyunzhu! I am nearly done with making the changes, but I have a query. There is an unit test which is dependent on the functionality of the old function `_dispatch_p_adjust`:

```
    def test_dispatch_p_adjust(self):
        self.assertIsNone(_dispatch_p_adjust(None))
        self.assertEqual(_dispatch_p_adjust(
            "holm-bonferroni").__name__, "_holm_bonferroni")
        self.assertEqual(_dispatch_p_adjust(
            "benjamini-hochberg").__name__, "_benjamini_hochberg")
```

Given that we no longer return a function to compute FDR Correction and rather return the final result, this unit test is failing. Could you please advise me on how to address this issue?

qiyunzhu:
@aryankeluskar I would say we can remove this unit test. Because we are going to call `multipletests`, and the different options will become values of a parameter of it (like `holm`, `fdr_bh`), rather than separate functions, the function `_dispatch_p_adjust` cannot return functions. Instead it will directly calculate adjusted p-values by calling `multipletests`. So the corresponding unit test should test the actual result rather than the function name.

Another thought: if the parameter is set to `None`, then the function should return the original p-values, I thought.

skbio/stats/composition.py
```diff
@@ -1147,78 +1147,41 @@ def tree_basis(tree):
...
-    bh_p = sm_multipletests(pvals=p, alpha=0.05, method="fdr_bh")[1]
-    bh_p = list(bh_p)
-    return bh_p
-
-
-def _calc_p_adjust(name, p):
-    """Refer to a *p*-value correction function by the method's name."""
     if name is None:
-        return
+        return p
```
aryankeluskar, line 1178: This line returns the original list if name is None

skbio/stats/composition.py
```diff
@@ -2059,7 +2036,8 @@ def dirmult_ttest(
 
     # multiple comparison
     if p_adjust is not None:
```
qiyunzhu, line 2038: I think that you `_calc_p_adjust` function already takes care of `None` and upper case. Therefore these checks are not necessary. Same for the other instance where `_calc_p_adjust` is called.

skbio/stats/composition.py
```diff
@@ -1146,81 +1147,57 @@ def tree_basis(tree):
...
+    # Original options are kept for backwards compatibility
     if name_ in ("holm", "holm-bonferroni"):
-        return _holm_bonferroni
-    elif name_ in ("bh", "fdr_bh", "benjamini-hochberg"):
-        return _benjamini_hochberg
+        name_ = "holm"
+    if name_ in ("bh", "fdr_bh", "benjamini-hochberg"):
+        name_ = "fdr_bh"
+
+    if name_ in (
```
qiyunzhu, line 1187: Here, you explicitly listed all FDR correction methods available in statsmodels. This is great. However, it might be broken by future updates of statsmodels, if they decide to add/remove/change methods. How about the following:

```python
try:
    res = sm_multipletests(pvals=arr, alpha=0.05, method='xxx')
except ValueError as e:
    if 'method not recognized' in str(e)
        raise ValueError(f'{name} is not an available FDR correction method.')
    else:
        raise ValueError(f'Cannot perform FDR correction using the {name} method.')
else:
    return res[1]
```

skbio/stats/composition.py
```diff
@@ -1146,81 +1147,57 @@ def tree_basis(tree):
...
+        "sidak",
+        "holm-sidak",
+        "simes-hochberg",
+        "hommel",
+        "fdr_by",
+        "fdr_tsbh",
+        "fdr_tsbky",
+    ):
+        res_p = sm_multipletests(pvals=p, alpha=0.05, method=name_)[1]
+        return list(res_p)
```
qiyunzhu, line 1200: The docstring says an ndarray will be returned. Therefore, it is not necessary to apply `list`.

skbio/stats/composition.py
```diff
@@ -1146,81 +1147,57 @@ def tree_basis(tree):
...
-    multiple comparisons.
+def _calc_p_adjust(name, p):
+    """
+    Calculate the p-value adjustment for a given method.
 
     Parameters
-    ----------
-    p : ndarray of shape (n_tests,)
-        Original *p*-values.
+    -------
```
qiyunzhu, line 1155: There are some issues with the format of the docstring. Despite that it won't be linted because it belongs to a helper function (starting with `_`), it is still a good habit to write docstring in a nice way.

qiyunzhu (changes requested):
@aryankeluskar Good job! I noticed a few more minor issues. Once they are addressed we can merge!

skbio/stats/composition.py
aryankeluskar: Thank you for your feedback! I have reviewed the function and removed the redundant code.

skbio/stats/composition.py
aryankeluskar: Oh yes, I may have overlooked some aspects of the formatting. I have made the necessary changes and ensured they align with the rest.

skbio/stats/composition.py
aryankeluskar: I have inserted the code. Thanks for pointing that out!

skbio/stats/composition.py
aryankeluskar: I now see that applying list is unnecessary, I have made the necessary changes.

qiyunzhu:
@aryankeluskar Thank you! This looks great. I only have one last comment: Line 2034: `p_adjust = str(p_adjust).lower()` is not necessary because line 1181 `name_ = name.lower()` already does this conversion. Can you please make one last trivial edit?

qiyunzhu:
@aryankeluskar Thank you! Once tests pass I will merge.

qiyunzhu (approved):

PR #2079, opened 2024-07-04
Adding stdin support for FASTA/FASTQ

RaeedA:
Updated version of solution to Issue #2017, able to read both fasta and fastq files.
Will give an error if attempting to read unsupported filetypes


skbio/io/format/_base.py
```diff
@@ -227,13 +229,18 @@ def _line_generator(fh, skip_blanks=False, strip=True):
 def _too_many_blanks(fh, max_blanks):
     count = 0
     too_many = False
+    consumed = []
     for line in _line_generator(fh, skip_blanks=False):
+        consumed += [line]
```
qiyunzhu, line 234: Please do `consumed.append(line)` for the reason we discussed before.

skbio/io/_iosources.py
```diff
@@ -180,14 +180,17 @@ def get_writer(self):
...
+
     def can_read(self):
         if hasattr(self.file, "__iter__"):
             iterator = iter(self.file)
             head = next(iterator, None)
             if head is None:
                 self.repaired = []
                 return True
-            if isinstance(head, str):
+            if isinstance(head, str) or isinstance(head, bytes):
```
qiyunzhu, line 193: Can be: `if isinstance(head, (str, bytes)):`.

skbio/io/registry.py
```diff
@@ -514,24 +533,44 @@ def _read_gen(self, file, fmt, into, verify, kwargs):
         # kwargs should still retain the contents of io_kwargs because the
         # actual reader will also need them.
         with _resolve_file(file, **io_kwargs) as (file, _, _):
-            reader, kwargs = self._init_reader(
+            seekable = True
+            try:
+                file.tell()
+            except io.UnsupportedOperation:
+                seekable = False
+            if not seekable and (fmt is not None or not fmt.support_non_seekable):
```
qiyunzhu, line 541: Here is a problem. This `fmt` is a string. It doesn't have properties like `support_non_seekable`. Similar above.

skbio/io/registry.py
```diff
@@ -497,14 +504,26 @@ def read(self, file, format=None, into=None, verify=True, **kwargs):
...
         io_kwargs = self._find_io_kwargs(kwargs)
         with _resolve_file(file, **io_kwargs) as (file, _, _):
-            reader, kwargs = self._init_reader(
+            seekable = True
+            try:
+                file.tell()
+            except io.UnsupportedOperation:
+                seekable = False
+            if not seekable:
+                print(seekable)
```
qiyunzhu, line 519: These two lines are redundant?

qiyunzhu (changes requested):
@RaeedA Good job! However, I tested your code and found that it doesn't work in real-world cases. Assuming there is `script.py` to parse standard input, one would hope that the behavior is like:

```bash
cat fasta_file.fasta | python script.py
```

The content of `script.py` is like:

```python
...
for seq in skbio.io.read(sys.stdin, format='fasta'):
    print(str(seq))
```

However, it doesn't work. You can test and debug.

I found that the following works, which is the case in your unit test:

```python
...
for seq in skbio.io.read(io.StringIO(sys.stdin.read()), format='fasta'):
    print(str(seq))
```

So the difference appears to be `sys.stdin` vs `io.StringIO`. The former is an instance of `io.TextIOWrapper`, which somehow resembles `io.StringIO` but not exactly the same. The correct mechanism for parsing the former might be missing so far.

The 2nd solution cannot be accepted as a workaround, because it reads the entire standard input into memory prior to be parsed by scikit-bio. This voids the original purpose of standard input (streamlining data).

Could you please take another look? Thank you!

skbio/io/registry.py
```diff
@@ -515,10 +515,15 @@ def _read_ret(self, file, fmt, into, verify, kwargs):
                 file.tell()
             except io.UnsupportedOperation:
                 seekable = False
-            if not seekable:
-                print(seekable)
-            if not seekable and (fmt is not None or not fmt.support_non_seekable):
-                raise ValueError("Cannot parse non-seekable data of type %r", fmt.name)
+            if not seekable and fmt is None:
```
qiyunzhu, line 518: This works but `not seekable` is checked twice. It can be combined into one check.

```python
if not seekable:
    if fmt is None:
        raise ValueError("Cannot sniff filetype of non-seekable data.")
    for lookup in self._lookups:
        ...
```

skbio/io/format/_base.py
```diff
@@ -227,13 +229,18 @@ def _line_generator(fh, skip_blanks=False, strip=True):
...
             count += 1
             if count > max_blanks:
                 too_many = True
                 break
-    fh.seek(0)
-    return too_many
+    try:
+        fh.seek(0)
+    except io.UnsupportedOperation:
+        pass
```
wasade, line 245: If the `seek` doesn't matter, why do it?

wasade: From below, it looks like a call to `.seekable` may work rather than a try/except?

skbio/io/format/fasta.py
```diff
@@ -640,22 +640,26 @@ def _fasta_sniffer(fh):
...
+    consumed = []
     num_records = 10
     empty = True
     try:
         parser = _parse_fasta_raw(fh, _sniffer_data_parser, FASTAFormatError)
-        for _ in zip(range(num_records), parser):
+        # Check 10 records, record is in the form of (index, data)
+        # Data is returned as consumed from the parser
+        for record in zip(range(num_records), parser):
+            consumed += record[1][3]
```
wasade, line 657: I'm not really sure what this is doing or why, the comment doesn't add much information here

It may help to perform a tuple expansion within the for loop statement, but even then, it's not obvious why the third index of the data variable is important

skbio/io/format/fasta.py
```diff
@@ -940,7 +948,7 @@ def _parse_fasta_raw(fh, data_parser, error_type):
                 data_chunks.append(line)
         prev = line
     # yield last record in file
-    yield data_parser(data_chunks), id_, desc
+    yield data_parser(data_chunks), id_, desc, consumed
```
wasade, line 951: I'm worried about separation of concerns here. The parser is responsible for reporting to the callee about what data it consumed, but the callee already knows what data it is providing. This adds unnecessary bookkeeping to the parser when this could be managed higher in the stack

skbio/io/format/fastq.py
```diff
@@ -313,52 +314,38 @@ def _fastq_sniffer(fh):
...
-        not_empty = False
-        for _, seq in zip(range(10), _fastq_to_generator(fh, phred_offset=33)):
-            split_length = len(
-                (seq.metadata["id"] + seq.metadata["description"]).split(":")
-            )
-            description = seq.metadata["description"].split(":")
+        # Scan 10 lines of the file and make sure they're valid
+        fastqgen = _parse_fastq_raw(fh, FASTQFormatError, phred_offset=33)
+        for _, seq in zip(range(num_records), fastqgen):
+            consumed += seq[4]
```
wasade, line 330: same here with the separation of concerns -- the sniffer ideally does not need to be responsible for what it's consumed as that can be managed higher in the stack

skbio/io/format/fastq.py
```diff
@@ -313,52 +314,38 @@ def _fastq_sniffer(fh):
...
+            split_length = len((seq[1] + seq[2]).split(":"))
+            description = seq[2].split(":")
```
wasade, line 332: Tuple expansion would remove the magic numbers 

I don't recall `:` being part of the fastq format other than perhaps use in qual?

wasade: Oh I see, to guess at phred encoding. Ugh.

wasade: It is not safe for the parser or sniffer to assume `:` is present in the description

skbio/io/format/fastq.py
```diff
@@ -632,7 +623,39 @@ def _parse_quality_scores(fh, seq_len, variant, phred_offset, prev):
...
+                "Sequence (@) and quality (+) header lines do not match: "
+                "%r != %r" % (str(seq_header[1:]), str(qual_header[1:]))
+            )
+
+        phred_scores, seq_header, lines = _parse_quality_scores(
+            fh, len(seq), variant, phred_offset, qual_header
+        )
+        # Return full sequence data and reset consumed
+        consumed += lines
+        yield seq, id_, desc, phred_scores, consumed
```
wasade, line 657: Same here re: separation of concerns

skbio/io/format/fasta.py
RaeedA: Hey Daniel! Me and @qiyunzhu discussed this, and I wanted to clarify that every file format has a unique parser within the scikit-bio system. Due to this, every sniffer reads a different amount of data at a different time, which is hard to predict before getting to this sniffer. That's why I decided to do parsing at the sniffer level instead of higher up. Doing it higher up also runs the risk of incorrectly predicting how much to give to the sniffer, resulting in false negatives.

qiyunzhu: Re: @wasade I discussed with Raeed and am under impression that the current solution is a necessary evil. The entire I/O may be revisited, and Matt may be able to work on it. But at this moment, I can't think of a better way to handle stdin.

wasade: We shouldn't expect the sniffers to be perfect. There are a lot of pathological cases for the common formats here. We don't know _for sure_ that we can parse until we actually fully parse, but that is expensive which is why we sniff.  

Why not standardize on reading and sniffing the first 1024 bytes -- are there any sniffers which actually need more than that?

wasade: @qiyunzhu, something like [this](https://github.com/wasade/mxdx/blob/main/mxdx/_io.py#L285-L301) doesn't work?

qiyunzhu: @wasade I am afraid that some sniffers need more than that, and it is hard to predict which one unless we examine every existing sniffer. It could also be that some sniffers (like csv-based ones) assumes complete table structure and could break if the file content is truncated to the first 1024 bytes. I agree that your solution is ideal, and I prefer that too. But to adopt it, we need to investigate every single existing sniffer in scikit-bio and propose a revision plan. This is probably beyond Raeed's capacity, and should be reassigned to Matt.

mataton:
Closing in favor of #2185

Issue #1948, opened 2024-02-22
Differential abundance analysis via Dirichlet Multinomial 


mortonjt:
After talking to @qiyunzhu , it is becoming clear that skbio would need to have a differential abundance method that can at least provide directionality (i.e. features increase / decrease in abundance). The simplest methodology would be to leverage the Dirichilet Multinomial approach, similar to what is done with aldex2. This implemented method wouldn't be "competitive", nor will it replicate the aldex2 functionality.  But it will provide a barebones differential abundance framework that we can iterate on.

This methodology would serve two purposes
1. It would provide an API placeholder for standard GLM-style differential abundance analyses.  If we wanted to have interoperability with other differential abundance methods, there would be a well-defined API that can be followed as a guideline.
2. It would provide a means to link the metadata object to the biom table object

mortonjt:
Next steps

-  [ ] Implement `dirmult_glm` for generalized linear model support. Requires [statsmodels glm](https://www.statsmodels.org/stable/glm.html) instead of [welch_ttest](https://github.com/scikit-bio/scikit-bio/blob/main/skbio/stats/composition.py#L1773). See unit tests for dirmult_ttest [here](https://github.com/scikit-bio/scikit-bio/blob/main/skbio/stats/tests/test_composition.py#L1348)
- [ ] Implement `dirmult_lme` for linear mixed effects. Requires [MixedLm](https://www.statsmodels.org/stable/generated/statsmodels.regression.mixed_linear_model.MixedLM.html)
- [ ] Implement `DifferentialSummary` object, that wraps resulting pandas dataframe. We need to discuss standardizing the outputs from `dirmult_ttest`, `dirmult_glm` and `dirmult_lme`.
- [ ] log ratio methods. Given two sets of microbial ids, compute the log ratios per sample. See log ratio figure in [Figure 1a](https://www.nature.com/articles/s41467-019-10656-5/figures/1). See here and [here](https://github.com/knightlab-analyses/american-gut-analyses/blob/master/ipynb/western-nonwestern-balances.ipynb) for examples on how to compute this
- [ ] imports of DESeq2 / ANCOM-BC (optional)

PR #2080, opened 2024-07-05
Implement linear mixed effects with Dirichlet-multinomial distribution

aryankeluskar:
Addressing #1948, this PR is the second step towards differential abundance analysis. I have added support for fitting a linear mixed effects model in a Dirichlet-multinomial distribution by using classes and functions from [statsmodels.regression.mixed_linear_model.MixedLM](https://www.statsmodels.org/stable/generated/statsmodels.regression.mixed_linear_model.MixedLM.html). Since I'm using some of the documentation from the `statsmodels` package, I have added its license as well.

I do request your assistance in these matters:
1) Could you please advise me on which folder and method for importing CSV data files for unit tests? 
2) What would you recommend including in the notes section of the new `dirmult_lme` function?


skbio/stats/composition.py
```diff
@@ -2058,3 +2059,91 @@ def dirmult_ttest(
...
+    ...                     [1, 7, 14.8], [1, 14, 16.1]],
+    ...                     index=[0, 1, 2, 3],
+    ...                     columns=["patient_id", "time_point",
+    ...                             "response"])
+    >>> result = dirmult_lme("response ~ time_point", groups="patient_id", data=data)
+    >>> print(result.summary())
+              Mixed Linear Model Regression Results
+    =========================================================
+    Model:             MixedLM  Dependent Variable:  response
+    No. Observations:  4        Method:              REML
```
aryankeluskar, line 2127: The trailing whitespace is automatically inserted by the `result.summary()`  function but is removed by ruff, which is causing a failure in doctests. I will find other outputs to replace this one.

qiyunzhu: A quick search found [this](https://github.com/astral-sh/ruff/issues/10275). There seems to have no easy solution now. A quick pack would be adding `# doctest: +SKIP` to the end of `print(result.summary())`. This will instruct Sphinx to skip this doctest, regardless the output is correct or not.

aryankeluskar: Thank you so much for your suggestion! I have appended it to the line, and it's no longer posing a problem.

skbio/stats/composition.py
```diff
@@ -2058,3 +2060,93 @@ def dirmult_ttest(
...
+    formula, data, groups, reml=True, method=None, fit_kwargs=None, **kwargs
+):
+    r"""Fit a Dirichlet Multinomial linear mixed effects model.
+
+    The Dirichlet-multinomial distribution is a compound distribution that
+    combines a Dirichlet distribution over the probabilities of a multinomial
+    distribution. This distribution is used to model the distribution of
+    species abundances in a community.
+
+    To perform
```
qiyunzhu, line 2075: This sentence is incomplete.

skbio/stats/composition.py
```diff
@@ -2058,3 +2060,93 @@ def dirmult_ttest(
...
+    pass
+
+    Examples
+    --------
+    >>> data = pd.DataFrame([[1, 0, 10.2], [1, 1, 12.5],
+    ...                     [1, 7, 14.8], [1, 14, 16.1]],
+    ...                     index=[0, 1, 2, 3],
+    ...                     columns=["patient_id", "time_point",
+    ...                             "response"])
+    >>> result = dirmult_lme("response ~ time_point", groups="patient_id", data=data)
```
qiyunzhu, line 2125: The function can take an R formula? That is cool!

qiyunzhu (commented):
Hi @aryankeluskar Thank you for making this effort! I am not sure if I have the necessary expertise to review this code. Calling @mortonjt . But my first impression is that the current code simply calls statsmodels but without other treatment. What is the "Dirichlet-multinomial distribution" ingradient in this code? I also think that we need to synchronize the interface of `dirmult_ttest` and `dirmult_lme` such that they return similar data structure (i.e., a pandas dataframe), rather than a statsmodels object. My overall feeling is that it is a good attempt, but more investigations are needed. I understand that this is a tough statistics task for you, and I can't help much... If @mortonjt you can help that will be great.

aryankeluskar:
Hello Qiyun, thank you for your feedback on my attempt. I have looked up more details on designing linear effects models with the dirichlet distribution, and will be implementing them shortly. My understanding so far is that the function should start by fitting a Dirichlet-multinomial distribution, then draw samples from it, and then I could use statsmodels to construct a linear effects model.

qiyunzhu:
@aryankeluskar I think your understanding is correct. Thank you!

skbio/stats/composition.py
```diff
@@ -2058,3 +2060,93 @@ def dirmult_ttest(
...
+    =========================================================
+    <BLANKLINE>
+
+    """
+
+    # Fit the Dirichlet Multinomial linear mixed effects model
+    model = MixedLM.from_formula(formula=formula, data=data, groups=groups, **kwargs)
+    if fit_kwargs is None:
+        fit_kwargs = {}
+    results = model.fit(reml=reml, method=method, **fit_kwargs)
```
mortonjt, line 2150: I think something was lost in translation here. 

There should be a double nested loop over [posterior draws](https://github.com/scikit-bio/scikit-bio/blob/main/skbio/stats/composition.py#L2003-L2009) and [features](https://github.com/scikit-bio/scikit-bio/blob/main/skbio/stats/composition.py#L2011-L2016) -- it should be almost an exact mirror of the dirmult_ttest, except with the ttest swapped out with the MixedLM function

skbio/stats/composition.py
mortonjt: I think the formula here is also off. The `'response'` here should be each microbe. But because we are dealing with many (i.e. hundreds / thousands) of microbes, it means that we are going to need to fix hundreds / thousands of MixedLM models. 

The `dirmult_lme` shouldn't make the user specify `response`, since it should be implicit with the input table.

aryankeluskar: Hello Jamie, thank you for your feedback. I was previously unaware of this aspect, but now I understand that `response` would be each microbe. For the latter half, could you please give me any tips or recommendations on how to implement this change?

mortonjt:
@aryankeluskar try using this as a template : https://github.com/mortonjt/differential/blob/main/differential/regression/_mixedlm.py

aryankeluskar:
Hello @mortonjt and @qiyunzhu, I have made some changes to the code which address your suggestions. I'm especially grateful for your code snippet Jamie, it helped me resolve multiple mathematical errors/warnings so thank you so much! Additionally, I have adjusted the handling of the formula, so users are no longer required to specify `response`.

However, I am currently facing an issue with converting the output into a DataFrame; temporarily, the code returns a dictionary with all of relevant information but converting it to a DataFrame throws the following error. I would greatly appreciate your assistance in resolving this matter.

```
File "/Users/aryank/Developer/scikit-bio/../test2.py", line 242, in <module>
    result = dirmult_lme("response ~ time_point", groups="patient_id", data=data)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/aryank/Developer/scikit-bio/../test2.py", line 197, in dirmult_lme
    return pd.DataFrame(res)
           ^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/skbio-dev/lib/python3.12/site-packages/pandas/core/frame.py", line 778, in __init__
    mgr = dict_to_mgr(data, index, columns, dtype=dtype, copy=copy, typ=manager)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/skbio-dev/lib/python3.12/site-packages/pandas/core/internals/construction.py", line 503, in dict_to_mgr
    return arrays_to_mgr(arrays, columns, index, dtype=dtype, typ=typ, consolidate=copy)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/skbio-dev/lib/python3.12/site-packages/pandas/core/internals/construction.py", line 114, in arrays_to_mgr
    index = _extract_index(arrays)
            ^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/skbio-dev/lib/python3.12/site-packages/pandas/core/internals/construction.py", line 670, in _extract_index
    index = union_indexes(indexes)
            ^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/skbio-dev/lib/python3.12/site-packages/pandas/core/indexes/api.py", line 319, in union_indexes
    index = _unique_indices(indexes, dtype)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/skbio-dev/lib/python3.12/site-packages/pandas/core/indexes/api.py", line 252, in _unique_indices
    result = result.sort_values()
             ^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/skbio-dev/lib/python3.12/site-packages/pandas/util/_decorators.py", line 333, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/anaconda3/envs/skbio-dev/lib/python3.12/site-packages/pandas/core/indexes/base.py", line 5932, in sort_values
    _as = nargsort(
          ^^^^^^^^^
  File "/opt/anaconda3/envs/skbio-dev/lib/python3.12/site-packages/pandas/core/sorting.py", line 439, in nargsort
    indexer = non_nan_idx[non_nans.argsort(kind=kind)]
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '<' not supported between instances of 'str' and 'int'
```

mortonjt:
Hi @aryankeluskar I don't recommend using the R-style formulas this way. Instead, I recommend using the code that we had in the link that I sent

https://github.com/mortonjt/differential/blob/main/differential/regression/_mixedlm.py#L132-L142

qiyunzhu:
@mortonjt @aryankeluskar I think we can exploit [pasty](https://patsy.readthedocs.io/en/latest/) -- a dependency of statsmodels for parsing R-style formula into Python variables. I hope that this can power up many statistical analyses in scikit-bio, including differential abundance tests and multivariate tests like PERMANOVA.

aryankeluskar:
Hello @mortonjt, I was able to implement looping through the table columns, and creating a different formula for each one as specified in https://github.com/mortonjt/differential/blob/65752567ef4cf303471405b0a9be503eb10a0bbb/differential/regression/_mixedlm.py#L134-L137. 

However, I need your help in understanding how to return the model for the `output` variable that is supposed to be implicit in the input table. Is the function expected to return the fitted model for both `Y1` and `Y2` as dependent variables separately or should the dependent variable be `patient` ?

aryankeluskar:
@qiyunzhu Surely, we could use patsy to enhance scikit-bio. However, while browsing through its [pip webpage](https://pypi.org/project/patsy/), I found out that patsy is no longer under active development, and is replaced by [Formulaic](https://github.com/matthewwardrop/formulaic) so we could use the latter in our future developments .

mortonjt:
Hi, it looks like patsy is still a dependency of statsmodels, so if we are depending on statsmodels, we are stuck with patsy. I think we'd need a solid reason to add formulaic

qiyunzhu:
@aryankeluskar Following @mortonjt 's comment, I would recommend that we adopt neither patsy nor formulaic within this PR. Let's start with something basic. R-style formula is a beast and we can have a separate PR to tackle it.

qiyunzhu:
@aryankeluskar Please disregard my previous comment (now deleted).

aryankeluskar:
@mortonjt Thank you for your response! As per your suggestion, I have added a doctest, but your answer did resolve my query. The final result is now formatted as follows:

```
      FeatureID  Covariate  Log2(FC)   CI(2.5)  CI(97.5)    pvalue    qvalue
    0        Y1       time -0.210769 -0.731153  0.309650  0.403737  0.403737
    1        Y1  treatment -0.744093 -1.780094  0.291999  0.252057  0.252057
    2        Y2       time  0.210769 -0.309650  0.731153  0.403737  0.403737
    3        Y2  treatment  0.744093 -0.291999  1.780094  0.252057  0.252057
```

Please let me know if there are any further adjustments needed to this result.

skbio/stats/composition.py
```diff
@@ -2058,3 +2061,407 @@ def dirmult_ttest(
...
+def _type_cast_to_float(df):
+    """Attempt to cast all of the values in dataframe to float.
+
+    This will try to type cast all of the series within the
+    dataframe into floats.  If a column cannot be type casted,
+    it will be kept as is.
+
+    See Also
+    --------
+    differential.regression.mixedlm
```
mortonjt, line 2075: Let's remove this, since everything has been ported from that differential package

skbio/stats/composition.py
```diff
@@ -2058,3 +2061,407 @@ def dirmult_ttest(
...
+    >>> res = dirmult_lme(
+    ...     formula="time + treatment", data=table, metadata=metadata,
+    ...     groups="patient", seed=0
+    ... )
+    >>> res
+      FeatureID  Covariate  Log2(FC)   CI(2.5)  CI(97.5)    pvalue    qvalue
+    0        Y1       time -0.210769 -0.731153  0.309650  0.403737  0.403737
+    1        Y1  treatment -0.744093 -1.780094  0.291999  0.252057  0.252057
+    2        Y2       time  0.210769 -0.309650  0.731153  0.403737  0.403737
+    3        Y2  treatment  0.744093 -0.291999  1.780094  0.252057  0.252057
```
mortonjt, line 2302: Something seems strange with this doctest. The pvalue should not be equal to the qvalue (qvalue should be significantly larger)

skbio/stats/composition.py
```diff
@@ -2058,3 +2061,407 @@ def dirmult_ttest(
...
+        method=method,
+        fit_kwargs=fit_kwargs,
+        **kwargs,
+    )
+
+    res = _call_res[0]
+    list_of_vars = _call_res[2]
+
+    submodels = _call_res[1]
+    for i in submodels:
```
mortonjt, line 2335: Note for the future (not in the PR), we will want to have the option to offload this to either joblib or multiprocessing so that models can be independently fitted

skbio/stats/composition.py
```diff
@@ -2058,3 +2061,407 @@ def dirmult_ttest(
...
+            single_covar_data["CI(97.5)"]
+        )
+        res_dict[single_covar_data["FeatureID"]][single_covar_data["Covariate"]][3] = (
+            single_covar_data["qvalue"]
+        )
+        res_dict[single_covar_data["FeatureID"]][single_covar_data["Covariate"]][4] = (
+            single_covar_data["Log2(FC)"]
+        )
+
+    for i in range(1, draws):
```
mortonjt, line 2374: Something is weird with this ordering. Within each of the bootstraps, the model should be refitted.
Note the call to welch ttest : https://github.com/scikit-bio/scikit-bio/blob/main/skbio/stats/composition.py#L2011-L2016

skbio/stats/composition.py
```diff
@@ -2058,3 +2061,407 @@ def dirmult_ttest(
...
+            res_dict[curr_feature_id][curr_covariate][1] += single_covar_data["CI(2.5)"]
+            res_dict[curr_feature_id][curr_covariate][2] += single_covar_data[
+                "CI(97.5)"
+            ]
+            res_dict[curr_feature_id][curr_covariate][3] += single_covar_data["qvalue"]
+            res_dict[curr_feature_id][curr_covariate][4] += single_covar_data[
+                "Log2(FC)"
+            ]
+
+    # TODO: implement online averages to save memory
```
mortonjt, line 2420: Let's do this now. We don't want to save all of the bootstraps in memory.

See these lines on how to implement this
https://github.com/scikit-bio/scikit-bio/blob/main/skbio/stats/composition.py#L2019-L2024

skbio/stats/tests/test_composition.py
```diff
@@ -1486,8 +1487,7 @@ def test_dirmult_ttest_missing_values_in_table(self):
     def test_dirmult_ttest_inconsistent_indexes(self):
         self.table.index = ['a', 'b', 'c', 'd', 'e']  # Change table index
         with self.assertRaises(ValueError):
-            dirmult_ttest(self.table, self.grouping, self.treatment, self.reference)
-
+            dirmult_ttest(self.table, self.grouping, self.treatment, self.reference)
```
mortonjt, line 1490: I don't see a test for dirmult_lme, let's make sure to add in multiple tests for this, namely the following

- [ ] Test with the original data from dirmult_ttest, but with an additional column in the metadata with at least 3 subjects (it can be random vector with values [0, 1, 2])
- [ ] Test failure modes to make sure that the input types are checked (i.e. numpy / pandas)
- [ ] Test the reml bool input
- [ ] Test a real dataset (i.e. [moving pictures](https://github.com/qiime2/docs/blob/master/source/tutorials/moving-pictures.rst) to make sure that time, body site and host_subject_id effects can be characterized (body site and host_subject_id should have pvalues that are significant)
- [ ] Add additional tests for including the `re_formula` and possibly the `vc_formula` that are inputted into the statsmodels [mixedlm function](https://www.statsmodels.org/devel/generated/statsmodels.regression.mixed_linear_model.MixedLM.from_formula.html#statsmodels.regression.mixed_linear_model.MixedLM.from_formula). It may require a bit of digging, but I would start with [one of their tutorials](https://www.statsmodels.org/devel/examples/notebooks/generated/mixed_lm_example.html) to see how they use this. 
- [ ] We may need to simulate an example to stress test this. See [this function](https://github.com/biocore/songbird/blob/master/songbird/util.py#L11) for an example of fitting a multinomial GLM (it will require a minor modification to add in random effects, but it should be just a couple of additional lines of code).

mortonjt (changes requested):
Thanks for putting this together. I think we are getting warmer in terms of a working LME model. Major comments
- [ ] Make sure that the bootstrapping is correct. It looks like the loop order is off
- [ ] Need to make sure that the p-value correction is correct. Those look identical now.
- [ ] Add support for online averaging the bootstraps. Otherwise, it will become too memory intensive to run this
- [ ] Tests, tests, tests. We should start with a simulation test to make sure that the fitted values are correct (i.e. run simulation with ground truth parameters, run model and see if model coefficients are the same as those used in the simulation). [Here is an example](https://github.com/biocore/songbird/blob/master/songbird/tests/test_multinomial.py#L24-L39) of what one of these unittests could look like end-to-end

skbio/stats/composition.py
```diff
@@ -2058,3 +2061,396 @@ def dirmult_ttest(
...
+        fit_kwargs=fit_kwargs,
+        **kwargs,
+    )
+
+    res = _call_res[0]
+    list_of_vars = _call_res[2]
+
+    submodels = _call_res[1]
+    for i in submodels:
+        i = i.fit(reml=reml, method=method, **fit_kwargs)
```
mortonjt, line 2332: Why not replace this with _lme_call?

licenses/statsmodels.txt
```diff
@@ -0,0 +1,34 @@
+Copyright (C) 2006, Jonathan E. Taylor
```
qiyunzhu, line 1: After the remodeling, I wonder if including statsmodels' license file is still necessary, since the documentation text is now quite different?

skbio/stats/composition.py
```diff
@@ -2058,3 +2061,423 @@ def dirmult_ttest(
...
+        except (TypeError, ValueError):
+            raise TypeError(errmsg % "Data")
+
+    if data.ndim != 2:
+        try:
+            data = pd.DataFrame(data, index=data.dtype.names)
+        except (TypeError, ValueError):
+            raise TypeError(errmsg % "Data")
+
+    if not hasattr(data, "__getitem__"):
```
qiyunzhu, line 2346: This `if` block can be merged into line 2334.

wasade: Yes, although ideally a specific type or subclass is checked. But, is this check necessary though given that the if on 2334 ensures `data` is a `DataFrame`?

skbio/stats/composition.py
```diff
@@ -2058,3 +2061,423 @@ def dirmult_ttest(
...
+    ``statsmodels.regression.mixed_linear_model``
+
+    Parameters
+    ----------
+    formula : str or generic Formula object
+        The formula specifying the model
+    data : DataFrame
+        The data for the model. data must define __getitem__ with the keys in
+        the formula terms args and kwargs are passed on to the model
+        instantiation. E.g., a numpy structured or rec array, a dictionary,
```
wasade, line 2217: The described allowed types appears different than the stated type expectation on line 2214

skbio/stats/composition.py
```diff
@@ -2058,3 +2061,423 @@ def dirmult_ttest(
...
+
+    # Test if metadata and data have the same index, regardless of order
+    if not data.index.sort_values().equals(metadata.index.sort_values()):
+        raise ValueError("Data and metadata must have the same index.")
+
+    rng = get_rng(seed)
+
+    posterior = [
+        rng.dirichlet(data.values[i] + pseudocount) for i in range(data.shape[0])
+    ]
```
wasade, line 2374: I think that should be equivalent too the below. But, note that the operation may be more efficient if the frame is expressed columnar as that is how Pandas stores data

```suggestion
    def dirichlet_with_pseudocount(pseudocount, row):
        return rng.dirichlet(row + pseudocount)
    dirichlet_f = functools.partial(dirichlet_with_pseudocount, pseudocount)
    posterior = data.apply(dirichlet_f, axis=1) 
```

skbio/stats/composition.py
```diff
@@ -2058,3 +2061,423 @@ def dirmult_ttest(
...
+    res_dict = {}
+
+    for feature in data.columns:
+        res_ = res_dict[feature] = {}
+        for covar in list_of_vars:
+            res_[covar] = [0] * 5
+
+    for single_covar_data in res:
+        group = res_dict[single_covar_data["FeatureID"]][single_covar_data["Covariate"]]
+        for i, key in enumerate(("pvalue", "CI(2.5)", "CI(97.5)", "Log2(FC)")):
```
wasade, line 2404: Expressing the common strings as static strings can reduce chance of certain bugs, while also improving both maintainability and readability

skbio/stats/composition.py
```diff
@@ -2058,3 +2061,423 @@ def dirmult_ttest(
...
+            group[i] = single_covar_data[key]
+
+    for i in range(1, draws):
+        posterior = [
+            rng.dirichlet(data.values[i] + pseudocount) for i in range(data.shape[0])
+        ]
+        dir_table = pd.DataFrame(clr(posterior), index=data.index, columns=data.columns)
```
wasade, line 2411: Suggested change above could also apply here, and I think also removes the need to construct the `df`. I suspect the `clr` operation could be performed within the `apply` as well

skbio/stats/composition.py
```diff
@@ -2058,3 +2061,423 @@ def dirmult_ttest(
...
+            fit_kwargs=fit_kwargs,
+            **kwargs,
+        )
+
+        ires = _call_ires[0]
+
+        if ires is None:
+            continue
+
+        # online average to avoid holding all of the results in memory
```
wasade, line 2429: If the amount of data being tracked is well under expected system memory, like 1GB for instance, then doing the aggregate statistics after the fact rather than online could simplify the work

skbio/stats/composition.py
```diff
@@ -2058,3 +2061,423 @@ def dirmult_ttest(
...
+            res_[1] = np.minimum(res_[1], single_covar_data["CI(2.5)"])
+            res_[2] = np.maximum(res_[2], single_covar_data["CI(97.5)"])
+            res_[3] = (i * res_[3] + single_covar_data["Log2(FC)"]) / (i + 1)
+
+    # convert all log fold changes to base 2
+    log2_ = np.log(2)
+    for feature in data.columns:
+        for covar in list_of_vars:
+            res_ = res_dict[feature][covar]
+            for i in range(1, 4):
```
wasade, line 2444: I think these relate to the indices above. But if the data are a common type, and the operation to perform is common, why not store in a contiguous `np.array` or `DataFrame` and apply in bulk?

skbio/stats/composition.py
```diff
@@ -2058,3 +2061,423 @@ def dirmult_ttest(
...
+                    "CI(97.5)": ci975,
+                    "pvalue": results.pvalues[var_name],
+                }
+
+            output.append(individial_results)
+
+    return (output, submodels, list_of_vars)
+
+
+def dirmult_lme(
```
wasade, line 2178: There are a few opportunities to decompose blocks of code in this method, which could simplify it substantially

skbio/stats/composition.py
mortonjt: Unfortunately the CLR function does not work with apply. I'm not sure why, but I have never been able to get it to work.  We got it to work after transposing the dataframe, but not directly. See code block 13 here : https://github.com/scikit-bio/scikit-bio-tutorials/blob/main/06-marker-inference/06-marker-inference.ipynb

If we *really* want to have the apply function, we can do the following
```
dir_table = data.apply(lambda x: np.log(rng.dirichlet(data.values[i] + pseudocount)) - np.log(rng.dirichlet(data.values[i] + pseudocount)).mean())
```

skbio/stats/composition.py
mortonjt: Unfortunately, the amount of data being tracked could be well over system memory.

Say you have 10K samples by 100K features by 10K draws, you'd be tracking 10 trillion entries.

skbio/stats/tests/test_composition.py
```diff
@@ -1486,8 +1488,214 @@ def test_dirmult_ttest_missing_values_in_table(self):
...
+        for each animal: a random intercept, and a random slope (with respect to time).
+        The model will have a better fit. This is observed by increase in CI 97.5% and
+        Coefficients (ie: Log2(FC)), and a decrease in CI 2.5% for this dataset.
+        Similarly, vc_formula describes the random effects of the LME model. The third
+        model will be fitted with the variance component of Feed variable being
+        described. The model will have a better fit. This is observed by increase in
+        CI 97.5% and Coefficients (ie: Log2(FC)), and a decrease in CI 2.5% for this dataset.
+        """
+        data = smdatasets.get_rdataset('dietox', 'geepack').data
+        table = data[["Weight"]]
```
mortonjt, line 1648: Let's try to replace this with the table object with many different species, not `weight`.

mortonjt: Hi, following up with this -- this test doesn't make sense, since this data is not realistic. It is probably better for us to just drop this test.

mortonjt:
Let's remove the last test added, since it isn't relevant. I think this is pretty close to merging, so let's aim to get the test passing again.

aryankeluskar:
@qiyunzhu on manual inspection, this PR still has some merge conflicts (eg: it's not able to detect changes made by #2110) so I'll try to fetch and merge again or open a new PR with `dirmult_lme`

qiyunzhu:
@aryankeluskar I see. It seems that the easiest fix is to rebuild the development environment from scratch and manually copy the edits you made into the relevant files (`composition.py` and `test_composition.py`), then recreate a new PR

PR #2094, opened 2024-07-25
Wrapping SciPy's UPGMA and WPGMA functions (WIP)

christapo:
SciPy's functions 'average' and 'weighted' implement UPGMA and WPGMA, respectively. These functions are included into scikit-bio while producing a TreeNode object instead of the linkage matrix produced by SciPy.

This is WIP. Remaining tasks are:
-unit testing
-changelog

### DIFF
--- a/skbio/tree/_upgma.py
+++ b/skbio/tree/_upgma.py
@@ -8,12 +8,12 @@
 
 import numpy as np
 
-from scipy.cluster.hierarchy import average, weighted
+from scipy.cluster.hierarchy import linkage
 from skbio import TreeNode
 from skbio.stats.distance import DistanceMatrix
 
 
-def upgma(dm: DistanceMatrix, weighted=False) -> TreeNode:
+def upgma(dm, weighted=False):
     r"""
     This function implements the Unweighted Pair Group Method with Arithmetic
     mean (UPGMA) algorithm as well as the weighted variant (WPGMA) for phylogenetic
@@ -36,8 +36,8 @@ def upgma(dm: DistanceMatrix, weighted=False) -> TreeNode:
     -----
     UPGMA is a hierarchical clustering method appearing as the 'average' function
     in the SciPy package, where the linkage matrix produced by 'average' is used
-    to construct a TreeNode object. A weighted variant known as WPGMA is attributed
-    to Sokal and Michener [1]_.
+    to construct a TreeNode object. A weighted variant is known as WPGMA, and both
+    variants are due to Sokal and Michener [1]_.
 
     References
     ----------
@@ -79,10 +79,14 @@ def upgma(dm: DistanceMatrix, weighted=False) -> TreeNode:
 
     # If weighted is set to 'False', UPGMA is performed
     if weighted is False:
-        linkage_matrix = average(dm.condensed_form())
+        linkage_matrix = linkage(
+            dm.condensed_form(), method="average", metric="euclidean"
+        )
     # Otherwise, WPGMA is performed
     else:
-        linkage_matrix = weighted(dm.condensed_form())
+        linkage_matrix = linkage(
+            dm.condensed_form(), method="weighted", metric="euclidean"
+        )
 
     # Construct the TreeNode from the linkage matrix
     tree = TreeNode.from_linkage_matrix(linkage_matrix, dm.ids)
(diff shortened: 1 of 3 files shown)


skbio/tree/_upgma.py
```diff
@@ -0,0 +1,90 @@
...
+# ----------------------------------------------------------------------------
+
+import numpy as np
+
+from scipy.cluster.hierarchy import average, weighted
+from skbio import TreeNode
+from skbio.stats.distance import DistanceMatrix
+
+
+def upgma(dm: DistanceMatrix, weighted=False) -> TreeNode:
```
qiyunzhu, line 16: Here, you are performing a very good practice: [typing](https://docs.python.org/3/library/typing.html). However, scikit-bio is backward compatible with as early as Python 3.8, and typing wasn't widely adopted by the time. Therefore, scikit-bio so far doesn't universally have typing in its codebase. Therefore we may skip typing until further discussions.

skbio/tree/_upgma.py
```diff
@@ -0,0 +1,90 @@
...
+    -------
+    TreeNode
+        A TreeNode object with estimated edge values
+
+    Notes
+    -----
+    UPGMA is a hierarchical clustering method appearing as the 'average' function
+    in the SciPy package, where the linkage matrix produced by 'average' is used
+    to construct a TreeNode object. A weighted variant known as WPGMA is attributed
+    to Sokal and Michener [1]_.
```
qiyunzhu, line 40: I thought both UPGMA and WPGMA were invented by Sokal and Michener?

skbio/tree/_upgma.py
```diff
@@ -0,0 +1,90 @@
+# ----------------------------------------------------------------------------
```
qiyunzhu, line 1: I thought a bit. Initially I thought that wrappers are trivial and don't have to be a separate module, and WPGMA can be a separate function from UPGMA ("U" stands for "unweighted"). But then I thought that beyond SciPy's implementation, there may more efficient implementations of UPGMA (e.g., [here](https://github.com/DavidCain/upgma)), which we may consider adopt in the future. Being the fastest tree-building method, UPGMA deserves attention (optimization). Therefore, having it in a separate module is justified.

skbio/tree/_upgma.py
```diff
@@ -0,0 +1,90 @@
...
+    >>> print(tree)
+    (c:1.25,(a:0.5,b:0.5):0.75);
+
+    """
+    # Ensure the input is a DistanceMatrix object
+    if not isinstance(dm, DistanceMatrix):
+        raise ValueError("Input must be a DistanceMatrix object.")
+
+    # If weighted is set to 'False', UPGMA is performed
+    if weighted is False:
```
qiyunzhu, line 81: This is good. Just for information, another way of writing is:

```python
func = average if weighted is False else weighted
linkage_matrix = func(dm.condensed_form())
```

qiyunzhu: On another thought: There is a namespace contamination issue here: `weighted` is both an imported SciPy function and a parameter in this function. The program doesn't know which one is which. A remedy would be:

```python
from scipy.cluster.hierarchy import average as sp_average
from scipy.cluster.hierarchy import weighted as sp_weighted
```

qiyunzhu: On a third thought: `average` and `weighted` are but wrappers of `linkage` in SciPy (see [here](https://github.com/scipy/scipy/blob/87c46641a8b3b5b47b81de44c07b840468f7ebe7/scipy/cluster/hierarchy.py#L408)). Then why don't we directly call `linkage`? Such that the namespace contamination issue can be resolved.

skbio/tree/_upgma.py
```diff
@@ -0,0 +1,92 @@
...
+    ---------|
+             |          /-a
+              \--------|
+                        \-b
+
+    The tree also has estimated edge values assigned to each edge.
+
+    >>> print(tree)
+    (c:1.25,(a:0.5,b:0.5):0.75);
+
```
qiyunzhu, line 72: To resolve the doctest error, you can literally add a line of `<BLANKLINE>` here. @christapo

qiyunzhu:
@christapo I meant adding a line which says `<BLANKLINE>`, not a blank line.

qiyunzhu (approved):
@christapo and I discussed in person. This PR is good as-is. Let's merge.

Issue #2083, opened 2024-07-07
TreeNode cache mechanism needs revision


qiyunzhu:
This issue is relevant to #2082 but needs to be addressed separately. `skbio.tree.TreeNode` has two mechanisms of caching information as attributes of the nodes to facilitate certain operations. 1) caching a name-to-node dictionary to facilitate searching by name, 2) caching certain node attributes in a postorder manner to facilitate some calculations. These mechanisms are useful but they make other operations inefficient. For example, operations like `append`, `extend`, `remove`, `pop` and `shuffle` automatically call the method `invalidate_caches`, which clears out caches of the tree. This operation involves at minimum climbing to the root, and at maximum traversing the entire tree. Additionally, caches complicate tree operations that requires subsetting, rerooting, etc. There isn't a robust (or well-documented) mechanisms in the codebase to address these challenges. In #2082 and #2073 I fixed some of these issues to make the operations more performant (e.g., `extend` no longer requires one call of `invalidate_caches` per child). However, I think that it will be useful to carefully revisit the caching mechanism overall.

PR #2082, opened 2024-07-06
TreeNode: Improved time and memory efficiency by making caches lazy

qiyunzhu:
I examined the mechanism of loading and storing phylogenetic trees in scikit-bio, and realized that a major bottleneck is the following [three lines](https://github.com/scikit-bio/scikit-bio/blob/0131e023a72cc45d650032a1237c35f290c23805/skbio/tree/_tree.py#L93) during the initiation of a `TreeNode` instance:

```python
self._tip_cache = {}
self._non_tip_cache = {}
self._registered_caches = set()
```

These three private attributes significantly increased the memory and runtime burden.

Here, `_tip_cache` and `_non_tip_cache` are dictionaries to facilitate searching by node name. They are only needed at the root node of a tree (see `create_caches`). However, they are created for every single node, which causes a huge waste. `_registered_caches` is very useful in customizing node attributes in a postorder style. But it is not used elsewhere in scikit-bio, and its creation is done through the public method `cache_attr`, which a user will use.

The current PR made these three private attributes created-by-need. This simple change significantly improved the time and memory efficiency of tree loading, especially with extremely large trees.

Benchmark on a tree with 212,293 tips:

Tree object:
- Memory (MB): 275.8 -> 126.8 (**54%** decrease)

Load Newick tree:
- Time (sec): 5.585 -> 3.822 (**32%** decrease)

Find 100 tips:
- Time (sec): 0.5700 ->0.4572 (**20%** decrease)

***

Please complete the following checklist:

* [x] I have read the [contribution guidelines](https://scikit.bio/contribute.html).

* [ ] I have documented all public-facing changes in the [changelog](https://github.com/scikit-bio/scikit-bio/blob/main/CHANGELOG.md).

* [ ] **This pull request includes code, documentation, or other content derived from external source(s).** If this is the case, ensure the external source's license is compatible with scikit-bio's [license](https://github.com/scikit-bio/scikit-bio/blob/main/LICENSE.txt). Include the license in the `licenses` directory and add a comment in the code giving proper attribution. Ensure any other requirements set forth by the license and/or author are satisfied.

  - **It is your responsibility to disclose** code, documentation, or other content derived from external source(s). If you have questions about whether something can be included in the project or how to give proper attribution, include those questions in your pull request and a reviewer will assist you.

* [x] This pull request does not include code, documentation, or other content derived from external source(s).

Note: [This document](https://scikit.bio/devdoc/review.html) may also be helpful to see some of the things code reviewers will be verifying when reviewing your pull request.

PR #2099, opened 2024-08-04
Improved tree node cache functionality (2 of 3)

qiyunzhu:
This PR follows #2082 and partially addresses #2083. It improves the caching mechanism of `TreeNode`.

Previously, every node in a tree has three private attributes: `_tip_cache`, `_non_tip_cache` and `_registered_caches`. They are largely empty or duplicate. This is very inefficient, especially in memory. #2082 removed the necessity of the former two. The current PR removed the last one. Now they are attached to the root node only in a lazy mode.

Specifically, `_tip_cache`, `_non_tip_cache` are lookup tables which facilitate searching nodes by name. They are automatically invoked in `find` and `find_all`. Meanwhile, `_registered_caches` stores names of attributes assigned to each node. These attributes are assigned in a post order traversal using the [`cache_attr`](https://scikit.bio/docs/latest/generated/skbio.tree.TreeNode.cache_attr.html) method.

The `cache_attr` method is very useful in saving tree traversal time for many applications. So far, it is only causally used in the [`majority_rule`](https://scikit.bio/docs/latest/generated/skbio.tree.majority_rule.html) function to calculate a consensus tree given multiple trees. The current PR significantly expanded its functionality. Now, one can calculate node attributes such as: 1) number of nodes descending from each node, 2) total sum of branch lengths of each clade, 3) accumulative distances from all tips to each node. Calculating these properties is much faster with the improved `cache_attr` than calling alternative methods individually for multiple times. The docstring of this method has been enriched to provide these examples.

**TODO**: The last important thing is to make cache clearance optional in every single tree manipulation operation. Currently, even appending a single node requires traversing the entire tree to remove the caches. This is very inefficient. Some methods even requires many calls of the clearance operation. All of them need to be optimized to save unnecessary traversals.

At least the following methods need to be improved: `append`, `extend`, `insert`, `remove`, `unroot`, `pop`, `shuffle`. Each method is also used by other methods and they need to be optimized as well. For example, `append` is used in `copy`, `bifurcate`, `unpack_by_func`, `from_taxonomy` and `from_linkage_matrix`.

***


skbio/tree/_tree.py
```diff
@@ -2398,53 +2530,118 @@ def find(self, name):
...
+        All internal nodes (including root) and tips with the given name will be
+        returned, with the former placed before the latter in the returned list.
+
+        The first call to ``find_all`` will cache all nodes in the tree on the
+        assumption that additional calls to ``find_all`` will be made.
+
+        Examples
+        --------
+        >>> from skbio.tree import TreeNode
+        >>> tree = TreeNode.read(["((a,b)c,(d,e)d,(f,g)c);"])
```
mataton, line 2590: May be illustrative to include the tree ASCII art here.

mataton (approved):
My only comments are minor changes to the documentation. The actual code looks good to me.

qiyunzhu:
@mataton Thank you! I have addressed your comments.

PR #2105, opened 2024-08-13
Adding versions of tree construction and tree rearrangement functions based on Balanced Minimum Evolution 

christapo:
Versions of the GME and NNI functions that follow the Balanced Minimum Evolution estimation model are added. In particular, the function for Balanced Minimum Evolution (BME) has been added along with an implementation of Balanced Nearest Neighbor Interchange (BNNI) as an option for NNI.


skbio/tree/_bme.py
```diff
@@ -0,0 +1,352 @@
...
+
+def bme(dm, allow_edge_estimation=True):
+    r"""Perform balanced minimum evolution (BME) for phylogenetic reconstruction.
+
+    Parameters
+    ----------
+    dm : skbio.DistanceMatrix
+        Input distance matrix containing distances between taxa.
+    allow_edge_estimation : bool, optional
+        Whether to perform an OLS-based estimation of edge values (``True``, default)
```
qiyunzhu, line 24: I guess "edge value" means branch length?

skbio/tree/_bme.py
```diff
@@ -0,0 +1,352 @@
...
+        or return a tree without edge values assigned (``False``).
+
+    Returns
+    -------
+    TreeNode
+        Reconstructed phylogenetic Tree with estimated edge
+        values (if ``allow_edge_estimation`` is ``True``).
+
```
qiyunzhu, line 32: Maybe add a "See Also" section with `nj` and `gme`. Also add `bme` to the "See Also" section of `nj`.

skbio/tree/_bme.py
```diff
@@ -0,0 +1,352 @@
...
+        raise ValueError(
+            "Distance matrix must be at least 3x3 to "
+            "generate a minimum evolution tree."
+        )
+
+    # count the number of taxa
+    n = len(dm.ids)
+
+    # create initial nodes
+    root_name, node_a_name, node_b_name = dm.ids[0], dm.ids[1], dm.ids[2]
```
qiyunzhu, line 90: There is a trick:

```python
root_name, node_a_name, node_b_name, *_ = dm.ids
```

skbio/tree/_bme.py
```diff
@@ -0,0 +1,352 @@
...
+                                 |          /-chicken
+                                  \--------|
+                                            \-rat
+
+    Notice that, unlike neighbor joining, the tree is rooted at a taxa/leaf node.
+    This will allow it to have nearest neighbor interchange performed on it without
+    needing to re-root the tree.
+
+    """
+    if dm.shape[0] < 3:
```
qiyunzhu, line 80: I think this can be `n = dm.shape[0]`, and remove line 87.

skbio/tree/_bme.py
```diff
@@ -0,0 +1,352 @@
...
+        raise ValueError(
+            "Distance matrix must be at least 3x3 to "
+            "generate a minimum evolution tree."
+        )
+
+    # count the number of taxa
+    n = len(dm.ids)
+
+    # create initial nodes
```
qiyunzhu, line 89: Because `dm.ids` will be accessed many times during the iteration, it will be more efficient to do `taxa = dm.ids` before the iteration, and use `taxa` within the iteration.

skbio/tree/_bme.py
```diff
@@ -0,0 +1,352 @@
...
+    root_child = TreeNode("")
+    node_a = TreeNode(node_a_name)
+    node_b = TreeNode(node_b_name)
+    root.append(root_child)
+    root_child.append(node_a)
+    root_child.append(node_b)
+
+    # loop over rest of nodes for k = 4 to n
+    for k in range(3, n):
+        taxa = root.count(tips=True) + 1
```
qiyunzhu, line 101: `root.count` is expensive. Is my guess correct that each iteration adds one taxon to the tree? If so, `taxa` should be equal to `k`.

skbio/tree/_bme.py
```diff
@@ -0,0 +1,352 @@
...
+        # calculate average distance matrix between subtrees of T_(k-1)
```
qiyunzhu, line 102: The maximum line width is 88.

skbio/tree/_bme.py
```diff
@@ -0,0 +1,352 @@
...
+        # however the updating step requires all subtrees to be updated
+        # in contrast to GME, so instead the matrix is recomputed after
+        # every iteration.
+        adm = _balanced_average_matrix(root, dm)
+        # create node for taxa k
+        node_k_name = dm.ids[k]
+        connecting_node = TreeNode("")
+        node_k = TreeNode(node_k_name)
+        connecting_node.append(node_k)
+        # connecting_node = TreeNode.read(["(%s);" % (node_k_name)])
```
qiyunzhu, line 114: Here are two lines of unused code. If useless, can delete.

skbio/tree/_bme.py
```diff
@@ -0,0 +1,352 @@
...
+        ordered = list(root.postorder(include_self=False))
+        lowerlist = _balanced_lower(ordered, node_k, dm)
+        upperlist = _balanced_upper(ordered, node_k, taxa, dm)
+        # Initialize the list of edge parent nodes for computation.
+        # The tuple is a pair of a TreeNode object and the value of the tree
+        # length for attaching at the edge consisting of the node and its parent.
+        # Here, we start with setting the initial length value of the edge
+        # defined by the root and its unique descendant to 0.
+        edge_list = [(root_child, 0)]
+        for a, i in edge_list:
```
qiyunzhu, line 126: This code is risky. You are extending a list while iterating it. Are you sure it works as expected? (Maybe yes, but just double-check).

skbio/tree/_bme.py
```diff
@@ -0,0 +1,352 @@
...
+            )
+            value2 = _balanced_attach_length(
+                a2, lowerlist, upperlist, ordered, taxa, adm
+            )
+            e1 = (a1, value1 + i)
+            e2 = (a2, value2 + i)
+            edge_list.append(e1)
+            edge_list.append(e2)
+        # find the edge with minimum length after edge attachment
+        minimum_child = sorted(edge_list, key=itemgetter(1))[0][0]
```
qiyunzhu, line 141: Repeated call of `itemgetter(1)` is expensive. You may add a line `key = itemgetter(1)` outside the for loop, and use it during iteration.

skbio/tree/_bme.py
```diff
@@ -0,0 +1,352 @@
...
+        minimum_parent.append(connecting_node)
+        connecting_node.append(minimum_child)
+    if allow_edge_estimation:
+        _bal_ols_edge(root, dm)
+    return root
+
+
+def _balanced_lower(ordered, node, dm):
+    """Return the list of values representing the change in tree length
+    after attaching a taxa at an edge not the root edge. The subtree
```
qiyunzhu, line 153: The singular form of "taxa" is "taxon".

skbio/tree/_bme.py
```diff
@@ -0,0 +1,352 @@
...
+    involved is a lower subtree.
+
+    This list is indexed by the tree's postorder.
+
+    """
+    lower_list = []
+    for i, a in enumerate(ordered):
```
qiyunzhu, line 160: The variable `i` is not used. So this line may be `for a in ordered:`.

skbio/tree/_bme.py
```diff
@@ -0,0 +1,352 @@
...
+def _balanced_average_matrix(tree, dm):
+    """Return the matrix of distances between pairs of subtrees.
+
+    Here, the definition of average distance coincides with the balanced minimum
+    evolution problem.
+
+    """
+    # start from unique descendant
+    name = tree.name
+    (tree,) = tree.children
```
qiyunzhu, line 227: This line can be `tree, = tree.children`

skbio/tree/_bme.py
```diff
@@ -0,0 +1,352 @@
...
+                    i3 = index
+            edge_node.length = 0.5 * (adm[i2, i1] + adm[i3, i1] - adm[i2, i3])
+        # calculate edge lengths for external edges
+        elif edge_node.is_tip():
+            a = parent.parent
+            if a.is_root():
+                for child in a.children:
+                    a = child
+            b = edge_node.siblings()[0]
+            for index, node in enumerate(ordered):
```
qiyunzhu, line 322: This `index` may be `i` because the word "index" may refer to row names in a dataframe. Whereas `i` is the typical way of denoting an index. Same below.

skbio/tree/_bme.py
```diff
@@ -0,0 +1,352 @@
...
+                elif node is b:
+                    i3 = index
+            edge_node.length = 0.5 * (adm[i2, i1] + adm[i3, i1] - adm[i2, i3])
+        # calculate edge lengths for external edges
+        elif edge_node.is_tip():
+            a = parent.parent
+            if a.is_root():
+                for child in a.children:
+                    a = child
+            b = edge_node.siblings()[0]
```
qiyunzhu, line 321: If you are certain that `edge_node` has only one sibling, you can do `b, = edge_node.siblings()`

skbio/tree/_bme.py
```diff
@@ -0,0 +1,352 @@
...
+                if node is a:
+                    i2 = index
+                elif node is b:
+                    i3 = index
+            edge_node.length = 0.5 * (adm[i2, i1] + adm[i3, i1] - adm[i2, i3])
+        # calculate edge lengths for external edges
+        elif edge_node.is_tip():
+            a = parent.parent
+            if a.is_root():
+                for child in a.children:
```
qiyunzhu, line 319: Can this be `a, = a.children`?

skbio/tree/_nni.py
```diff
@@ -110,19 +123,28 @@ def nni(tree, dm, allow_edge_estimation=True, inplace=True):
...
+    if balanced:
+        adm = _balanced_average_matrix(tree, dm)
+    else:
+        adm = _average_distance_matrix(tree, dm)
     while True:
         # create heap of possible swaps and then swapping subtrees
         # until no more swaps are possible.
         adm = _average_distance_matrix(tree, dm)
-        heap = _swap_heap(tree, adm)
+        if balanced:
```
qiyunzhu, line 134: These four lines may simply be: `heap = _swap_heap(tree, adm, balanced)`

skbio/tree/_nni.py
```diff
@@ -110,19 +123,28 @@ def nni(tree, dm, allow_edge_estimation=True, inplace=True):
...
+            heap = _swap_heap(tree, adm)
+        else:
+            heap = _swap_heap(tree, adm, balanced=False)
         if not heap:
```
qiyunzhu, line 138: By changing the order of code, this loop can be modified into `while heap:` in the beginning, without explicitly checking `if not heap:`. But this is trivial and no need to spend time on changing the current code.

qiyunzhu (commented):
@christapo Good job! I added some minor comments.

qiyunzhu:
@christapo I see that you have addressed all my comments. Well done! Is it ready to merge?

christapo:
> @christapo I see that you have addressed some of my comments. Well done! Do you think we can merge it or are you still working on it?

Thank you! This PR should be ready to merge.

skbio/tree/_bme.py
qiyunzhu: @christapo There are still a few unaddressed comments. This one is most important. If my guess is correct, it can save lots of compute.

skbio/tree/_bme.py
christapo: I agree that this code is not best practice and prone to error. My idea is to instead use a separate list of all non-tip/root nodes and then to loop through this list to compute the corresponding tuples for edge_list. This will avoid having to mutate a list during iteration.

qiyunzhu: @christapo Mutating a list while visiting may not always be a bad practice, as long as it is done right. For example [here](https://github.com/scikit-bio/scikit-bio/blob/e6417514c47fce9e140e3017d7c9c7f7e4ca1728/skbio/tree/_tree.py#L697). But in your case, I fear that it will fall into an infinite loop, as demonstrated in the following code:

```python
lst = [1, 2, 3]
for x in lst:
    lst.append(x * 2)
```

If you did not encounter infinite loop, then maybe this piece of code wasn't trigger at all. You may want to double check to see if that's the case.

To make this code safe, you can iterate the list by a pre-defined range of indices, such as:

```python
for k in range(len(edge_list)):
    a, i = edge_list[k]
```

qiyunzhu: @christapo On another thought, please take my previous words with caution. I see in your code that there is `if a.is_tip(): continue`, meaning that the iteration may not infinitely extend `edge_list`. If you are certain that the current code is correct, you don't need to change it.

PR #2113, opened 2024-08-23
Implement linear mixed effects with Dirichlet-multinomial distribution (without merge conflicts)

aryankeluskar:
Addressing #1948, this PR is the second step towards differential abundance analysis. I have added support for fitting a linear mixed effects model in a Dirichlet-multinomial distribution by using classes and functions from [statsmodels.regression.mixed_linear_model.MixedLM](https://www.statsmodels.org/stable/generated/statsmodels.regression.mixed_linear_model.MixedLM.html). Since I'm using some of the documentation from the `statsmodels` package, I have added its license as well.

To resolve the merge conflicts, I detached the head and created a new branch from the commit right before we had merge conflicts (ie: [35fb17d](https://github.com/scikit-bio/scikit-bio/pull/2080/commits/35fb17d6438d83a84cb03bfec914732a3b9a313b)). Following this, I executed `git fetch` and `git merge` to get the latest changes. On manual inspection, it looks like recently merged PRs are reflected in this unlike #2080


aryankeluskar:
I'll wait for the tests to pass, and then manually re-inspect to verify new code is present.

skbio/stats/composition.py
```diff
@@ -2058,3 +2060,489 @@ def dirmult_ttest(
...
+        ]
+        dir_table = pd.DataFrame(
+            posterior, index=data.index, columns=data.columns
+        ).apply(clr, axis=1)
+        dir_table = pd.DataFrame(
+            dir_table.tolist(), index=dir_table.index, columns=data.columns
+        )
+    else:
+        posterior = rng.dirichlet(data.values.flatten() + pseudocount)
+        dir_table = np.log(posterior)
```
mortonjt, line 2226: This doesn't look correct. It should be CLR, not log

skbio/stats/composition.py
```diff
@@ -2058,3 +2060,489 @@ def dirmult_ttest(
...
+                    }
+
+                    individual_results.update(measures)
+
+            output.append(individual_results)
+
+    return (output, submodels, _covariate_list)
+
+
+def _obtain_dir_table(data, pseudocount, rng):
```
mortonjt, line 2213: Can we do a refactor of the existing `dirmult_test` and replace all of the posterior calculations there?

https://github.com/scikit-bio/scikit-bio/blob/ffc47bf6af7d7948249e03e6b61d7848161323fc/skbio/stats/composition.py#L1993-L1996

https://github.com/scikit-bio/scikit-bio/blob/ffc47bf6af7d7948249e03e6b61d7848161323fc/skbio/stats/composition.py#L2006-L2011

It will add another sanity check to make sure that this code is working as expected. Plus it will help clean up the dirmult_ttest as well (there will be less redundancy)

skbio/stats/composition.py
```diff
@@ -2058,3 +2060,489 @@ def dirmult_ttest(
...
+        model = MixedLM.from_formula(
+            formula=stats_formula,
+            data=merged_data,
+            groups=groups,
+            re_formula=re_formula,
+            vc_formula=vc_formula,
+            **kwargs,
+        )
+
+        results = model.fit(reml=reml, method=method, **fit_kwargs)
```
mortonjt, line 2148: I don't believe that we need to explicitly expose `reml` in the args. If we just remove this in all of the function signatures, users should still be able to pass `reml` in to the `fit` method via the `fit_kwargs`.

I suggest deleting `reml` from all of the function signatures, but keep the tests the same to confirm.

skbio/stats/composition.py
```diff
@@ -2058,3 +2060,489 @@ def dirmult_ttest(
...
+                    FEATUREID: response_var,
+                    COVARIATE: var_name,
+                    LOG2FC: _lme_coef,
+                    CI25: _lme_25,
+                    CI975: _lme_975,
+                    PVALUE: results.pvalues[var_name],
+                }
+            except Exception as e:
+                if type(e) is not ValueError:
+                    print("Something went really wrong, details below: ")
```
mortonjt, line 2168: This could give users a negative reaction. Let's rename this to 

```suggestion
                    print(f"LME fit failed for covariate {var_name} and `response_var`, outputting nans.")
```

skbio/stats/composition.py
```diff
@@ -2058,3 +2060,489 @@ def dirmult_ttest(
...
+
+    return dir_table
+
+
+def dirmult_lme(
+    formula,
+    data,
+    metadata,
+    groups=None,
+    reml=True,
```
mortonjt, line 2237: Let's remove this and see users can still pass in `reml` (aka check if the tests still pass)

skbio/stats/composition.py
```diff
@@ -2058,3 +2060,489 @@ def dirmult_ttest(
...
+            continue
+    return df
+
+
+def _lme_call(
+    formula,
+    metadata,
+    table,
+    groups,
+    reml=True,
```
mortonjt, line 2096: Same here, let's see if we can exclude `reml` to see if users can still pass it in via `fit_kwargs`.

Looking over this more carefully, it may require `fit_kwargs = {'reml': True}`

skbio/stats/composition.py
```diff
@@ -2058,3 +2060,489 @@ def dirmult_ttest(
...
+        contain duplicate indices.
+    metadata: array-like
+        The metadata for the model. If metadata is a pd.DataFrame, it must contain
+        the covariates in metadata.columns. If metadata is not a pd.DataFrame,
+        it must contain the covariates in indices of metadata. metadata can
+        be a numpy structured array, or a numpy recarray, or a dictionary. It must
+        not contain duplicate indices.
+    groups : str
+        The column name in data that identifies the grouping variable
+    reml : bool
```
mortonjt, line 2284: Let's try to remove this

mortonjt (changes requested):
Hi @aryankeluskar thanks for putting this together.  This is *really* close to merging.

I added a few requests to further improve the user experience (i.e. better messages, less complex function by removing `reml`), as well as a refactoring request for dirichlet posterior function to add more confirmation that it works as expected.

Once those 3 comments are addressed, I think this is good to merge.

skbio/stats/composition.py
aryankeluskar: True, but I noticed that when a feature table with only one column is passed through this function, all values are modified to 0, resulting in the posterior containing only 1s. From my understanding, `closure` ensures that all elements in a row sum to 1, so when the table has only one column, all of the values become 1. This is significantly deviating the results compared to just using `statsmodels`. Meanwhile, `data` with more than one column is handled as before.

If this outcome is expected with the use of Dirichlet-Multinomial Distribution, please let me know and I will change it back to handling data of all dimensions with `clr`.

skbio/stats/composition.py
aryankeluskar: Thank you so much for this suggestion! It does describes the error in a much better way.

skbio/stats/composition.py
aryankeluskar: Yes, I have refactored `dirmult_ttest` to use the new function and removed the redundant code. It continues to work as expected.

skbio/stats/composition.py
aryankeluskar: I have made the necessary changes by removing `reml` from all function signatures as recommended. The functionality remains intact, and the updated function call is 
```python
fit_kwargs = {"reml": False}
res_ml = dirmult_lme(formula=formula, data=self.table, metadata=self.metadata, groups='Covar1', seed=0, p_adjust=None, fit_kwargs=fit_kwargs)
```

skbio/stats/composition.py
aryankeluskar: `reml` is set to True by default in `statsmodels` functions, so users wouldn't necessarily need to pass it as `fit_kwargs = {'reml': True}`. But I did test it out and the functionality remains intact.

skbio/stats/composition.py
mortonjt: I don't think we should even consider this scenario. perhaps in the else statement, we can raise a ValueError instead of computing the log. Perhaps this warrants another unittest to confirm that this error is raised as expected.

skbio/stats/tests/test_composition.py
```diff
@@ -1635,5 +1635,13 @@ def test_dirmult_lme_inconsistent_indexes(self):
             formula = "Covar2 + Covar3"
             dirmult_lme(formula=formula, data=self.table, metadata=self.metadata, groups='Covar1', seed=0, p_adjust="sidak")
 
+    def test_dirmult_lme_single_feature(self):
+        with self.assertRaises(ValueError):
+            # keep only the first column in the table
+            self.table = self.table.iloc[:, :1]
+            print(self.table)
```
mortonjt, line 1642: Let's remove this print statement

mortonjt:
This is basically ready to merge. Let's remove that print statement and update the changelog with the proposed changes

qiyunzhu:
@mortonjt How does it look now?

mortonjt:
Looks good. Merging.

PR #2116, opened 2024-08-27
Improving documentation of `dirmult_lme`

qiyunzhu:
Hi @aryankeluskar Congratulations on completing the `dirmult_lme` function! This PR exposes your work in the documentation of scikit-bio, such that users will see it. Like this:

image

@mortonjt I also cleaned up the docstring a little bit to make some terms more accurate or linking to correct sources. However, there are a few things I am unsure:

1. There are multiple `print` calls in the code. Are these intentional?
2. When I ran the doctest manually, it threw a ton of warnings like: "ConvergenceWarning: Maximum Likelihood optimization failed to converge. Check mle_retvals". It might reduce user experience. I wonder if it is an issue of the toy dataset?
3. Wrapper parameter inconsistency: In the current function, `method` is a parameter of `MixedLM.fit`. `re_formula` and `vc_formula` are parameters of `MixedLM.from_formula` (PS: they were not in the docstring; I added them). `fit_kwargs` is for `MixedLM.fit`. `**kwargs` are for `MixedLM.from_formula`. This might be confusing.
  - How about renaming `**kwargs` as `model_kwargs`, merging `method` into `fit_kwargs`, and merging `re_formula` and `vc_formula` into `model_kwargs`? Such that we have -3 extra parameters.
4. Interface inconsistency. In the relevant function [`dirmult_ttest`](https://scikit.bio/docs/dev/generated/skbio.stats.composition.dirmult_ttest.html), the parameters are table, grouping, treatment, reference... But in the current function, the parameters are formula, data, metadata, groups... I think that a consistent interface will be helpful. But I understand that lme calls statsmodels therefore the interface may have to be different.

***


skbio/stats/composition.py
```diff
@@ -2327,6 +2341,7 @@ def dirmult_lme(
     --------
     statsmodels.formula.api.mixedlm
     statsmodels.regression.mixed_linear_model.MixedLM
+    statsmodels.stats.multitest.multipletests
     differential.regression.mixedlm
```
mortonjt, line 2345: Should we remove this?
```suggestion
```

Also, shouldn't we add `skbio.stats.composition.dirmult_ttest`?

mortonjt:
```suggestion
    differential.regression.mixedlm

  References
    ----------
    .. [1] Fernandes et al. "Unifying the analysis of
       high-throughput sequencing datasets: characterizing RNA-seq,
       16S rRNA gene sequencing and selective growth experiments by
       compositional data analysis." Microbiome (2014).
```
I'm also wondering if we should add in the aldex2 reference for the Dirichlet multinomial reference

mortonjt (approved):
Overall I think this is good to go. See my comments and once you think it is OK I can merge.

qiyunzhu:
Hi @mortonjt Thanks for reviewing! I have addressed your comments and edits. Can you please also take a look at my comments on top of this PR? Thanks

aryankeluskar:
@qiyunzhu Thank you very much for the kind words! I think I have an answer to the ConvergenceWarning. It indeed is stemming from the toy dataset, since using the sample datasets (such as `dietox` from R) provided by statsmodels didn't throw these warnings.

aryankeluskar:
Regarding the `print` calls, I believe they are intentional. All of them are elaborating in case of an error and will not be invoked in most cases. First set of them are used to print the indices of `data` and `metadata` if they don't match from 2413-2414. Second set print the exact error message in case it is beyond our handling from 2174-2179. The latter exists in case the model returns a value other than a valid number or `NaN`, which is very unlikely.

qiyunzhu:
@aryankeluskar Thanks for the explanation on `print`. My understanding of the typical practices of developing a library is that it shouldn't print anything at all. The library means to be programmatically called by other code, instead of interacting with an end user. Any feedback should be delivered in the form of [exception](https://docs.python.org/3/tutorial/errors.html) or [warning](https://docs.python.org/3/library/warnings.html). But I would like to consult with @mortonjt on this too.

mortonjt:
Thanks for catching this. Yes we should have warnings thrown
https://github.com/scikit-bio/scikit-bio/blob/3697ec490031e69e573361e68830a900ae664738/skbio/stats/composition.py#L2179-L2184

qiyunzhu:
@aryankeluskar and I discussed today. I had better understanding of the rationale behind this function. The additional edits should address all issues. Most importantly, the modifications ensure the greatest possible consistency between the interfaces of `dirmult_lme` and `dirmult_ttest`. Also, the original unit tests were very slow due to 128 draws. They are accelerated with fewer draws in this modification. Finally, the drawing process has been optimized to avoid format-conversion overheads.

qiyunzhu:
Let me merge this because otherwise the unit test is lengthy and unreadable.

PR #2117, opened 2024-08-29
Added `TreeNode.is_bifurcating`

qiyunzhu:

### DIFF
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -1801,11 +1801,14 @@ class TreeNode(SkbioObject):
         See Also
         --------
         prune
+        is_bifurcating
 
         Notes
         -----
         This method does not modify single-child nodes. These nodes can be collapsed
         using :meth:`prune` prior to this method to create a strictly bifurcating tree.
+
+        This method modifies the subtree under the current node.
 
         Examples
         --------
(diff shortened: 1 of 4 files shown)


skbio/tree/_tree.py
```diff
@@ -3116,6 +3119,54 @@ def assign_supports(self):
...
+        if strict:
+            for node in self.non_tips(include_self=True):
+                if len(node.children) != 2:
+                    return False
+            return True
+        else:
+            for node in self.traverse(include_self=True):
+                if len(node.children) > 2:
+                    return False
+            return True
```
wasade, line 3168: A redundant return can be removed with:

```suggestion
        else:
            for node in self.traverse(include_self=True):
                if len(node.children) > 2:
                    return False
        return True
```

A further reduction in replication could be done with the below, though for something like this I think it's really just a stylistic item. 

```python
import operator

if strict:
    test = operator.ne
else:
    test = operator.gt

for node in self.non_tips(include_self=True):
    if test(len(node.children), 2):
        return False

return True
```

wasade (approved):
Just a minor comment. And, I think it would be nice to note in the ChangeLog

skbio/tree/_tree.py
qiyunzhu: @wasade Interesting! Your suggestion is very elegant. I think that `test` as a function call may be slower than a straight operator, but perhaps Python has optimization for this. Let me benchmark it.

wasade: Possibly though this isn't a performance critical method unless there is a reasonable need to run `is_bifurcating` a very large number of times on the same tree?

qiyunzhu:
Hi @wasade I have benchmarked, and realized that the true bottleneck is `non_tips`, whereas the operator functions you suggested has minimum impact on the performance. Therefore, I have updated the PR to adopt your suggestion, with slight modifications. Here are the benchmarks:

Version 1:
```python
if strict:
    for node in self.traverse(include_self=True):
        if node.children and len(node.children) != 2:
            return False
else:
    for node in self.traverse(include_self=True):
        if len(node.children) > 2:
            return False
return True
```

Version 2:
```python
test = ne if strict else gt
for node in self.non_tips(include_self=True):
    if test(len(node.children), 2):
        return False
return True
```

Version 3:
```python
test = ne if strict else gt
for node in self.traverse(include_self=True):
    if (children := node.children) and test(len(children), 2):
        return False
return True
```

Benchmarks on the 10.5k-tip tree:

```
v1: 1.83 ms ± 44.2 μs
v2: 3.66 ms ± 96.6 μs
v3: 1.88 ms ± 7.13 μs
v1, strict: 1.81 ms ± 7.42 μs
v2, strict: 3.61 ms ± 72.8 μs
v3, strict: 1.95 ms ± 42.2 μs
```

Issue #2059, opened 2024-06-20
The IO registry sniffer is being used unexpectedly, and it is very expensive


wasade:
In the example below, a tree is being loaded into `skbio.TreeNode`. A format is not provided, so the IO registry is attempting to determine the format. The kernel was interrupted after a few minutes, and as can be seen below, the code is bound up in Python's CSV sniffer (this was observed on multiple kernel interruptions).

If specifying `format='newick'`, the code completes in seconds.

The code was run using skbio 0.6.1 on Python 3.11.9. This is new. What I suspect is that the `csv` module in Python may have changed in recent versions with its sniffer. What I believe is going on is that newick files are typically very long single lines, and that is likely quite expensive with the iterated regex's in the `csv` sniffer. To resolve this, perhaps constraining skbio's sniifer to the first N characters would suffice. An additional improvement would be to, when `into=` is specified, to limit the specific formats tested -- in the below, skbio is testing whether the data are `lsmat` which as far as I know is not a container to represent a tree.

```python
> sktree = skbio.read('qiita_study_13114-2022.10.phylogeny.asv.nwk', into=skbio.TreeNode)
---------------------------------------------------------------------------
KeyboardInterrupt                         Traceback (most recent call last)
Cell In[43], line 1
----> 1 sktree = skbio.read('qiita_study_13114-2022.10.phylogeny.asv.nwk', into=skbio.TreeNode)

File ~/miniconda3/envs/sk061/lib/python3.11/site-packages/skbio/io/registry.py:1177, in read(file, format, into, verify, **kwargs)
   1174 @wraps(IORegistry.read)
   1175 def read(file, format=None, into=None, verify=True, **kwargs):
   1176     """Read data from file using specified format."""
-> 1177     return io_registry.read(file, format=format, into=into, verify=verify, **kwargs)

File ~/miniconda3/envs/sk061/lib/python3.11/site-packages/skbio/io/registry.py:500, in IORegistry.read(self, file, format, into, verify, **kwargs)
    498         return (x for x in [])
    499 else:
--> 500     return self._read_ret(file, format, into, verify, kwargs)

File ~/miniconda3/envs/sk061/lib/python3.11/site-packages/skbio/io/registry.py:505, in IORegistry._read_ret(self, file, fmt, into, verify, kwargs)
    503 io_kwargs = self._find_io_kwargs(kwargs)
    504 with _resolve_file(file, **io_kwargs) as (file, _, _):
[...]

PR #2128, opened 2024-09-16
Make IO registry sniffer only sniff relevant formats for a given skbio object

mataton:
This PR partially addresses issue #2059 by making the IO registry sniffer only try formats which make sense for a given skbio object passed to the `into` parameter of `skbio.read`.

For example, if a user is trying to read a file into a `TreeNode` object, it doesn't make sense for the sniffer to check if the file is FASTA format or any other format which cannot represent a tree.

* [] This pull request does not include code, documentation, or other content derived from external source(s).

### DIFF
--- a/skbio/io/registry.py
+++ b/skbio/io/registry.py
@@ -353,7 +353,7 @@ class IORegistry:
                 if cls in getattr(format, lookup_name):
                     yield format.name
 
-    def sniff(self, file, **kwargs):
+    def sniff(self, file, into=None, **kwargs):
         r"""Detect the format of a given file and suggest kwargs for reading.
 
         Parameters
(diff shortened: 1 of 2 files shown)


skbio/io/registry.py
```diff
@@ -422,8 +422,14 @@ def sniff(self, file, **kwargs):
 
         return matches[0]
 
-    def _find_matches(self, file, lookup, **kwargs):
+    def _find_matches(self, file, lookup, into=None, **kwargs):
         matches = []
+        # Only run sniffers for the relevant formats for the given skbio object.
+        if into is not None:
```
qiyunzhu, line 428: I am thinking if this code block can be elevated into `sniff`, such that we don't have to override a mandatory parameter (`lookup`)?

ebolyen (approved):
Looks pretty good to me :)

qiyunzhu:
@ebolyen Thanks for the review! @mataton Good job! I think we can merge now. If there is room to add unit tests, you may pursue that next.

PR #2131, opened 2024-09-20
Adding path function for TreeNode

christapo:

### DIFF
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -636,19 +636,19 @@ class TreeNode(SkbioObject):
         --------
         >>> from skbio import TreeNode
         >>> tree = TreeNode.read(["((a,b)c,(d,e)f)root;"])
-        >>> node_1, node_2 = tree.find('a'), tree.find('b')
+        >>> node_1, node_2 = tree.find('a'), tree.find('d')
         >>> path = node_1.path(node_2)
         >>> print(path)
         ['c', 'root', 'f']
         >>> path_2 = node_1.path(node_2, include_ends=True)
         >>> print(path_2)
-        ['a', 'c', 'root', 'f', 'e']
+        ['a', 'c', 'root', 'f', 'd']
 
         """
         # find root
         root = self.root()
 
-        # find the lowest common ancestor
+        # find the lowest common ancestor if not ancestors of each other
         lca = (
             self
             if self in other.ancestors()
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -617,6 +617,72 @@ class TreeNode(SkbioObject):
 
     lca = lowest_common_ancestor  # for convenience
 
+    def path(self, other, include_ends=False):
+        r"""Return the list of node names in the path from one node to another.
+
+        Parameters
+        ----------
+        other : TreeNode
+            Final node of path.
+        include_ends: bool, optional
+            Whether to include the initial and final nodes in the list.
+
+        Returns
+        -------
+        list
+            List of names of TreeNode objects.
+
+        Examples
+        --------
+        >>> from skbio import TreeNode
+        >>> tree = TreeNode.read(["((a,b)c,(d,e)f)root;"])
+        >>> node_1, node_2 = tree.find('a'), tree.find('b')
+        >>> path = node_1.path(node_2)
+        >>> print(path)
+        ['c', 'root', 'f']
+        >>> path_2 = node_1.path(node_2, include_ends=True)
+        >>> print(path_2)
+        ['a', 'c', 'root', 'f', 'e']
+
+        """
+        # find root
+        root = self.root()
+
+        # find the lowest common ancestor
+        lca = (
+            self
+            if self in other.ancestors()
+            else other
+            if other in self.ancestors()
+            else root.lowest_common_ancestor([self, other])
+        )
+
+        # create tree rooted at lca
+        lca_tree = lca.copy()
+
+        # find path from each node to the lowest common ancestor.
+        node_1, node_2 = lca_tree.find(self.name), lca_tree.find(other.name)
+        list_1, list_2 = (
+            [node.name for node in node_1.ancestors()],
+            [node.name for node in node_2.ancestors()],
+        )
+
+        # reverse the path of final node to ensure correct order
+        list_2.reverse()
+
+        # remove the lca from the first list since the lca
+        # is also contained in the second list.
+        list_1 = list_1[:-1]
+
+        # combine the lists
+        path_list = list_1 + list_2
+
+        # insert initial and final nodes if desired
+        if include_ends is True:
+            path_list.insert(0, self.name)
+            path_list.append(other.name)
+        return path_list
+
     # ------------------------------------------------
     # Tree traversal
     # ------------------------------------------------
--- a/skbio/tree/tests/test_tree.py
+++ b/skbio/tree/tests/test_tree.py
@@ -276,6 +276,27 @@ class TreeTests(TestCase):
         with self.assertRaises(ValueError):
             t1.lowest_common_ancestor([])
 
+    def test_path(self):
+        """List of TreeNode object names in path between nodes"""
+        t1 = TreeNode.read(["((a,(b,c)d)e,f,(g,h)i)j;"])
+        t2 = t1.copy()
+        t3 = t1.copy()
+        init1 = t1.find("a") # tip
+        init2 = t2.find("j") # root node
+        init3 = t3.find("e") # internal node
+        fin1 = t1.find("g") # tip
+        fin2 = t2.find("c") # tip
+        fin3 = t3.find("i") # internal node
+        exp1 = ['e', 'j', 'i']
+        exp2 = ['j', 'e', 'd']
+        exp3 = ['j']
+        obs1 = init1.path(fin1) # tip-to-tip
+        obs2 = init2.path(fin2) # root-to-tip
+        obs3 = init3.path(fin3) # internal-to-internal
+        self.assertEqual(obs1, exp1)
+        self.assertEqual(obs2, exp2)
+        self.assertEqual(obs3, exp3)
+
     # ------------------------------------------------
     # Tree traversal
     # ------------------------------------------------


skbio/tree/tests/test_tree.py
```diff
@@ -276,6 +276,35 @@ def test_lowest_common_ancestor(self):
...
+        t2 = t1.copy()
+        t3 = t1.copy()
+        init1 = t1.find("a") # tip
+        init2 = t2.find("j") # root node
+        init3 = t3.find("e") # internal node
+        fin1 = t1.find("g") # tip
+        fin2 = t2.find("c") # tip
+        fin3 = t3.find("i") # internal node
+        exp1 = ['e', 'j', 'i']
+        exp2 = ['j', 'e', 'd']
```
qiyunzhu, line 291: This result is incorrect -- since `include_ends` is not set, the initial node "j" should not be in the path.

skbio/tree/_tree.py
```diff
@@ -617,6 +617,72 @@ def lowest_common_ancestor(self, tipnames):
 
     lca = lowest_common_ancestor  # for convenience
 
+    def path(self, other, include_ends=False):
+        r"""Return the list of node names in the path from one node to another.
```
qiyunzhu, line 621: There is an exception to handle: What if there is no path connecting self and other? This could happen if they are not in the same tree.

skbio/tree/_tree.py
```diff
@@ -617,6 +617,72 @@ def lowest_common_ancestor(self, tipnames):
...
+        ----------
+        other : TreeNode
+            Final node of path.
+        include_ends: bool, optional
+            Whether to include the initial and final nodes in the list.
+
+        Returns
+        -------
+        list
+            List of names of TreeNode objects.
```
qiyunzhu, line 633: I think that it should return a list of TreeNodes instead of their names. This is mainly because internal nodes usually don't have unique names, or name at all.

skbio/tree/_tree.py
```diff
@@ -617,6 +617,72 @@ def lowest_common_ancestor(self, tipnames):
...
+        lca = (
+            self
+            if self in other.ancestors()
+            else other
+            if other in self.ancestors()
+            else root.lowest_common_ancestor([self, other])
+        )
+
+        # create tree rooted at lca
+        lca_tree = lca.copy()
```
qiyunzhu, line 661: Copying a subtree is an expensive operation. We should avoid it wherever possible. Additionally, once you copy a tree, you lose reference to original nodes.

skbio/tree/_tree.py
```diff
@@ -617,6 +617,72 @@ def lowest_common_ancestor(self, tipnames):
...
+        >>> path_2 = node_1.path(node_2, include_ends=True)
+        >>> print(path_2)
+        ['a', 'c', 'root', 'f', 'd']
+
+        """
+        # find root
+        root = self.root()
+
+        # find the lowest common ancestor if not ancestors of each other
+        lca = (
```
qiyunzhu, line 652: There are two `ancestor` and one `lca` calls, which are expensive. I tend to think that we can identify the LCA of the two nodes from their ancestors, without explicitly calling `lca`. Specifically, you can interate over the two list of ancestors from root, and get the last pair of elements that are identical. For example:

```python
anc1 = self.ancestors()
anc2 = other.ancestors()
for n1, n2 in zip(reversed(anc1), reversed(anc2)):
    if n1 is n2:
        lca = n1
    else:
        break
```

skbio/tree/_tree.py
```diff
@@ -617,6 +617,72 @@ def lowest_common_ancestor(self, tipnames):
...
+            else root.lowest_common_ancestor([self, other])
+        )
+
+        # create tree rooted at lca
+        lca_tree = lca.copy()
+
+        # find path from each node to the lowest common ancestor.
+        node_1, node_2 = lca_tree.find(self.name), lca_tree.find(other.name)
+        list_1, list_2 = (
+            [node.name for node in node_1.ancestors()],
```
qiyunzhu, line 666: No need to call `ancestors` again since you did it already.

qiyunzhu (changes requested):
@christapo Good attempt! Though there is room to improve.

skbio/tree/_tree.py
```diff
@@ -617,6 +617,75 @@ def lowest_common_ancestor(self, tipnames):
...
+        # create list of ancestors for both nodes
+        anc1, anc2 = (
+            [node for node in self.ancestors()],
+            [node for node in other.ancestors()],
+        )
+
+        # find lca and delete nodes from intersection of ancestors
+        for n1, n2 in zip(reversed(anc1), reversed(anc2)):
+            if n1 is n2:
+                del anc1[anc1.index(n1)]
```
christapo, line 662: Should I have a copy of the ancestor lists to modify instead?

skbio/tree/_tree.py
```diff
@@ -617,6 +617,77 @@ def lowest_common_ancestor(self, tipnames):
...
+
+        """
+        # check to see if nodes are on same tree
+        root1, root2 = self.root(), other.root()
+        if root1 is not root2:
+            raise TypeError("Could not find path between nodes.")
+
+        # create list of ancestors for both nodes
+        anc1, anc2 = (
+            [node for node in self.ancestors()],
```
qiyunzhu, line 655: `ancestors` already returns a list of nodes. Therefore there is no need to do `node for node in`.

skbio/tree/_tree.py
```diff
@@ -617,6 +617,77 @@ def lowest_common_ancestor(self, tipnames):
...
+        >>> path = node_1.path(node_2)
+        >>> print(len(path))
+        3
+        >>> path_2 = node_1.path(node_2, include_ends=True)
+        >>> print(len(path_2))
+        5
+
+        """
+        # check to see if nodes are on same tree
+        root1, root2 = self.root(), other.root()
```
qiyunzhu, line 649: Both `root` and `ancestor` involves traversing from the current node to the root. If a tree has n taxa, the time complexity of traversing to the root would be $O(log_2 n)$, which is expensive. Therefore, rather than calling `root` and `ancestor` separately, you may only call `ancestor`, the last node of which must be the root.

qiyunzhu: Additionally, you don't even need to explicitly compare `anc1[-1] is anc2[-1]`. The `for` loop below already covers this. If it doesn't produce an `lca`, then the two nodes are not connected.

skbio/tree/_tree.py
```diff
@@ -617,6 +617,77 @@ def lowest_common_ancestor(self, tipnames):
...
+            raise TypeError("Could not find path between nodes.")
+
+        # create list of ancestors for both nodes
+        anc1, anc2 = (
+            [node for node in self.ancestors()],
+            [node for node in other.ancestors()],
+        )
+
+        # find lowest common ancestor
+        for n1, n2 in zip(reversed(anc1), reversed(anc2)):
```
qiyunzhu, line 660: Missing `lca = None` above.

skbio/tree/_tree.py
```diff
@@ -617,6 +617,77 @@ def lowest_common_ancestor(self, tipnames):
...
+            if n1 is n2:
+                lca = n1
+            else:
+                break
+
+        # delete nodes from intersection of ancestors
+        del anc1[anc1.index(lca) :]
```
qiyunzhu, line 667: `del` is expensive because you are mutating a list in place. `index` is expensive too because you are searching a list, which is $O(n)$. I think a better way is to modify the above `for` loop into:

```python
for i, (n1, n2) in enumerate(zip(reversed(anc1), reversed(anc2))):
    ...
```

This will return the index of the lca without the need to find it in the subsequent code. Then you might do something like:

```python
path = anc1[:len(anc1) - i] + [n1] + anc2[:len(anc2) - i][::-1]
```

skbio/tree/_tree.py
```diff
@@ -617,6 +617,77 @@ def lowest_common_ancestor(self, tipnames):
...
+        del anc2[anc2.index(lca) :]
+
+        # reverse the second list for correct order
+        anc2.reverse()
+
+        # handle cases for nodes that are ancestors of each other
```
qiyunzhu, line 673: I am not sure if the instances of nodes being ancestors of each other need to be singled out. What if we don't?

skbio/tree/_tree.py
```diff
@@ -617,6 +617,77 @@ def lowest_common_ancestor(self, tipnames):
...
+        elif other in anc1:
+            del anc1[anc1.index(other)]
+        else:
+            anc1 = anc1 + [lca]
+
+        # create list of nodes in path
+        path_list = anc1 + anc2
+
+        # insert initial and final nodes if desired
+        if include_ends is True:
```
qiyunzhu, line 685: It should be cheaper to slice a list than to insert an element to the beginning of a list. For example:

```python
if include_ends is False:
    path_list = path_list[1:-1]
```

qiyunzhu (changes requested):

skbio/tree/_tree.py
qiyunzhu: A Python list is a [stack](https://www.geeksforgeeks.org/difference-between-stack-and-queue-data-structures/). Adding / deleting elements to the end is efficient, but that to the beginning is inefficient.

qiyunzhu: On a 2nd thought -- if `path_list` doesn't already contain head and tail, you may disregard my suggestion above.

skbio/tree/_tree.py
christapo: The path between a node and an ancestor will be the complement of the intersection of the two ancestor lists. Otherwise, the path will be the union of the two ancestor lists and subtracting their intersection while keeping the lowest common ancestor. 

These two constructions are not equivalent. For example, the lowest common ancestor is not in the path between a node and an ancestor, but does appear in the path otherwise.

qiyunzhu: My understanding is that if node A is ancestral to node B, then the LCA of A and B is A, and the path from A to B is A straight down to B. So LCA is indeed in the path, as long as `include_ends=True`. I think the logic of your code below is correct. But I wonder even without the two `if`s above, it will still work. I am not sure if my guess is correct (because I didn't code). Maybe you can try?

skbio/tree/_tree.py
```diff
@@ -617,6 +617,75 @@ def lowest_common_ancestor(self, tipnames):
...
+        """
+        # create list of ancestors including nodes themselves
+        anc1, anc2 = (
+            [self] + self.ancestors(),
+            [other] + other.ancestors(),
+        )
+
+        # initialize variables
+        lca = None
+        lca_i = None
```
qiyunzhu, line 656: A trick is that even without pre-defining `lca_i`, you can just use `i` after the for loop.

skbio/tree/_tree.py
```diff
@@ -617,6 +617,75 @@ def lowest_common_ancestor(self, tipnames):
...
+                break
+
+        # check to see if nodes are on same tree
+        if lca is None:
+            raise TypeError("Could not find path between nodes.")
+
+        # handle cases for nodes that are ancestors of each other
+        if self in anc2:
+            path = anc2[: len(anc2) - lca_i][::-1]
+        elif other in anc1:
```
qiyunzhu, line 674: These `other in anc1` can be replaced with `self is lca`, which is faster. But as I guessed above, you probably don't need this check.

skbio/tree/_tree.py
christapo: Your idea is correct with my code as I add the nodes themselves to the beginning of their ancestor lists in lines 650-651, since the ancestor function does not include the node itself in the list. 

However, maybe I should note that the lowest common ancestor function in TreeNode does not have this behavior, and will return the parent of A rather than A in your example.

qiyunzhu: You are right. I tested your code and mine and found that your code works.

skbio/tree/_tree.py
```diff
@@ -617,6 +617,66 @@ def lowest_common_ancestor(self, tipnames):
...
+
+        Examples
+        --------
+        >>> from skbio import TreeNode
+        >>> tree = TreeNode.read(["((a,b)c,(d,e)f)root;"])
+        >>> node_1, node_2 = tree.find('a'), tree.find('d')
+        >>> path = node_1.path(node_2)
+        >>> print(len(path))
+        3
+        >>> path_2 = node_1.path(node_2, include_ends=True)
```
qiyunzhu, line 643: Can add:

```
>>> print('-'.join(x.name for x in path))
c-root-f
```

Same below.

skbio/tree/_tree.py
```diff
@@ -617,6 +617,66 @@ def lowest_common_ancestor(self, tipnames):
...
+        for i, (n1, n2) in enumerate(zip(reversed(anc1), reversed(anc2))):
+            if n1 is n2:
+                lca = n1
+                lca_i = i
+            else:
+                break
+
+        # check to see if nodes are on same tree
+        if lca is None:
+            raise TypeError("Could not find path between nodes.")
```
qiyunzhu, line 667: This may be `TreeError`.

qiyunzhu (approved):
Good job! I have a few last minor comments. Otherwise it is good to go.

christapo:
> Good job! I have a few last minor comments. Otherwise it is good to go.

Thank you! I have added and committed the recommended changes.

qiyunzhu:
@christapo Thanks! Let me merge.

PR #2135, opened 2024-09-24
Optimized `TreeNode.ancestors`

qiyunzhu:
This PR adds parameter `include_self` to `TreeNode.ancestors` to optionally include the initial node in the path (default: False). This logic is consistent with the newly added `path`, which has `include_ends`. With this parameter, one can easily generate a lineage from any taxon to the root.

@christapo I guess this addition can also make `path` simpler. What do you think?

This PR also improved the performance of `ancestors`. In my benchmarks:

Before: 40.3 ms ± 565 μs
After: 21.8 ms ± 194 μs
After (with `include_self=True`): 22.1 ms ± 278 μs


christapo:
I think this option is a good idea! I think it will simplify the`path` function as well as other potential functions that will utilize `TreeNode.ancestors`.

qiyunzhu:
@christapo Thanks! Let's see!

PR #2136, opened 2024-10-04
WIP:  Remove `monkey_patch` in favor of descriptors for generating `read` and `write` methods

mataton:
# Overview
This PR implements new strategies for managing scikit-bio's `read` and `write` methods for registered classes, as well adding functionality for removing  methods from the registry. 

# Motivation
The motivation here is that the `monkey_patch` method modifies scikit-bio classes (such as `TreeNode` or `TabularMSA`) after runtime, increasing complexity and decreasing maintainability. The `read` and `write` methods for various scikit-bio objects are not explicitly defined within the class, making it unclear where these methods are coming from. Furthermore, with the `monkey_patch` method, all the `read` and `write` methods **and** the documentation for every registered object are generated at runtime, i.e. during `import skbio`. This introduces a large and often unnessecary amount of overhead, as most users are not making use of the `read` and `write` methods for every type of object that scikit-bio offers.

Comparatively, the descriptor approach is more explicit, with the `read` and `write` methods for each registered object being explicitly defined within that class. Additionally, the methods are generated lazily, meaning that they are not generated until a user calls the `read` or `write` method for a particular scikit-bio object. In order to avoid generating the method and its docstring on every call to `read` or `write`, the descriptor protocol has been set up to attach the given method and its documentation to the respective scikit-bio object on the first call of that method. In theory, this should reduce the execution time of `import skbio`, but further testing is required to assess this.

# To Do
Unit tests still need to be developed for the refactored code, and benchmarking to determine performance changes is still pending. Documentation may be improved as well.


skbio/io/__init__.py
```diff
@@ -325,4 +325,4 @@ class of Python's standard library. The goal of a ``sniffer`` is two-fold: to
 
 # Now that all of our I/O has loaded, we can add the object oriented methods
 # (read and write) to each class which has registered I/O operations.
-io_registry.monkey_patch()
+# io_registry.monkey_patch()
```
wasade, line 328:
```suggestion
```

skbio/io/registry.py
```diff
@@ -244,6 +245,31 @@ def add_format(self, format_object):
...
+
+        Parameters
+        ----------
+        format : str or Format object
+            The format to be removed from the registry. It can be a name or the Format
+            object itself.
+
+        """
+        if isinstance(format, skbio.io.registry.Format):
+            # print(format)
```
wasade, line 259:
```suggestion
```

skbio/io/registry.py
```diff
@@ -244,6 +245,31 @@ def add_format(self, format_object):
...
+            format = format.name
+        success = False
+        for lookup in self._lookups:
+            try:
+                if format in lookup:
+                    del lookup[format]
+                    success = True
+                    break
+            except KeyError():
```
wasade, line 268: Can a test of membership be explicit?

wasade: Actually, I don't think this would arise because the membership test is done?

wasade (commented):
Just some minor items, this is great!!!

skbio/io/registry.py
mataton: Good point! The try-except block I don't think was necessary so I removed it.

skbio/tree/_tree.py
```diff
@@ -211,6 +215,9 @@ def _copy(self, deep, memo):
         if hasattr((root := self.root()), "_registered_caches"):
             exclude_attrs = exclude_attrs | root._registered_caches
 
+        # exclude dynamically generated methods
+        exclude_attrs = exclude_attrs | {"_write_method"}
```
qiyunzhu, line 219: This edit rings an alarm to me. I am not sure if this will work or not. But can you try to make a `TreeNode` object, copy it, and see if the copy still has the `write` method?

skbio/alignment/_tabular_msa.py
```diff
@@ -143,6 +144,9 @@ class TabularMSA(MetadataMixin, PositionalMetadataMixin, SkbioObject):
     default_write_format = "fasta"
     __hash__ = None
 
+    read = Read()
```
qiyunzhu, line 147: I am thinking if we can further simplify the code by modifying `SkbioObject` such that all classes inheriting from it will have `read` and `write` methods? This may or may not be a good idea though. What do you think?

skbio/sequence/_dna.py
```diff
@@ -148,6 +149,9 @@ class DNA(GrammaredSequence, NucleotideMixin):
 
     """
 
+    read = Read()
```
qiyunzhu, line 152: I tent to guess it is not necessary because `Sequence`, which is a parent class of `DNA`, already has `read` and `write` added. Can you test it? Also see my other comment.

skbio/stats/distance/_base.py
```diff
@@ -1020,6 +1024,9 @@ class DistanceMatrix(DissimilarityMatrix):
     # Override here, used in superclass __str__
     _matrix_element_name = "distance"
 
+    read = Read()
```
qiyunzhu, line 1027: Also here. Since `DissimilarityMatrix` is a parent class of `DistanceMatrix`, I wonder if these two lines only need to exist in the former. See also my other comment.

qiyunzhu (commented):
@mataton Great work! I have some further thoughts after seeing the code, which itself should be correct.

skbio/alignment/_tabular_msa.py
mataton: My initial thought is that having the methods explicitly defined in each class is actually a good thing, because it makes it very clear where these methods are coming from. But I will test out putting the read and write descriptors in parent classes and let you know what I find.

mataton:
@qiyunzhu I've done some initial testing on the speed of `import skbio` with the code from this PR and the code that's currently in the `main` branch. The speeds are roughly the same. Upon further investigation it appears that all of the [import](https://github.com/scikit-bio/scikit-bio/blob/9264f7c3a81f4353bbe39c6d62fabb3e1830461d/skbio/io/__init__.py#L302) statements in the `skbio/io/__init__.py` file are contributing significantly to the execution time of `import skbio`. 

I'm going to try to come up with a solution to dynamically import a given format the first time it is needed by the user, instead of importing them all upfront. Basically, what scikit-bio IO currently does is build this whole `io_registry` object upfront, and I think we could speed things up a lot if we delay building the pieces of `io_registry` until they are actually needed. Once I have a solution I'll push the code to this PR, so let's hold off merging for now.

qiyunzhu:
@mataton Thanks for benchmarking this. I agree with you that lazy loading would be the way to go forward.

mataton:
@qiyunzhu This should be ready to merge now.

PR #2140, opened 2024-10-11
Categorized composition functions

qiyunzhu:
Organized functions under `skbio.stats.composition`. Like this:

view

***

### DIFF
--- a/skbio/stats/composition.py
+++ b/skbio/stats/composition.py
@@ -32,33 +32,81 @@ The multiplicative replacement technique [3]_ can be used to substitute these
 zeros with small pseudocounts without introducing major distortions to the
 data.
 
-Functions
----------
+
+Differential abundance
+----------------------
+
+Statistical tests for the differential abundance (DA) of components among groups of
+compositions.
 
 .. autosummary::
    :toctree:
 
+   ancom
+   dirmult_ttest
+   dirmult_lme
+
+
+Aitchison arithmetics
+---------------------
+
+Manipulate compositional data within the Aitchison space.
+
+.. autosummary::
+   :toctree:
+
+   centralize
    closure
-   multi_replace
-   multiplicative_replacement
+   inner
    perturb
    perturb_inv
    power
-   inner
+
+
+Log-ratio transformation
+------------------------
+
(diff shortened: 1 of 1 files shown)


mortonjt:
I like it, but am confused about the zero handling. Shouldnt we only have 1 function?

Arithmetics ->  Arithmetic operations

there are missing references like aldex2

vlr functions should belong to a miscellaneous section. May want consider putting zero handling methods under there as well.

qiyunzhu:
@mortonjt Thanks for the comments!

`multi_replace` is just an abbreviation for `multiplicative_replacement`, which is too long. I saw other [packages](https://www.rdocumentation.org/packages/zCompositions/versions/1.5.0-4/topics/multRepl) call it `multRepl`. I think that "zero handling" is a topic that average users would be interested in. What do you think?

I don't understand the purpose of `vlr` as its documentation is very brief. It will be helpful if you can briefly describe it.

mortonjt:
Right, but I thought we agreed on just 1 function rather than 2 for zero handling. I’m OK with keeping that subheading as is

VLR is for variance log-ratio. It is used for correlation analysis (ie microbe-microbe correlation). So maybe a good subheading is Correlation

mortonjt (approved):
LGTM thanks @qiyunzhu

qiyunzhu:
Hi @mortonjt I addressed your other comments. Regarding the abbreviation of `multiplicative_replacement`, I did discussed with you before making the change (#1988). There are multiple places in skbio that functions have abbreviations (such as `lowest_common_ancestor` => `lca`). It will be the best if we can let long and short names of the same function stay in the same line of the documentation, like

`multi_replace` | `multiplicative_replacement`: Replace all zeros with small non-zero values.

But so far I haven't figured out a way to let Sphinx do that in a toctree. I can spend some extra time in the near future for this.

Regarding the citation of ALDEx2, I think that it exists inside the [documentation page](https://scikit.bio/docs/latest/generated/skbio.stats.composition.dirmult_ttest.html) of the`dirmult_ttest` function. Same for ANCOM. Do you think we should leverage them to the main page of "composition"?

mortonjt:
Thanks @qiyunzhu I think this is fine as is.

PR #2144, opened 2024-11-01
Improved topological analysis and comparison

qiyunzhu:
Following our internal discussions on tree comparison metrics, I have implemented or updated multiple common metrics, following their accurate definitions in the literature. The results were compared with those of popular programs to ensure consistency. A strong focus was placed on the flexibility of the metrics, such that a larger spectrum of measurements other than published ones may be calculated. Meanwhile, the original default behaviors of the relevant functions are largely preserved. Additionally, documentation has been significantly enriched to inform the users how to use use metrics and what to expect.

A summary of code added or changed in this large PR:

- The **Robinson-Foulds (RF) distance** (`compare_rfd`) was updated to match its original definition, which is based on bipartitions in unrooted trees. The rationales were discussed separately. To accomondate this, new methods `bipart`, `biparts` and `compare_biparts` were added, which mirror `subset`, `subsets` and `compare_subsets`. Specifically, the former category is for unrooted trees while the latter category is for rooted trees (the only original behavior). The method `compare_rfd` automatically determines which category to use, but the user can override this.

- The underlying algorithm for RF distance has been updated and uniformized. Previously, `compare_subsets` actually calculated the rooted RF distance, with the same output, but with different (and less efficient) implementation. Now they all use the same implementation under a new helper function `_compare_topology`. The main improvement is that the previous implementation sheared both trees to match the common taxon set, which is expensive because it involves copying the entire trees. But the new implementation directly identifies subsets or bipartitions of a given taxon set in the original trees. This resulted in about 5x performance gain.

- Meanwhile, the function `_compare_topology` and its frontends allow great flexibility, such that it can base the calculation on shared or all taxa, symmetric or asymmetric, normalized or not, weighted by branch length (i.e., weighted RF, see below) or not, including terminal branches or not, etc. The algorithm was carefully optimized such that this flexibility doesn't introduce much overhead.

- A new method `compare_wrfd` was added to calculate the **weighted Robinson-Foulds (RF) distance**. Basically, it is the difference in bipartitions weighted by their branch lengths. This is comparable to Bray-Curtis vs Jaccard in community ecology. Depending on the choice of a pairwise distance metric, it can also calculate the **Kuhner-Felsenstein (KF) distance**, or a unit distance within [0, 1], or other custom metrics.

- The method `compare_tip_distances` was restructured, following our internal discussion. It retains the default behavior (calculating a unit distance within [0, 1] to indicate the correlation distance between evolutionary distances among taxa). But the new parameters allow it to calculate two published metrics: **path distance** and **path-length distance**, both are based on the Euclidean distance. Although "path distance" or "patristic distance" are widely used in the literature, I think "tip distance" is a reasonable term because it is more intuitive to general users.

- The algorithm underlying `get_max_distance` (i.e., diameter) was re-written. The new implementation is about 5x faster that the previous version, and avoids some undesired issues with the previous version.

- The algorithm for finding a path between two nodes was optimized. This impacts methods `distance` and `path`.

- Other minor tweaks that enhance the flexibility of the `TreeNode` class.

***

### DIFF
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -31,30 +31,6 @@ from skbio.util._warning import _warn_deprecated
 from skbio.io.registry import Read, Write
 
 
-def distance_from_r(m1, m2):
-    r"""Estimate distance as (1-r)/2: neg correl = max distance.
-
-    .. deprecated:: 0.6.3
-        This function will become a private member in version 0.7.0. It has never
-        been exposed in the documentation, and it has a very specific usage in this
-        module.
-
-    Parameters
-    ----------
-    m1 : DistanceMatrix
-        First distance matrix to compare.
-    m2 : DistanceMatrix
-        Second distance matrix to compare.
-
-    Returns
-    -------
-    float
-        The distance between m1 and m2.
-
-    """
-    return spdist.correlation(m1.data.flat, m2.data.flat) / 2
-
-
 # ----------------------------------------------------------------------------
 # Important note: The TreeNode class has a large number of methods. They are
 # organized under several categories, which are defined in this script as well
@@ -3865,7 +3841,7 @@ class TreeNode(SkbioObject):
             )
 
     def distance(self, other, use_length=True, missing_as_zero=False):
-        """Calculate the distance between self and another node.
+        r"""Calculate the distance between self and another node.
 
         Parameters
         ----------
@@ -3879,7 +3855,8 @@ class TreeNode(SkbioObject):
 
         missing_as_zero : bool, optional
             When a node without an associated branch length is encountered, raise an
-            error (False, default) or use 0 (True). Applicable when `length` is True.
+            error (False, default) or use 0 (True). Applicable when ``use_length`` is
+            True.
 
             .. versionadded:: 0.6.3
 
@@ -4401,11 +4378,11 @@ class TreeNode(SkbioObject):
         The Robinson-Foulds (RF) distance, a.k.a. symmetric difference, is a measure of
         topological dissimilarity between two trees. It was originally described in
         [1]_. It is calculated as the number of bipartitions that differ between two
-        unrooted trees. It is equivalent to :meth:`compare_bipartitions`.
+        unrooted trees. It is equivalent to :meth:`compare_biparts`.
 
         .. math::
 
-           RF(T_1, T_2) = |S_1 \triangle S_2| = |(S_1 \setminus S_2) \cup (S_2
+           \text{RF}(T_1, T_2) = |S_1 \triangle S_2| = |(S_1 \setminus S_2) \cup (S_2
            \setminus S_1)|
 
         where :math:`S_1` and :math:`S_2` are the sets of bipartitions of trees
@@ -4514,7 +4491,7 @@ class TreeNode(SkbioObject):
         rooted : bool, optional
             Whether to consider the trees as rooted or unrooted. If None (default),
             this will be determined based on whether self is rooted. However, one
-            can override it by explicitly specifying True (rooted) or False (unrooted).
+            can override it by explicitly setting True (rooted) or False (unrooted).
             See :meth:`compare_rfd` for details.
         include_single : bool, optional
             Whether to include single-taxon biparitions (terminal branches) in the
@@ -4540,7 +4517,7 @@ class TreeNode(SkbioObject):
 
         .. math::
 
-           wRF(T_1, T_2) = \sum_{s \in S_1 \cup S_2} |l_1(s) - l_2(s)|
+           \text{wRF}(T_1, T_2) = \sum_{s \in S_1 \cup S_2} |l_1(s) - l_2(s)|
 
         where :math:`S_1` and :math:`S_2` are the sets of bipartitions of trees
         :math:`T_1` and :math:`T_2`, respectively. :math:`l_1` and :math:`l_2` are the
@@ -4553,7 +4530,7 @@ class TreeNode(SkbioObject):
 
         .. math::
 
-           KF(T_1, T_2) = \sqrt{\sum_{s \in S_1 \cup S_2} (l_1(s) - l_2(s))^2}
+           \text{KF}(T_1, T_2) = \sqrt{\sum_{s \in S_1 \cup S_2} (l_1(s) - l_2(s))^2}
 
         Only taxa shared between the two trees are considered. Taxa unique to either
         tree are excluded from the calculation.
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -3381,6 +3381,8 @@ class TreeNode(SkbioObject):
         if (size := len(bipart)) > (th := len(full) * 0.5):
             bipart = full - bipart
         elif size == th:
+            # sort the elements of each part by lexicographic order, then order the two
+            # parts and pick the smaller part
             bipart, _ = sorted([bipart, full - bipart], key=sorted)
         return bipart
 
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -3221,14 +3221,18 @@ class TreeNode(SkbioObject):
             within = frozenset(within)
 
         # initiate result
+        subsets = []
+        subsets_append = subsets.append
         if map_to_length:
-            subsets = {}
-            subsets_get = subsets.get
-        else:
-            subsets = []
-            subsets_append = subsets.append
+            lengths = []
+            lengths_append = lengths.append
 
-        for node in self.postorder(include_self=include_full):
+        # If the current subset has been encountered during postorder traversal, it
+        # must be the immediately previous subset. This happens when a single-child
+        # node is encountered after refining taxa to the "within" set.
+        last = None
+
+        for node in self.postorder(include_self=True):
             # tip: create a one-taxon set
             if not node.children:
                 if not within or node.name in within:
@@ -3245,21 +3249,28 @@ class TreeNode(SkbioObject):
 
             # add to result
             if subset and (include_single or len(subset) > 1):
-                if map_to_length:
-                    subsets[subset] = subsets_get(subset, 0.0) + (node.length or 0.0)
-                else:
-                    subsets_append(subset)
+                if subset != last:
+                    subsets_append(last := subset)
+                    if map_to_length:
+                        lengths_append(node.length or 0.0)
+                elif map_to_length:
+                    lengths[-1] += node.length or 0.0
 
             node._subset = subset
 
         # final clean up
-        if include_full:
-            del self._subset
-        else:
-            for child in self.children:
-                del child._subset
+        del self._subset
 
-        return subsets if map_to_length else frozenset(subsets)
+        # remove the full set
+        if not include_full:
+            subsets = subsets[:-1]
+            if map_to_length:
+                lengths = lengths[:-1]
+
+        if map_to_length:
+            return dict(zip(subsets, lengths))
+        else:
+            return frozenset(subsets)
 
     def bipart(self):
         r"""Return a bipartition of the tree at the current branch.
@@ -4415,8 +4426,9 @@ class TreeNode(SkbioObject):
         By specifying ``proportion=True``, a unit distance will be returned, ranging
         from 0 (identical) to 1 (completely different).
 
-        Only taxa shared between the two trees are considered. Taxa unique to either
-        tree are excluded from the calculation.
+        This method operates on the subtrees below the given nodes. Only taxa shared
+        between the two trees are considered. Taxa unique to either tree are excluded
+        from the calculation.
 
         See Also
         --------
@@ -4548,8 +4560,9 @@ class TreeNode(SkbioObject):
 
            \text{KF}(T_1, T_2) = \sqrt{\sum_{s \in S_1 \cup S_2} (l_1(s) - l_2(s))^2}
 
-        Only taxa shared between the two trees are considered. Taxa unique to either
-        tree are excluded from the calculation.
+        This method operates on the subtrees below the given nodes. Only taxa shared
+        between the two trees are considered. Taxa unique to either tree are excluded
+        from the calculation.
 
         See Also
         --------
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -4284,6 +4284,9 @@ class TreeNode(SkbioObject):
         proportion : bool, optional
             Whether to return count (False) or proportion (True, default) of different
             subsets.
+
+            .. versionadded:: 0.6.3
+
         symmetric : bool, optional
             Whether to calculate the symmetric difference between self and other (True,
             default), or only the difference from self to other (False).
@@ -4294,16 +4297,10 @@ class TreeNode(SkbioObject):
             Alias of ``shared_only`` for backward compatibility. Deprecated and to be
             removed in a future release.
 
-            .. deprecated:: 0.6.3
-
         Returns
         -------
         int or float
             The count or proportion of subsets that differ between the trees.
-
-        .. versionchanged:: 0.6.3
-            The algorithm is now identical to that of :meth:`compare_rfd` for rooted
-            trees.
 
         See Also
         --------
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -14,7 +14,7 @@ from collections import defaultdict, deque
 
 import numpy as np
 import pandas as pd
-from scipy.spatial.distance import correlation
+import scipy.spatial.distance as spdist
 
 from skbio._base import SkbioObject
 from skbio.stats.distance import DistanceMatrix
@@ -52,7 +52,7 @@ def distance_from_r(m1, m2):
         The distance between m1 and m2.
 
     """
-    return correlation(m1.data.flat, m2.data.flat) / 2
+    return spdist.correlation(m1.data.flat, m2.data.flat) / 2
 
 
 # ----------------------------------------------------------------------------
@@ -3089,7 +3089,7 @@ class TreeNode(SkbioObject):
         --------
         tips
         subsets
-        bipartition
+        bipart
 
         Notes
         -----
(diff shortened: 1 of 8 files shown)


skbio/tree/_tree.py
```diff
@@ -3415,34 +3864,60 @@ def descending_branch_length(self, tip_subset=None):
...
+        missing_as_zero : bool, optional
+            When a node without an associated branch length is encountered, raise an
+            error (False, default) or use 0 (True). Applicable when `length` is True.
+
+            .. versionadded:: 0.6.3
 
         Returns
         -------
-        float
+        float or int
```
wasade, line 3865: Same comment w/ return type. It is not incorrect to force this value to be `float`, but if `int` is important then I recommend considering expressing it as a distinct method so the function signature is fixed

skbio/tree/_tree.py
```diff
@@ -3451,84 +3926,58 @@ def distance(self, other):
...
 
-        This method was originally implemented by Julia Goodrich with the
-        intent of being able to determine max tip to tip distances between
-        nodes on large trees efficiently. The code has been modified to track
-        the specific tips the distance is between
-        """
-        maxkey = itemgetter(0)
+        Returns
+        -------
+        float or int
```
wasade, line 3935: Same comment here

skbio/tree/_tree.py
```diff
@@ -3451,84 +3926,58 @@ def distance(self, other):
...
-            encountered
-
-        See Also
-        --------
-        distance
-        tip_tip_distances
-        compare_tip_distances
+        References
+        ----------
+        .. [1] Mai, U., & Mirarab, S. (2018). TreeShrink: fast and accurate detection
```
wasade, line 3955: I'm a bit puzzled by this reference -- this method predates that paper, and the authors didn't contribute the code?

wasade: Or is it using their implementation?

skbio/tree/_tree.py
```diff
@@ -3650,168 +4149,540 @@ def update_result():
...
                 continue
-            # subtree with solved child wedges
+            # subtree with solved child edges
             # can possibly use np.zeros
-            starts, stops = [], []  # to calc ._start and ._stop for curr node
-            for child in node.children:
-                length = child.length
-                if length is None:
+            starts, stops = [], []
+            for child in (children := node.children):
```
wasade, line 4132: It looks like the use of the walrus is to leak the `children` alias for use outside of the for loop. Doesn't seem needed?

skbio/tree/_tree.py
```diff
@@ -3650,168 +4149,540 @@ def update_result():
...
+        weighted : bool, optional
+            Weight by branch length.
+        metric : str or callable, optional
+            Pairwise distance metric.
 
         Returns
         -------
-        float
-            The distance between the trees
+        int or float
```
wasade, line 4198: :/

skbio/tree/_tree.py
```diff
@@ -3650,168 +4149,540 @@ def update_result():
...
+        compare_subsets
+        compare_biparts
+
+        """
+        topo1, topo2 = getattr(self, method), getattr(other, method)
+        kwargs = dict(include_single=include_single, map_to_length=weighted)
+        if shared_only:
+            set1, set2 = self.subset(), other.subset()
+            n_shared = len(shared := set1 & set2)
+            sets1 = topo1(within=(shared if len(set1) > n_shared else None), **kwargs)
```
wasade, line 4214: Simple if/else?

skbio/tree/_tree.py
```diff
@@ -3650,168 +4149,540 @@ def update_result():
...
+            sets2 = topo2(within=(shared if len(set2) > n_shared else None), **kwargs)
+        else:
+            sets1, sets2 = topo1(**kwargs), topo2(**kwargs)
+
+        # unweighted (set difference)
+        if not weighted:
+            method = ("symmetric_" if symmetric else "") + "difference"
```
wasade, line 4221: I don't think the use of ternary is common in this codebase (or at least previously wasn't), and these can harm readability...

skbio/tree/_tree.py
```diff
@@ -3650,168 +4149,540 @@ def update_result():
...
+
+            # normalize result to unit range [0, 1]
+            # if total is 0, return 1 (dist = 1 means saturation)
+            if proportion:
+                total = len(sets1) + (symmetric and len(sets2))
+                result = result / total if total else 1.0
+
+        # branch length weighted (vector distance)
+        else:
+            union = frozenset(sets1.keys()).union(sets2.keys())
```
wasade, line 4232:
```suggestion
            union = frozenset(sets1).union(sets2)
```

skbio/tree/_tree.py
```diff
@@ -3650,168 +4149,540 @@ def update_result():
...
+
+            .. versionadded:: 0.6.3
+
+        exclude_absent_taxa : bool, optional
+            Alias of ``shared_only`` for backward compatibility. Deprecated and to be
+            removed in a future release.
+
+        Returns
+        -------
+        int or float
```
wasade, line 4279: I'm going to stop flagging these

skbio/tree/_tree.py
```diff
@@ -3650,168 +4149,540 @@ def update_result():
...
-        if proportion:
-            total_subsets = len(tree1_sets) + len(tree2_sets)
-            dist /= total_subsets
+        >>> tree1.compare_rfd(tree2)
+        4
 
-        return dist
+        """
+        if rooted is None:
+            rooted = len(self.children) == 2
```
wasade, line 4463:
```suggestion
            rooted = len(self.root().children) == 2
```

wasade: ...or would it be accurate for a subtree to be considered rooted/unrooted?

wasade (commented):
Hi @qiyunzhu, I added comments where I could but did not review the docstrings much. This is a very large pull request. Two things stands out about the changes, which I'm unsure how they fit into the broader project. First is the use of execution-dependent function signatures. In general, these are harder to reason about than consistent signatures, and they currently seem to make the underlying code more complex (as opposed to decomposing if/when needed). Second, there is a lot of use of ternary operators or unnecessary tuple expansion. These compress a lot of content into a single statement, which can negatively affect readability and debugging with tracebacks. There should not be any performance difference with being more verbose. I recommend considering these two items, in particular as as far as I know, they also deviate from the typical coding style used in the rest of the codebase

skbio/tree/_tree.py
```diff
@@ -1127,9 +1143,9 @@ def tips(self, include_self=False):
         e
 
         """
-        for n in self.postorder(include_self=include_self):
-            if n.is_tip():
-                yield n
+        for node in self.postorder(include_self=include_self):
+            if not node.children:
```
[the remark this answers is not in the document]
qiyunzhu: No it is not. But a method call like `is_tip` has an overhead, which cannot be neglected in a performance demanding function like this. Before I started to work on `TreeNode`, I saw a mixture of `not node.children` and `is_tip` in the codebase. For sake of performance, I am uniformly using the former.

skbio/tree/_tree.py
```diff
@@ -3053,67 +3069,500 @@ def count(self, tips=False):
...
+            within = frozenset(within)
+
+        # initiate result
+        if map_to_length:
+            subsets = {}
+            subsets_get = subsets.get
+        else:
+            subsets = []
+            subsets_append = subsets.append
+
```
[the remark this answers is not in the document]
qiyunzhu: I see the rationale here and your comment above (`if node.name in within:`). However, I think that it will be more performant to skip checking when `within` is None. Even hash table search is _O_(1), it is still not zero, especially when the tree is huge.

The most performant approach would be separating `within is None` and `within is not None` into two independent loops. However, this can make the code lengthy and hard to maintain, especially when this method also has other flags (i.e., multiple combinations of flags).

While I was coding, I benchmarked this scenario and others, and found that have an `if within` check in the loop has minimum impact on the performance.

skbio/tree/_tree.py
```diff
@@ -3053,67 +3069,500 @@ def count(self, tips=False):
...
+            # internal node: merge sets of children
+            else:
+                subset = frozenset()
+                for child in node.children:
+                    subset |= child._subset
+                    del child._subset
+
+            # add to result
+            if subset and include_single or len(subset) > 1:
```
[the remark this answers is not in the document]
qiyunzhu: Good point! Let me revise.

skbio/tree/_tree.py
```diff
@@ -3053,67 +3069,500 @@ def count(self, tips=False):
...
+            # flipping status.
             else:
-                leaf_set = reduce(or_, [c.__leaf_set for c in i.children])
-                if len(leaf_set) > 1:
-                    sets_append(leaf_set)
-                i.__leaf_set = leaf_set
-        return frozenset(sets)
+                aboves, belows = [], []
+                for child in node.children:
+                    (aboves if child._flip else belows).append(child._bipart)
```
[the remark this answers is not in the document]
qiyunzhu: Let me revise this one.

skbio/tree/_tree.py
```diff
@@ -3053,67 +3069,500 @@ def count(self, tips=False):
...
+                    del child._bipart
+                    del child._flip
+
+                if aboves:
+                    bipart, flip = (
```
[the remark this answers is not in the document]
qiyunzhu: Let me change.

skbio/tree/_tree.py
```diff
@@ -3053,67 +3069,500 @@ def count(self, tips=False):
...
+                        frozenset.intersection(*aboves).difference(*belows),
+                        True,
+                    )
+                else:
+                    bipart, flip = frozenset().union(*belows), False
+                    if (size := len(bipart)) >= th:
+                        other = full - bipart
+                        if size > th or sorted(bipart) > sorted(other):
+                            bipart, flip = other, True
```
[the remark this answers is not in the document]
qiyunzhu: Changed.

skbio/tree/_tree.py
qiyunzhu: The paper explicitly introduced the notion of "diameter". But I think there should be lots of mentionings of this term prior to that paper. Given your concern, let me remove this citation.

skbio/tree/_tree.py
qiyunzhu: This is a good point. scikit-bio was vague about whether a method affects the subtree or the entire tree. In some of the method I revised, I added a sentence like "This method modifies the subtree under the current node." to the docstring. But not all of them has that.

I designed the mechanism of deeming a tree is rooted or not based on whether it has two children. But I indeed overlooked the point you raised, which is a good point. To fix this, we need another look into the rooted/unrooted mechanism, and revise the documentation.

skbio/tree/_tree.py
qiyunzhu: I recall that I benchmarked this (either myself or delegated to AI) and found that `keys` is faster than without. But my memory could be incorrect (either I didn't or I found the other way around).

skbio/tree/_tree.py
qiyunzhu: Edited as suggested.

skbio/tree/_tree.py
qiyunzhu: Changed this into simple if/else.

skbio/tree/_tree.py
```diff
@@ -3053,67 +3069,500 @@ def count(self, tips=False):
...
+            .. versionadded:: 0.6.3
 
         Returns
         -------
-        frozenset
-            A frozenset of frozensets of str
+        frozenset of frozenset of str, or
+            All subsets of taxa defined by nodes descending from self. Returned if
+            `map_to_length` is False.
+        dict of {frozenset of str: float}
```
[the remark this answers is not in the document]
qiyunzhu: I see your point and I tend to agree in other cases. However, `TreeNode` already has too many methods (77 in total). To make the documentation easier to follow, I even added [subtitles](https://github.com/scikit-bio/scikit-bio/blob/main/doc/source/_templates/TreeNode.rst) to it.

I thought that it isn't uncommon to have different return types in one function. For example NumPy's [`unique`](https://numpy.org/doc/stable/reference/generated/numpy.unique.html).

skbio/tree/_tree.py
qiyunzhu: Tested. Your suggestion is faster. Will change.

```python
>>> lst = list(range(10000))
>>> dict1 = {x: None for x in random.sample(lst, 5000)}
>>> dict2 = {x: None for x in random.sample(lst, 5000)}
>>> %%timeit
>>> _ = frozenset(dict1).union(dict2)
198 μs ± 6.35 μs per loop
>>> %%timeit
>>> _ = frozenset(dict1.keys()).union(dict2.keys())
259 μs ± 4.02 μs per loop
```

skbio/tree/_tree.py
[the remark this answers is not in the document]
qiyunzhu: Tested and it seems that the class idea is slower.

```python
>>> class _UniversalSet(frozenset):
...     def __contains__(self, other):
...         return True

>>> 3 in _UniversalSet()
True

>>> %%timeit universe = frozenset()
>>> for num in range(100000):
...     _ = num in universe
1.47 ms ± 34 μs per loop

>>> %%timeit universe = frozenset()
>>> for num in range(100000):
...     _ = not universe or num in universe
1.47 ms ± 43.3 μs per loop

>>> %%timeit universe = _UniversalSet()
>>> for num in range(100000):
...     _ = num in universe
3.78 ms ± 40 μs per loop
```

skbio/tree/_tree.py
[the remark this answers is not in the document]
wasade: Python uses a `dict` internally to resolve methods so the number won't appreciably affect runtime. For comparison, the namespace of a Pandas DataFrame is > 400 entries.

The problem with an unstable type signature is that a static analyzer cannot determine what the return type will be as it is based off the values not types provided. I don't know why numpy does that with unique and it's a surprise every time I encounter it.

One example of where this reasoning can go wrong is below. Despite type annotation, and despite using a code path which returns `int`, the `jit`d code returns `float:

```python
In [8]: def foo(a: int) -> tuple[int, float]:
   ...:     if a > 10:
   ...:         return a
   ...:     else:
   ...:         return a / 3
   ...: 

In [9]: foo(15)
Out[9]: 15

In [10]: foo(5)
Out[10]: 1.6666666666666667

In [11]: @numba.jit()
    ...: def foo(a: int) -> tuple[int, float]:
    ...:     if a > 10:
    ...:         return a
    ...:     else:
    ...:         return a / 3
    ...: 

In [12]: foo(15)
Out[12]: 15.0

In [13]: foo(5)
Out[13]: 1.6666666666666667

In [14]:
```

Ideally doing this is an exception not a norm

skbio/tree/_tree.py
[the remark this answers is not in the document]
wasade: Remember, Python is not a fast language. 

This seems to show that the absolute difference for an individual invocation is about 2 nanoseconds. The example does not demonstrate a hotspot in a function of interest, and it's not clear whether the example is correct as CPython could easily implement a short circuit here making this an apples / oranges comparison. In the absence of evidence this very specific operation should be incrementally faster, isn't the readable and more explicit solution better?

skbio/tree/_tree.py
wasade: Ok, but the suggestion was because the calls to `.keys` here are not necessary, not for the purpose of a micro-optimization (which I recommend against focusing on in this PR).

skbio/tree/_tree.py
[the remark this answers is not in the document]
qiyunzhu: Hi @wasade Thanks for the additional comment. I thought a bit and tend to agree with you on this. Namely, methods like `distance` and `compare_rfd` should return a `float` even it counts the number of edges. I have an additional concern which I will communicate with you in the email.

skbio/tree/_tree.py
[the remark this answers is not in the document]
qiyunzhu: For this case I tend to respectfully disagree. In my opinion, creating a class just to solve a very simple problem is unnecessarily convoluted. The current method indeed needs performance, because trees can be large. Additionally tests without a potential CPython short circuit still show that class is significantly slower.

```python
>>> falses = [False] * 100000
>>> %%timeit universe = frozenset()
>>> for num in range(100000):
>>>     _ = not falses[num] or num in universe
1.77 ms ± 11.8 μs per loop
```

wasade:
This doesn’t demonstrate it is a meaningful performance gain for the method. And creating types is normal for OOP?On Nov 14, 2024, at 18:32, Qiyun Zhu ***@***.***> wrote:﻿
@qiyunzhu commented on this pull request.



In skbio/tree/_tree.py:
> +
+        >>> {'a', 'b'} in subsets
+        False
+
+        """
+        if within and not isinstance(within, (set, frozenset, dict)):
+            within = frozenset(within)
+
+        # initiate result
+        if map_to_length:
+            subsets = {}
+            subsets_get = subsets.get
+        else:
+            subsets = []
+            subsets_append = subsets.append
+

For this case I tend to respectfully disagree. In my opinion, creating a class just to solve a very simple problem is unnecessarily convoluted. The current method indeed needs performance, because trees can be large. Additionally tests without a potential CPython short circuit still show that class is significantly slower.
>>> falses = [False] * 100000
>>> %%timeit universe = frozenset()
>>> for num in range(100000):
>>>     _ = not falses[num] or num in universe
1.77 ms ± 11.8 μs per loop

qiyunzhu:
Fixed a bug that affected both the previous code and the code I presented prior to commit [f39d8fe](https://github.com/scikit-bio/scikit-bio/pull/2144/commits/f39d8fe8ce34f409f0f63ef00b8f4f8721485566): A "subset" is supposed to be a proper set of all taxa in a tree. This was reflected in the original `subsets` method and was inherited in the current PR. However, when a custom set of taxa is given, a problem emerged: If one of the basal clades consist of only taxa not within the custom set, the resulting subsets will contain the full custom set. Otherwise it will not. This problem arose because of the way the tree is traversed. It affects the previous `compare_subsets` methods. Now it is fixed.

PR #2145, opened 2024-11-04
Allowed PCoA's number of dimensions to accept float values to capture fractional cumulative variance

nirmithah:
This pull request allows the number of dimensions parameter in the PCoA function to accept float values between 0 and 1, in addition to integers. This will allow the user to input fractional numbers to compute percentage variance. 

### DIFF
--- a/skbio/stats/ordination/_principal_coordinate_analysis.py
+++ b/skbio/stats/ordination/_principal_coordinate_analysis.py
@@ -83,7 +83,7 @@ def pcoa(
 
     Notes
     -----
-    If the distance is not euclidean (for example if it is a
+    If the distance is not Euclidean (for example if it is a
     semimetric and the triangle inequality doesn't hold),
     negative eigenvalues can appear. There are different ways
     to deal with that problem (see Legendre & Legendre 1998, \S
--- a/skbio/stats/ordination/_principal_coordinate_analysis.py
+++ b/skbio/stats/ordination/_principal_coordinate_analysis.py
@@ -162,7 +162,7 @@ def pcoa(
             )
             num_dimensions = matrix_data.shape[0]
         eigvals, eigvecs = _fsvd(matrix_data, num_dimensions, seed=seed)
-        long_method_name = "Approximate Principal Coordinate Analysis " "using FSVD"
+        long_method_name = "Approximate Principal Coordinate Analysis using FSVD"
     else:
         raise ValueError(
             "PCoA eigendecomposition method {} not supported.".format(method)
--- a/skbio/stats/ordination/_principal_coordinate_analysis.py
+++ b/skbio/stats/ordination/_principal_coordinate_analysis.py
@@ -6,12 +6,14 @@
 # The full license is in the file LICENSE.txt, distributed with this software.
 # ----------------------------------------------------------------------------
 
+from numbers import Integral
+from warnings import warn
+
 import numpy as np
 import pandas as pd
 from numpy import dot, hstack
 from numpy.linalg import qr, svd
 from scipy.linalg import eigh
-from warnings import warn
 
 from skbio.util import get_rng
 from skbio.stats.distance import DistanceMatrix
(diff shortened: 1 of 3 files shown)


qiyunzhu:
Hello @wasade This is the updated code by my student @nirmithah . The original and its rationale are in #2141 . Would you like to review it? Thanks!

skbio/stats/ordination/_principal_coordinate_analysis.py
```diff
@@ -89,14 +83,26 @@ def pcoa(
...
     -----
-    .. note:: If the distance is not euclidean (for example if it is a
-        semimetric and the triangle inequality doesn't hold),
-        negative eigenvalues can appear. There are different ways
-        to deal with that problem (see Legendre & Legendre 1998, \S
-        9.2.3), but none are currently implemented here.
-        However, a warning is raised whenever negative eigenvalues
-        appear, allowing the user to decide if they can be safely
-        ignored.
+    If the distance is not euclidean (for example if it is a
```
wasade, line 86:
```suggestion
    If the distance is not Euclidean (for example if it is a
```

skbio/stats/ordination/_principal_coordinate_analysis.py
```diff
@@ -137,12 +148,27 @@ def pcoa(
...
+        num_dimensions = number_of_dimensions
+        if 0 < number_of_dimensions < 1:
+            warn(
+                "FSVD: since value for number_of_dimensions is specified as float, "
+                "PCoA for all dimensions will be computed, which may "
+                "result in long computation time if the original "
+                "distance matrix is large."
+                "Consider specifying an integer value to optimize performance.",
+                RuntimeWarning,
+            )
```
wasade, line 162: Why not progressively compute new dimensions until the cumulative variance is met?

skbio/stats/ordination/_principal_coordinate_analysis.py
```diff
@@ -196,6 +222,14 @@ def pcoa(
         sum_eigenvalues = np.sum(eigvals)
 
     proportion_explained = eigvals / sum_eigenvalues
+    if 0 < number_of_dimensions < 1:
+        cumulative_variance = np.cumsum(proportion_explained)
+        num_dimensions = (
+            np.searchsorted(cumulative_variance, number_of_dimensions, side="left") + 1
```
wasade, line 228:
```suggestion
            np.searchsorted(cumulative_variance, number_of_dimensions, side="right")
```

?

skbio/stats/ordination/_principal_coordinate_analysis.py
```diff
@@ -137,12 +148,27 @@ def pcoa(
...
+                "FSVD: since value for number_of_dimensions is specified as float, "
+                "PCoA for all dimensions will be computed, which may "
+                "result in long computation time if the original "
+                "distance matrix is large."
+                "Consider specifying an integer value to optimize performance.",
+                RuntimeWarning,
+            )
+            num_dimensions = matrix_data.shape[0]
+        eigvals, eigvecs = _fsvd(matrix_data, num_dimensions, seed=seed)
         long_method_name = "Approximate Principal Coordinate Analysis " "using FSVD"
```
wasade, line 165:
```suggestion
        long_method_name = "Approximate Principal Coordinate Analysis using FSVD"
```

wasade (changes requested):
Thanks! This is really cool, but I think the missing magic would be to allow fsvd to progressively compute to meet the desired variance. A little thought needs to go into the strategy of how many dimensions to additionally calculate, but it's fast enough that the strategy doesn't need to be perfect

skbio/stats/ordination/_principal_coordinate_analysis.py
qiyunzhu: We tested this and found that left + 1 produces the correct result at edge cases (e.g., the cumulative variance equal to the parameter).

qiyunzhu:
@wasade Thank you for reviewing the code! Regarding your thought, we found that it is impossible to predict how many dimensions are needed to meet a target cumulative variance before running FSVD. scikit-learn also does full SVD in this scenario ([code](https://github.com/scikit-learn/scikit-learn/blob/6e9039160f0dfc3153643143af4cfdca941d2045/sklearn/decomposition/_pca.py#L541)).

But I think your suggestion is also reasonable. To clarify, you suggested running FSVD on dimensions = 1, 2, 3,... until the taget is met? I guess this will take much more time in the worst case (first few PCs have low variance), but meanwhile could save memory? Or, you are suggesting improving the FSVD algorithm itself to make it progressive?

wasade:
Right, not predict but compute. I'd be more aggressive, as it's fast. Start at 10 dimensions, if the amount of variance is captured stop and select top N as done with `eigh`. If the amount of variance is not met, compute to 20, etc. 

The exact strategy to grow the number of dimensions could be made smarter, but given how fast this stuff is and the typical size of data being used, I recommend doing a simple thing now and only make it better later if it becomes important

PR #2147, opened 2024-11-12
Optimized the neighbor joining (NJ) tree building algorithm

qiyunzhu:
This PR optimizes [**neighbor joining (NJ)**](https://en.wikipedia.org/wiki/Neighbor_joining), a classical algorithm for phylogenetic tree building based on a distance matrix among taxa. The new algorithm is 20-30x faster than the old one, and is competitive when compared with popular implementations in the community.

FYI @christapo @RaeedA 

The old implementation was already quite efficient, due to the use of NumPy vectorization throughout the algorithm. The new algorithm further achieves performance gain because of:

1. The agglomerative clustering process does not create intermediate matrices and arrays during each iteration. Instead, it re-uses the same block of memory space over and over. This not only improves memory efficiency but also reduces the time overheads of allocating new memory space.
2. One time-consuming step: Building a "Q-matrix" and finding the minimum value, is Cythonized. It is possible to use NumPy to find the minimum non-diagonal value in a distance matrix, but the process is quite convoluted. The Cython code is more efficient in this.
3. Instead of building a Newick string incrementally during iteration, the new code builds a linkage matrix, which is purely numeric and can be pre-allocated. The linkage matrix data structure resembles [that of SciPy's](https://docs.scipy.org/doc/scipy/reference/generated/scipy.cluster.hierarchy.linkage.html), but with some differences to reflect the unequal branch lengths of the left and right children.
5. Multiple other tweaks of the math.

Note: The entire function works with numbers only, and can be completely Cythonized. However, Cythonization does not deliver much performance gain in the test (roughly 5-20%), likely because the code already uses NumPy in every line. But the code may be further optimized given its independence from any Pythonic operations.

Another potentional major optimization is to directly manipulate the input distance matrix rather than making a copy of it. This will be useful when working with extremely large distance matrices. However, this will destroy the input distance matrix, therefore requires more thoughts on the `DistanceMatrix` data structure before implementing.

***

Benchmarks on a distance matrix of **485** taxa, derived from real-world 16S rRNA sequences.

- scikit-bio (old): runtime = 0.441 sec, memory = 12,580 KB
- scikit-bio (new): runtime = 0.0161 sec, memory = 2,220 KB

Additional benchmarks were performed on Xeon Platinum 8358 CPUs:

485 taxa:

| Program | Runtime (sec) |
| --- | ---: |
| scikit-bio (new) | 0.03833 |
| scikit-bio (old) | 0.76472 |
| DendroPy 5.0.1 | 3.92156 |
| Biotite 1.0.1 | 0.13615 |
| BioPython 1.84 | 43.41698 |
| QuickTree 2.5 | 0.07855 |
| RapidNJ 2.3.3 | 0.05281 |

9232 taxa:

| Program | Runtime (sec) |
| --- | ---: |
| scikit-bio (new) | 140.638 |
| scikit-bio (old) | 7409.659 |
| Biotite 1.0.1 | 1209.335 |
| QuickTree 2.5 | 256.255 |
| RapidNJ 2.3.3 | 10.226 |

Notes:

- Runtime for Python libraries scikit-bio, DendroPy, Biotite and BioPython only includes the execution of the NJ algorithm, whereas runtime for command-line tools QuickTree and RapidNJ also include file input / output. Therefore, the benchmark is unfair to the latter category, especially the smaller test. scikit-bio isn't fast in reading a big distance matrix from a file. This needs further improvement.
- If I understand correctly, RapidNJ is heuristic, wheras others are optimal (NJ itself is heuristic for minimum evolution, but it is another story).
- I couldn't figure out how to install DecentTree, which is a very recent and performant NJ package.

***

### DIFF
--- a/skbio/tree/_nj.py
+++ b/skbio/tree/_nj.py
@@ -12,31 +12,50 @@ import numpy as np
 
 from skbio.stats.distance import DistanceMatrix
 from skbio.tree import TreeNode
+from skbio.util._warning import _warn_deprecated
+from ._cutils import nj_minq_cy
 
 
-def nj(dm, disallow_negative_branch_length=True, result_constructor=None):
+def nj(
+    dm,
+    clip_to_zero=True,
+    result_constructor=None,
+    disallow_negative_branch_length=None,
+):
     r"""Perform neighbor joining (NJ) for phylogenetic reconstruction.
 
     Parameters
     ----------
     dm : skbio.DistanceMatrix
-        Input distance matrix containing distances between taxa.
-    disallow_negative_branch_length : bool, optional
-        Neighbor joining can result in negative branch lengths, which don't
-        make sense in an evolutionary context. If `True`, negative branch
-        lengths will be returned as zero, a common strategy for handling this
-        issue that was proposed by the original developers of the algorithm.
+        Input distance matrix containing pairwise distances among taxa.
+    clip_to_zero : bool, optional
+        If True (default), convert negative branch lengths into zeros.
+
+        .. versionadded:: 0.6.3
+
     result_constructor : function, optional
         Function to apply to construct the result object. This must take a
-        newick-formatted string as input. The result of applying this function
-        to a newick-formatted string will be returned from this function. This
-        defaults to ``lambda x: TreeNode.read(StringIO(x), format='newick')``.
+        newick-formatted string as input. Deprecated and to be removed in a
+        future release.
+
+        .. deprecated:: 0.6.3
+
+    disallow_negative_branch_length : bool, optional
+        Alias of ``clip_to_zero`` for backward compatibility. Deprecated and to be
+        removed in a future release.
+
+        .. deprecated:: 0.6.3
 
     Returns
     -------
     TreeNode
-        By default, the result object is a `TreeNode`, though this can be
-        overridden by passing `result_constructor`.
+        Reconstructed phylogenetic tree.
+
+        .. versionchanged:: 0.6.3
+            The NJ algorithm has been optimized. The output may be slightly different
+            from the previous one in root placement, node ordering, and the numeric
+            precision of branch lengths. However, the overall tree topology and branch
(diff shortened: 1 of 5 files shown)


skbio/tree/_nj.py
```diff
@@ -86,209 +112,232 @@ def nj(dm, disallow_negative_branch_length=True, result_constructor=None):
...
-    return i_to_u, j_to_u
+        taxon1, taxon2, length1, length2
+
+    """
+    # allocate node list
+    nodes = [TreeNode(name=x) for x in taxa] + [TreeNode() for _ in range(len(lm))]
+
+    # build tree incrementally
+    idx = len(taxa)
+    for c1, c2, l1, l2 in lm if rooted else lm[:-2]:
```
mataton, line 319: `l1`, `l2` are hard to read in my opinion. They look very close to the numbers `11` and `12`. I'd suggest naming the variables something more descriptive or that avoids the use of a lot of lowercase `l`'s. A simple change to capital `L` would also suffice.

skbio/tree/_nj.py
```diff
@@ -86,209 +112,232 @@ def nj(dm, disallow_negative_branch_length=True, result_constructor=None):
...
+
+    # final treatment of an unroot tree
+    if not rooted:
+        # this code assumes that the first element of the last row is the node
+        # see the end of _nj
+        c0, _, l0, _ = lm[-1]
+        c1, c2, l1, l2 = lm[-2]
+        c0, c1, c2 = nodes[int(c0)], nodes[int(c1)], nodes[int(c2)]
+        if clip_to_zero:
+            c0.length = l0 if l0 >= 0 else 0.0
```
mataton, line 337: Same here, regarding readability issues of lowercase `l`.

mataton (changes requested):
Besides the minor readability issues, it all looks good to me!

qiyunzhu:
@mataton Thanks! I have addressed your comments.

PR #2161, opened 2024-11-26
Optimized the calculation of total branch lengths of a clade

qiyunzhu:
This PR improves the performance and customizability of `TreeNode.total_length` (renamed from `descending_branch_length`; old name kept as an alias). Test on a tree with 16k tips:

Entire tree:
- Before: 23.8 ms ± 474 μs
- After: 10.9 ms ± 404 μs

5000 tips
- Before: 56.8 ms ± 2.09 ms
- After: 10.7 ms ± 724 μs

2000 tips
- Before: 38.7 ms ± 809 μs
- After: 4.79 ms ± 118 μs

1000 tips
- Before: 32.5 ms ± 1.09 ms
- After: 2.81 ms ± 47.9 μs

***

### DIFF
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -546,13 +546,16 @@ class TreeNode(SkbioObject):
         else:
             return [n for n in nodes if n is not ignore]
 
-    def lowest_common_ancestor(self, tipnames):
+    def lowest_common_ancestor(self, nodes=None, **kwargs):
         r"""Find the lowest common ancestor of a list of nodes.
 
         Parameters
         ----------
-        tipnames : iterable of TreeNode or str
-            The nodes of interest.
+        nodes : iterable of TreeNode or str
+            Instances or names of the nodes of interest.
+
+            .. versionchanged:: 0.6.3
+                Renamed from ``tipnames``. The old name is preserved as an alias.
 
         Returns
         -------
@@ -561,13 +564,18 @@ class TreeNode(SkbioObject):
 
         Raises
         ------
-        ValueError
-            If no tips could be found in the tree, or if not all tips were found.
+        MissingNodeError
+            If some nodes cannot be found in the tree.
 
         Notes
         -----
-        Despite the parameter is named as ``tipnames``, it can accept both tips and
-        internal nodes.
+        Both tips and internal nodes may be provided in ``nodes``. If internal node
+        names are provided, it is the user's responsibility to ensure that they are
+        unique in the tree.
+
+        This method considers the entire tree rather than the subtree below self.
+        Therefore, if some nodes are not descendants of self, the LCA of nodes will be
+        ancestral to self.
 
         Examples
         --------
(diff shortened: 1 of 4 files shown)


skbio/tree/_tree.py
```diff
@@ -583,36 +591,45 @@ def lowest_common_ancestor(self, tipnames):
         root
 
         """
-        nodes = [self.find(x) for x in tipnames]
+        if kwargs and "tipnames" in kwargs:
+            nodes = kwargs["tipnames"]
```
mataton, line 595: Does this design assume that `nodes` and `tipnames` would never be passed at the same time? If the two parameters are passed at the same time, `tipnames` will override `nodes`, which may not be desired behavior. Probably better to ensure only one or the other is passed. However, I don't think this is a likely scenario, so it may not be necessary to address.

skbio/tree/_tree.py
```diff
@@ -3795,70 +3812,159 @@ def accumulate_to_ancestor(self, ancestor):
...
-            lca = self.lowest_common_ancestor(tip_subset)
-            ancestors = {}
-            for tip in tip_subset:
-                curr = self.find(tip)
-                while curr is not lca:
-                    ancestors[curr.id] = curr.length if curr.length is not None else 0.0
-                    curr = curr.parent
-            return sum(ancestors.values())
+        if kwargs and "tip_subset" in kwargs:
+            nodes = kwargs["tip_subset"]
```
mataton, line 3904: Similar concern here regarding potential passing of `tip_subset` and `nodes` at the same time.

skbio/tree/_tree.py
```diff
@@ -3795,70 +3812,159 @@ def accumulate_to_ancestor(self, ancestor):
...
+        other_paths = []
+        other_paths_append = other_paths.append
+        for curr in nodes:
+            while not hasattr(curr, "_unique"):
+                other_paths_append(curr)
+                curr._unique = True
+                curr = curr.parent
+            curr._unique = False
+
+        # Iterative the first path in reverse order (from root to starting node) and
```
mataton, line 3936: "Iterative" could be "Iterate over".

skbio/tree/_tree.py
```diff
@@ -3795,70 +3812,159 @@ def accumulate_to_ancestor(self, ancestor):
...
+        for node in other_paths:
+            del node._unique
+
+        # If all nodes are descendants of self, LCA must also be self or one of its
+        # descendants, and self must be identified when iterating the first path.
+        if i_self is None:
+            raise MissingNodeError("Some nodes are not descendants of self.")
+
+        # Identify the range of nodes to be included in calculation depending on the
+        # parameter setting
```
mataton, line 3958: Need period at end of sentence here.

mataton (changes requested):
@qiyunzhu Just a couple minor grammatical fixes, and an easy fix in the case of a user passing the old and new parameter names at the same time if you want to address it. I don't think that one is completely necessary though.

skbio/tree/_tree.py
qiyunzhu: This means if `tipnames` is passed, it will override `nodes`. This design isn't perfect. But I am not aware of an elegant way to alias a parameter of a Python function. To avoid the unwanted scenario you pointed out, I think we can do this:

```python
if nodes is None and kwargs and "tipnames" in kwargs:
```

mataton (approved):
@qiyunzhu This looks good to me now. Is it ready to merge?

PR #2166, opened 2024-12-09
Added drivers for multiple tree distance calculations

qiyunzhu:


skbio/tree/_dists.py
```diff
@@ -0,0 +1,395 @@
...
+    sets = [getattr(x, method)(within=y) for x, y in zip(trees, withins)]
+    lens = [len(s) for s in sets]
+
+    result = np.zeros((n_trees := len(trees), n_trees))
+    for i, j in combinations(range(n_trees), 2):
+        rf = len(sets[i] ^ sets[j])
+        if proportion:
+            total = lens[i] + lens[j]
+            rf = rf / total if total else 1.0
+        result[i, j] = result[j, i] = rf
```
wasade, line 133: It looks like this may be a re-implementation of RFD? I understand that this may be faster than to filter the trees and use `TreeNode.compare_rfd`, and suspect that's why it was done. However, it comes at the expense of maintenance burden, the potential for differing results based on the exact implementation used, and may violate the DRY principle. 

If the existing `compare_rfd` method is used, then there is a single point in the codebase to maintain for RFD, and a single block of code to consider optimizing in the future.

skbio/tree/_dists.py
```diff
@@ -0,0 +1,395 @@
...
+    withins = _withins(trees)
+    maps = [
+        getattr(x, method)(within=y, include_single=include_single, map_to_length=True)
+        for x, y in zip(trees, withins)
+    ]
+    gets = [x.get for x in maps]
+    sets = [frozenset(x) for x in maps]
+
+    result = np.zeros((n_trees := len(trees), n_trees))
+    for i, j in combinations(range(n_trees), 2):
```
wasade, line 247: Same comment here, it seems like there is a missed opportunity to funnel this work all through the same implementation.

skbio/tree/_dists.py
```diff
@@ -0,0 +1,395 @@
...
+     [ 13.7113092    0.          19.5192213 ]
+     [ 11.87434209  19.5192213    0.        ]]
+
+    """
+    _check_ids(trees, ids)
+
+    if sample is not None:
+        shuffler = _check_shuffler(shuffler)
+
+    if pairwise:
```
wasade, line 357: There is a fair amount of common boiler plate code here that likely can be abstracted out. This branch is could probably be a function with the signature along the lines of `_pairwise(tree_function: Callable, ids: Iterable[str], **arguments) -> skbio.DistanceMatrix`?

skbio/tree/_dists.py
```diff
@@ -0,0 +1,395 @@
...
+        metric = partial(
+            TreeNode.compare_cophenet,
+            sample=sample,
+            metric=metric,
+            shuffler=shuffler,
+            use_length=use_length,
+            ignore_self=True,
+        )
+        return DistanceMatrix.from_iterable(
+            trees, metric=metric, keys=ids, validate=False
```
wasade, line 367: `metric` here I _think_ would be the `repr` of the `partial` function from line 358 which is likely not what's intended

skbio/tree/_tree.py
```diff
@@ -4306,9 +4310,9 @@ def _compare_topology(
         # unweighted (set difference)
         if not weighted:
             if symmetric:
-                result = sets1.symmetric_difference(sets2)
+                result = sets1 ^ sets2
```
wasade, line 4313: Many developers won't recognize what operation is being performed here and will need to google it. Can the explicit form be retained?

wasade (commented):
Added a few comments, thanks!

skbio/tree/_dists.py
```diff
@@ -0,0 +1,395 @@
...
+    if n > len(set(ids)):
+        raise ValueError(f"IDs contain duplicates.")
+
+
+def _shared(trees):
+    """Get shared taxa among multiple trees."""
+    taxon_sets = [t.subset() for t in trees]
+    shared = frozenset.intersection(*taxon_sets)
+    if not shared:
+        raise ValueError("No taxon is shared across all trees.")
```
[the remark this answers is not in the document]
qiyunzhu: The trees will br shrinked to only shared taxa prior to the calculation of Robinson-Foulds distances. If the trees are empty, the RF distances will become 0, and the normalized RF distances will be 0 / 0 = 1. This is technically doable. I am not sure how justified it is though. Let me think...

skbio/tree/_dists.py
```diff
@@ -0,0 +1,395 @@
...
+    .. versionadded:: 0.6.3
+
+    Parameters
+    ----------
+    trees : list of TreeNode
+        Input trees.
+    ids : list of str, optional
+        Unique identifiers of input trees. If omitted, will use incremental integers
+        "0", "1", "2",...
+    pairwise : bool, optional
```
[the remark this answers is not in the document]
qiyunzhu: Good point. I hesitated between `pairwise` and `shared_by_all`. Given your vote, let me go with the latter.

skbio/tree/_dists.py
qiyunzhu: Good point. I am still in the process of refining the code. Ideally, each algorithm is only implemented once. There are some discrepancies between two tree RF and multi-tree RF, because the former also takes care of some non-RF metrics. But I might be able to overcome this.

skbio/tree/_dists.py
qiyunzhu: Good suggestion!

skbio/tree/_tree.py
qiyunzhu: Makes sense! Let me add a comment after each line of code.

skbio/tree/_dists.py
wasade: I wonder whether a single implementation of RF that accepts `Iterable[set | frozen set]` (i.e., agnostic to `TreeNode`, etc), possibly an implementation specific for wRF, should be provided. And then this code, as well as the respective `TreeNode.compare_*` methods call that single stable implementation?

skbio/tree/_dists.py
wasade: FWIW, I've recently shifted to using PyRight as an LSP, alongside ruff. It would be painful to adapt in skbio because it actually uses typing within the IDE, and I think would require we move to type the project as a whole. But, it is very good at catching these types of issues, among others, so there may be long term benefit.

qiyunzhu: Great thought! I guess that typing can facilitate JIT by future Python interpreters. Since we ditch Python 3.8 now, I think that Python 3.9+ has better built-in support for type hinting. One the other hand, enforcing typing will require lots of efforts though, and some may not be straightforward (therefore cannot be directly outsourced to an undergrad). SciPy and scikit-learn haven't adopted typing, but pandas has.

skbio/tree/_dists.py
qiyunzhu: I have implemented your suggestion. Specifically, in `_compare.py` there is a driver function `_topo_dists` which calls `_setdiff` or `_mapdiff` depending on whether trees are rooted.

skbio/tree/_dists.py
qiyunzhu: Done.

qiyunzhu:
@wasade Thank you for reviewing and suggesting! I think I have implemented all of the suggestions you made. Most importantly, the algorithms are now implemented only once, with all public-facing functions calling them to achieve different outputs.

skbio/tree/_dists.py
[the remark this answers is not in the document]
qiyunzhu: I implemented your suggestion. Specifically, if two trees share no taxon, RF will be 0 and normalized RF will be 1. In this way, any input trees will return values rather than throwing errors.

Issue #1972, opened 2024-03-17
OrdinationResults formatting assumes pd.Series on certain attributes


wasade:
The `OrdinationResults` object indicates in its docstring that `eigvals` and `proportion_explained` are to be `pd.Series` but this is not verified. A consequence is that, while `np.ndarray` can be provided and is a reasonable data type to use for those attributes, the IO registry fails on write as it correctly assumes that these attributes are `pd.Series`. Example traceback is below.

One reasonable path forward would be to relax the doctring, and cast on instantiation of `OrdinationResults`. Or, to gracefully error if the expected information isn't provided. 

It may be the case that additional verification would be appropriate if for example there are index assumptions with the `eigvals` attribute

PR #2187, opened 2025-01-27
Proof-of-concept dispatcher to change underlying data structures from pandas to other options

mataton:
Currently, many objects in scikit-bio rely on `pd.DataFrame` or `pd.Series` under the hood. In a lot of cases, numpy arrays make sense for a given attribute but scikit-bio does not support that. See #1972 for an example. Furthermore, in cases where a `pd.DataFrame` or `pd.Series` object is ideal, users may want to be able to pass in `polars` objects or indeed tabular data structures of any sort, not only `pandas` objects. 

This PR is a proof-of-concept of a centralized dispatcher which the user may set to alter the underlying data structures which are used by scikit-bio. The current goal of this PR focuses on the `OrdinationResults` object, and aims to allow users to decide whether they want the underlying data to be represented as pandas, numpy, or polars objects.

For example, if you wanted to use the `cca` function from scikit-bio (which returns an `OrdinationResults` object), but wanted it to accept and return numpy data structures, the usage looks like this, although this is subject to change:

```python
from skbio.stats.ordination import cca
from skbio._config import set_option

set_option("table_backend", "numpy")

res = cca(y, x)
```

So far, this PR is just illustrating the potential functionality. There are other places in scikit-bio where it would make sense to allow users to decide which backend they want scikit-bio to use:  numpy, pandas, polars, etc.


wasade:
This may be or interest if haven’t seen it https://github.com/narwhals-dev/narwhals

qiyunzhu (approved):
@mataton and I discussed this offline quite a bit. We both think it is ready to be merged. We will further expand and reorganize this functionality prior to the next release.

Issue #2020, opened 2024-05-04
DOC : default_write_format is not documented


mortonjt:
It could be very useful for the html docs to include the `default_write_format` in the doctoring.
Right now, none of the objects specify this, even though all of the objects have a `default_write_format` attribute : https://scikit.bio/docs/latest/generated/skbio.sequence.DNA.html

For instance, all `Sequence` subclasses have a `fasta` format as a `default_write_format`, but it isn't specified. It could be helpful to new users to have the `default_write_format` specified in the docs.

PR #2188, opened 2025-01-31
Add docstrings for `default_write_format` attributes

mataton:
Addresses #2020

### DIFF
--- a/skbio/alignment/_tabular_msa.py
+++ b/skbio/alignment/_tabular_msa.py
@@ -142,7 +142,7 @@ class TabularMSA(MetadataMixin, PositionalMetadataMixin, SkbioObject):
     """
 
     default_write_format = "fasta"
-    """The default write format for this object. Defaults to ``fasta``."""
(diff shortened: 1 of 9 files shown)


qiyunzhu:
@mataton I don't quite understand it -- will the added line to each class be displayed in the documentation page of that class?

I am further thinking if there is an automated way to achieve this. For example, by modifying [this file](https://github.com/scikit-bio/scikit-bio/blob/main/doc/source/_templates/autosummary/class.rst?plain=1). But this is complicated and probably not worth the time.

mataton:
@qiyunzhu This indeed does modify the documentation as desired, but I agree that it may not be the most elegant solution. Modifying the file you pointed out may work, or an alternative solution may be to use `@property` to make `default_write_format` a read-only attribute. This is similar to the approach used for other class attributes in scikit-bio, and users shouldn't be able to modify what is considered the "default" behavior of the package.

Here is what the documentation looks like currently:

Screenshot 2025-02-03 081515

PR #2190, opened 2025-02-18
Integrate Phylomix into Scikit-bio

liaodisen:
This PR is trying to integrate the phylomix into the scikit-bio, this is an initial version. We can further discuss the desired output and supported functionalities

### DIFF
--- a/skbio/table/_augment.py
+++ b/skbio/table/_augment.py
@@ -210,6 +210,9 @@ class Augmentation(SkbioObject):
             The augmented matrix.
         augmented_label : numpy.ndarray
             The augmented label, in one-hot encoding.
+            if the user want to use the augmented label for regression,
+            users can simply call ``np.argmax(aug_label, axis=1)``
+            to get the discrete labels.
 
         Notes
         -----
@@ -288,6 +291,9 @@ class Augmentation(SkbioObject):
             The augmented matrix.
         augmented_label : numpy.ndarray
             The augmented label, in one-hot encoding.
+            if the user want to use the augmented label for regression,
+            users can simply call ``np.argmax(aug_label, axis=1)``
+            to get the discrete labels.
 
         Examples
         --------
@@ -404,7 +410,8 @@ class Augmentation(SkbioObject):
         augmented_matrix : numpy.ndarray
             The augmented matrix.
         augmented_label : numpy.ndarray
-            The augmented label, the label is in 0 and 1.
+            The augmented label, the label is 1D array.
+            User can use the 1D label for both classification and regression.
 
         Examples
         --------
@@ -501,6 +508,9 @@ class Augmentation(SkbioObject):
             The augmented matrix.
         augmented_label : numpy.ndarray
             The augmented label, in one-hot encoding.
+            if the user want to use the augmented label for regression,
+            users can simply call ``np.argmax(aug_label, axis=1)``
+            to get the discrete labels.
 
         Examples
         --------
--- a/skbio/table/_augment.py
+++ b/skbio/table/_augment.py
@@ -15,30 +15,26 @@ import pandas as pd
 
 
 class Augmentation(SkbioObject):
-    """Augmentation of a table with metadata
-    The output of the Augmentation should is a matrix and a label.
+    """Data Augmentation of a table,
+    for predictive microbiome models.
 
     Parameters
     ----------
     table : skbio.table.Table
         The table to augment.
-    method : str
-        The method to use to augment the table.
-    tree [Optional[skbio.tree.TreeNode]]: skbio.tree.TreeNode
+    label : numpy.ndarray, optional
+        The label of the table.
+    tree : skbio.tree.TreeNode, optional
         The tree to use to augment the table.
     """
 
-    def __init__(self, table, method, label=None, tree=None):
+    def __init__(self, table, label=None, tree=None):
         self.table = table
-        self.method = method
         self.tree = tree
         self.dataframe = self.table.to_dataframe()
         self.matrix = self.dataframe.values
         self.label = label
-        if self.label is not None and len(self.label.shape) == 1:
-            self.one_hot_label = pd.get_dummies(self.label).values.astype(int)
-        else:
-            self.one_hot_label = self.label
+        self.one_hot_label = np.eye(2)[self.label]
 
     def __str__(self):
         return f"Augmentation(method={self.method})"
@@ -74,21 +70,46 @@ class Augmentation(SkbioObject):
         -------
         numpy.ndarray
             The result of Aitchison multiplication.
+
         """
         return (x_to_lam := x**lam) / np.sum(x_to_lam)
 
-    def _get_all_possible_pairs(self):
+    def _get_all_possible_pairs(self, intra_class=False):
         r"""Get all possible pairs of samples that can be used for augmentation
+        if intra_class is True, only return pairs of samples within the same class
+        if intra_class is False, return all possible pairs of samples
+
+        Parameters
+        ----------
+        intra_class : bool
+            Whether to get all possible pairs of samples within the same class.
 
         Returns
         -------
         numpy.ndarray
             An array of all possible pairs of samples.
+
         """
         possible_pairs = []
-        for idx1 in range(self.table.shape[1]):
-            for idx2 in range(idx1 + 1, self.table.shape[1]):
-                possible_pairs.append((idx1, idx2))
+        if intra_class:
+            if self.label is None:
+                raise ValueError("label is required for intra-class augmentation")
+            matrix_cls0_indices = np.where(self.label == 0)[0]
+            matrix_cls1_indices = np.where(self.label == 1)[0]
+            for idx1 in range(matrix_cls0_indices.shape[0]):
+                for idx2 in range(idx1 + 1, matrix_cls0_indices.shape[0]):
+                    possible_pairs.append(
+                        (matrix_cls0_indices[idx1], matrix_cls0_indices[idx2])
+                    )
+            for idx1 in range(matrix_cls1_indices.shape[0]):
+                for idx2 in range(idx1 + 1, matrix_cls1_indices.shape[0]):
+                    possible_pairs.append(
+                        (matrix_cls1_indices[idx1], matrix_cls1_indices[idx2])
+                    )
+        else:
+            for idx1 in range(self.table.shape[1]):
+                for idx2 in range(idx1 + 1, self.table.shape[1]):
+                    possible_pairs.append((idx1, idx2))
         return np.array(possible_pairs)
 
     def mixup(self, n_samples, alpha=2, seed=None):
@@ -96,11 +117,19 @@ class Augmentation(SkbioObject):
         Randomly select two samples :math:`s_1` and :math:`s_2` from the OTU table,
         and generate a new sample :math:`s` by a linear combination
         of :math:`s_1` and :math:`s_2`, as follows:
+
         .. math::
-            s = \lambda * s1 + (1 - \lambda) * s2
+
+            s = \lambda \cdot s_1 + (1 - \lambda) \cdot s_2
+
         where :math:`\lambda` is a random number sampled from a beta distribution
         with parameters :math:`\alpha` and :math:`\alpha`.
+        The label is computed as the linear combination of
+        the two labels of the two samples
 
+        .. math::
+
+            y = \lambda \cdot y_1 + (1 - \lambda) \cdot y_2
 
         Parameters
         ----------
@@ -112,7 +141,21 @@ class Augmentation(SkbioObject):
             A user-provided random seed or random generator instance. See
             :func:`details <skbio.util.get_rng>`.
 
-            .. versionadded:: 0.6.3
+        Examples
+        --------
+        >>> from skbio.table import Table
+        >>> from skbio.table.augment import Augmentation
+        >>> data = np.arange(40).reshape(10, 4)
+        >>> sample_ids = ['S%d' % i for i in range(4)]
+        >>> observ_ids = ['O%d' % i for i in range(10)]
+        >>> table = Table(data, observ_ids, sample_ids)
+        >>> label = np.random.randint(0, 2, size=4)
+        >>> augmentation = Augmentation(table, label)
+        >>> aug_matrix, aug_label = augmentation.mixup(n_samples=5)
+        >>> print(aug_matrix.shape)
+        (10, 9)
+        >>> print(aug_label.shape)
+        (9, 2)
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


skbio/table/_augment.py
```diff
@@ -0,0 +1,292 @@
...
+        Returns
+        -------
+        augmented_matrix : numpy.ndarray
+            The augmented matrix.
+        augmented_label : numpy.ndarray
+            The augmented label, in one-hot encoding.
+        """
+        possible_pairs = self._get_all_possible_pairs()
+        selected_pairs = possible_pairs[
+            np.random.randint(0, len(possible_pairs), size=n_samples)
```
mataton, line 120: For random number generation, we [recently switched](https://github.com/scikit-bio/scikit-bio/issues/2096) to numpy's [Generator](https://numpy.org/doc/stable/reference/random/generator.html) object. We have a function, [get_rng()](https://scikit.bio/docs/dev/generated/skbio.util.get_rng.html), which may be used to integrate this new functionality. As an example, to modify this line of code you may do the following:
1) add `from skbio.util import get_rng` to the top of this module
2) add a parameter, `seed=None`, to the function signature
3) add the line `rng = get_rng(seed)` to the beginning of this function
4) line 120 becomes `rng.integers(0, len(possible_pairs), size=n_samples)`

Feel free to look through the `skbio.stats` submodule for other examples on how to implement this functionality.

skbio/table/_augment.py
```diff
@@ -0,0 +1,292 @@
...
+        ]
+
+        indices1 = selected_pairs[:, 0]
+        indices2 = selected_pairs[:, 1]
+        lambdas = np.random.beta(alpha, alpha, size=(1, n_samples))
```
mataton, line 125: See previous comment about replacing `np.random` with `rng`.

skbio/table/_augment.py
```diff
@@ -0,0 +1,292 @@
...
+            augmented_label = None
+
+        return augmented_matrix, augmented_label
+
+    def aitchison_mixup(self, n_samples, alpha=2):
+        r"""Implementation of Aitchison mixup,
+        which is essentially the vanilla mixup in the Aitchison geometry.
+        This mixup method only works on the Compositional data.
+        where a set of datapoints are living in the simplex:
+        x_i > 0, and \sum_{i=1}^{p} x_i = 1
```
mataton, line 148: See previous comments for math in docstrings.

skbio/table/_augment.py
```diff
@@ -0,0 +1,292 @@
...
+        See paper:
+        Data Augmentation for Compositional Data:
+        Advancing Predictive Models of the Microbiome
```
mataton, line 151: This sentence may be added to a `Notes` section of the docstring, and the reference itself may be added to a `References` section of the docstring. Please see [`skbio.stats.ordination.pcoa`](https://scikit.bio/docs/latest/generated/skbio.stats.ordination.pcoa.html) as an example.

skbio/table/_augment.py
```diff
@@ -0,0 +1,292 @@
...
+        Returns
+        -------
+        augmented_matrix : numpy.ndarray
+            The augmented matrix.
+        augmented_label : numpy.ndarray
+            The augmented label, in one-hot encoding.
+        """
+        possible_pairs = self._get_all_possible_pairs()
+        selected_pairs = possible_pairs[
+            np.random.randint(0, len(possible_pairs), size=n_samples)
```
mataton, line 169: See previous comments on `np.random`

skbio/table/_augment.py
```diff
@@ -0,0 +1,292 @@
...
+        ]
+
+        augmented_matrix = []
+        augmented_label = []
+        for idx1, idx2 in selected_pairs:
+            _lambda = np.random.beta(alpha, alpha)
```
mataton, line 175: `np.random` comments here

skbio/table/_augment.py
```diff
@@ -0,0 +1,292 @@
...
+            augmented_label = None
+
+        return augmented_matrix, augmented_label
+
+    def phylomix(self, tip_to_obs_mapping, n_samples, alpha=2):
+        r"""Data Augmentation by phylomix
+
+        Jiang, Y., Liao, D., Zhu, Q., & Lu, Y. Y. (2025).
+        PhyloMix: Enhancing microbiome-trait association prediction through
+        phylogeny-mixing augmentation. Bioinformatics, btaf014.
```
mataton, line 206: See previous comment regarding placement of references into a `References` section.

skbio/table/_augment.py
```diff
@@ -0,0 +1,292 @@
...
+
+        if set(tip_to_obs_mapping.keys()) != set(leave_names):
+            raise ValueError("tip_to_obs_mapping must contain all tips in the tree")
+
+        all_nodes = [node for node in self.tree.levelorder()]
+        num_leaves = len(leave_names)
+
+        possible_pairs = self._get_all_possible_pairs()
+        selected_pairs = possible_pairs[
+            np.random.randint(0, len(possible_pairs), size=n_samples)
```
mataton, line 241: `np.random` usage

skbio/table/_augment.py
```diff
@@ -0,0 +1,292 @@
...
+        ]
+        feature_dict = {
+            feature_name: idx for idx, feature_name in enumerate(self.tree.tips())
+        }
+        augmented_matrix = []
+        augmented_label = []
+        for pair in selected_pairs:
+            x1, x2 = self.matrix[:, pair[0]], self.matrix[:, pair[1]]
+            _lambda = np.random.beta(alpha, alpha)
```
mataton, line 250: `np.random` usage

mataton (changes requested):
@liaodisen This looks promising! I've made some comments, mostly regarding documentation and the use of `np.random`. 

In order to test the changes to your documentation, you will need to modify the `skbio/table/__init__.py` file as follows:

Add this section to the docstring (it may be added after the `BIOM Table` section):
```
Data Augmentation
-----------------

.. autosummary::
   :toctree: generated/

   Augmentation
```

Also, add the following line to the import section in that file:
`from skbio.table._augment import Augmentation`

And finally, add `"Augmentation"` to the `__all__` list in that file.

With these changes, you can run `make doc` to generate an html copy of scikit-bio's documentation with your changes included. See the [contributing](https://scikit.bio/contribute.html#html-rendering) guides for more info.

The logic of your PR looks sound at first glance, but I will do a second review focusing on this aspect in the near future. Thank you for your contribution!

skbio/table/_augment.py
```diff
@@ -0,0 +1,292 @@
...
+        v : numpy.ndarray
+            The second sample.
+
+        Returns
+        -------
+        numpy.ndarray
+            The result of Aitchison addition.
+        """
+        sum_xv = np.sum(x * v)
+        return (x * v) / sum_xv
```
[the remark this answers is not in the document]
qiyunzhu: Well, this can be `return (xv := x * v) / np.sum(xv)` such that `x * v` doesn't need to be calculated twice.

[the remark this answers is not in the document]
qiyunzhu: Also, `xv.sum()` is slightly faster than `np.sum(xv)`.

liaodisen:
Addressed those comments, and added one more method, please review them.

skbio/table/_augment.py
```diff
@@ -0,0 +1,545 @@
...
+        selected_pairs = possible_pairs[
+            rng.integers(0, len(possible_pairs), size=n_samples)
+        ]
+
+        augmented_matrix = []
+        augmented_label = []
+        for idx1, idx2 in selected_pairs:
+            x1, x2 = self.matrix[:, idx1], self.matrix[:, idx2]
+            _lambda = rng.uniform(0, 1)
+            indicator_binomial = rng.binomial(self.matrix.shape[0], _lambda)
```
mataton, line 390: I believe this line should be `indicator_binomial = rng.binomial(1, _lambda, size=self.matrix.shape[0])`. This ensures that `indicator_binomial` is an array of the correct shape containing only 0s or 1s. As it is currently written, `indicator_binomial` is a single integer which may be greater than 1, leading to the output possibly containing negative values, which does not make sense for count data. My understanding of the algorithm may not be entirely correct though.

Here is some example output which was generated by running the example code contained in the docstring for this function:

Screenshot 2025-02-24 133548

skbio/table/_augment.py
```diff
@@ -0,0 +1,545 @@
...
+            )
+            augmented_matrix = np.concatenate([self.matrix, augmented_x], axis=1)
+            augmented_label = np.concatenate([self.one_hot_label, augmented_y])
+        else:
+            augmented_matrix = np.concatenate([self.matrix, augmented_x], axis=1)
+            augmented_label = None
+
+        return augmented_matrix, augmented_label
+
+    def aitchison_mixup(self, n_samples, alpha=2, seed=None):
```
mataton, line 194: I'm not sure why, but in my functional tests, the output of this function generates samples whose values are all very close to zero. I don't think this is the intended behavior. If you run the example code from the docstring you may see this result:

```python
        >>> from skbio.table import Table
        >>> from skbio.table import Augmentation
        >>> data = np.arange(40).reshape(10, 4)
        >>> sample_ids = ['S%d' % i for i in range(4)]
        >>> observ_ids = ['O%d' % i for i in range(10)]
        >>> table = Table(data, observ_ids, sample_ids)
        >>> label = np.random.randint(0, 2, size=4)
        >>> augmentation = Augmentation(table, label)
        >>> aug_matrix, aug_label = augmentation.aitchison_mixup(n_samples=5)
```

I've been using heatmaps to quickly visualize the output to see if they look reasonable or not. Here is the heatmap generated from the above code. 

Screenshot 2025-02-24 133237

skbio/table/_augment.py
```diff
@@ -0,0 +1,545 @@
...
+            augmented_label.append(label)
+
+        augmented_matrix = np.array(augmented_matrix).T
+        augmented_label = np.array(augmented_label)
+        augmented_matrix = np.concatenate([self.matrix, augmented_matrix], axis=1)
+        augmented_label = np.concatenate([self.label, augmented_label])
+        return augmented_matrix, augmented_label
+
+    def phylomix(self, tip_to_obs_mapping, n_samples, alpha=2, seed=None):
+        r"""Data Augmentation by phylomix
```
mataton, line 403: Period to end this sentence as well.

mataton (changes requested):
@liaodisen Thank you for addressing previous comments! I ran some functional tests and found unexpected results for the `compositional_cutmix` and `aitchison_mixup` functions.

skbio/table/_augment.py
liaodisen: This is because Aitchison mixup assumes all samples live in a simplex. That means we have to normalize each sample to sum to 1, which is why many values are close to 1.

skbio/table/_augment.py
liaodisen: That is a bug, added a unit test to ensure each entry of the augmented data is within (min(table), max(table)).

skbio/table/_augment.py
```diff
@@ -0,0 +1,560 @@
...
+        .. [1] Jiang, Y., Liao, D., Zhu, Q., & Lu, Y. Y. (2025).
+            PhyloMix: Enhancing microbiome-trait association prediction through
+            phylogeny-mixing augmentation. Bioinformatics, btaf014.
+
+        """
+        rng = get_rng(seed)
+        if self.tree is None:
+            raise ValueError("tree is required for phylomix augmentation")
+
+        self.tree.bifurcate()
```
qiyunzhu, line 490: This is good as it ensures that the tree is binary. However, the bifurcation pattern may have a strong impact on the result. Therefore, we should state this more explicitly in the documentation. For example, we can point the user to `isbifurcating` and `bifurcate`.

I am also thinking if phylomix can be generalized to non-binary trees? Just like the generalization of binary search tree to B-tree. Of course, this requires benchmarking and probably shouldn't been done right now.

skbio/table/_augment.py
```diff
@@ -0,0 +1,560 @@
...
+            Data augmentation for compositional data: Advancing predictive
+            models of the microbiome. Advances in Neural Information Processing
+            Systems, 35, 20551-20565.
+
+        """
+        if not np.allclose(np.sum(self.matrix, axis=0), 1):
+            warnings.warn("The data is not compositional, it will be normalized first.")
+            if np.any(np.sum(self.matrix, axis=0) == 0):
+                self.matrix = self.matrix + 1e-5
+            self.matrix = self.matrix / np.sum(self.matrix, axis=0)
```
qiyunzhu, line 293: scikit-bio has a function [`closure`](https://scikit.bio/docs/latest/generated/skbio.stats.composition.closure.html) for this. It doesn't require an arbitrary pseudocount like 1e-5. Use it with caution though.

qiyunzhu:
@liaodisen Great job! I also added some comments. Matt and I just discussed and we will communicate by email on some overarching questions.

skbio/table/_augment.py
liaodisen: I think the current implementation also supports when the phylogeny tree is non-binary, bifurcating here is just for matching with paper. I think I can make bifurcate optional for user.

qiyunzhu: @liaodisen Great to know! This will remove a major obstacle to microbiome research. I would suggest that we skip bifurcation in the code, and instead note in the documentation that if the user intends to strictly match the paper, they should run meth:`~skbio.tree.TreeNode.bifurcate` on the tree to make sure it is binary. My rationales are that 1) Many phylogenetic trees are not binary, letting alone non-phylogenetic hierarchical structures like taxonomy and gene ontology. 2) scikit-bio is a library not a workflow, so user may expect more flexibility. But I am open to counter-opinions. What do you think?

qiyunzhu:
@liaodisen Another thought: In your code you refer to features as "observations". This term is consistent with BIOM. However, in [anndata](https://anndata.readthedocs.io/en/stable/) which is popular for single-cell data analysis, "observations" are samples (individual cells) whereas "variables" are features. So I am thinking if replacing "observations" with "features" in the code can make it more explicit.

skbio/table/_augment.py
liaodisen: I agree with you, I will remove the bifurcate, and I will add instruction for user to bifurcate in the note

qiyunzhu (commented):
Me and Matt have discussed and we think it is in good shape. In addition to Matt's comments, a minor comment of mine is that we don't need to overemphasize microbiome in the documentation. I think these methods are as well effective on other types of hierarchical data. scikit-bio can support a variety of omic data analyses, not just microbiome.

qiyunzhu:
@liaodisen Thanks for the additional work. Matt and I discussed and we think this PR is ready to be merged. I will click the green button. Thank you for the contribution!

We have identified several minor issues that still need to be addressed. These are primarily engineering issues instead of algorithmic. Matt and I will write code to address those issues, and ask you for code review. We will follow up with an email shortly.

PR #2199, opened 2025-03-25
Added 2D functionality + unit tests

nirmithah:

### DIFF
--- a/skbio/stats/ordination/_ordination_results.py
+++ b/skbio/stats/ordination/_ordination_results.py
@@ -172,7 +172,7 @@ class OrdinationResults(SkbioObject, PlottableMixin):
         self,
         df=None,
         column=None,
-        axes=(0, 1),
+        axes=None,
         axis_labels=None,
         title="",
         cmap=None,
@@ -299,6 +299,9 @@ class OrdinationResults(SkbioObject, PlottableMixin):
         # instead be added to EMPeror (http://biocore.github.io/emperor/).
         # Only bug fixes and minor updates should be made to this method.
 
+        if axes is None:
+            axes = np.arange(len(self.eigvals))
+
         self._get_mpl_plt()
 
         # print(df)
@@ -314,7 +317,6 @@ class OrdinationResults(SkbioObject, PlottableMixin):
             df, column, self.sample_ids, cmap
         )
         self._validate_plot_axes(coord_matrix, axes)
-
         if len(axes) == 3:  # 3d functionality
             fig = self.plt.figure()
             ax = fig.add_subplot(projection="3d")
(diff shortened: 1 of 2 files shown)


skbio/stats/ordination/_ordination_results.py
```diff
@@ -298,6 +299,9 @@ def plot(
         # instead be added to EMPeror (http://biocore.github.io/emperor/).
         # Only bug fixes and minor updates should be made to this method.
 
+        if axes is None:
+            axes = np.arange(len(self.eigvals))
```
mataton, line 303: @nirmithah I'm doing some functional tests and found these lines to be problematic. It may be the case that there are numerous (>3) values in `self.eigvals`, in which case we can't reliably use `self.eigvals` to determine what `axes` should be.

I think a simpler and more robust approach would be to have a boolean parameter called `n_dims` or something similar, which would allow the user to explicitly determine whether they want the plot to be 3D or 2D. The default behavior should be to keep things 3D. 

In the 3D case, `axes` should be `(0, 1, 2)`, and in the 2D case, `axes` should be `(0, 1)`.

skbio/stats/ordination/_ordination_results.py
```diff
@@ -313,34 +317,51 @@ def plot(
...
 
         if axis_labels is None:
             axis_labels = ["%d" % axis for axis in axes]
-        elif len(axis_labels) != 3:
+        elif len(axis_labels) != len(axes):
             raise ValueError(
-                "axis_labels must contain exactly three elements "
+                "axis_labels must contain exactly two or three elements "
                 "(found %d elements)." % len(axis_labels)
             )
```
mataton, line 354: We could change this error message to say something like `f"axis_labels ({len(axis_labels} elements) must contain the same number of elements as axes ({len(axes) elements).

skbio/stats/ordination/tests/test_ordination_results.py
```diff
@@ -330,5 +330,204 @@ def test_plot_categorical_legend(self):
...
+        labels = [t.get_text() for t in legend.get_texts()]
+        npt.assert_equal(sorted(labels), ['bar', 'foo'])
+
+        colors = [line.get_color() for line in legend.get_lines()]
+        npt.assert_equal(sorted(colors), ['green', 'red'])
+        
+
 if __name__ == '__main__':
-    unittest.main()
+    unittest.main(buffer=False)
```
mataton, line 533: When you're done altering the unit tests remember to remove the `buffer=False` from within the parentheses.

mataton (changes requested):
@nirmithah This is almost ready, a few more minor changes and then we should be able to merge!

mataton (approved):
@nirmithah This looks good, I'll merge it now!

PR #2220, opened 2025-05-11
add _netembed utils

jianshu93:
Dear skbio team (to Qiyun and Mat specifically),

I followed the _plotting.py in skbio/util folder to add this _netembed.py utility class, so that graphembed_rs can be imported and used. Not sure whether I need to also add some tests, but this add is to import the grapuembed library only when asked. By default it does not change anything.  I am not a python person per se,  so let me know your thoughts. 

Thanks,
Jianshu

### DIFF
--- a/skbio/util/_netembed.py
+++ b/skbio/util/_netembed.py
@@ -1,0 +1,137 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2013--, scikit-bio development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE.txt, distributed with this software.
+# ----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


jianshu93:
Hi Qiyun and Mat, how about this one. I do not understand why it failed only in Ubuntu but not macOS. -Jianshu

jianshu93:
Hi @mataton, Thanks I will add some basic tests. However the data has to be in a CSV format from a file. In that case, the test will need to read an additional file, where do you think I can add the files (a very small CSV file).

qiyunzhu:
Hi both, do we have necessary code in scikit-bio to interface with graph embedding code? Like input / output data structures?

jianshu93:
For input we do not because I streamed the edge list. For output yes， output is just numpy vectors.

PR #2239, opened 2025-06-02
Initiation for 4 log ratio-transform functions to support GPU and Array standard

songyesog2000:
Hi Scikit-team,

This is Sam, I have been working on this issue as I want to help Jamie on the development also a part of practice to improve my coding.

This is just an initialization for the pull&request, as I have finish the adaption of codes for GPU and array standard api,  I will finish the test.py next two days.

There are a few points I need to consult your team about the functional behaviors:

1. typing: I used Array (a dummy class of class, implemented in array standard api), however, I notice that composition data can be a certain class/data type, if you want to add the class, please give me a hint where to add and what property/method this class should embrace.
2. closure: I noticed closure function is automatically operated within the 4 functions, however, log-ratio transformed is not necessary coupled with the closure, will it be a good idea to avoid having the closure inside these functions?
3. shape of returns: the input maybe in various shapes (vectors, matrices, singleton), should the return shape be consistent or corresponding the shape of input?

Best regards,
Sam Mingsong 

Issue reference: 
https://github.com/scikit-bio/scikit-bio/issues/2210

=============================================

### DIFF
--- a/skbio/stats/composition.py
+++ b/skbio/stats/composition.py
@@ -157,6 +157,42 @@ from statsmodels.regression.mixed_linear_model import MixedLM
 from statsmodels.tools.sm_exceptions import ConvergenceWarning
 from patsy import dmatrix
 
+import array_api_compat as aac
+
+Array = object
+    
+# # codes from array_api_compat.common._typing
+# #===============================================================
+# from typing import Protocol, TypeAlias, Any
+# from typing_extensions import TypeIs, TypeVar
+
+# _T_co = TypeVar("_T_co", covariant=True)
+# class SupportsArrayNamespace(Protocol[_T_co]):
+#     def __array_namespace__(self, /, *, api_version: str | None) -> _T
+
+# import dask.array as da
+# import jax
+# import ndonnx as ndx
+# import numpy as np
+# import numpy.typing as npt
+# import sparse  # pyright: ignore[reportMissingTypeStubs]
+# import torch
+
+#     # TODO: import from typing (requires Python >=3.13)
+
+# _CupyArray: TypeAlias = Any  # cupy has no py.typed
+
+# _ArrayApiObj: TypeAlias = (
+#     npt.NDArray[Any]
+#     | da.Array
+#     | jax.Array
+#     | ndx.Array
+#     | sparse.SparseArray
+#     | torch.Tensor
+#     | SupportsArrayNamespace[Any]
+#     | _CupyArray
+# )
+
 
 def closure(mat):
     """Perform closure to ensure that all elements add up to 1.
@@ -449,7 +485,7 @@ def inner(x, y):
     return a.dot(b.T)
 
 
-def clr(mat):
+def clr(mat:Array)-> Array:
     r"""Perform centre log ratio transformation.
 
     This function transforms compositions from Aitchison geometry to the real
@@ -489,13 +525,23 @@ def clr(mat):
     array([-0.79451346,  0.30409883,  0.5917809 , -0.10136628])
 
     """
-    mat = closure(mat)
-    lmat = np.log(mat)
-    gm = lmat.mean(axis=-1, keepdims=True)
-    return (lmat - gm).squeeze()
+    # xp is the namespace wrapper for different array libraries
+    xp = aac.array_namespace(mat)
+    
+    # ensure last dimension is not singleton
+    if xp.shape(mat)[-1] == 1:
+        for i in range(xp.shape(mat),0,-1):
+            if xp.shape(mat)[i] == 1:
+                mat = xp.squeeze(mat, axis=i)
+            else:
+                break
+    
+    lmat = xp.log(mat)
+    gm = xp.mean(lmat, axis=-1, keepdims=True)
+    return xp.subtract(lmat, gm)
 
 
-def clr_inv(mat):
+def clr_inv(mat:Array) -> Array:
     r"""Perform inverse centre log ratio transformation.
 
     This function transforms compositions from the real space to Aitchison
@@ -533,12 +579,20 @@ def clr_inv(mat):
 
     """
     # for numerical stability (aka softmax trick)
-    mat = np.atleast_2d(mat)
-    emat = np.exp(mat - mat.max(axis=-1, keepdims=True))
-    return closure(emat)
+    xp = aac.array_namespace(mat)
+    
+    # ensure last dimension is not singleton
+    if xp.shape(mat)[-1] == 1:
+        for i in range(xp.shape(mat),0,-1):
+            if xp.shape(mat)[i] == 1:
+                mat = xp.squeeze(mat, axis=i)
+            else:
+                break
+    
+    gm = xp.mean(mat, axis=-1, keepdims=True)
+    return xp.exp(mat + gm)
 
-
-def ilr(mat, basis=None, check=True):
+def ilr(mat:Array, basis:Array|None=None, /, check:bool=True)-> Array:
     r"""Perform isometric log ratio transformation.
 
     This function transforms compositions from Aitchison simplex to the real
@@ -590,23 +644,25 @@ def ilr(mat, basis=None, check=True):
     where rows represent basis vectors, and the columns represent proportions.
 
     """
-    mat = closure(mat)
+    xp = aac.array_namespace(mat)
     if basis is None:
         d = mat.shape[-1]
         basis = _gram_schmidt_basis(d)  # dimension (d-1) x d
-    else:
-        if len(basis.shape) != 2:
-            raise ValueError(
-                "Basis needs to be a 2D matrix, "
-                "not a %dD matrix." % (len(basis.shape))
-            )
-        if check:
-            _check_orthogonality(basis)
+    
+    assert isinstance(basis, xp.ndarray)
+    
+    if len(basis.shape) != 2:
+        raise ValueError(
+            "Basis needs to be a 2D matrix, "
+            "not a %dD matrix." % (len(basis.shape))
+        )
+    if check:
+        _check_orthogonality(basis)
 
-    return clr(mat) @ basis.T
+    return mat @ basis.T
 
 
-def ilr_inv(mat, basis=None, check=True):
+def ilr_inv(mat:Array, basis:Array|None=None, check:bool=True)-> Array:
     r"""Perform inverse isometric log ratio transform.
 
     This function transforms compositions from the real space to Aitchison
@@ -658,26 +714,28 @@ def ilr_inv(mat, basis=None, check=True):
     where rows represent basis vectors, and the columns represent proportions.
 
     """
-    mat = np.atleast_2d(mat)
+    xp = aac.array_namespace(mat)
+    original_shape = xp.shape(mat)
+    
     if basis is None:
         # dimension d-1 x d basis
         basis = _gram_schmidt_basis(mat.shape[-1] + 1)
-    else:
-        if len(basis.shape) != 2:
-            raise ValueError(
-                "Basis needs to be a 2D matrix, "
-                "not a %dD matrix." % (len(basis.shape))
-            )
-        if check:
-            _check_orthogonality(basis)
+    assert isinstance(basis, xp.ndarray)
+    
+    if len(basis.shape) != 2:
+        raise ValueError(
+            "Basis needs to be a 2D matrix, "
+            "not a %dD matrix." % (len(basis.shape))
+        )
+    if check:
+        _check_orthogonality(basis)
         # this is necessary, since the clr function
         # performs np.squeeze()
-        basis = np.atleast_2d(basis)
 
-    return clr_inv(mat @ basis)
+    return mat @ basis
 
 
-def alr(mat, denominator_idx=0):
+def alr(mat:Array, denominator_idx:int|Array=0, /):
     r"""Perform additive log ratio transformation.
 
     This function transforms compositions from a D-part Aitchison simplex to
@@ -721,22 +779,24 @@ def alr(mat, denominator_idx=0):
     array([ 1.09861229,  1.38629436,  0.69314718])
 
     """
-    mat = closure(mat)
-    if mat.ndim == 2:
+    xp = aac.array_namespace(mat)
+    if xp.ndim(mat) == 2:
+        M,N = xp.shape(mat)
         mat_t = mat.T
-        numerator_idx = list(range(0, mat_t.shape[0]))
-        del numerator_idx[denominator_idx]
-        lr = np.log(mat_t[numerator_idx, :] / mat_t[denominator_idx, :]).T
-    elif mat.ndim == 1:
-        numerator_idx = list(range(0, mat.shape[0]))
-        del numerator_idx[denominator_idx]
-        lr = np.log(mat[numerator_idx] / mat[denominator_idx])
+        numerator_matrix = mat_t[xp.repeat(xp.arange(N),M,axis=0) != denominator_idx] 
+        denominator_vector = mat_t[:, denominator_idx]
+        lr = xp.log(numerator_matrix / denominator_vector).T
+    elif mat.ndim == 1 and isinstance(denominator_idx, int):
+        M = len(mat)
+        numerator = mat_t[xp.arange(N) != denominator_idx] 
+        denominator = mat[denominator_idx]
+        lr = xp.log(numerator /denominator)
     else:
-        raise ValueError("mat must be either 1D or 2D")
+        raise ValueError("mat must be either 1D or 2D and denominator_idx must be an integer for 1D array")
     return lr
 
 
-def alr_inv(mat, denominator_idx=0):
+def alr_inv(mat:Array, denominator_idx:int|Array=0, /):
     r"""Perform inverse additive log ratio transform.
 
     This function transforms compositions from the non-isometric real space of
@@ -782,25 +842,30 @@ def alr_inv(mat, denominator_idx=0):
     array([ 0.1,  0.3,  0.4,  0.2])
 
     """
-    mat = np.array(mat)
+    xp = aac.array_namespace(mat)
     if mat.ndim == 2:
-        mat_idx = np.insert(mat, denominator_idx, np.repeat(0, mat.shape[0]), axis=1)
-        comp = np.zeros(mat_idx.shape)
-        comp[:, denominator_idx] = 1 / (np.exp(mat).sum(axis=1) + 1)
-        numerator_idx = list(range(0, comp.shape[1]))
-        del numerator_idx[denominator_idx]
-        for i in numerator_idx:
-            comp[:, i] = comp[:, denominator_idx] * np.exp(mat_idx[:, i])
-    elif mat.ndim == 1:
-        mat_idx = np.insert(mat, denominator_idx, 0, axis=0)
-        comp = np.zeros(mat_idx.shape)
-        comp[denominator_idx] = 1 / (np.exp(mat).sum(axis=0) + 1)
-        numerator_idx = list(range(0, comp.shape[0]))
-        del numerator_idx[denominator_idx]
-        for i in numerator_idx:
-            comp[i] = comp[denominator_idx] * np.exp(mat_idx[i])
+        M, N_minus_1 = mat.shape
+        N = N_minus_1 + 1
+        comp = xp.zeros((M, N), dtype=mat.dtype)
+        mat = xp.exp(mat)
+        # x_N / (x_1 + ... + x_{N-1} + x_N)
+        comp[:, denominator_idx] = 1 / (xp.sum(mat, axis=1) + 1)
+        # x_i/(x_1 + ... + x_{N-1} + x_N) = \frac{x_i}{x_N} * \frac{x_N}{x_1 + ... + x_{N-1} + x_N}
+        comp[xp.repeat(xp.arange(N),M, axis=0) != denominator_idx] = \
+                    mat[:, N_minus_1] * mat[:, :denominator_idx]
+    elif mat.ndim == 1 and isinstance(denominator_idx, int):
+        N_minus_1 = len(mat)
+        N = N_minus_1 + 1
+        M = 1
+        comp = xp.zeros((M, N), dtype=mat.dtype)
+        mat = xp.exp(mat)
+        # x_N / (x_1 + ... + x_{N-1} + x_N)
+        comp[:, denominator_idx] = 1 / (xp.sum(mat, axis=1) + 1)
+        # x_i/(x_1 + ... + x_{N-1} + x_N) = \frac{x_i}{x_N} * \frac{x_N}{x_1 + ... + x_{N-1} + x_N}
+        comp[xp.repeat(xp.arange(N),M,axis=0) != denominator_idx] = \
+                    mat[:, N_minus_1] * mat[:, :denominator_idx]
     else:
-        raise ValueError("mat must be either 1D or 2D")
+        raise ValueError("mat must be either 1D or 2D and denominator_idx must be an integer for 1D array")
     return comp
 
 


skbio/stats/composition.py
```diff
@@ -489,13 +525,23 @@ def clr(mat):
...
     """
-    mat = closure(mat)
-    lmat = np.log(mat)
-    gm = lmat.mean(axis=-1, keepdims=True)
-    return (lmat - gm).squeeze()
+    # xp is the namespace wrapper for different array libraries
+    xp = aac.array_namespace(mat)
+    
+    # ensure last dimension is not singleton
+    if xp.shape(mat)[-1] == 1:
```
qiyunzhu: I wonder if this can be `mat.shape[-1]`? In my experience with NumPy, `ndarray.some_method` is usually more efficient than `np.some_function`.

skbio/stats/composition.py
```diff
@@ -533,12 +579,20 @@ def clr_inv(mat):
...
+            if xp.shape(mat)[i] == 1:
+                mat = xp.squeeze(mat, axis=i)
+            else:
+                break
+    
+    gm = xp.mean(mat, axis=-1, keepdims=True)
+    return xp.exp(mat + gm)
 
-def ilr(mat, basis=None, check=True):
+def ilr(mat:Array, basis:Array|None=None, /, check:bool=True)-> Array:
```
qiyunzhu: scikit-bio still supports Python 3.9, which doesn't allow `|` in type hinting. So `Array|None` needs to be `Optional[Array]` instead. Likewise, `int|Array` below needs to be `Union[int, Array]`.

skbio/stats/composition.py
```diff
@@ -721,22 +779,24 @@ def alr(mat, denominator_idx=0):
     array([ 1.09861229,  1.38629436,  0.69314718])
 
     """
-    mat = closure(mat)
-    if mat.ndim == 2:
+    xp = aac.array_namespace(mat)
+    if xp.ndim(mat) == 2:
```
qiyunzhu: I wonder if `mat.ndim` still works. I could be wrong though.

mortonjt:
RE closure, we don't need to call the original function, but we should have asserts place, namely making sure that there are no negative entries
https://github.com/scikit-bio/scikit-bio/blob/main/skbio/stats/composition.py#L195-L196

And making sure that the rows don't sum to zero
https://github.com/scikit-bio/scikit-bio/blob/main/skbio/stats/composition.py#L199-L200

skbio/stats/composition.py
```diff
@@ -489,13 +525,23 @@ def clr(mat):
...
+    # xp is the namespace wrapper for different array libraries
+    xp = aac.array_namespace(mat)
+    
+    # ensure last dimension is not singleton
+    if xp.shape(mat)[-1] == 1:
+        for i in range(xp.shape(mat),0,-1):
+            if xp.shape(mat)[i] == 1:
+                mat = xp.squeeze(mat, axis=i)
+            else:
+                break
```
mortonjt: Where is the break coming from? Can we have a unittest illustrating this use case?

skbio/stats/composition.py
```diff
@@ -489,13 +525,23 @@ def clr(mat):
     array([-0.79451346,  0.30409883,  0.5917809 , -0.10136628])
 
     """
-    mat = closure(mat)
-    lmat = np.log(mat)
-    gm = lmat.mean(axis=-1, keepdims=True)
-    return (lmat - gm).squeeze()
+    # xp is the namespace wrapper for different array libraries
```
mortonjt, line 531: Recommended to put asserts here so that we don't need to call closure.
```
    if np.any(mat < 0):
        raise ValueError("Cannot have negative proportions")
    if np.all(mat == 0, axis=1).sum() > 0:
        raise ValueError("Input matrix cannot have rows with all zeros")
```
Since we are dealing with tensors, the shape considerations need special care.

skbio/stats/composition.py
```diff
@@ -533,12 +579,20 @@ def clr_inv(mat):
 
     """
     # for numerical stability (aka softmax trick)
-    mat = np.atleast_2d(mat)
-    emat = np.exp(mat - mat.max(axis=-1, keepdims=True))
-    return closure(emat)
```
mortonjt, line 538: Same here, add the asserts

skbio/stats/composition.py
```diff
@@ -533,12 +579,20 @@ def clr_inv(mat):
...
-
+    xp = aac.array_namespace(mat)
+    
+    # ensure last dimension is not singleton
+    if xp.shape(mat)[-1] == 1:
+        for i in range(xp.shape(mat),0,-1):
+            if xp.shape(mat)[i] == 1:
+                mat = xp.squeeze(mat, axis=i)
+            else:
+                break
```
mortonjt: This is a repeating pattern, move this over to a separate function called `_squeeze_tensor`

skbio/stats/composition.py
```diff
@@ -533,12 +579,20 @@ def clr_inv(mat):
...
+    
+    gm = xp.mean(mat, axis=-1, keepdims=True)
+    return xp.exp(mat + gm)
```
mortonjt: This should identical to softmax, probably just want to compare against scipy softmax https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.softmax.html

This needs to sum to one. We don't need gm, but according to best practice, you should subtract the max for numerical error. Something like as follows works

```python
gm = xp.max(mat, axis=-1, keepdims=True)
return xp.exp(mat - gm) / xp.exp(mat - gm).sum(axis=-1)
```

skbio/stats/composition.py
```diff
@@ -590,23 +644,25 @@ def ilr(mat, basis=None, check=True):
     where rows represent basis vectors, and the columns represent proportions.
 
     """
-    mat = closure(mat)
```
mortonjt, line 593: Make sure to add assertions here

skbio/stats/composition.py
```diff
@@ -590,23 +644,25 @@ def ilr(mat, basis=None, check=True):
...
+    if len(basis.shape) != 2:
+        raise ValueError(
+            "Basis needs to be a 2D matrix, "
+            "not a %dD matrix." % (len(basis.shape))
+        )
+    if check:
+        _check_orthogonality(basis)
 
-    return clr(mat) @ basis.T
+    return mat @ basis.T
```
mortonjt, line 708: bring back CLR, this is incorrect.

skbio/stats/composition.py
```diff
@@ -658,26 +714,28 @@ def ilr_inv(mat, basis=None, check=True):
...
+            "not a %dD matrix." % (len(basis.shape))
+        )
+    if check:
+        _check_orthogonality(basis)
         # this is necessary, since the clr function
         # performs np.squeeze()
-        basis = np.atleast_2d(basis)
 
-    return clr_inv(mat @ basis)
+    return mat @ basis
```
mortonjt: Same thing here, bring back clr_inv

skbio/stats/composition.py
```diff
@@ -590,23 +644,25 @@ def ilr(mat, basis=None, check=True):
...
+        )
+    if check:
+        _check_orthogonality(basis)
 
-    return clr(mat) @ basis.T
+    return mat @ basis.T
 
 
-def ilr_inv(mat, basis=None, check=True):
+def ilr_inv(mat:Array, basis:Array|None=None, check:bool=True)-> Array:
```
mortonjt: let's rename this to validate_args to be consistent with other packages (i.e. PyTorch). Every function will need to have this

skbio/stats/composition.py
```diff
@@ -658,26 +714,28 @@ def ilr_inv(mat, basis=None, check=True):
...
         # this is necessary, since the clr function
         # performs np.squeeze()
-        basis = np.atleast_2d(basis)
 
-    return clr_inv(mat @ basis)
+    return mat @ basis
 
 
-def alr(mat, denominator_idx=0):
+def alr(mat:Array, denominator_idx:int|Array=0, /):
```
mortonjt: What is `/`? Can we remove it?

skbio/stats/composition.py
mortonjt: Let's not forget to remove the `/` here

skbio/stats/composition.py
```diff
@@ -721,22 +779,24 @@ def alr(mat, denominator_idx=0):
...
+        M,N = xp.shape(mat)
         mat_t = mat.T
-        numerator_idx = list(range(0, mat_t.shape[0]))
-        del numerator_idx[denominator_idx]
-        lr = np.log(mat_t[numerator_idx, :] / mat_t[denominator_idx, :]).T
-    elif mat.ndim == 1:
-        numerator_idx = list(range(0, mat.shape[0]))
-        del numerator_idx[denominator_idx]
-        lr = np.log(mat[numerator_idx] / mat[denominator_idx])
+        numerator_matrix = mat_t[xp.repeat(xp.arange(N),M,axis=0) != denominator_idx]
```
mortonjt: This should be more similar to `clr`, we should just replace with `gm=mat[denominator_idx]`
We will want to force `denominator_idx` to be integer

skbio/stats/composition.py
```diff
@@ -721,22 +779,24 @@ def alr(mat, denominator_idx=0):
...
+        denominator = mat[denominator_idx]
+        lr = xp.log(numerator /denominator)
     else:
-        raise ValueError("mat must be either 1D or 2D")
+        raise ValueError("mat must be either 1D or 2D and denominator_idx must be an integer for 1D array")
     return lr
 
 
-def alr_inv(mat, denominator_idx=0):
+def alr_inv(mat:Array, denominator_idx:int|Array=0, /):
```
mortonjt: Don't forget about the `/`. Also use `validate_args` if we want to stay consistent

skbio/stats/composition.py
```diff
@@ -782,25 +842,30 @@ def alr_inv(mat, denominator_idx=0):
...
+        N_minus_1 = len(mat)
+        N = N_minus_1 + 1
+        M = 1
+        comp = xp.zeros((M, N), dtype=mat.dtype)
+        mat = xp.exp(mat)
+        # x_N / (x_1 + ... + x_{N-1} + x_N)
+        comp[:, denominator_idx] = 1 / (xp.sum(mat, axis=1) + 1)
+        # x_i/(x_1 + ... + x_{N-1} + x_N) = \frac{x_i}{x_N} * \frac{x_N}{x_1 + ... + x_{N-1} + x_N}
+        comp[xp.repeat(xp.arange(N),M,axis=0) != denominator_idx] = \
+                    mat[:, N_minus_1] * mat[:, :denominator_idx]
```
mortonjt: Same here, we want to force denominator_idx to be integer. Again, this should be more or less identical to clr_inv. This should dramatically simplify this code.

skbio/stats/composition.py
qiyunzhu: The use of slash (`/`) might be fine and recommended in recent Python packages. It means that every parameter left to is has to be positional. Therefore one can do `alr_inv(arr, 0)` but they cannot do `alr_inv(mat=arr, denominator_idx=0)`. See: https://realpython.com/python-asterisk-and-slash-special-parameters/ Whether this design is good or bad will be your call.

mortonjt: @qiyunzhu @songyesog2000 I don't have a super strong opinion as long as it doesn't introduce unforeseen complexity.

skbio/stats/composition.py
qiyunzhu: A slightly faster verison would be:

```python
gm = mat.max(axis=-1, keepdims=True)
diff = xp.exp(mat - gm)
return diff / diff.sum(axis=-1)
```

Or they can be in just one line.

skbio/stats/composition.py
qiyunzhu: In other functions of scikit-bio this parameter is often called `validate`. [Example 1](https://scikit.bio/docs/latest/generated/skbio.diversity.beta_diversity.html) [Example 2](https://scikit.bio/docs/latest/generated/skbio.stats.distance.DissimilarityMatrix.html). I explored around and found that there isn't a consensus outside deep learning community. For example, SciPy's [curve_fit](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.curve_fit.html) uses `check_finite`. scikit-learn doesn't have those flags but have standalone functions like [check_array](https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_array.html), [check_X_y](https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_X_y.html), etc.

skbio/stats/composition.py
qiyunzhu: @mortonjt  I learned about this only recently after we were asked to add type hinting... I don't have a strong opinion either. In fact, before I learned about this, I found the documentations with `*` or `/` confusing.

skbio/stats/composition.py
songyesog2000: I think 

> I wonder if this can be `mat.shape[-1]`? In my experience with NumPy, `ndarray.some_method` is usually more efficient than `np.some_function`.

Agree that there will be some sacrifice of efficiency, however, inevitable ish.
mat can be the ndarray of any array library (say PyTorch, numpy), xp is a wrapper of correspondent library, where xp rewrite the functions but following Array API standards.

So mat.some_function will bring some risk as functional behavior maybe different, for instance, dim argument in PyTorch while axis used in numpy.

For mat.shape I think it's okay as it is likely universal for all libraries, however, the improvement on efficiency is very limited.

songyesog2000: Maybe casting will be a good idea:
https://docs.scipy.org/doc/scipy/dev/api-dev/array_api.html
https://data-apis.org/array-api/latest/API_specification/generated/array_api.asarray.html

And specifying the internal function _clr and clr, using _clr to run based on Array API and clr will have the casting ? since the casting should be put before the call of _clr

songyesog2000:
Haven't finished yet, outstanding tasks:
1. add dependency/ test decencies into CI
2. a new unit test for arries of different array-consuming libraries.
3. efficience improvement.

tests/test_composition.py has been passed, that is, backward compatibility is ensured (i.e., works for numpy and the result is consistent)
the computational efficiency is a little compromised 

20.5s for testing the new version
Screenshot 2025-06-05 at 9 12 30 PM

19s for testing the initial version
image

songyesog2000:
updated the CI config file for the dependencies/ test dependencies, CI should be complete
added a new file for unit testing for  PyTorch (cpu, cuda) array compatibility
updated composition as basis for irv needed to be consistent on the same gpu, dtype
made clousure function compatible with Array API compat.

Outstanding work:
1. extend the unit test to JAX, CuPy, xarray
2. Redo the Documentation

songyesog2000:
> @songyesog2000 There seems to be some dependency issue. Specifically, the combination of PyTorch, Python 3.13 and Intel MacOS is not available on PyPI (seemingly). If this issue cannot be resolved, a quick patch would be excluding PyTorch from the test requirement, and only test it locally and manually. [Here is how this can be done](https://github.com/scikit-bio/scikit-bio/blob/4872c6c73a0c151cada2139e4d849994e3b2121c/skbio/stats/distance/tests/test_base.py#L555).

Thank you for the heads up, will look into, also, document and unit test for Jax is almost done, will push them tonight!

songyesog2000:
Updated for CI/documentation/jax.numpy

However, the testing show Jax.numpy runs extremely slow also have to reduce atol/rtol from 1e-7 to 1e-4 to pass assert_allclose.
Probably because of the default dtype (is float16?) , JIT compile, or arrays are immutable  in Jax.

qiyunzhu:
@songyesog2000 Error again... due to missing torch, and linting issues. For the latter, you can resolve it automatically using the [Ruff pre-commit hook](https://scikit.bio/contribute.html#style-code)

songyesog2000:
Sorry, my careless, should be good to go now

ci/conda_requirements.txt
```diff
@@ -7,4 +7,5 @@ scipy >= 1.9.0
 h5py >= 3.6.0
 biom-format >= 2.1.16
 statsmodels >= 0.14.0
-patsy >= 0.5.0
\ No newline at end of file
+patsy >= 0.5.0
+rray-api-compat>=1.3,<2.0
```
qiyunzhu: Is `array-api-compat` accidentally truncated in the beginning?

pyproject.toml
```diff
@@ -44,6 +44,7 @@ dependencies = [
     "biom-format >= 2.1.16",
     "statsmodels >= 0.14.0",
     "patsy >= 0.5.0",
+    "array-api-compat>=1.3,<2.0"
```
qiyunzhu: I think we may not need to cap the maximum version. Doing so is usually a good thing, but we haven't with other dependencies of scikit-bio.

ci/conda_requirements.txt
songyesog2000: Yep, somehow the beginning "a" is missed.

pyproject.toml
songyesog2000: Will remove the upper bound version.

songyesog2000:
Minor change is done as per suggested, thanks!

skbio/stats/composition.py
```diff
@@ -566,8 +657,14 @@ def ilr(mat, basis=None, check=True):
...
         orthonormal basis.
-    check : bool
+    axis: int, default -1
+        Should be ignored, as two-dimension mat is concerned, otherwise indicating
+        which dimension of mat the ilr will be applied.
+    validate : bool, default True
         Check to see if basis is orthonormal.
+    is_aitchison: bool, default False
+        Should be ignored, True when the given mat is known to be aitchison, otherwise
+        closre will be performed.
```
mortonjt:
```suggestion
        closure will be performed.
```

skbio/stats/composition.py
```diff
@@ -191,15 +217,24 @@ def closure(mat):
...
-    if np.any(mat < 0):
+    # mat = np.atleast_2d(mat)
+
+    if not aac.is_array_api_obj(mat):
+        mat = np.array(mat)
+    xp = aac.array_namespace(mat)
+    if xp.any(mat < 0):
         raise ValueError("Cannot have negative proportions")
     if mat.ndim > 2:
         raise ValueError("Input matrix can only have two dimensions or less")
```
mortonjt, line 227: @songyesog2000 would you be willing to submit a separate pull request to remove this? We absolutely want tensor support for these compositional operations.

skbio/stats/composition.py
```diff
@@ -449,7 +484,8 @@ def inner(x, y):
     return a.dot(b.T)
 
 
-def clr(mat):
+def clr(mat: Array, axis=-1,
+        validate:bool=True, is_aitchison:bool=False) -> Array:
```
mortonjt: Let's ditch `is_aitchison`. We don't need it

skbio/stats/composition.py
```diff
@@ -474,6 +510,15 @@ def clr(mat):
...
         A matrix of proportions.
+    axis: int, default -1
+        Ignored for the backward compactibility, as the mat
+        is assumed to be a (n_compositions, n_components) matrix.
+    validate: bool, default True
+        Checking of the mat is proportional.
+    is_aitchison: bool, default False
+        should be ignored for the backward compactibility, use True
+        when the input is known to be proportions. Otherwise closure
+        operation will be applied.
```
mortonjt: Don't forget to remove

skbio/stats/composition.py
```diff
@@ -474,6 +510,15 @@ def clr(mat):
     ----------
     mat : array_like of shape (n_compositions, n_components)
         A matrix of proportions.
+    axis: int, default -1
+        Ignored for the backward compactibility, as the mat
+        is assumed to be a (n_compositions, n_components) matrix.
+    validate: bool, default True
+        Checking of the mat is proportional.
```
mortonjt:
```suggestion
        Checking of the mat is a legitimate  composition with elements strictly greater than 0.
```

skbio/stats/composition.py
```diff
@@ -489,13 +534,35 @@ def clr(mat):
...
+    # assert xp.all(mat>0), 'Input must be of positive values.'
+
+    if is_aitchison:
+        # this input is known to be a composition
+        if validate:
+            # check if the input is a composition
+            _composition_check(mat, axis=axis)
+    else:
+        # if the input is not a composition, run closure
+        mat = closure(mat)
```
mortonjt: I suggest replacing this entire nested if statement with the following

```python
if validate:
     # check if the input is a composition
     _composition_check(mat, axis=axis)
```

All of these compositional functions are scale invariant, so we really don't need closures. Its the inverse functions that need closures.

skbio/stats/composition.py
```diff
@@ -517,6 +584,16 @@ def clr_inv(mat):
     ----------
     mat : array_like of shape (n_compositions, n_components)
         A matrix of clr-transformed data.
+    axis: int, default -1
+        Should be ignored, as the mat is assumed to be a
+        (n_compositions, n_components) matrix.
+    validate: bool, defautl True
```
mortonjt:
```suggestion
    validate: bool, default True
```

skbio/stats/composition.py
```diff
@@ -517,6 +584,16 @@ def clr_inv(mat):
...
+        Should be ignored for backward compactibility,the flag of
+        checking whether the mat is centered at 0.
+    is_normalized: bool, default False
+        Should be ignored for the backward compactibility, using True
+        when the input is known to be centered
+        otherwise the data will be shifted to have 0 as center.
```
mortonjt: Here is `is_normalized` being called? Can we remove it?

skbio/stats/composition.py
```diff
@@ -532,13 +609,27 @@ def clr_inv(mat):
...
+        xp = aac.array_namespace(mat)
+    if is_normalized:
+        if validate:
+            # assert xp.all(xp.sum(mat, axis=-1) == 0), \
+                # 'Input matrix is not in the clr range.'
+            pass
+    else:
+        # NOTE: for backward compatibility
+        # mat = mat - xp.sum(mat, axis=-1, keepdims=True)
+        pass
```
mortonjt: I'm really confused about this nested if statement - can we completely remove?

skbio/stats/composition.py
```diff
@@ -566,8 +657,14 @@ def ilr(mat, basis=None, check=True):
     basis : ndarray or sparse matrix, optional
         Orthonormal basis for Aitchison simplex. Defaults to J. J. Egozcue
         orthonormal basis.
-    check : bool
+    axis: int, default -1
+        Should be ignored, as two-dimension mat is concerned, otherwise indicating
+        which dimension of mat the ilr will be applied.
+    validate : bool, default True
         Check to see if basis is orthonormal.
+    is_aitchison: bool, default False
```
mortonjt: I think we should remove `is_aitchison`

skbio/stats/composition.py
```diff
@@ -590,23 +687,53 @@ def ilr(mat, basis=None, check=True):
...
+    except TypeError as e:
+        warnings.warn(
+            "Input is not supported by array_namespace, converting to numpy array. ",
+            UserWarning,
+        )
+
+        mat = np.asarray(mat)
+        xp = aac.array_namespace(mat)
+
+    if is_aitchison:
```
mortonjt: Not a fan of this argument (`is_aitchison` also doesn't make semantic sense). 
I recommend ditching it completely and just sticking with validate.

skbio/stats/composition.py
```diff
@@ -590,23 +687,53 @@ def ilr(mat, basis=None, check=True):
...
+        # this input is known to be a composition
+        if validate:
+            # check if the input is a composition
+            _composition_check(mat, axis=axis)
     else:
+        # if the input is not a composition, run closure
+        mat = closure(mat)
```
mortonjt: Again, let's replace the entire if statement with just

```python
        if validate:
            # check if the input is a composition
            _composition_check(mat, axis=axis)
```

skbio/stats/composition.py
```diff
@@ -590,23 +687,53 @@ def ilr(mat, basis=None, check=True):
...
+                        device=mat.device,
+                        dtype=mat.dtype)
+
+    if axis != -1:
+        switch_tuple = tuple(i for i in range(mat.ndim) if i != axis) + (axis,)
+        mat = mat.permute_dims(switch_tuple)
+        return xp.permute_dims(clr(mat, validate=validate, is_aitchison=True) \
+            @ basis.T,switch_tuple)
+    else:
+        return clr(mat, validate=validate, is_aitchison=True) @ basis.T
```
mortonjt: Let's remove all of the `is_aitchison` arguments

skbio/stats/composition.py
```diff
@@ -590,23 +687,53 @@ def ilr(mat, basis=None, check=True):
...
-
-    return clr(mat) @ basis.T
+        basis = xp.asarray(basis,
+                        device=mat.device,
+                        dtype=mat.dtype)
+
+    if axis != -1:
+        switch_tuple = tuple(i for i in range(mat.ndim) if i != axis) + (axis,)
+        mat = mat.permute_dims(switch_tuple)
+        return xp.permute_dims(clr(mat, validate=validate, is_aitchison=True) \
```
mortonjt: Let's remove all of the `is_aitchison` arguments

skbio/stats/composition.py
```diff
@@ -590,23 +687,53 @@ def ilr(mat, basis=None, check=True):
...
-        if check:
-            _check_orthogonality(basis)
-
-    return clr(mat) @ basis.T
+        basis = xp.asarray(basis,
+                        device=mat.device,
+                        dtype=mat.dtype)
+
+    if axis != -1:
+        switch_tuple = tuple(i for i in range(mat.ndim) if i != axis) + (axis,)
```
mortonjt: Let's add some documentation on what this `switch_tuple` is. This is a little hard to read.

skbio/stats/composition.py
```diff
@@ -721,22 +870,54 @@ def alr(mat, denominator_idx=0):
...
+                        high dimensional alr is new feature",
+                UserWarning,
+            )
+            raise ValueError(
+                f"matrix must be 1d or 2d"
+            )
+            # for backward compactibility
+        if mat.shape[axis] < 2:
+            raise ValueError(
+                f"dimesnion D{mat.ndim + axis} of the input matrix is singleton"
```
mortonjt:
```suggestion
                f"dimension D{mat.ndim + axis} of the input matrix is singleton"
```

skbio/stats/composition.py
```diff
@@ -721,22 +870,54 @@ def alr(mat, denominator_idx=0):
...
+            warnings.warn(
+                "the input matrix has more than 2 dimensions, \
+                        high dimensional alr is new feature",
+                UserWarning,
+            )
+            raise ValueError(
+                f"matrix must be 1d or 2d"
+            )
+            # for backward compactibility
+        if mat.shape[axis] < 2:
```
mortonjt, line 863: Let's revisit in another pull request -- we'll want tensor support soon.

skbio/stats/composition.py
```diff
@@ -721,22 +870,54 @@ def alr(mat, denominator_idx=0):
...
+    if axis != -1:
+        switch_tuple = tuple(i for i in range(mat.ndim) if i != axis) + (axis,)
+        mat = mat.permute_dims(switch_tuple)
+
+    # no matter (n,) or (n, w, z), it will return n
+    N = mat.shape[-1]
+    numerator_indexs = tuple(i for i in range(N) if i != denominator_idx)
+    numerator_matrix = mat[..., numerator_indexs]
+    denominator_vector = mat[..., [denominator_idx]]
+    mat = xp.log(numerator_matrix / denominator_vector)
```
mortonjt: Not numerically stable. We want 
```suggestion
    mat = xp.log(numerator_matrix) - xp.log(denominator_vector)
```

skbio/stats/composition.py
```diff
@@ -782,26 +966,60 @@ def alr_inv(mat, denominator_idx=0):
...
+    except TypeError as e:
+        warnings.warn(
+            "Input is not supported by array_namespace, converting to numpy array. ",
+            UserWarning,
+        )
+
+        mat = np.asarray(mat)
+        xp = aac.array_namespace(mat)
+
+    if mat.ndim > 2:
```
mortonjt, line 937: Revisit in future PR

skbio/stats/composition.py
```diff
@@ -782,26 +966,60 @@ def alr_inv(mat, denominator_idx=0):
...
+        # warnings.warn(
+        #     "the input matrix has more than 2 dimensions, \
+        #               high dimensional alr is new feature",
+        #     UserWarning,
+        # )
+
+
+    if mat.shape[axis] < 2:
+        raise ValueError(
+            f"dimesnion D{mat.ndim + axis} of the input matrix is singleton"
```
mortonjt:
```suggestion
            f"dimension D{mat.ndim + axis} of the input matrix is singleton"
```

skbio/stats/tests/test_composition_torch_compat.py
```diff
@@ -0,0 +1,1071 @@
...
+    #     s = np.apply_along_axis(func1d=lambda x: np.sum(x < 0),
+    #                             axis=1, arr=sbp)
+    #     psi = np.zeros(sbp.shape)
+    #     for i in range(0, sbp.shape[0]):
+    #         for j in range(0, sbp.shape[1]):
+    #             if sbp[i, j] == 1:
+    #                 psi[i, j] = np.sqrt(s[i]/(r[i]*(r[i]+s[i])))
+    #             elif sbp[i, j] == -1:
+    #                 psi[i, j] = -np.sqrt(r[i]/(s[i]*(r[i]+s[i])))
+    #     npt.assert_allclose(psi, sbpbasis)
```
mortonjt: Should we try to get this code working?

skbio/stats/composition.py
qiyunzhu: What does it mean by "be aitchison"? Shall we explain it somewhere in the documentation?

skbio/stats/composition.py
songyesog2000: the issue is that the nested list format of array is used in the unit test, however, "xp = aac.array_namespace(mat)" will lead to error when "mat" is passed into as a list object

skbio/stats/tests/test_composition_torch_compat.py
songyesog2000: Will ditch this as we discussed in the zoom

skbio/stats/composition.py
songyesog2000: Shown in the attached picture, the domain of clr is S^D (the aitchison). However, the initial code also accept R_+^D, which will run closure automatically to map into S^D. This makes sense when the raw data is counts, however, skipping this automation can speed up the calculation when the data is preprocessed or known to be aitchison.
image

mortonjt: Hi @qiyunzhu , @songyesog2000 and I had an offline discussion. We should kick out this parameter, since it is completely unnecessary and will be confusing to users.

skbio/stats/composition.py
mortonjt: Correct. Which is why I'm proposing to not touch it in this pull request -- we should revisit in a future pull request to enable tensor manipulations.

skbio/stats/composition.py
mortonjt: @songyesog2000 I think perhaps a warning is more appropriate here. This is a valid operation that is more telling the user that their CLR array isn't a true CLR array since it doesn't sum to zero. But it is still valid, since the proportions returned from clr_inv will still be valid.

songyesog2000: > @songyesog2000 I think perhaps a warning is more appropriate here. This is a valid operation that is more telling the user that their CLR array isn't a true CLR array since it doesn't sum to zero. But it is still valid, since the proportions returned from clr_inv will still be valid.

Yep, this is a better idea, will modify it.

songyesog2000:
This RP regarding the compatibility with array-consuming lib is done.
The adoption to ND tensor will be in another PR.

skbio/stats/composition.py
```diff
@@ -634,7 +752,9 @@ def ilr_inv(mat, basis=None, check=True):
     basis : ndarray or sparse matrix, optional
         Orthonormal basis for Aitchison simplex. Defaults to J. J. Egozcue
         orthonormal basis.
-    check : bool
+    axis: int, default -1
+        ignored for backward compactibility
```
mortonjt: I'm not crazy about introducing the axis into the function signature at this moment in time. It will make sense once we have tensor support, but for now , it will just confuse users.

I propose to have axis=-1 hard-coded within the function, so that it is not exposed to the user interface. @qiyunzhu do you agree?

skbio/stats/composition.py
```diff
@@ -701,10 +836,15 @@ def alr(mat, denominator_idx=0):
...
     mat : array_like of shape (n_compositions, n_components)
         A matrix of proportions.
-    denominator_idx : int
+    denominator_idx : int, default 0
         The index of the column (2-D matrix) or position (vector) of ``mat``
         which should be used as the reference composition. Default is 0 which
         specifies the first column or position.
+    axis: int, default -1
+        Should be ignored for backward compactibility, the index of dimension the
+        operation will be applied.
```
mortonjt: Let's move axis=-1 to the inside of the function

skbio/stats/composition.py
```diff
@@ -763,10 +935,13 @@ def alr_inv(mat, denominator_idx=0):
...
     mat : array_like of shape (n_compositions, n_components - 1)
         A matrix of alr-transformed data.
-    denominator_idx : int
+    denominator_idx : int, default 0
         The index of the column (2-D matrix) or position (vector) of ``mat``
         which should be used as the reference composition. Default is 0 which
         specifies the first column or position.
+    axis: int, default -1
+        Should be ignored for the backward compactivity, the index of dimension
+        that the operation will be applied on.
```
mortonjt: Same here, move axis to inside of function

skbio/stats/composition.py
```diff
@@ -782,26 +957,60 @@ def alr_inv(mat, denominator_idx=0):
...
+
+
+    if mat.shape[axis] < 2:
+        raise ValueError(
+            f"dimension D{mat.ndim + axis} of the input matrix is singleton"
+        )
+    # switch the wanted axis to the last one
+    if axis != -1:
+        switch_tuple = tuple(i for i in range(mat.ndim) if i != axis) + (axis,)
+        mat = mat.permute_dims(switch_tuple)
```
mortonjt: If axis=-1 is hard-coded, we can completely remove this line.

skbio/stats/composition.py
```diff
@@ -590,23 +676,55 @@ def ilr(mat, basis=None, check=True):
...
+            raise ValueError("Cannot have negative proportions")
+        if mat.ndim > 2:
+            raise ValueError("Input matrix can only have two dimensions or less")
+
+        # check if the input is a composition
+        _composition_check(mat, axis=axis)
+
+    if axis != -1:
+        switch_tuple = tuple(i for i in range(mat.ndim) if i != axis) + (axis,)
+        mat = mat.permute_dims(switch_tuple)
```
mortonjt: Remove since axis=-1 is always enforced

skbio/stats/composition.py
```diff
@@ -590,23 +676,55 @@ def ilr(mat, basis=None, check=True):
...
-        if check:
-            _check_orthogonality(basis)
+        basis = xp.asarray(basis,
+                        device=mat.device,
+                        dtype=mat.dtype)
 
-    return clr(mat) @ basis.T
+    # switch the wanted axis to the last one
+    if axis != -1:
+        return xp.permute_dims(mat @ basis.T, switch_tuple)
```
mortonjt: Remove since axis=-1

skbio/stats/composition.py
```diff
@@ -658,26 +778,41 @@ def ilr_inv(mat, basis=None, check=True):
...
-    return clr_inv(mat @ basis)
+    # if not isinstance(basis, xp.array):
+    basis = xp.asarray(basis,
+                       device=mat.device,
+                       dtype=mat.dtype)
+    if axis != -1:
+        switch_tuple = tuple(i for i in range(mat.ndim) if i != axis) + (axis,)
+        mat = mat.permute_dims(switch_tuple)
+        return xp.permute_dims(clr_inv(mat @ basis, validate=validate),
+                               switch_tuple)
```
mortonjt: Remove since axis=-1

skbio/stats/composition.py
```diff
@@ -721,22 +861,54 @@ def alr(mat, denominator_idx=0):
...
+            # for backward compactibility
+        if mat.shape[axis] < 2:
+            raise ValueError(
+                f"dimension D{mat.ndim + axis} of the input matrix is singleton"
+            )
+        if xp.any(mat <= 0):
+            raise ValueError("Input matrix must be positive")
+    if axis != -1:
+        switch_tuple = tuple(i for i in range(mat.ndim) if i != axis) + (axis,)
+        mat = mat.permute_dims(switch_tuple)
```
mortonjt: Remove since axis=-1

skbio/stats/composition.py
```diff
@@ -721,22 +861,54 @@ def alr(mat, denominator_idx=0):
...
+    # no matter (n,) or (n, w, z), it will return n
+    N = mat.shape[-1]
+    numerator_indexs = tuple(i for i in range(N) if i != denominator_idx)
+    numerator_matrix = mat[..., numerator_indexs]
+    denominator_vector = mat[..., [denominator_idx]]
+    mat = xp.log(numerator_matrix)-xp.log(denominator_vector)
+
+    # recover the permutation
+    if axis != -1:
+        mat = mat.permute_dims(switch_tuple)
```
mortonjt: remove since axis=-1

skbio/stats/composition.py
```diff
@@ -782,26 +957,60 @@ def alr_inv(mat, denominator_idx=0):
...
+        # do not support in-place item assignment.
+        # Instead of x[idx] = y, use x = x.at[idx].set(y)
+        comp = comp.at[..., numerator_indexs].set(xp.exp(mat))
+    else:
+        # in-place item assignment
+        comp[..., numerator_indexs] = xp.exp(mat)
+
+    # recover the permutation
+    if axis != -1:
+        comp = comp.permute_dims(switch_tuple)
```
mortonjt: Remove since axis=-1

skbio/stats/tests/test_composition_jnp_compat.py
```diff
@@ -0,0 +1,1045 @@
...
+    #     s = np.apply_along_axis(func1d=lambda x: np.sum(x < 0),
+    #                             axis=1, arr=sbp)
+    #     psi = np.zeros(sbp.shape)
+    #     for i in range(0, sbp.shape[0]):
+    #         for j in range(0, sbp.shape[1]):
+    #             if sbp[i, j] == 1:
+    #                 psi[i, j] = np.sqrt(s[i]/(r[i]*(r[i]+s[i])))
+    #             elif sbp[i, j] == -1:
+    #                 psi[i, j] = -np.sqrt(r[i]/(s[i]*(r[i]+s[i])))
+    #     npt.assert_allclose(psi, sbpbasis)
```
mortonjt, line 566: Make sure to remove these comments

mortonjt (commented):
Almost there. Let's have the following done before merging

1. Update the changelog
2. Remove all axis arguments in this pull request
3. Remove all switch_tuple variables

We can always revisit permute dims and axis arguments in the next pull request when dealing with N-D tensors

songyesog2000:
modified as suggested, ruff check passed.
Did not involve with changelog, looked at CHANGELOG.md feeling not the thing I should touch.

skbio/stats/composition.py
```diff
@@ -590,23 +668,48 @@ def ilr(mat, basis=None, check=True):
...
+        warnings.warn(
+            "Input is not supported by array_namespace, converting to numpy array. ",
+            UserWarning,
+        )
+
+        mat = np.asarray(mat)
+        xp = aac.array_namespace(mat)
+    if validate:
+        # assert from closure() while it is removed in latest version
+        if xp.any(mat < 0):
```
mortonjt, line 685: isn't this redundant? This check is contained inside of `_composition_check`

skbio/stats/composition.py
```diff
@@ -489,13 +527,36 @@ def clr(mat):
...
+    except Exception as e:
+        mat = np.asarray(mat)
+        xp = aac.array_namespace(mat)
 
 
-def clr_inv(mat):
+    if validate:
+        # assert from closure() while it is removed in latest version
+        if xp.any(mat < 0):
+            raise ValueError("Cannot have negative proportions")
```
mortonjt, line 544: Same here, isn't his redundant since this check is contained inside of `_composition_check`?

skbio/stats/composition.py
```diff
@@ -721,22 +836,46 @@ def alr(mat, denominator_idx=0):
...
+            raise ValueError(
+                f"matrix must be 1d or 2d"
+            )
+            # for backward compactibility
+        if mat.shape[axis] < 2:
+            raise ValueError(
+                f"dimension D{mat.ndim + axis} of the input matrix is singleton"
+            )
+        if xp.any(mat <= 0):
+            raise ValueError("Input matrix must be positive")
```
mortonjt, line 868: Shouldn't we use `_composition_check` here instead for consistency?

skbio/stats/composition.py
```diff
@@ -782,26 +921,52 @@ def alr_inv(mat, denominator_idx=0):
...
+
+    if mat.ndim > 2:
+        # NOTE: backward compatibility
         raise ValueError("mat must be either 1D or 2D")
-    return comp
+        # warnings.warn(
+        #     "the input matrix has more than 2 dimensions, \
+        #               high dimensional alr is new feature",
+        #     UserWarning,
+        # )
```
mortonjt, line 944: Remove these comments

mortonjt (commented):
Hi @songyesog2000 , last round of comments. Don't forget to update the changelog -- we have a habit of not merging PRs until the changelog is updates

https://github.com/scikit-bio/scikit-bio/blob/main/CHANGELOG.md

See here for more details why this is important https://scikit.bio/contribute.html#changelog

skbio/stats/composition.py
songyesog2000: Yep, this is redundant, actually, not for the _composition_check, but for clr(mat,validate).
mat will be checked again in clr if validate is true.

mortonjt (approved):

skbio/stats/composition.py
```diff
@@ -782,26 +921,52 @@ def alr_inv(mat, denominator_idx=0):
...
+        xp = aac.array_namespace(mat)
+    except TypeError as e:
+        warnings.warn(
+            "Input is not supported by array_namespace, converting to numpy array. ",
+            UserWarning,
+        )
+
+        mat = np.asarray(mat)
+        xp = aac.array_namespace(mat)
+
```
songyesog2000, line 936: @mortonjt how do you think about adding validate in alr_inv and other inv functions. Initially there is not such setting.

mortonjt: lets not - not clear what we are validating

songyesog2000:
Sorry... I don't know how this PR work, but I committed  the new one resolved the comments.
My bad, should not mark the comment resolved before pushing the commit.

skbio/stats/composition.py
```diff
@@ -157,6 +157,30 @@
 from statsmodels.tools.sm_exceptions import ConvergenceWarning
 from patsy import dmatrix
 
+import array_api_compat as aac
+import warnings
+from typing import Optional
+
+Array = object
+
+def supports_array_api(obj):
```
qiyunzhu, line 166: @songyesog2000 Quick question: What is this `supports_array_api` function for? It isn't used anywhere in the codebase. Can this be removed? (Just a question not a request. I will remove it if it can.)

songyesog2000: > @songyesog2000 Quick question: What is this `supports_array_api` function for? It isn't used anywhere in the codebase. Can this be removed? (Just a question not a request. I will remove it if it can.)

@qiyunzhu  it is a function to check whether the library of the object supports array_api (standards) , some package, e.g. numpy, have already adopted array_api under the __array_namespace__ . Please remove it, it is redundant  as it is implicitly integrated into the wrapper  array_api_compat already.

My initial intention is to detect the case when the input is of list type to avoid using "try", e.g. code line 841-850. However, I just realize using  `supports_array_api` will also skip some libraries that  array_api_compat supports, so discarded.

qiyunzhu: @songyesog2000 Thanks for the clarification! Will do.

skbio/stats/composition.py
```diff
@@ -489,13 +527,36 @@ def clr(mat):
...
+        if xp.any(mat < 0):
+            raise ValueError("Cannot have negative proportions")
+        if mat.ndim > 2:
+            raise ValueError("Input matrix can only have two dimensions or less")
+
+        # check if the input is a composition
+        _composition_check(mat)
+
+    original_shape = mat.shape
+    # squeeze the singleton dimensions
```
qiyunzhu, line 552: @songyesog2000 I think that `squeeze` is part of the Python API standard ([source](https://data-apis.org/array-api/latest/API_specification/generated/array_api.squeeze.html)) and it automatically removes all singleton dimensions. So the following line can just be `mat = xp.squeeze(mat)`?

skbio/stats/composition.py
qiyunzhu: @mortonjt If I understand correctly, you are suggesting that the last dimension of a multidimensional array is the composition. So the axis closure is working on should become -1. The parameter signature should be `mat : array_like of shape (..., n_components)`. Is my understanding correct? In the current code, `clr` already has tensor support (axis=-1).

skbio/stats/composition.py
songyesog2000: @qiyunzhu yes, we can use squeeze instead of reshape, but no that  `xp.squeeze(mat)` can not be used.
Python API standard ([source](https://data-apis.org/array-api/latest/API_specification/generated/array_api.squeeze.html))  states `squeeze(x: array, /, axis: int | Tuple[int, ...]) → array`, where no default value is given for the axis and even "None" is not usable shown by the typing hints.
If `squeeze` is the better option, then `mat = squeeze(mat,  tuple(i for i,d in enumerate(mat.shape) if d ==1 )) ` is the case.

PR #2244, opened 2025-06-23
Removed miniforge version and patched checkout in CI workflow

qiyunzhu:
In the previous status, when the GitHub Actions workflows were executed, a warning was raised:

> 'Mambaforge' variants are now equivalent to 'Miniforge3'. In the future, we will ignore with a warning and use 'Miniforge3'. Eventually, using 'Mambaforge' will throw an error. Please change to 'Miniforge3' at your earliest convenience.

Therefore, this PR aims to resolve this warning.

According to [setup-miniconda/action.yml](https://github.com/conda-incubator/setup-miniconda/blob/main/action.yml), if both `miniconda-version` and `miniforge-variant` are not specified, they will default to the latest Miniforge3. This is what we desire.

***

The 2nd commit also addresses a (maybe temporary) Sphinx problem: It reads the git timestamp for some reason (although I don't know what reason), whereas GitHub Actions' "checkout" by default checks out only the last commit and ignore previous git history. So when `make doc` is executed, it warns: "Git clone too shallow [git.too_shallow]", and this warning is treated as an error. One solution is to add `fetch-depth: 0` to `actions/checkout`, as is done here. Another solution is to ignore this or all warnings in Sphinx.

***

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,8 @@ jobs:
         shell: bash -l {0}
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
       - uses: conda-incubator/setup-miniconda@v3
         with:
           auto-update-conda: true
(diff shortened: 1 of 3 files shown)


qiyunzhu:
Let me self-merge as this will unblock things.

PR #2247, opened 2025-06-30
Add support for scikit-bio-binaries

sfiligoi:
Add the needed glue code for using the scikit-bio-binaries package, which contains optimized versions of pcoa_fsvd and permanova.
The integration is fully optional.
If the needed shared library, i.e. libskbb, is detected, it will attempt to use the optimized versions of the functions.
If either the shared library is not found, or the invocation of the optimized functions fails, the standard logic will be used, instead.

### DIFF
--- a/skbio/__init__.py
+++ b/skbio/__init__.py
@@ -23,8 +23,8 @@ from skbio.table import Table
 from skbio.metadata import SampleMetadata
 from skbio.util import set_config, get_config
 import skbio.diversity  # noqa
-import skbio.stats.evolve  # noqa
 import skbio.binaries  # noqa
+import skbio.stats.evolve  # noqa
 
 __all__ = [
     "Sequence",


sfiligoi:
@qiyunzhu The PR should be ready for review.

skbio/binaries/_util.py
```diff
@@ -0,0 +1,173 @@
...
+        if (seed_or_generator<0):
+             raise ValueError("seed must be a non-negative number")
+        else:
+             return seed_or_generator # was already a valid seed
+    else:
+        # assume it is a generator, get a new int seed
+        maxint = np.iinfo(np.int32).max + 1
+        if isinstance(seed, np.random.Generator):
+            return seed_or_generator.integers(0, maxint)
+        elif isinstance(seed, np.random.RandomState):
```
mataton: The two instances of `seed` should be `seed_or_generator`, correct?

skbio/binaries/_ordination.py
```diff
@@ -0,0 +1,189 @@
...
+):
+    r"""Perform Principal Coordinate Analysis (PCoA).
+
+    PCoA is an ordination method similar to Principal Components Analysis (PCA), with
+    the difference that it operates on distance matrices, calculated using meaningful
+    and typically non-Euclidian methods.
+
+    Parameters
+    ----------
+    distance_matrix : DistanceMatrix
```
mataton: This documentation suggests only `DistanceMatrix` objects may be passed but the code can actually handle `ndarray` objects as well, given the conditional `if isinstance(distance_matrix,np.ndarray):` on line 136

skbio/binaries/_ordination.py
```diff
@@ -0,0 +1,189 @@
...
+
+    """
+    if _skbb_get_api_version()>=1: # minimum version that support pcoa
+        if not isinstance(number_of_dimensions, Integral):
+            raise ValueError("number_of_dimensions must be an integer value")
+        if number_of_dimensions<1:
+            raise ValueError("number_of_dimensions must be a positive number")
+        int_seed = py_to_bin_random_seed(seed)
+        if isinstance(distance_matrix,np.ndarray):
+            # already a raw matrix, jsut use
```
mataton: Typo here, `just`

skbio/binaries/__init__.py
```diff
@@ -0,0 +1,49 @@
+r"""Optimized libraries from scikit-bio-binaries
+============================================
+
+.. currentmodule:: skbio.binaries
```
mataton, line 4: Currently this documentation won't actually be built if one runs `make doc`. To add this to the documentation there needs to be a new file added at `doc/source/binaries.rst`, if we want the documentation to be built as a top level submodule.

skbio/binaries/_distance.py
```diff
@@ -0,0 +1,178 @@
...
+        A user-provided random seed or random generator instance.
+
+    Returns
+    -------
+    boolean
+        If False, the permanova function will raise an exception.
+
+    Note
+    ----
+    Internally it uses caching, to minimize overhead
```
mataton: Need period at end of sentence here.

skbio/binaries/_distance.py
```diff
@@ -0,0 +1,178 @@
...
+            # already a raw matrix, just use
+            distance_matrix_data = distance_matrix
+        else:
+            # we are assuming it is a DistanceMatrix object
+            # get internal representations
+            distance_matrix_data = distance_matrix.data
+        distance_matrix_shape0 = distance_matrix_data.shape[0]
+        if (len(distance_matrix_data.shape)!=2 or
+            distance_matrix_data.shape[1] != distance_matrix_shape0):
+               raise TypeError("distance_matrix not square")
```
mataton, line 143: This may not be enough input validation. For example I was able to do this:
```python
from skbio.binaries._distance import permanova
import numpy as np
dm = np.array([[1, 2, 3],
               [4, 5, 6],
               [7, 8, 9]], dtype='float')
groups = np.array([0, 1, 0])
permanova(distance_matrix=dm, grouping=groups, permutations=999)
```

The input `dm` here isn't actually a distance matrix at all, so the results of permanova I think are meaningless in this case. I'm not sure that a user would ever do this, and I understand that I am calling the skbb version of permanova directly, but it is possible. There is the [`is_symmetric_and_hollow`](https://github.com/scikit-bio/scikit-bio/blob/a6f1b1bda0379eae3f51b7c2f18f573330e08092/skbio/stats/distance/_utils.py#L15) function already in the package which may be of use here.

skbio/binaries/_ordination.py
```diff
@@ -0,0 +1,189 @@
...
+        A user-provided random seed or random generator instance.
+
+    Returns
+    -------
+    boolean
+        If False, the pcoa_fsvd function will raise an exception.
+
+    Note
+    ----
+    Internally it uses caching, to minimize overhead
```
mataton: Period at end here as well.

skbio/binaries/_ordination.py
```diff
@@ -0,0 +1,189 @@
...
+            # already a raw matrix, jsut use
+            distance_matrix_data = distance_matrix
+        else:
+            # we are assuming it is a DistanceMatrix object
+            # get internal representations
+            distance_matrix_data = distance_matrix.data
+        distance_matrix_shape0 = distance_matrix_data.shape[0]
+        if (len(distance_matrix_data.shape)!=2 or
+            distance_matrix_data.shape[1] != distance_matrix_shape0):
+               raise TypeError("distance_matrix not square")
```
mataton, line 151: Similar concerns here regarding input validation. It is possible to pass a non-symmetric and non-hollow square numpy array and the function will work.

skbio/binaries/_ordination.py
```diff
@@ -0,0 +1,189 @@
...
+        if number_of_dimensions>distance_matrix_shape0:
+            raise ValueError("number_of_dimensions cannot be larger that matrix size")
```
mataton: `than` instead of `that` here.

mataton (changes requested):
Some concerns regarding input validation, documentation building, but other than that things look pretty good. Thanks @sfiligoi

skbio/binaries/_distance.py
sfiligoi: I have added a comment specifying that the result is undefined if a ndarray that is not a distance matrix is passed.
I don't think protecting users from their own mistakes is worth the added cost of the check, when it is just a matter of interpreting the results.

skbio/binaries/_util.py
sfiligoi: Thanks for catching this.
Fixed.

skbio/binaries/_ordination.py
sfiligoi: Updated the comment.

skbio/binaries/_ordination.py
sfiligoi: Fixed.

skbio/binaries/_distance.py
sfiligoi: Fixed.

skbio/binaries/_ordination.py
sfiligoi: Fixed.

skbio/binaries/_ordination.py
sfiligoi: See above. I have added a comment specifying that the result is undefined if a ndarray that is not a distance matrix is passed.

skbio/binaries/_ordination.py
sfiligoi: Fixed.

skbio/binaries/__init__.py
sfiligoi: @mataton The binaries sub-module should be internal to the library.
It is not meant to be called by users directly, apart from tests and during debugging.
Please advise how to best achieve this.

mataton: Hey @sfiligoi I discussed with @qiyunzhu and we both agree that it should be internal to the library, therefore the documentation does not need to be public facing. So we can leave it as is.

mataton (approved):
@sfiligoi Looks good to me, ready to merge?

PR #2250, opened 2025-07-08
WIP Improving the composition module

qiyunzhu:
This PR intends to provide a further upgrade of the `skbio.stats.composition` module based on @songyesog2000 and @mortonjt 's previous work. Specifically:

1. Log-ratio transformations:
  - All functions support input with an arbitrary number of dimensions (e.g., high-dimensional tensors).
  - All functions support transformation on an arbitrary axis (not necessarily the last axis).
  - Removed redundant code such as namespace and validation.
    - Specifically, a global `_ingest_array` function now handles namespace. A modified `_check_composition` function handles validation.
  - Refined edge cases, validation check, etc.
    - e.g., Input for `closure` can have zero; input for `clr` cannot have zero.
  - All functions comply with the Python array API standard, and support GPU.

2. Differential abundance testing:
  - Optimized `ancom` and `dirmult_ttest` such that they are multiple folds faster than the previous versions, whereas the results remain unchanged (i.e., with the same seed, the resulting numbers are precisely identical).
    - Replaced the logic of pandas DataFrame with arrays, which accelerated the analyses.
    - Vectorized steps such as Dirchlet-multinomial sampling.
  - Fixed some bugs and issues.

***

### DIFF
--- a/skbio/util/_array.py
+++ b/skbio/util/_array.py
@@ -1,0 +1,100 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2013--, scikit-bio development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+"""Array-like objects and namespaces."""
+
+from typing import Any, Tuple, TYPE_CHECKING
+
+import numpy as np
+import array_api_compat as aac
+
+
+if TYPE_CHECKING:  # pragma: no cover
+    from types import ModuleType
+    from ._typing import ArrayLike
+
+_ = aac.array_namespace(np.array([]))
+
+
+def _ingest_array(
+    arr: "ArrayLike", /, *, to_numpy: bool = False
+) -> Tuple["ModuleType", Any]:
+    r"""Convert an array-like variable into an array object and its namespace.
+
+    Parameters
+    ----------
+    arr : array_like
+        Input object.
+    to_numpy : bool, optional
+        If True, make sure the returned array is a NumPy array.
+
+    Returns
+    -------
+    xp : namespace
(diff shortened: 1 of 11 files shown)


skbio/util/_array.py
```diff
@@ -0,0 +1,100 @@
...
+if TYPE_CHECKING:  # pragma: no cover
+    from types import ModuleType
+    from ._typing import ArrayLike
+
+_ = aac.array_namespace(np.array([]))
+
+
+def _ingest_array(
+    arr: "ArrayLike", /, *, to_numpy: bool = False
+) -> Tuple["ModuleType", Any]:
```
songyesog2000: @qiyunzhu maybe a good idea to return the device and dtype as well?

qiyunzhu: Device and dtype don't require aac I guess. One can do `arr.device` and `arr.dtype`.

skbio/util/_typing.py
```diff
@@ -6,17 +6,46 @@
...
+    ._array._ingest_array
+
+    References
+    ----------
+    .. [1] https://data-apis.org/array-api/latest/
+
+    """
+
+    def __array_namespace__(self, api_version: Optional[str] = None): ...
+
```
songyesog2000, line 42: maybe worthy to add other universal methods of array-consuming libraries?
```
    @property
    def shape(self) -> Tuple[int, ...]: ...
    
    @property
    def ndim(self) -> int: ...
    
    @property
    def dtype(self) -> Any: ...
```

qiyunzhu: This sounds a good idea. These three properties are part of the Python array API standards. So every compliant array object should have those. Enforcing them again in scikit-bio can may the static typing process more smooth.

qiyunzhu:
This PR has been discussed with Jamie and Sam. We now need to merge and move on.

Issue #2275, opened 2025-08-11
Add support for unifrac-binaries compatible DM file format


sfiligoi:
The unifrac-binaries package has long had support for HDF5-based DM file format.
Using a binary-based format is both faster (about 100x) and uses less space (about 1/5th) than the existing text format.

sfiligoi:
Turns out that the existing binary_dm implementation is already unifrac-binary compatible, but the implementation was broken.

sfiligoi:
PRs #2282 and #2283 implemented the needed functionality, so closing the issue.

PR #2282, opened 2025-08-20
Fix binary_dm reader

sfiligoi:

### DIFF
--- a/skbio/io/format/binary_dm.py
+++ b/skbio/io/format/binary_dm.py
@@ -112,26 +112,28 @@ def _binary_dm_sniffer(fh):
 
 @binary_dm.reader(DissimilarityMatrix)
 def _binary_dm_to_dissimilarity(fh):
-    return _h5py_mat_to_skbio_mat(fh)
+    return _h5py_mat_to_skbio_mat(DissimilarityMatrix,fh)
 
 
 @binary_dm.reader(DistanceMatrix)
 def _binary_dm_to_distance(fh):
-    return _h5py_mat_to_skbio_mat(fh)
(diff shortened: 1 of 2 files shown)


sfiligoi:
Fixed the syntax errors in the binary_dm source code, limited to the reader.
This allows to read a unifrac-compatible HDF5 into a Distance Matrix object.

It does not fix the writer, as the current mechanism of passing a compressed stream to the binary_dm functions does not play well with h5py.
A refactoring of the direct-to-file handling in the io module is necessary to properly support that.

sfiligoi:
Partially addresses #2275

mataton:
I "vibe coded" a quite ugly workaround for the writers. Not sure if you're interested at all in it but here it is:

```python
def _skbio_mat_to_h5py_mat_stream(obj, fh):
    import tempfile
    import shutil
    
    # Create temporary file that h5py can work with
    with tempfile.NamedTemporaryFile(delete=False) as tmp:
        tmp_path = tmp.name
    
    try:
        # Write to temporary file using h5py
        with h5py.File(tmp_path, "w") as f:
            _skbio_mat_to_h5py_mat(obj, f)
        
        # Copy the temporary file contents to the output file handle
        with open(tmp_path, 'rb') as tmp_file:
            shutil.copyfileobj(tmp_file, fh)
    finally:
        # Clean up temporary file
        import os
        if os.path.exists(tmp_path):
            os.unlink(tmp_path)
```

It works, but again, pretty rough and the IO cost is significant. A better redesign of the IO system is probably warranted for this.

The code for the readers looks good to me, so I'm okay with merging even though the writers are still an issue.

PR #2285, opened 2025-08-28
Drastically speed up default eigh pcoa

sfiligoi:

### DIFF
--- a/skbio/stats/ordination/_principal_coordinate_analysis.py
+++ b/skbio/stats/ordination/_principal_coordinate_analysis.py
@@ -152,6 +152,14 @@ def pcoa(
                 "distance matrix is large.",
                 RuntimeWarning,
             )
+        elif method == "eigh" and distmat.data.shape[0] > 10:
+            warn(
+                "EIGH: since no value for number_of_dimensions is specified, "
+                "PCoA for all dimensions will be computed, which may "
+                "result in long computation time if the original "
+                "distance matrix is large.",
+                RuntimeWarning,
+            )
 
         # distmat is guaranteed to be square
         dimensions = distmat.data.shape[0]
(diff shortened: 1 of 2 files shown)


sfiligoi:
This new code can compute a 25k (EMP) DM and dimensions=10 in less than 150s on my M2Pro MacBook.
Before, it would not complete in half a day.

For a 5k DM and dimensions=10, the time on my M2Pro MacBook went from almost 9s to about 1.5s.

sfiligoi:
The PCoA of a 50k DM now completes in about 990s on my M2Pro.

sfiligoi:
FYI: I  managed to run the 25k DM on a 128-core Xeon, and it took 26.5k seconds (7.5 hours)
With the new code, it finishes in 43 seconds.
So, a 600x speedup.
(suspect the 50k speedup would be an order of magnitude larger still, but do not plan to try)

BTW: A quick check shows that the eigenvalues are indeed identical to at least 6 digits:
```
# PR2285					 		        # skbio 0.7.0
PC1     949.005493					      |	PC1     949.005615
PC2     544.519104					      |	PC2     544.519043
PC3     409.926331					        PC3     409.926331
PC4     337.943268					        PC4     337.943268
PC5     277.664062					      |	PC5     277.664093
PC6     214.412384					      |	PC6     214.412415
PC7     145.958344					      |	PC7     145.958374
PC8     140.763458					        PC8     140.763458
PC9     124.495651					      |	PC9     124.495689
PC10     95.257599					      |	PC10     95.257652
```

rob-knight:
Wow, that’s amazing!

Any other major stuff we are leaving on the table through our lack of linear algebra skills?

skbio/stats/ordination/_principal_coordinate_analysis.py
```diff
@@ -152,6 +152,14 @@ def pcoa(
                 "distance matrix is large.",
                 RuntimeWarning,
             )
+        elif method == "eigh" and distmat.data.shape[0] > 10:
```
qiyunzhu, line 155: I am not sure if this warning is necessary, or if the triggering threshold (n > 10) is too small. It is quite common that a distance matrix with hundreds of samples is fed to PCoA. That won't cause long computation time, will it?

sfiligoi: I followed the same exact logic we had to fsvd.
If this is not desirable, we should update both, for consistency.

That said: Anyone really needs more than 10 dimensions?
(and are not just throwing them away in post-processing?)

qiyunzhu: There are several `matrix_data.shape[0] > 10` lines in the added code. I thought that it means the input distance matrix has more than 10 samples, instead of the user specified more than 10 dimensions. Is this understanding correct?

sfiligoi: This specific test is within `if dimensions == 0` block.
Which has the implict meaning of  "convert dimensions to size of DM".
So testing on either would be equivalent (if we tested after the assignment; the test actually happens before dimensions is changed).

I.e., we are telling the users who **do not explicitly** provide an integer `dimension` that we will compute all the eigenvalues, and that this will be very expensive.
(If a user does provide an integer dimension value, we assume they do know what they are doing)

mortonjt: @sfiligoi depends on the use-case. If we are going to be using eigh as a proxy for embedding search, then yes absolutely -- the dimensionality dictates distance approximation accuracy (which will affect your search results).

But we don't currently support this type of functionality yet, so maybe we can revisit

sfiligoi: It's just a warning in case a user does not explicitly provide the number of dimensions it wants.
(So, nothing breaks in those cases)
Moreover, if they do provide the number of dimensions, there is no warning, ever.

That said, I have no strong opinion if we want/need a warning or not.

But would definitely prefer if we threated "eight" and "fsvd" the same, as the cost considerations are now identical.

qiyunzhu: I have reviewed the code again. Given @mortonjt 's input, and that the threshold 10 is already used in the current [code](https://github.com/scikit-bio/scikit-bio/blob/74e1e14dcefbf381bae5ae6d7ae6048060f95856/skbio/stats/ordination/_principal_coordinate_analysis.py#L147), I'd like to support @sfiligoi 's addition of this warning.

@sfiligoi , I'd like to suggest using the [`_warn_once`](https://github.com/scikit-bio/scikit-bio/blob/74e1e14dcefbf381bae5ae6d7ae6048060f95856/skbio/util/_warning.py#L14) function, in case the user knows what they are doing and want to stick to it for many calls.

qiyunzhu:
@sfiligoi Given that both me and Jamie reviewed the `subset_by_index` approach and found no problem, I'd think that this improvement can be safely shipped. Thank you very much for making this simple but powerful acceleration!

PR #2289, opened 2025-09-09
WIP:  Implement `PairwiseMatrix`, `SymmetricMatrix`, and `DistanceMatrix` classes

mataton:
This PR is a work in progress, not ready for review.

The goal of this PR is to implement a new class hierarchy in `skbio.stats.distance._base`. This is the first step in the plan to refactor the `DistanceMatrix` class in scikit-bio. The work will be broken up across several PR's to make reviews and progress more manageable.

The previous `DissimilarityMatrix` will be renamed `PairwiseMatrix`. The old name will be kept as an alias for backwards compatibility.

A new class `SymmetricMatrix` will be introduced to fill the gap between the old `DissimilarityMatrix` and `DistanceMatrix` classes.

`DistanceMatrix` will remain largely unchanged for now.

### DIFF
--- a/skbio/io/format/lsmat.py
+++ b/skbio/io/format/lsmat.py
@@ -17,7 +17,7 @@ Format Support
 +------+------+---------------------------------------------------------------+
 |Reader|Writer|                          Object Class                         |
 +======+======+===============================================================+
-|Yes   |Yes   |:mod:`skbio.stats.distance.DissimilarityMatrix`                |
+|Yes   |Yes   |:mod:`skbio.stats.distance.PairwiseMatrix`                |
 +------+------+---------------------------------------------------------------+
 |Yes   |Yes   |:mod:`skbio.stats.distance.DistanceMatrix`                     |
 +------+------+---------------------------------------------------------------+
@@ -79,7 +79,7 @@ import csv
 
 import numpy as np
 
-from skbio.stats.distance import DissimilarityMatrix, DistanceMatrix
+from skbio.stats.distance import PairwiseMatrix, DistanceMatrix
 from skbio.io import create_format, LSMatFormatError
 
 
@@ -106,9 +106,9 @@ def _lsmat_sniffer(fh):
     return False, {}
 
 
-@lsmat.reader(DissimilarityMatrix)
+@lsmat.reader(PairwiseMatrix)
 def _lsmat_to_dissimilarity_matrix(fh, delimiter="\t"):
-    return _lsmat_to_matrix(DissimilarityMatrix, fh, delimiter)
+    return _lsmat_to_matrix(PairwiseMatrix, fh, delimiter)
 
 
 @lsmat.reader(DistanceMatrix)
@@ -116,7 +116,7 @@ def _lsmat_to_distance_matrix(fh, delimiter="\t"):
     return _lsmat_to_matrix(DistanceMatrix, fh, delimiter)
 
 
-@lsmat.writer(DissimilarityMatrix)
+@lsmat.writer(PairwiseMatrix)
 def _dissimilarity_matrix_to_lsmat(obj, fh, delimiter="\t"):
     _matrix_to_lsmat(obj, fh, delimiter)
 
@@ -156,7 +156,7 @@ def _lsmat_to_matrix(cls, fh, delimiter):
             # We've hit a nonempty line after we already filled the data
             # matrix. Raise an error because we shouldn't ignore extra data.
             raise LSMatFormatError(
-                "Encountered extra row(s) without corresponding IDs in " "the header."
+                "Encountered extra row(s) without corresponding IDs in the header."
             )
 
         num_vals = len(row_data)
--- a/skbio/stats/distance/_base.py
+++ b/skbio/stats/distance/_base.py
@@ -33,7 +33,7 @@ from skbio.util._misc import resolve_key
 from skbio.util._plotting import PlottableMixin
 from skbio.io.descriptors import Read, Write
 
-from ._utils import is_symmetric_and_hollow
+from ._utils import is_symmetric_and_hollow, is_symmetric
 from ._utils import distmat_reorder, distmat_reorder_condensed
 
 if TYPE_CHECKING:  # pragma: no cover
@@ -43,19 +43,19 @@ if TYPE_CHECKING:  # pragma: no cover
     from skbio.util._typing import SeedLike
 
 
-class DissimilarityMatrixError(Exception):
+class PairwiseMatrixError(Exception):
     """General error for dissimilarity matrix validation failures."""
 
     pass
 
 
-class DistanceMatrixError(DissimilarityMatrixError):
+class DistanceMatrixError(PairwiseMatrixError):
     """General error for distance matrix validation failures."""
 
     pass
 
 
-class MissingIDError(DissimilarityMatrixError):
+class MissingIDError(PairwiseMatrixError):
     """Error for ID lookup that doesn't exist in the dissimilarity matrix."""
 
     def __init__(self, missing_id):
@@ -63,28 +63,28 @@ class MissingIDError(DissimilarityMatrixError):
         self.args = ("The ID '%s' is not in the dissimilarity matrix." % missing_id,)
 
 
-class DissimilarityMatrix(SkbioObject, PlottableMixin):
-    """Store dissimilarities between objects.
+class PairwiseMatrix(SkbioObject, PlottableMixin):
+    """Store pairwise relationships between objects.
 
-    A `DissimilarityMatrix` instance stores a square, hollow, two-dimensional
-    matrix of dissimilarities between objects. Objects could be, for example,
+    A `PairwiseMatrix` instance stores a square, two-dimensional
+    matrix of relationships between objects. Objects could be, for example,
     samples or DNA sequences. A sequence of IDs accompanies the
-    dissimilarities.
+    relationships.
 
-    Methods are provided to load and save dissimilarity matrices from/to disk,
-    as well as perform common operations such as extracting dissimilarities
+    Methods are provided to load and save pairwise matrices from/to disk,
+    as well as perform common operations such as extracting values
     based on object ID. Additionally, the
     :meth:`plot` method provides
     convenient built-in plotting functionality.
 
     Parameters
     ----------
-    data : array_like or DissimilarityMatrix
-        Square, hollow, two-dimensional ``numpy.ndarray`` of dissimilarities
+    data : array_like or PairwiseMatrix
+        Square, two-dimensional ``numpy.ndarray`` of pairwise relationships
         (floats), or a structure that can be converted to a ``numpy.ndarray``
-        using ``numpy.asarray`` or a one-dimensional vector of dissimilarities
+        using ``numpy.asarray`` or a one-dimensional vector of pairwise relationships
         (floats), as defined by `scipy.spatial.distance.squareform`. Can
-        instead be a `DissimilarityMatrix` (or subclass) instance,
+        instead be a `PairwiseMatrix` (or subclass) instance,
         in which case the instance's data will be used.
         Data will be converted to a float ``dtype`` if necessary. A copy will
         *not* be made if already a ``numpy.ndarray`` with a float ``dtype``.
@@ -95,7 +95,7 @@ class DissimilarityMatrix(SkbioObject, PlottableMixin):
         starting from zero, e.g., ``('0', '1', '2', '3', ...)``.
     validate : bool, optional
         If `validate` is ``True`` (the default) and data is not a
-        DissimilarityMatrix object, the input data will be validated.
+        PairwiseMatrix object, the input data will be validated.
 
     See Also
     --------
@@ -104,9 +104,9 @@ class DissimilarityMatrix(SkbioObject, PlottableMixin):
 
     Notes
     -----
-    The dissimilarities are stored in redundant (square-form) format [1]_.
+    The values are stored in redundant (square-form) format [1]_.
 
-    The data are not checked for symmetry, nor guaranteed/assumed to be
+    The data are not checked for symmetry, hollowness, nor guaranteed/assumed to be
     symmetric.
 
     References
@@ -128,7 +128,7 @@ class DissimilarityMatrix(SkbioObject, PlottableMixin):
             np.ndarray,
             Sequence[float],
             Sequence[Sequence[float]],
-            "DissimilarityMatrix",
+            "PairwiseMatrix",
         ],
         ids: Optional[Sequence[str]] = None,
         validate: bool = True,
@@ -137,7 +137,7 @@ class DissimilarityMatrix(SkbioObject, PlottableMixin):
         validate_shape = False
         validate_ids = False
 
-        if isinstance(data, DissimilarityMatrix):
+        if isinstance(data, PairwiseMatrix):
             if isinstance(data, self.__class__):
                 # Never validate when copying from an object
                 # of the same type
@@ -212,8 +212,9 @@ class DissimilarityMatrix(SkbioObject, PlottableMixin):
         metric: Callable,
         key: Optional[Any] = None,
         keys: Optional[Iterable[Any]] = None,
-    ) -> "DissimilarityMatrix":
-        """Create DissimilarityMatrix from an iterable given a metric.
+        validate: bool = True,
+    ) -> "PairwiseMatrix":
+        """Create PairwiseMatrix from an iterable given a metric.
 
         Parameters
         ----------
@@ -231,10 +232,17 @@ class DissimilarityMatrix(SkbioObject, PlottableMixin):
         keys : iterable, optional
             An iterable of the same length as `iterable`. Each element will be
             used as the respective key.
+        validate : boolean, optional
+            If ``True``, all pairwise relationships are computed, including upper
+            and lower triangles and the diagonal. If ``False``, `metric` is
+            assumed to be hollow and symmetric and only the lower triangle
+            (excluding the diagonal) is computed. Pass ``validate=False`` if
+            you are sure `metric` is hollow and symmetric for improved
+            performance.
 
         Returns
         -------
-        DissimilarityMatrix
+        PairwiseMatrix
             The `metric` applied to all pairwise elements in the `iterable`.
 
         Raises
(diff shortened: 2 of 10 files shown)


skbio/stats/distance/_base.py
```diff
@@ -253,12 +261,18 @@ def from_iterable(
...
-                dm[i, j] = metric(a, b)
-
-        return cls(dm, keys_)  # type: ignore[operator]
+        if validate:
+            dm = np.empty((len(iterable),) * 2)
+            for i, a in enumerate(iterable):
+                for j, b in enumerate(iterable):
+                    dm[i, j] = metric(a, b)
+            return cls(dm, keys_)  # type: ignore[operator]
+        else:
```
qiyunzhu: The following process will be faster if we do `np.empty` instead of `np.zeros`, then fill the diagonal with 0 using `np.fill_diagonal`.

skbio/stats/distance/_base.py
```diff
@@ -1039,23 +1051,126 @@ def _is_id_pair(self, index: tuple) -> bool:
...
+
+        if not data_sym:
+            raise DistanceMatrixError("Data must be symmetric and cannot contain NaNs.")
+
+    def permute(
+        self,
+        condensed: bool = False,
+        seed: Optional["SeedLike"] = None,
+    ) -> Union["SymmetricMatrix", np.ndarray]:
+        """Randomly permute both rows and columns in the matrix.
```
qiyunzhu: Missing an `r` before the docstring. This `r` will render the docstring correctly.

skbio/stats/distance/_base.py
```diff
@@ -361,7 +361,7 @@ def size(self) -> int:
 
         """
         if self._flags["VECTOR"]:
-            return self.shape[0] * self.shape[1]
+            return self.shape[0] ** 2
```
qiyunzhu: Is this correct? I guess `shape` is the same regardless of vector or square form, whereas `size` is the number of elements actually stored in the array, which is n(n-1)/2 != 2n.

skbio/stats/distance/_base.py
```diff
@@ -1660,9 +1689,9 @@ def _build_results(
     )
 
 
-def _vec_to_size(vec_length: int) -> float:
+def _vec_to_size(vec: int) -> float:
     """Calculate the redundant size of a matrix given its condensed vector."""
-    return int((1 + np.sqrt(1 + 8 * vec_length)) / 2)
+    return int((1 + np.sqrt(1 + 8 * len(vec))) / 2)
```
qiyunzhu, line 2230: The return value is int whereas the function signature is float. Also, I wonder if the loss of floating-point precision can cause `np.sqrt(a**2) != a`. I tested a few and spotted none. But it will be great if we can get a firm answer from elsewhere that the current code is indeed the recommended way of calculating the size of a vector form matrix.

skbio/stats/distance/_base.py
mataton: Do we want size to reflect the total number of elements in the redundant form matrix? Or do we want it to reflect the number of elements in the condensed array? The way I have it written assumes that the user would want to know the total number of elements in the redundant form matrix, `n^2`.

mataton:
@sfiligoi @jianshu93 Your reviews on this PR would be appreciated! It is still work in progress, but I feel it is ready for a first round of review.

skbio/stats/distance/_base.py
qiyunzhu: I tend to think `size` should be the size of the underlying array in the memory. But I am not sure...

skbio/stats/distance/_base.py
```diff
@@ -1558,3 +1914,131 @@ def _build_results(
...
+
+    Returns
+    -------
+    int
+        Index in the condensed form vector.
+    """
+    # can do this because of symmetry
+    if i > j:
+        i, j = j, i
+    return i * n + j - ((i + 2) * (i + 1)) // 2
```
qiyunzhu: If this method needs to be called many times, we should find a way to vectorize it.

skbio/stats/distance/_base.py
```diff
@@ -176,34 +197,20 @@ def __init__(
...
-            ids = tuple(str(i) for i in range(data_.shape[0]))
-            # I just created the ids, so no need to re-validate them
-            validate_ids = False
-        ids = tuple(ids)
-
-        if validate_full:
-            self._validate(data_, ids)
+    def _generate_ids(self, data):
+        if data.ndim != 1:
+            return tuple(str(i) for i in range(data.shape[0]))
```
wasade, line 202: Strings which look like numbers lead to surprising bugs later. Can this instead be `ID-%d` or something?

skbio/stats/distance/_base.py
```diff
@@ -176,34 +197,20 @@ def __init__(
...
         else:
-            if validate_shape:
-                self._validate_shape(data_)
-            if validate_ids:
-                self._validate_ids(data_, ids)
+            return tuple(str(i) for i in range(_vec_to_shape(data)))
```
wasade, line 204: Same here

skbio/stats/distance/_base.py
wasade: The distinction here is I think between `shape` and `size`. Size should be the number of elements represented, where as `shape` is the logical size of the matrix. Does that make sense?

skbio/stats/distance/_base.py
```diff
@@ -334,25 +348,26 @@ def T(self) -> "DissimilarityMatrix":
...
-        DissimilarityMatrix
-            Transpose of the dissimilarity matrix. Will be the same type as
-            `self`.
+        PairwiseMatrix
+            Transpose of the matrix. Will be the same type as ``self``.
 
         """
         # Note: Skip validation, since we assume self was already validated
-        return self.__class__(self.data.T.copy(), deepcopy(self.ids), validate=False)
+        return self.__class__(self._data.T.copy(), deepcopy(self.ids), validate=False)
```
wasade: Code duplication can be reduced some with:

```suggestion
        return self._copy(transpose=True)
```

skbio/stats/distance/_base.py
```diff
@@ -398,44 +410,47 @@ def redundant_form(self) -> np.ndarray:
...
 
         """
         # We deepcopy IDs in case the tuple contains mutable objects at some
         # point in the future.
         # Note: Skip validation, since we assume self was already validated
-        return self.__class__(self.data.copy(), deepcopy(self.ids), validate=False)
+        # We deepcopy IDs in case the tuple contains mutable objects at some
+        # point in the future.
+        # Note: Skip validation, since we assume self was already validated
+        return self.__class__(self._data.copy(), deepcopy(self.ids), validate=False)
```
wasade: ...related to suggestion in `transpose()`

```suggestion
        return self._copy(transpose=False)
        
    def _copy(self, transpose: bool):
        """Copy support.
        
        Parameters
        ----------
        transpose : bool
            Transpose the data on copy.
            
        Returns
        -------
        PairwiseMatrix
            Deep copy of the matrix. Will be the same type as `self`.
        """
        data = self._data.copy()
        if transpose:
            data = data.T
        return self.__class__(self, data, deepcopy(self.ids), validate=False)
        
```

skbio/stats/distance/_base.py
```diff
@@ -700,33 +753,31 @@ def _subset_to_dataframe(
     def plot(
```
wasade, line 740: Should this be deleted? It's only useful for tiny distance matrices

skbio/stats/distance/_base.py
```diff
@@ -1155,20 +1350,54 @@ def from_iterable(
...
-        for i, a in enumerate(iterable):
-            for j, b in enumerate(iterable[:i]):
-                dm[i, j] = dm[j, i] = metric(a, b)
+        dm = np.empty((len(iterable),) * 2)
+        if diagonal is not None:
+            np.fill_diagonal(dm, diagonal)
+        else:
+            np.fill_diagonal(dm, 0.0)
+        if validate:
+            for i, a in enumerate(iterable):
```
wasade, line 1417: The nested for loops won't work if the iterable is a generator. 

Why is the `dm` being created within `validate`?

wasade: ...same comment for the nested for loop on line 1366

Could these be pushed to one of the pw distance methods of scipy?

skbio/stats/distance/_base.py
```diff
@@ -1229,31 +1465,155 @@ def permute(
...
+                permuted_condensed = distmat_reorder_condensed_py(self._data, order)
+            else:
+                permuted_condensed = distmat_reorder_condensed(self._data, order)
             return permuted_condensed
         else:
             # Note: Skip validation, since we assume self was already validated
-            permuted = distmat_reorder(self._data, order)
-            return self.__class__(permuted, self.ids, validate=False)
+            if self._flags["VECTOR"]:
+                permuted = distmat_reorder_condensed_py(self._data, order)
```
wasade, line 1603: Is the use of reorder correct here? Won't this produce an isomorphic distance matrix rather than the _same_ distance matrix with only the IDs permuted as would be used for a Monte Carlo procedure? Maybe I'm misunderstanding the intent of this method or something here

skbio/stats/distance/_base.py
mataton: I agree the `ID-%d` approach is better, but the old `DistanceMatrix` class created IDs this way, so it will be a change from existing public facing components. Not sure how big a deal that is to the users though.

skbio/stats/distance/_base.py
mataton: Changed to make `size` reflect total number of elements represented.

skbio/stats/distance/_base.py
mataton: Here's the mathematical justification:
Screenshot 2025-10-01 113841


It works as long as the vector is indeed the correct length (binomial coefficient). It could be the case that someone could do:
```python
sm = SymmetricMatrix([1, 2, 3, 4], validate=False, condensed=True)
```
That object could be created. But I'm not sure if we want to throw errors here if the user is explicitly setting `validate=False`?

skbio/stats/distance/_base.py
qiyunzhu: A long time ago typing `dm` will automatically render the plot. In a previous version we muted this such that the user will not accidentally find their computer stuck rendering a huge distance matrix. But it is still good to let the user plot if they want.

skbio/stats/distance/_base.py
```diff
@@ -43,71 +43,82 @@
...
-        *not* be made if already a ``numpy.ndarray`` with a float ``dtype``.
+        Data will be converted to a float `dtype` if necessary. A copy will
+        *not* be made if already a `numpy.ndarray` with a float `dtype`.
     ids : sequence of str, optional
         Sequence of strings to be used as object IDs. Must match the number of
-        rows/cols in `data`. If ``None`` (the default), IDs will be
+        rows/cols in `data`. If `None` (the default), IDs will be
         monotonically-increasing integers cast as strings, with numbering
-        starting from zero, e.g., ``('0', '1', '2', '3', ...)``.
+        starting from zero, e.g., `('0', '1', '2', '3', ...)`.
```
qiyunzhu: Why this and below were changed from double ticks into single ticks?

When a docstring is prepended by "r", it will be rendered as [reStructuredText](https://docutils.sourceforge.io/rst.html) (RST), in which code blocks should be wrapped by double ticks, whereas single ticks represent italic instead.

Also, I think values like None, True, False don't need to be in a code block.

skbio/stats/distance/_base.py
```diff
@@ -684,12 +707,42 @@ def _subset_to_dataframe(
         # included here so that np.hstack works in the event that either i_ids
         # or j_ids is empty.
         values = [np.array([])]
-        for i_idx in i_indices:
-            i.extend([self.ids[i_idx]] * j_length)
-            j.extend(j_labels)
+        if self._flags["CONDENSED"]:
```
qiyunzhu: Is this code under `PairwiseMatrix` or `SymmetricMatrix`? I assume the later, because the former doesn't have a condensed form. Let me know if I am wrong.

skbio/stats/distance/_base.py
```diff
@@ -684,12 +707,42 @@ def _subset_to_dataframe(
...
+            diagonal_val = getattr(self, "_diagonal", 0.0)
+            n = self.shape[0]
+
+            # precompute to avoid repeated calculations
```
qiyunzhu: This part of code looks inefficient to me. But I don't immediately have a better idea either. Dictionary is less efficient than an array of indices, if you can. But since the input indices are merely a subset of all indices, dictionary might be more justified. However, I guess there is a way that utilizes NumPy fancy indexing to accelerate this. Maybe this can be saved to a future pull request.

skbio/stats/distance/_base.py
```diff
@@ -1067,33 +1116,183 @@ class DistanceMatrix(DissimilarityMatrix):
...
+
+    def _validate_diagonal(
+        self, data: np.ndarray, diagonal: Optional[np.ndarray] = None
+    ) -> None:
+        """Validate the diagonal of the matrix."""
+        if diagonal is not None:
+            # if it's a single value, it doesn't need to be validated
+            if not np.isscalar(diagonal):
+                diagonal = np.asarray(diagonal)
+                # if it's a nd.array it needs to be 1d
```
qiyunzhu, line 1245: I think you meant "ndarray" or "np.array".

skbio/stats/distance/_base.py
```diff
@@ -1558,3 +1956,131 @@ def _build_results(
...
+    n_filtered = len(reorder_vec)
+
+    out_size = n_filtered * (n_filtered - 1) // 2
+    if out_size == 0:
+        return np.array([], dtype=in_mat.dtype)
+
+    old_indices = np.zeros(out_size, dtype=np.intp)
+    out_idx = 0
+
+    for i in range(n_filtered - 1):
```
qiyunzhu: There is much room for future optimization.

skbio/stats/distance/_base.py
```diff
@@ -1558,3 +1956,131 @@ def _build_results(
...
+    condensed_data: np.ndarray, i: int, j: int, n: int
+) -> float:
+    """Get a single element from condensed storage.
+
+    Parameters
+    ----------
+    condensed_data : np.ndarray
+        1-dimensional vector form of the matrix.
+    i : int
+        Row index
```
qiyunzhu: The line should end with a period.

skbio/stats/distance/tests/test_base.py
```diff
@@ -847,15 +934,36 @@ def test_from_iterable_with_key(self):
...
+        iterable = (x for x in range(4))
+
+        exp = self.matobj([[0, 1, 2, 3],
+                           [1, 0, 1, 2],
+                           [2, 1, 0, 1],
+                           [3, 2, 1, 0]], ['0', '1', '4', '9'], condensed=True)
+        res = self.matobj.from_iterable(iterable, lambda a, b: abs(b - a),
+                                        key=lambda x: str(x**2),
+                                        condensed=True)
+        self.assertEqual(res, exp)
```
qiyunzhu, line 1398: I wonder if `assertEqual` will work as intended since the two objects are matrix classes. I have no problem if it does.

mataton: Indeed `assertEqual` will use the matrix class's own `__eq__` method under the hood, which we have defined with our own desired behavior, so it should be okay.

skbio/stats/distance/_base.py
mataton: I moved the code now, so `PairwiseMatrix` only has the original version, the modified version now exists in `SymmetricMatrix`.

skbio/stats/distance/_base.py
mataton: The nested for loops are the current approach, and I believe they actually do work if iterable is a generator. See this unit test which passes:
```python
    def test_from_iterable_with_keys(self):
        iterable = (x for x in range(4))

        exp = self.matobj([[0, 1, 2, 3],
                           [1, 0, 1, 2],
                           [2, 1, 0, 1],
                           [3, 2, 1, 0]], ['0', '1', '4', '9'])
        res = self.matobj.from_iterable(iterable,
                                        lambda a, b: abs(b - a),
                                        keys=iter(['0', '1', '4', '9'])
                                        )
        self.assertEqual(res, exp)
```

I would be in favor of using `pdist` from scipy, but we'd need to convert all input to 2D for that to work, because `pdist` throws an error if the input isn't 2D.

mataton: Notably, `pdist` itself doesn't accept generators as input.

mataton: For the above reasons I'm in favor of keeping the current approach.

skbio/stats/distance/_base.py
mataton: It does indeed produce an isomorphic distance matrix instead of the same distance matrix with IDs permuted. Speaking to the intent of the method, it looks very much like this was the original desired behavior. See this note in the original docstring:  `"This method does not modify the distance matrix that it is called on."` as well as the use of `assertFalse` these unit tests, which expect that the result of `permute` is indeed a completely different object than the original input matrix.

```python
    def test_permute_not_condensed(self):
        obs = self.dm_1x1.permute()
        self.assertEqual(obs, self.dm_1x1)
        self.assertFalse(obs is self.dm_1x1)

        obs = self.dm_2x2.permute()
        self.assertEqual(obs, self.dm_2x2)
        self.assertFalse(obs is self.dm_2x2)
```

My approach generally is to preserve existing functionality as much as possible, because I personally may not have the requisite statistical/mathematical knowledge to know if changing the functionality will break things or not.

skbio/io/format/lsmat.py
```diff
@@ -107,17 +109,27 @@ def _lsmat_sniffer(fh):
 
 
 @lsmat.reader(PairwiseMatrix)
-def _lsmat_to_dissimilarity_matrix(fh, delimiter="\t"):
+def _lsmat_to_pairwise_matrix(fh, delimiter="\t"):
     return _lsmat_to_matrix(PairwiseMatrix, fh, delimiter)
 
 
+@lsmat.reader(SymmetricMatrix)
```
qiyunzhu, line 116: I was exploring the `io` module and noticed an issue: It seems that one has to specify the reader for each and every class, even if some classes are inherited from other classes. For example, if `PairwiseMatrix` already has a working `@lsmat.reader`, I would expect that `SymmetricMatrix` and `DistanceMatrix` would inherit it without needing to explicitly specifying their own readers. However, this isn't the case in the current scikit-bio. This not only makes code redundant, but also causes a problem that if a user subclasses a scikit-bio class, their subclass won't have the reader and writer interface. Is my understanding correct?

mataton: Indeed that is correct. For any class regardless of inheritance, for the current io system to be able to handle it it must have explicitly defined reader and writers. It's a pretty big bottleneck to extensibility.

qiyunzhu: After doing some research, I think that this can be achieved by modifying the [`_get_rw` method](https://github.com/scikit-bio/scikit-bio/blob/f863136d2c4f1b17c35100ea383c59682d2404c3/skbio/io/registry.py#L328) of the IO registry. Specifically, one can recursively trace the parent class of the current class (using `cls.__bases__`) until reaching `SkbioObject`. Would you like to try this in a separate PR?

mataton: Sure, let me wrap up the documentation on this PR and then I'll check that out.

PR #2301, opened 2025-10-14
Enable inheritance of IO methods on scikit-bio objects

mataton:
Currently, the way scikit-bio's IO system is set up means that each class must have explicitly defined readers and writers in the IO registry for the IO operations to work. In effect, this means that a subclass of a scikit-bio object will **not** inherit the IO operations of the parent class. This is a big problem for extensibility, both for users and developers.

With this PR you can subclass any scikit-bio class and the IO operations (readers, writers, and sniffers) will be inherited. This means you can now define your own custom classes on scikit-bio objects and not have to worry about the IO portions!

### DIFF
--- a/skbio/io/format/newick.py
+++ b/skbio/io/format/newick.py
@@ -268,7 +268,7 @@ def _newick_sniffer(fh):
 
 
 @newick.reader(TreeNode)
-def _newick_to_tree_node(fh, convert_underscores=True, cls=None):
+def _newick_to_tree_node(fh, cls=None, convert_underscores=True):
     if cls is None:
         cls = TreeNode
     tree_stack = []


skbio/io/format/binary_dm.py
```diff
@@ -119,18 +119,24 @@ def _binary_dm_sniffer(fh):
 
 
 @binary_dm.reader(PairwiseMatrix)
-def _binary_dm_to_pairwise(fh):
-    return _h5py_mat_to_skbio_mat_stream(PairwiseMatrix, fh)
+def _binary_dm_to_pairwise(fh, cls=None):
+    if cls is None:
```
qiyunzhu, line 123: Is there a chance that the function is called without specifying `cls`? If not, `cls` can be a required parameter.

mataton: In the case that a user calls a reader from the registry then `cls` will always be specified. But there is a very large number of existing unit tests which call these functions directly, in which case `cls` may not be specified, so I think it's good to keep `cls=None` to avoid refactoring a bunch of unit tests.

PR #2335, opened 2025-11-12
Enable condensed operations with `permanova`

mataton:
This code enables `permanova` to be run directly on a condensed form distance matrix. Similarly to the modifications to `mantel` the only change was to use condensed form indexing instead of the old redundant form indexing.

### DIFF
--- a/skbio/stats/distance/_cutils.pyx
+++ b/skbio/stats/distance/_cutils.pyx
@@ -435,7 +435,7 @@ def permanova_f_stat_sW_condensed_cy(TReal[::1] distance_matrix,
     cdef double local_s_W
     cdef double val
 
-    cdef Py_ssize_t row, col, rowi, coli
+    cdef Py_ssize_t row, col, rowi, coli, row1
     cdef Py_ssize_t in_n_2 = n//2
 
     for rowi in prange(in_n_2, nogil=True):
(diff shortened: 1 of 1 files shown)


skbio/stats/distance/_cutils.pyx
```diff
@@ -307,6 +341,58 @@ def permanova_f_stat_sW_cy(TReal[:, ::1] distance_matrix,
...
+    cdef Py_ssize_t row, col, rowi, coli
+    cdef Py_ssize_t in_n_2 = n//2
+
+    for rowi in prange(in_n_2, nogil=True):
+        # since columns get shorter, combine first and last
+        row=rowi
+        local_s_W = 0.0
+        group_idx = grouping[row]
+        for coli in range(n-row-1):
+            col = coli+row+1
```
qiyunzhu: Can pre-calculate `row + 1` such that you can save one operation per loop.

skbio/stats/distance/_cutils.pyx
```diff
@@ -307,6 +341,58 @@ def permanova_f_stat_sW_cy(TReal[:, ::1] distance_matrix,
...
+            if grouping[col] == group_idx:
+                idx_ = condensed_index(row, col, n)
+                val = distance_matrix[idx_]
+                local_s_W = local_s_W + val * val
```
qiyunzhu: Can be `local_s_W += val * val` to save one temporary variable.

skbio/stats/distance/_cutils.pyx
```diff
@@ -307,6 +341,58 @@ def permanova_f_stat_sW_cy(TReal[:, ::1] distance_matrix,
...
+        row = n-rowi-2
+        if row!=rowi: # don't double count
+            local_s_W = 0.0
+            group_idx = grouping[row]
+            for coli in range(n-row-1):
+                col = coli+row+1
+                if grouping[col] == group_idx:
+                    idx_ = condensed_index(row, col, n)
+                    val = distance_matrix[idx_]
+                    local_s_W = local_s_W + val * val
```
qiyunzhu: Looks like `val` is always squared. In this case, do `distance_matrix **= 2` in NumPy before feeding into Cython can save compute. But this will destroy original matrix data. Since you are doing permutation, it should probably be fine. Just a thought. No need to change.

PR #2337, opened 2025-11-19
Implement PCA in stats.ordination

aj-pinto:

* [X] I have read the [contribution guidelines](https://scikit.bio/contribute.html).


skbio/stats/ordination/tests/test_principal_component_analysis.py
```diff
@@ -0,0 +1,190 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2013--, scikit-bio development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import numpy as np
```
qiyunzhu: As a convention, the import of third-party packages (like `numpy`) should follow built-in packages (like `unittest`) and prior to `skbio` modules.

skbio/stats/ordination/__init__.py
```diff
@@ -26,6 +26,13 @@
    pcoa
    pcoa_biplot
 
+Dimensionality reduction
```
qiyunzhu, line 29: A new section "Dimensionality reduction" sounds a good idea to me. Although this requires more considerations, which can be done after merging this PR. Currently, the sections are "Multidimensional scaling" (methods that start with a distance matrix), "Correspondence analysis" (methods that takes a contingency table as input), "Canonical analysis" (methods that combine two tables). So it seems that "Dimensionality reduction" can describe a new category of methods that takes any table (not necessarily contingency table) as input.

skbio/stats/ordination/_principal_component_analysis.py
```diff
@@ -0,0 +1,207 @@
...
+    of the data along that direction.
+    sigma_i^2 = var(X_centered**w**_i)
+
+    Parameters
+    ----------
+    X : table_like
+        Samples by features table (n, p). See :ref:`supported formats <table_like>`.
+    method : str, optional
+        Matrix decomposition method to use. Default is
+        ####################################################          for review \/
```
qiyunzhu: This is probably what caused the documentation CI workflow to fail.

skbio/stats/ordination/_principal_component_analysis.py
```diff
@@ -0,0 +1,207 @@
...
+    *p* features. Each sample is then a row of the *n*x*p* data matrix *X*,
+    whose columns are the features of the data.
+    The goal of PCA is to find directions **w**_i in feature space along which
+    the variance of the samples is maximized, the so-called principle components
+    of the data matrix:
+    max_**w**_i var(X_centered**w**_i)
+
+    The principle components of *X* are the unit eigenvectors of the
+    covariance matrix:
+    C = 1/(n-1) X_centered^T X_centered    < can use sigma_XX instead
```
qiyunzhu: This should be a nicely rendering Latex equation. Use of `X_centered` (which is a nice Python variable) might be replaced with mathematical symbols.

skbio/stats/ordination/_principal_component_analysis.py
```diff
@@ -0,0 +1,207 @@
...
+import numpy as np
+from scipy.linalg import eigh, svd
+from scipy.sparse.linalg import svds
+from ._ordination_results import OrdinationResults
+from skbio.table._tabular import _create_table, _create_table_1d, _ingest_table
+
+
+def pca(
+    X,
+    method="svd",
```
qiyunzhu: I am thinking that the default method should be 'eigh', just to be consistent with `pcoa`, and also because sklearn's PCA defaults to SVD, and our method can be different by default.

skbio/stats/ordination/_principal_component_analysis.py
```diff
@@ -0,0 +1,207 @@
...
+       into principal components. Journal of Educational Psychology, 24(6),
+       417–441, 498–520.
+
+    """
+
+    # Ingestion of input data matrix
+    X, row_ids, column_ids = _ingest_table(X, sample_ids, feature_ids)
+
+    # Center the data matrix
+    X_centered = X - np.mean(X, axis=0)
```
qiyunzhu, line 155: I am not sure if we should do this as part of the function. The responsibility of data preprocessing should be on the user. Standardization like `X_std = (X - np.mean(X, axis=0)) / np.std(X, axis=0)` (can use SciPy's `zscore` or sklearn's `StandardScaler` instead) is common but not always executed, depending on the statistical properties of the input data. This should be the user's freedom. Maybe note that in the documentation.

skbio/stats/ordination/_principal_component_analysis.py
```diff
@@ -0,0 +1,207 @@
...
+    elif method == "svd":
+        # The eigenvalues of the covariance matrix are the right singular values
+        # of the centered data matrix scaled by the total degrees of freedom
+        # (number of samples - 1)
+        if dimensions is None:
+            # Perform dense SVD with LAPACK if dimension is not specified
+            _, singular_values, components = svd(X_centered)
+        else:
+            # Perform sparse SVD with ARPACK if dimension is specified
+            _, singular_values, components = svds(X_centered, k=dimensions)
```
qiyunzhu, line 214: Interesting! I didn't know `svds`. Maybe this can be used for PCoA too.

skbio/stats/ordination/_principal_component_analysis.py
aj-pinto: As I understand, centering the data matrix is a necessary step to computing the covariance matrix, which would be different from standardizing by standard deviations?

If the matrix has already been centered by the user, then this would have no effect. If it has not been centered, the function will not work properly.

qiyunzhu: @mortonjt Do you know?

aj-pinto:
The documentation should be in a working state now, with properly rendered LaTeX. If that much mathematical detail is unnecessary, it can be cut. I also wasn't sure if it was preferred for some parts to be rendered with italics rather than LaTeX.

I added an "iterative" parameter that makes the function use SciPy's sparse "eigsh" or "svds" methods instead of the dense "eigh" or "svd" methods. It is False by default. I haven't added unit testing for that yet in case it would be changed, although I've confirmed it works.

qiyunzhu:
Let's merge! To ship `pca` elegantly, we should also implement `rclr` (#2366) and `rpca` (#2363). Otherwise people will question what's new in our offering vs scikit-learn. @aj-pinto I think the next step is to get the `rclr` PR done.

Issue #2342, opened 2025-11-28
Constraint the number of parallel processes?


reox:
I'm running a PERMANOVA test on a shared server machine with 128 CPUs. Apparently, scikit-bio assumes it should use all CPUs - which is bad, as I have to keep my number of parallel processes low, as other people are working on that machine too.
I wasn't able to find a parameter that can be set to reduce the number of parallel threads for that function and there are so many different ways for parallelism in python, I do not want to try each and every environment variable that it may could be ;)

I'm using skbio 0.7.0 with Python 3.9.
Here is a mmve:
```python
import numpy as np
from skbio.stats.distance import permanova, DistanceMatrix
from scipy.spatial import distance_matrix

x = np.array([[ 0.34648083,  0.06266357, -0.40914441],
[ 0.34010614, -0.09446391, -0.24564223],
[ 0.38013652, -0.16329463, -0.2168419 ],
[ 0.34041802,  0.04336297, -0.38378099],
[ 0.34514699, -0.11880025, -0.22634674],
[ 0.37161071, -0.08090807, -0.29070265],
[ 0.26388796,  0.04569214, -0.3095801 ],
[ 0.34224386,  0.02099661, -0.36324047],
[ 0.27958363, -0.02720403, -0.2523796 ],
[ 0.32250293,  0.01471942, -0.33722235]])
permanova(DistanceMatrix(distance_matrix(x, x)), ['a'] * 5 + ['b'] * 5)
```

reox:
Thanks! Where in the documentation did you find that?

<img width="1226" height="293" alt="Image" src="https://github.com/user-attachments/assets/7be645fb-4ffa-45d8-9890-793c1244d0b3" />

That works fine and it also way faster now. There must be significant overhead when running with 128 threads instead with just 4...

PR #2343, opened 2025-12-05
Improve documentation for parallelization acceleration techniques

mataton:
In light of #2342, this PR improves documentation for parallelization techniques. It also creates links between the documentation and installation page, enabling keyword search to work more effectively.

### DIFF
--- a/web/install.rst
+++ b/web/install.rst
@@ -110,7 +110,7 @@ Many scikit-bio functions automatically take advantage of parallel processing to
     from skbio import some_function
 
     # Alternatively, set it before launching your script.
-    export OMP_NUM_THREADS=4 python script.py
+    OMP_NUM_THREADS=4 python script.py
 
 If more granular control over thread use is desired, we recommend using `threadpoolctl <https://github.com/joblib/threadpoolctl>`_::
 


web/install.rst
```diff
@@ -92,13 +92,28 @@ For a more robust verification of the installation you may run the scikit-bio's
...
 
-Many scikit-bio functions automatically take advantage of parallel processing to improve performance. By default, they utilize all available CPU cores when possible. There is currently no per-function parameter to control the number of threads used. You can set the environment variable ``OMP_NUM_THREADS`` to control the global parallelization behavior, or use `threadpoolctl <https://github.com/joblib/threadpoolctl>`_ to achieve more granular control within Python code.
+Many scikit-bio functions automatically take advantage of parallel processing to improve performance. By default, they utilize all available CPU cores when possible. There is currently no per-function parameter to control the number of threads used. You can set the environment variable ``OMP_NUM_THREADS`` to control the global parallelization behavior::
 
-Multiple compute-intensive algorithms in scikit-bio are implemented in Cython and use OpenMP for parallel execution. If your system does not support OpenMP, or if you prefer to disable it, you can build from the scikit-bio source code with OpenMP disabled:
+    # Setting OMP_NUM_THREADS inside a script
+    import os
+    os.environ["OMP_NUM_THREADS"] = "4"
```
qiyunzhu, line 109: There is a trick: This `os.environ` command works only when it is executed before importing the parallelizable function. So if you do:

```python
import os
import numpy as np
os.environ["OMP_NUM_THREADS"] = "4"
...
```

Then it won't work for NumPy functions.

Therefore, to indicate this, we may add `from skbio import some_function` underneath this line.

web/install.rst
```diff
@@ -92,13 +92,28 @@ For a more robust verification of the installation you may run the scikit-bio's
...
+
+If more granular control over thread use is desired, we recommend using `threadpoolctl <https://github.com/joblib/threadpoolctl>`_::
+
+    from skbio import some_function
+    from threadpoolctl import thread_limits
+
+    with thread_limits(limits=1):
+        some_function()
+
+Multiple compute-intensive algorithms in scikit-bio are implemented in Cython and use OpenMP for parallel execution. If your system does not support OpenMP, or if you prefer to disable it, you can build from the scikit-bio source code with OpenMP disabled::
```
qiyunzhu: `OMP_NUM_THREADS` is specifically for controlling OpenMP. Therefore, I think that we should mention OpenMP earlier.

skbio/stats/distance/_base.py
```diff
@@ -508,6 +508,12 @@ def filter(
         MissingIDError
             If an ID in ``ids`` is not in the object's list of IDs.
 
+        Notes
+        -----
+        This function uses parallel computation for improved performance.
```
qiyunzhu, line 514: This note looks good. In the future, there might be two improvements: 1) Store this entire sentence somewhere as a constant, which can be re-used in individual documentation pages. This will avoid copying & pasting the same sentence over and over. 2) Some styled text can make it more standing out to the user. For example SciPy [has a table](https://docs.scipy.org/doc/scipy/reference/generated/scipy.cluster.hierarchy.linkage.html) to indicate Python array API support. In our case, the easiest approach is to use the "..note::" box in Sphinx.

qiyunzhu (changes requested):

web/install.rst
```diff
@@ -92,13 +92,28 @@ For a more robust verification of the installation you may run the scikit-bio's
...
+
+    # Alternatively, set it before launching your script
+    export OMP_NUM_THREADS=4 python script.py
+
+If more granular control over thread use is desired, we recommend using `threadpoolctl <https://github.com/joblib/threadpoolctl>`_::
+
+    from skbio import some_function
+    from threadpoolctl import thread_limits
+
+    with thread_limits(limits=1):
```
qiyunzhu: It is probably better to do `limits=4` for consistency with the previous example.

PR #2347, opened 2025-12-12
Prevent garbage values in `DistanceMatrix` diagonal

mataton:
As pointed out in [this scikit-bio-tutorials issue](https://github.com/scikit-bio/scikit-bio-tutorials/issues/48), the `from_iterable` methods of the `DistanceMatrix` and related classes were filling the diagonals with garbage values. This PR replaces the garbage values with zeros by using `np.zeros` instead of `np.empty`.

### DIFF
--- a/skbio/stats/distance/_base.py
+++ b/skbio/stats/distance/_base.py
@@ -1469,7 +1469,8 @@ class SymmetricMatrix(PairwiseMatrix):
             Store the data in a 2-D redundant form (False, default) or a 1-D condensed
             form (True).
         diagonal : float or array_like, optional
-            Value(s) with which to fill the diagonal of the matrix.
+            Value(s) with which to fill the diagonal of the matrix. Relevant only when
+            ``validate`` is False.
 
         Returns
         -------
--- a/skbio/stats/distance/tests/test_base.py
+++ b/skbio/stats/distance/tests/test_base.py
@@ -1505,6 +1505,23 @@ class DistanceMatrixTestBase(PairwiseMatrixTestData):
         )
         self.assertEqual(res, exp)
 
+    def test_from_iterable_with_invalid_diagonal(self):
+        iterable = (x for x in range(4))
+        with self.assertRaises(DistanceMatrixError) as e:
+            self.matobj.from_iterable(iterable, lambda a, b: a * b, diagonal=1, validate=False)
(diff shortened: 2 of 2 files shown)


skbio/stats/distance/_base.py
```diff
@@ -261,7 +261,7 @@ def from_iterable(
         iterable = list(iterable)
         keys_ = _get_keys(iterable, key, keys)
 
-        dm = np.empty((len(iterable),) * 2)
+        dm = np.zeros((len(iterable),) * 2)
```
qiyunzhu: I am not sure if this is the right fix -- First, for `PairwiseMatrix`, all cell values are filled, therefore there is no need to pre-populate it with zero values. Pre-population is only needed for `SymmetricMatrix` with `validate=False`. Second, instead of pre-populating the entire matrix with 0, which is costly, you can only post-populate the diagonal line with 0 using `np.fill_diagonal`.

qiyunzhu (changes requested):

skbio/stats/distance/_base.py
```diff
@@ -1499,6 +1492,7 @@ def from_iterable(
             for i, a in enumerate(iterable):
                 for j, b in enumerate(iterable[:i]):
                     dm[i, j] = dm[j, i] = metric(a, b)
+            np.fill_diagonal(dm, 0)
```
qiyunzhu: I think it is almost right. But there is one thing: `SymmetricMatrix` has a `diagonal` parameter, which allows users to specify non-zero diagonal values. A more robust fix would be adding the `diagonal` parameter to `SymmetricMatrix.from_iterable`. It is relevant only when `validate=False`.

qiyunzhu (approved):
Looks good to me! Will merge once tests pass.

skbio/stats/distance/tests/test_base.py
```diff
@@ -1505,6 +1505,23 @@ def test_from_iterable_validate_false_condensed(self):
         )
         self.assertEqual(res, exp)
 
+    def test_from_iterable_with_invalid_diagonal(self):
+        iterable = (x for x in range(4))
```
qiyunzhu, line 1509: This can be `iterable = tuple(range(4))` or simply `iterable = range(4)`. But that's trivial.

PR #2357, opened 2025-12-25
Converted all `from skbio import` calls into `from skbio.module import`

qiyunzhu:
Traditionally, scikit-bio's public-facing data structures are implemented in individual (sub)modules, but can be imported directly from `skbio` without specifying the module. For example:

```python
from skbio import DNA, TabularMSA, DistanceMatrix
```

This approach is convenient, but it makes the namespace ambiguous, and creates an obstacle to lazy importing which we intend to implement in the near future (for reference see #2142). Therefore, this PR uniformly converts all `import` calls in code and docstring into ones that specify modules. For example:

```python
from skbio.sequence import DNA
from skbio.alignment import TabularMSA
```

This change does not impact the behavior of scikit-bio. Users can still directly import data structures from `skbio`. No error or warning will be raised. Performance is not impacted. This behavior will not be changed in the near future.

Therefore, this change is mainly for reference purpose. It encourages users and contributors to explicitly specify modules when importing scikit-bio data structures.

===


mataton:
@qiyunzhu I think this one could be updated with `main` and then merged?

PR #2360, opened 2026-01-02
First pass at MMvec

mortonjt:

I've taken a stab at encoding MMvec to be included into skbio. There have been some optimizations that make it faster to run on CPU (i.e. remove autograd to compute analytical gradients, vectorize gradient operations). It is looking like it replicates the main functionality of MMvec per the unittests, so it is ready for additional stress testing.

Will update the changelog once we are in alignment with functionality, documentation and tests.

### DIFF
--- a/skbio/stats/ordination/_mmvec.py
+++ b/skbio/stats/ordination/_mmvec.py
@@ -659,7 +659,9 @@ class MMvecResults(SkbioObject):
         .. math::
 
             Q^2 = 1 - \frac{SS_{res}}{SS_{tot}}
-                = 1 - \frac{\sum(y - \hat{y})^2}{\sum(y - \bar{y})^2}
+                = 1 - \frac{\sum(y - \hat{y})^2}{\sum(y - \bar{y}_j)^2}
+
+        where :math:`\bar{y}_j` is the per-metabolite mean across samples.
 
         Parameters
         ----------
@@ -727,7 +729,7 @@ class MMvecResults(SkbioObject):
 
         # Q² = 1 - SS_res / SS_tot
         ss_res = np.sum((actual - predicted) ** 2)
-        ss_tot = np.sum((actual - actual.mean()) ** 2)
+        ss_tot = np.sum((actual - actual.mean(axis=0)) ** 2)
 
         if ss_tot == 0:
             # All actual values are the same - return 0 to avoid division by zero


skbio/stats/ordination/tests/_simulation.py
```diff
@@ -0,0 +1,165 @@
...
+        Standard deviation of noise in microbe composition.
+    uU : float
+        Mean of microbe embedding distribution.
+    sigmaU : float
+        Standard deviation of microbe embedding distribution.
+    uV : float
+        Mean of metabolite embedding distribution.
+    sigmaV : float
+        Standard deviation of metabolite embedding distribution.
+    seed : int
```
qiyunzhu, line 71: `seed` should use scikit-bio's universal seed interface. An example is [here](https://github.com/scikit-bio/scikit-bio/blob/ce50a6ab94ba847a2988a4304b3bf12861b29175/skbio/stats/distance/_mantel.py#L123).

skbio/stats/ordination/tests/_simulation.py
```diff
@@ -0,0 +1,165 @@
...
+
+import numpy as np
+import pandas as pd
+
+from skbio.stats.composition import clr_inv as softmax
+from skbio.stats.composition import ilr_inv
+from skbio.util import get_rng
+
+
+def random_multimodal(
```
qiyunzhu, line 19: Is this intended to be a public-facing function? Looks like a useful one. Maybe consider adding it and its single-modal version to `skbio.table`.

skbio/stats/multimodal/tests/_simulation.py
```diff
@@ -0,0 +1,165 @@
...
+        Number of latent dimensions for the embedding.
+    low : float
+        Lower bound of the sample gradient.
+    high : float
+        Upper bound of the sample gradient.
+    microbe_total : int
+        Total microbial counts per sample.
+    metabolite_total : int
+        Total metabolite counts per sample.
+    uB : float
```
qiyunzhu: I'd recommend (slightly) more descriptive parameter names over mathematic symbols (except for simple cases like `n` and `k`). Also, we should avoid camel case.

skbio/stats/multimodal/tests/_simulation.py
```diff
@@ -0,0 +1,165 @@
...
+    sigmaU=1,
+    uV=0,
+    sigmaV=1,
+    seed=0,
+):
+    """Generate synthetic microbe-metabolite co-occurrence data.
+
+    Parameters
+    ----------
+    num_microbes : int
```
qiyunzhu: The naming of "number of xxx" parameters didn't have a convention, but now we are trying to uniformize it into "xxx", especially where there is no confusion. For [example](https://github.com/scikit-bio/scikit-bio/blob/ce50a6ab94ba847a2988a4304b3bf12861b29175/skbio/stats/ordination/_principal_coordinate_analysis.py#L39), in PCoA, `number_of_dimensions` has been renamed as `dimensions`. So I would recommend `microbes` here.

skbio/stats/multimodal/tests/_simulation.py
```diff
@@ -0,0 +1,165 @@
...
+    sigmaU : float
+        Standard deviation of microbe embedding distribution.
+    uV : float
+        Mean of metabolite embedding distribution.
+    sigmaV : float
+        Standard deviation of metabolite embedding distribution.
+    seed : int
+        Random seed for reproducibility.
+
+    Returns
```
qiyunzhu: This return value is a very large tuple, and can be confusing if each component can only be referred to with their index. Ideally, we should create a named tuple or data class to store the result. In scikit-bio, there weren't such structures. But we started to implement some. For [example](https://github.com/scikit-bio/scikit-bio/blob/ce50a6ab94ba847a2988a4304b3bf12861b29175/skbio/alignment/_pair.py#L29), the return value of `pair_align` is `class PairAlignResult(NamedTuple)`.

skbio/stats/multimodal/tests/_simulation.py
```diff
@@ -0,0 +1,165 @@
...
+    -------
+    microbe_counts : pd.DataFrame
+        Count table of microbial abundances (num_samples, num_microbes).
```
qiyunzhu: Should be `(n_samples, n_microbes)` instead of `(num_samples, num_microbes)` to be consistent with the scikit-learn convension.

skbio/stats/multimodal/tests/_simulation.py
```diff
@@ -0,0 +1,165 @@
...
+    metabolite_counts : pd.DataFrame
+        Count table of metabolite abundances (num_samples, num_metabolites).
+    X : np.ndarray
```
qiyunzhu: I think we should avoid using `X` to describe a design matrix, because `X` means data in scikit-learn convension.

skbio/stats/ordination/tests/data/cf/lcms_nt.biom
qiyunzhu, line 1: I think we should avoid including binary files in the package, unless absolutely necessary, to avoid security holes (recall the xz case). How about use tsv/csv instead of biom?

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+        ... )
+        >>> predictions = result.predict(new_microbes)
+        >>> predictions.shape
+        (5, 8)
+        >>> np.allclose(predictions.sum(axis=1), 1.0)
+        True
+
+        """
+        # Convert to array if needed
+        if hasattr(microbes, "values"):
```
qiyunzhu, line 623: I see multiple cases of the same pattern as this one. Consider using the existing helper function: `skbio.table._tabular._ingest_table` ([link](https://github.com/scikit-bio/scikit-bio/blob/ce50a6ab94ba847a2988a4304b3bf12861b29175/skbio/table/_tabular.py#L111C5-L111C18)) to conveniently parse any "table-like" inputs (including BIOM, polars, and other array libraries).

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+            X = microbes.values.astype(np.float64)
+            sample_ids = list(microbes.index)
+        else:
+            X = np.asarray(microbes, dtype=np.float64)
+            sample_ids = [f"sample_{i}" for i in range(X.shape[0])]
+
+        # Normalize to proportions
+        row_sums = X.sum(axis=1, keepdims=True)
```
qiyunzhu, line 631: `X.sum(...)` may be re-written as `np.sum(X, ...)` for future compatibility with Python array API.

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+        --------
+        predict : Predict metabolite distributions.
+        probabilities : Get conditional probability matrix.
+
+        Examples
+        --------
+        >>> from skbio.stats.multimodal import mmvec
+        >>> import numpy as np
+        >>> import pandas as pd
+        >>> np.random.seed(42)
```
qiyunzhu, line 688: In 2026 we should use the new `rng` instead of `np.random.xxx`.

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+        >>> # Training data
```
qiyunzhu, line 689: Comments like this can be a new line (without `>>>` and `#`), such that it can be rendered as text instead of code.

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+        than the mean.
+
+        .. math::
+
+            Q^2 = 1 - \frac{SS_{res}}{SS_{tot}}
+                = 1 - \frac{\sum(y - \hat{y})^2}{\sum(y - \bar{y})^2}
+
+        Parameters
+        ----------
+        microbes : pd.DataFrame or array-like of shape (n_samples, n_microbes)
```
qiyunzhu, line 668: This can be `table_like of shape (n_samples, n_microbes)`. [Example](file:///home/drz/Dropbox/scikit-bio/github/doc/build/html/generated/skbio.stats.composition.ancombc.html). See details below.

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+            Test microbe abundance counts.
+        metabolites : pd.DataFrame or array-like of shape (n_samples, n_metabolites)
+            Test metabolite abundance counts.
+
+        Returns
+        -------
+        q2 : float
+            Q² score. Higher is better, with 1.0 being perfect prediction.
```
qiyunzhu, line 676: I'm bit wary about including non-ASCII characters in the source code. Shall we do `Q<sup>2</sup>` instead?

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+    ):
+        """Initialize MMvec model parameters.
+
+        Parameters
+        ----------
+        n_microbes : int
+            Number of microbes (d1).
+        n_metabolites : int
+            Number of metabolites (d2).
+        n_components : int
```
qiyunzhu, line 85: Should add `, optional` after optional parameters.

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+        loss : float
+            Negative log posterior.
+        grads : dict
+            Gradients for U, b_U, V, b_V.
+        """
+        # Convert to sparse COO format if needed
+        if issparse(X):
+            X_coo = X.tocoo()
+        else:
+            X_coo = coo_matrix(X)
```
qiyunzhu, line 188: I wonder if `coo_matrix` can be replaced with `coo_array` to reflect SciPy's new offerings.

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+    This class contains the learned embeddings and co-occurrence patterns
+    from fitting an MMvec model. The key outputs enable both interpretation
+    (which microbes co-occur with which metabolites) and prediction
+    (expected metabolites given a microbial community).
+
+    Attributes
+    ----------
+    microbe_embeddings : pd.DataFrame
+        Microbe coordinates in latent space.
+        Shape: (n_microbes, n_components + 1) where +1 is the bias term.
```
qiyunzhu, line 436: I think shape can be written in the same line after `microbe_embeddings : pd.DataFrame`

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+    clipnorm : float, optional
+        Gradient clipping threshold for Adam (global L2 norm). Ignored for
+        'lbfgs'. Default is 10.0.
+    batch_normalization : {'unbiased', 'legacy'}, optional
+        Method for scaling mini-batch likelihood in Adam. Ignored for 'lbfgs'.
+        Default is 'unbiased'.
+
+        - 'unbiased': Uses norm = sum(microbe_counts) / batch_size.
+        - 'legacy': Uses norm = n_samples / batch_size.
+    random_state : int or numpy.random.Generator, optional
```
qiyunzhu, line 1221: Maybe rename it as `seed` to follow scikit-bio convention ([example](https://github.com/scikit-bio/scikit-bio/blob/ce50a6ab94ba847a2988a4304b3bf12861b29175/skbio/stats/distance/_mantel.py#L123)), and alias it as `random_state` to be compatible with scikit-learn ([example](https://github.com/scikit-bio/scikit-bio/blob/ce50a6ab94ba847a2988a4304b3bf12861b29175/skbio/diversity/alpha/_pd.py#L53)).

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+    ...     np.random.randint(0, 100, size=(50, 15)),
+    ...     columns=[f'metabolite_{i}' for i in range(15)]
+    ... )
+    >>> result = mmvec(microbes, metabolites, n_components=2, max_iter=10)
+    >>> result.ranks.shape
+    (10, 15)
+
+    """
+    # Create RNG - accept either an int seed or an existing Generator
+    if isinstance(random_state, np.random.Generator):
```
qiyunzhu, line 1275: Do `rng = get_rng(seed)` instead.

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+        n_components=n_components,
+        u_prior_mean=u_prior_mean,
+        u_prior_scale=u_prior_scale,
+        v_prior_mean=v_prior_mean,
+        v_prior_scale=v_prior_scale,
+        rng=rng,
+    )
+
+    # Convert to sparse COO format
+    X_coo = coo_matrix(X)
```
qiyunzhu, line 1308: `coo_array`. If applicable. See above.

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+        if batch_normalization == "legacy":
+            norm = n_samples / batch_size
+        else:  # unbiased
+            norm = total_count / batch_size
+
+        # Sample batch weighted by abundance
+        weights = X_coo.data.astype(np.float64)
+        weights /= weights.sum()
+
+        if rng is None:
```
qiyunzhu, line 203: Pass `rng` to the method instead of generating a new one.

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+            rng = np.random.default_rng()
+        batch_idx = rng.choice(
+            len(X_coo.data), size=batch_size, replace=True, p=weights
+        )
+
+        sample_ids = X_coo.row[batch_idx]
+        microbe_ids = X_coo.col[batch_idx]
+
+        # Forward pass
+        Y_arr = np.asarray(Y)
```
qiyunzhu, line 213: No need to `asarray` since this is a private helper function and the input/output must have defined types.

skbio/stats/multimodal/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+    >>> # Predict metabolites for new samples
+    >>> predictions = result.predict(microbes.iloc[:5])
+
+    >>> # Evaluate on held-out data
+    >>> q2 = result.score(microbes.iloc[-10:], metabolites.iloc[-10:])
+
+
+References
+----------
+.. [1] Morton, J.T., et al. "Learning representations of microbe-metabolite
```
qiyunzhu: This [1] isn't cited anywhere in the documentation. I recall that it will run into a rendering error in older Sphinx. But as I tested with the latest Sphinx, the error wasn't raised. But I still think we should cite it somewhere just to be consistent with other scikit-bio documentation pages.

skbio/stats/multimodal/tests/_simulation.py
```diff
@@ -0,0 +1,165 @@
...
+import numpy as np
+import pandas as pd
+
+from skbio.stats.composition import clr_inv as softmax
+from skbio.stats.composition import ilr_inv
+from skbio.util import get_rng
+
+
+def random_multimodal(
+    num_microbes=20,
```
qiyunzhu: I wonder if it's easy or hard to add type hinting to all public-facing functions, if not all functions.

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+            )
+        microbe_props = X / row_sums
+
+        # Get conditional probabilities P(metabolite | microbe)
+        cond_probs = self.probabilities().values  # (n_microbes, n_metabolites)
+
+        # Marginal: sum_microbe P(metabolite | microbe) * P(microbe)
+        predicted = microbe_props @ cond_probs
+
+        return pd.DataFrame(
```
qiyunzhu, line 645: Likewise, consider using `skbio.table._tabular._create_table` ([code](https://github.com/scikit-bio/scikit-bio/blob/ce50a6ab94ba847a2988a4304b3bf12861b29175/skbio/table/_tabular.py#L20C5-L20C18) and [example](https://github.com/scikit-bio/scikit-bio/blob/ce50a6ab94ba847a2988a4304b3bf12861b29175/skbio/stats/ordination/_canonical_correspondence_analysis.py#L232)) to enable support for various table formats.

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+        "V": (np.zeros_like(model.V), np.zeros_like(model.V)),
+        "b_V": (np.zeros_like(model.b_V), np.zeros_like(model.b_V)),
+    }
+
+    # Compute number of iterations per epoch
+    nnz = len(X_coo.data)
+    iterations_per_epoch = max(1, nnz // batch_size)
+
+    t = 0
+    for epoch in range(max_iter):
```
qiyunzhu, line 1041: Looks like this nested for loop is the bottlenecking step, and `loss_and_gradients` is the bottleneck of bottleneck. I wonder if there are ways to further optimize it via vectorized operations. Not urgent though.

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+    batch_normalization="unbiased",
+    random_state=None,
+    verbose=False,
+):
+    r"""Multiomics Microbe-Metabolite Vectors (MMvec).
+
+    Learns joint embeddings of two feature sets from their co-occurrence
+    patterns using a multinomial likelihood model.
+
+    While the parameter names use "microbes" and "metabolites" following the
```
qiyunzhu, line 1161: This clarification is good. Wondering if it's a good or bad idea to make the terminology even less specific, like `X` and `Y`.

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+            dU.ravel(),
+            db_U.ravel(),
+            dV.ravel(),
+            db_V.ravel(),
+        ])
+
+        return loss, grad
+
+
+class MMvecResults(SkbioObject):
```
qiyunzhu, line 422: I wonder how similar it is to `OrdinationResults` (just a question, not a suggestion)

skbio/stats/ordination/tests/data/cf/metabolite-metadata.txt
qiyunzhu, line 1: Do we need a license for including this data?

skbio/stats/multimodal/__init__.py
```diff
@@ -0,0 +1,124 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2013--, scikit-bio development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+r"""Multimodal analysis (:mod:`skbio.stats.multimodal`)
```
qiyunzhu: I like the idea of adding an explicit `multimodal` submodule. Just wanted to note that currently [`cca`](file:///home/drz/Dropbox/scikit-bio/github/doc/build/html/generated/skbio.stats.ordination.cca.html) and [`rda`](file:///home/drz/Dropbox/scikit-bio/github/doc/build/html/generated/skbio.stats.ordination.rda.html) under `skbio.stats.ordination` also do multimodal analysis. Wondering if it makes sense to move them here. Also, Mantel test somehow does multimodal analysis too.

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1325 @@
...
+
+    Learns joint embeddings of two feature sets from their co-occurrence
+    patterns using a multinomial likelihood model.
+
+    While the parameter names use "microbes" and "metabolites" following the
+    original publication, this method is **generic** and can be applied to any
+    two omics modalities representable as compositional (count-based) data.
+    For example: microbes and host transcripts, proteins and metabolites, or
+    any pair of feature tables sharing the same samples.
+
```
qiyunzhu, line 1166: Should have a `    .. versionadded:: 0.7.2` label.

skbio/stats/multimodal/__init__.py
mortonjt: I could also move MMvec to the ordination module (it is technically an ordination method as well).

skbio/stats/ordination/tests/data/cf/metabolite-metadata.txt
mortonjt: It's published data? https://www.nature.com/articles/s41467-017-02356-9

skbio/stats/ordination/_mmvec.py
mortonjt: Very similar - the major difference here is the ability to do cross-validation with Q2 scores.
I'm OK with using OrdinationResults instead, and creating a helper function to manually compute these metrics.

skbio/stats/ordination/_mmvec.py
```diff
@@ -0,0 +1,1328 @@
...
+        if np.any(row_sums == 0):
+            raise ValueError(
+                "metabolites contains samples with all-zero counts. "
+                "Remove these samples before calling score."
+            )
+        actual = Y / row_sums
+
+        # Q² = 1 - SS_res / SS_tot
+        ss_res = np.sum((actual - predicted) ** 2)
+        ss_tot = np.sum((actual - actual.mean()) ** 2)
```
mataton: Should this be `(actual - actual.mean(axis=0))`?

From claude:
"""
Q² is defined as the multivariate analog of R². The "null model" baseline should be "predict the per-column mean for every sample" — i.e., the best constant predictor per metabolite. That requires axis=0 so you get a (1, n_metabolites) mean vector that broadcasts across rows.
Without axis=0, actual.mean() collapses the entire 2D array into a single scalar — the grand mean across all samples and all metabolites. That's a much weaker baseline, so SS_tot is artificially large, which inflates Q² and makes even a poor model look decent.
You can verify this with a quick thought experiment: if you had a model that just predicted the per-metabolite average for every sample, it should get Q² = 0 by definition ("no better than the mean"). With the current code it would get Q² > 0, which breaks the interpretation described in your own docstring.
"""

mortonjt: Thanks @mataton good catch. That has been resolved in the last commit.

PR #2363, opened 2026-01-05
ENH: Add CTF and RPCA ordination methods

mortonjt:
This commit adds two new ordination methods from the gemelli package:

1. Robust PCA (RPCA) - A compositional PCA method that uses:
   - Robust centered log-ratio (rclr) transformation
   - OptSpace matrix completion algorithm
   - Handles sparse compositional data (e.g., microbiome counts)

2. Compositional Tensor Factorization (CTF) - For longitudinal studies:
   - Handles multi-way data (subjects x timepoints x features)
   - Uses tensor rclr transformation
   - CP decomposition with ALS optimization
   - Returns separate ordinations for subjects and states

New modules:
- _rclr.py: Robust centered log-ratio transformations
- _optspace.py: OptSpace matrix completion algorithm
- _tf.py: Tensor factorization with CP decomposition
- _rpca.py: Robust PCA implementation
- _ctf.py: Compositional Tensor Factorization

Test coverage includes tests ported from gemelli to verify numerical accuracy and behavioral compatibility.

References:
- Martino et al. 2019. mSystems 4:e00016-19 (RPCA)
- Martino et al. 2020. Nature Biotechnology 39:165-168 (CTF)


qiyunzhu (commented):
Hello @mortonjt , I have browsed this PR and wanted to provide some overarching opinions:

1. This PR contains four features. I'd suggest splitting into 2-3 sequential PRs (RCLR -> RPCA -> (C)TF) to make the process more manageable.

2. `rclr` can go to `skbio.stats.composition` as it naturally belongs there. That script already has a set of helper functions to make the implementation of `rclr` much easier than it currently is in your PR. Specifically, the existing transformation functions like `clr` already support arbitrarily ranked arrays. So there is no need to separate `tensor_rclr` from `matrix_rclr`. The `closure` function also doesn't need to be re-written.

3. The other features do belong to `skbio.stats.ordination`. Let's think of `rpca` first. As @aj-pinto is implementing PCA (#2337), it will make more sense to add `rpca` after his `pca` PR is merged, and ensure that the interfaces of `rpca` and `pca` have a uniform flavor.

4. scikit-bio is traditionally function-based, not object-oriented. So I am wary of the new classes `OptSpace` and `TensorFactorization`. Classes should be added only after careful thoughts. In your PR, I noticed that `tf` is object-oriented whereas `ctf` is merely a function. This discrepancy is confusing. I would suggest making `tf` and optspace functions as well, since they don't need to be classes.

5. If we do decide to introduce scikit-learn flavored classes (those with `fit` and `transform`, as `TensorFactorization` looks like in your PR), we will want to better follow scikit-learn's [guidelines](https://scikit-learn.org/stable/developers/develop.html). For example, we may want to inherit its `BaseEstimator` mixin. I think this is an overkill. Instead of mimicking the scikit-learn interface, we can provide instructions in the documentation on how to convert scikit-bio functions into scikit-learn compatible estimators. These instructions will be generally useful (e.g., for PCoA as well).

6. `ctf` outputs `OrdinationResults` but `tf` doesn't. This is another example of discrepancy and it should be addressed.

7. Also, I guess `optspace` may not need to be exposed as a public API.

PR #2366, opened 2026-01-09
ENH: Add rclr transformation to composition module

mortonjt:
Add robust centered log-ratio (rclr) transformation to skbio.stats.composition, alongside existing clr, alr, and ilr transforms.

The rclr transformation extends CLR to handle sparse compositional data by only operating on non-zero values, making it suitable for microbiome count data without requiring pseudocount addition.

API mirrors the clr function:
- rclr(mat, axis=-1, validate=True)
- Zeros become NaN in output
- NaN values in input are preserved (missing entries)
- Geometric mean computed only over non-zero values


skbio/stats/composition/_base.py
```diff
@@ -546,6 +546,145 @@ def _clr_inv(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
...
+
+    The robust CLR transformation is similar to the standard CLR transformation,
+    but it only operates on observed (non-zero) values. This makes it suitable
+    for sparse compositional data such as microbiome count data.
+
+    For each composition, the transformation computes:
+
+    .. math::
+
+        rclr(x_i) = \log(x_i) - \frac{1}{|S|} \sum_{j \in S} \log(x_j)
```
qiyunzhu, line 560: `clr` and `alr`'s documentation uses `ln` instead of `log`. Shall we uniformly write them as `ln` for explicity?

skbio/stats/composition/_base.py
```diff
@@ -546,6 +546,145 @@ def _clr_inv(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
     return _closure(xp, diff, axis)
 
 
+def rclr(mat: ArrayLike, axis: int = -1, validate: bool = True) -> StdArray:
+    r"""Perform robust centre log ratio (rclr) transformation.
+
+    The robust CLR transformation is similar to the standard CLR transformation,
+    but it only operates on observed (non-zero) values. This makes it suitable
+    for sparse compositional data such as microbiome count data.
```
qiyunzhu, line 554: I am thinking of downplaying "microbiome" to make these functions more general.

skbio/stats/composition/_base.py
```diff
@@ -546,6 +546,145 @@ def _clr_inv(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
...
+           [ 0.   ,  0.   ,  0.   ,  0.   ]])
+
+    """
+    xp, mat = ingest_array(mat)
+    if validate:
+        _check_rclr_input(xp, mat)
+    return _rclr(xp, mat, axis)
+
+
+def _check_rclr_input(xp: ModuleType, mat: StdArray) -> None:
```
qiyunzhu, line 626: I am thinking that this new function can be replaced with the existing `_check_composition` function, pending a modification: Currently it raises if there is any NaN in the array, and it uses `xp.all(xp.isfinite(mat))` to check if there are any NaN or infinite. A Boolean flag `allnum` may be added to control this behavior. If True, the original code (`xp.isfinite`) is executed. If False (i.e., NaN is allowed), this code becomes `xp.any(xp.isinf(mat))`, which only checks for infinite while tolerating NaN.

Optionally, one may further add a Boolean flag `allfin` to control for whether infinite is tolerated.

skbio/stats/composition/_base.py
```diff
@@ -546,6 +546,145 @@ def _clr_inv(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
...
+    """Check if input is valid for rclr transformation.
+
+    NaN values are allowed as they represent missing entries.
+    """
+    if not xp.isdtype(mat.dtype, "numeric"):
+        raise TypeError("Input matrix must have a numeric data type.")
+    if xp.any(xp.isinf(mat)):
+        raise ValueError("Input matrix cannot have infinite values.")
+    # Check for negative values, excluding NaN (which represents missing data)
+    non_nan_mask = ~xp.isnan(mat)
```
qiyunzhu, line 636: A non-NaN mask may not be needed, at least in NumPy, because `np.nan < 0` returns False. Omitting this step can save some compute. One may need to check Python array API to ensure this behavior is universal to compliant libraries. AI says yes, quoting IEEE 754 (NaN is unorderable).

skbio/stats/composition/_base.py
```diff
@@ -546,6 +546,145 @@ def _clr_inv(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
...
+    # For closure, replace NaN with 0 temporarily
+    mat_for_closure = xp.where(
+        xp.isnan(mat_float), xp.asarray(0.0, dtype=float_dtype), mat_float
+    )
+
+    # Normalize to closure (rows sum to 1)
+    closed = _closure(xp, mat_for_closure, axis)
+
+    # Take log, zeros become -inf
+    with np.errstate(divide="ignore"):
```
qiyunzhu, line 660: Muting warnings may not be the ideal pattern in engineering...

skbio/stats/composition/_base.py
```diff
@@ -546,6 +546,145 @@ def _clr_inv(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
...
+    # Ensure float dtype for output (NaN requires float)
+    float_dtype = xp.float64
+    mat_float = xp.asarray(mat, dtype=float_dtype)
+
+    # Create mask for observed (non-zero and non-NaN) values
+    # NaN values in input are treated as missing and should remain NaN
+    observed_mask = (mat > 0) & ~xp.isnan(mat_float)
+
+    # For closure, replace NaN with 0 temporarily
+    mat_for_closure = xp.where(
```
qiyunzhu, line 652: There might be an easier and more performant trick: Instead of replacing NaN with 0, one should do the other way around: replacing 0 with NaN. This is because NaN is automatically ignored in most element-wise calculations. For example you can do `mat / x` or `xp.log(mat)`, and NaN positions in `mat` will remain NaN (without throwing an error or warning). Additionally, NumPy can do `np.nansum` which skips NaN, but `nansum` is not part of Python array API. To ensure compatibility, one should do `xp.sum` manually instead.

skbio/stats/composition/_base.py
```diff
@@ -546,6 +546,145 @@ def _clr_inv(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
...
+    # Avoid division by zero for all-zero/all-NaN compositions
+    n_observed_safe = xp.where(
+        n_observed > 0, n_observed, xp.asarray(1.0, dtype=float_dtype)
+    )
+    geo_mean_log = log_sum / n_observed_safe
+
+    # Center by geometric mean
+    result = log_closed - geo_mean_log
+
+    # Replace non-observed (zeros and NaN) with NaN
```
qiyunzhu, line 680: If we replace 0 with NaN above (see my previous comment), this can be much easier as `result[zero_mask] = 0.0`.

skbio/stats/composition/_base.py
```diff
@@ -546,6 +546,145 @@ def _clr_inv(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
...
+        on this axis is considered as a composition. Default is the last
+        axis (-1).
+    validate : bool, default True
+        Check if the matrix consists of non-negative, finite values.
+        NaN values are allowed as missing entries.
+
+    Returns
+    -------
+    ndarray of shape (..., n_components, ...)
+        rclr-transformed matrix. Zero values in the input become NaN.
```
qiyunzhu, line 582: "Zero values in the input become NaN." Is this the desired behavior? Just curious and want to double-confirm.

qiyunzhu (commented):
@mortonjt Thanks! I made several comments. Some of them are about low-level optimization and compatibility enhancement, and please don't feel obligated to address them (delivering perfect code is a secondary priority). The CI workflow went into error because of some dependency issue, which I am trying to address in #2367. Once that PR is merged, your PR should be able to enter the CI workflow normally.

skbio/stats/composition/_base.py
mortonjt: yes that's correct, this is because RPCA will mask the NANs

skbio/stats/composition/_base.py
mataton: `closure` can't have NaN in input because it uses `xp.sum`, so I think it makes sense to keep the current approach. But I'm not sure what you mean by do `xp.sum` manually?

PR #2368, opened 2026-01-11
ENH: add comprehensive exception import tests

### LINKED ISSUE #2026: TEST: Exception handling in skbio.io.format lacks test coverage
Discovered here : https://github.com/scikit-bio/scikit-bio/pull/2021/files

There is the potential to not detect import errors in `skbio.io.format.__init__.py`, hinting that these exceptions are currently not covered by the unittests. We should have a simple unittest that loops through all of the imported exceptions to confirm that at a minimum the imports are configured (but preferably confirm that all the file formats correctly raise exceptions).

csbiy:
## Summary

This PR adds comprehensive unit tests for exception handling in `skbio.io`, addressing #2026.

**Changes:**
- Add `test_all_io_exceptions_importable`: Tests ALL exception classes exported from `skbio.io`, not just those ending with `*FormatError`. This ensures exceptions like `IOSourceError` are also covered.
- Add `test_exception_module_exports_match_definitions`: Verifies that public exceptions defined in `_exception.py` are properly exported in `__all__`.
- Add `test_exception_inheritance_hierarchy`: Ensures format-specific exceptions inherit from `FileFormatError` as expected (except `BIOMFormatError` and `EmbedFormatError` which have different base classes by design).

## Test plan

- [x] All existing tests pass
- [x] New tests pass locally (5/5 passed)
- [x] Full `skbio/io/tests/` test suite passes (149 passed)

Closes #2026

### DIFF
--- a/skbio/io/tests/test_format_imports.py
+++ b/skbio/io/tests/test_format_imports.py
@@ -37,35 +37,6 @@ class ImportModuleTests(TestCase):
                     f"\n\n\nModule '{name}' not imported in skbio.io.__init__.py"
                 )
 
-    def test_format_exceptions_importable(self):
-        """Verify all format-specific exceptions are importable from skbio.io"""
-        # Get all format exception classes defined in the io module
-        io_exceptions = {
-            name: obj
-            for name, obj in vars(skbio.io).items()
-            if inspect.isclass(obj)
-            and issubclass(obj, Exception)
-            and name.endswith("FormatError")
-            and name != "FileFormatError"
-        }
-
-        failed_exceptions = []
-        for name, exception_cls in io_exceptions.items():
-            try:
-                # Test instantiation
-                ex = exception_cls()
-                # Test raising
-                raise ex
-            except exception_cls:
-                # This is expected - exception raised and caught correctly
-                pass
-            except Exception as e:
-                # Any other exception means something is wrong
-                failed_exceptions.append((name, str(e)))
-
-        if failed_exceptions:
-            self.fail(f"Failed to instantiate or raise exceptions: {failed_exceptions}")
-
     def test_all_io_exceptions_importable(self):
         """Verify all I/O exceptions are importable and functional.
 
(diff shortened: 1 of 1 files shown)


skbio/io/tests/test_format_imports.py
```diff
@@ -61,10 +62,110 @@ def test_format_exceptions_importable(self):
...
+        """Verify format-specific exceptions inherit from FileFormatError.
+
+        This ensures that users can catch all format-related errors
+        by catching FileFormatError.
+        """
+        format_exceptions = {
+            name: obj
+            for name, obj in vars(skbio.io).items()
+            if inspect.isclass(obj)
+            and issubclass(obj, Exception)
```
mataton: This check will fail if the custom error class does not inherit from anything. For example:
```python
class LSMatFormatError():
    """Raised when a ``lsmat`` formatted file cannot be parsed."""
    pass
```
would pass this test, because `issubclass(obj, Exception)` would be `False`. Then this would cause `TypeError`s if anyone tried to raise that error. So ideally this test should check whether the objects inherit from `Exception` at all, and then if they do make sure that they inherit from `FileFormatError` and not `Exception` directly.

skbio/io/tests/test_format_imports.py
```diff
@@ -61,10 +62,110 @@ def test_format_exceptions_importable(self):
             except Exception as e:
                 # Any other exception means something is wrong
                 failed_exceptions.append((name, str(e)))
-        
+
         if failed_exceptions:
             self.fail(f"Failed to instantiate or raise exceptions: {failed_exceptions}")
 
+    def test_all_io_exceptions_importable(self):
```
mataton, line 40: This is actually more thorough than the existing `test_format_exceptions_importable` test. So you could completely remove the old `test_format_exceptions_importable` and that would keep the test file cleaner and less redundant.

csbiy: Good point! I've removed the redundant test_format_exceptions_importable since test_all_io_exceptions_importable covers everything it did and more.

skbio/io/tests/test_format_imports.py
csbiy: Thanks for catching this!  I've fixed this by removing the `issubclass(obj, Exception)` filter from the dictionary comprehension, so now all classes ending with "FormatError" are included in the check.

  Then in the loop, I explicitly verify that each class inherits from Exception first before checking FileFormatError inheritance. This way, if someone accidentally defines an error class without inheriting from Exception (like your `LSMatFormatError()` example), the test will fail with a clear message instead of silently passing.

PR #2371, opened 2026-01-21
Skip specific unit tests failing with Pandas 3.0

mataton:
~~Pandas released version 3.0 today, introducing the new `StringDtype` to replace the `object` dtype which was formerly used to represent strings. Additionally, certain functions which were deprecated but still being used in scikit-bio are now gone. This PR will get skbio fully compatible with the latest version of Pandas.~~

Note: This PR does not address pandas 3.0 compatibility for the metadata module.

The metadata module uses "NaN boxing" to store missing value scheme info (e.g., "missing", "not applicable") in the bit representation of NaN floats. Pandas 3.0 no longer preserves these bit patterns. Operations like .apply() now normalize NaNs, destroying the encoded data.

I plan to refactor this in a follow-up PR by storing missing value labels in a parallel data structure instead. This is less memory-efficient but far more maintainable. Until then, users relying on INSDC missing value schemes should pin pandas<3.0.

See issue #2375 for details.


ci/conda_requirements.txt
```diff
@@ -2,7 +2,7 @@ requests >= 2.20.0
 decorator >= 3.4.2
 natsort >= 4.0.3
 numpy >= 2.0
-pandas >= 1.5.0
+pandas >= 3.0
```
qiyunzhu: @mataton I am not sure about this... Is there a way to let scikit-bio work with Pandas 1.5, 2.x and 3.0?

mataton: I'm only including this temporarily for testing purposes, once it's ready to merge I'm going to remove it.

qiyunzhu:
Merging to unblock development. But we will need to address the "NaN boxing" issue sooner or later. @mortonjt @ebolyen Do you have some thoughts on this?

qiyunzhu:
Hi @mortonjt @ebolyen Thanks for the insights! Agreed on the mask idea. @mataton and I discussed last week and we came to a similar idea. Python's dictionary also has a second list attached to it (if I recall correctly). So I think it is the norm in industry and won't significantly slow down the code.

PR #2386, opened 2026-02-16
Implement RCLR

aj-pinto:

### DIFF
--- a/skbio/stats/composition/_base.py
+++ b/skbio/stats/composition/_base.py
@@ -1380,7 +1380,7 @@ def tree_basis(tree):
 
     Returns
     -------
-    scipy.sparse.coo_matrix
+    scipy.sparse.coo_array
         The ilr basis required to perform the ilr_inv transform. This is also
         known as the sequential binary partition. Note that this matrix is
         represented in clr coordinates.
@@ -1403,7 +1403,7 @@ def tree_basis(tree):
            [-0.70710678,  0.70710678,  0.        ]])
 
     """
-    from scipy.sparse import coo_matrix
+    from scipy.sparse import coo_array
 
     # Specifies which child is numerator and denominator
     # within any given node in a tree.
@@ -1480,7 +1480,7 @@ def tree_basis(tree):
         i += 1
         nodes.append(n.name)
 
-    basis = coo_matrix((value, (row, col)), shape=(D - 1, D))
+    basis = coo_array((value, (row, col)), shape=(D - 1, D))
 
     return basis, nodes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
 
 * Fixed `permdisp` mutating the input `OrdinationResults` object by adding a `"grouping"` column to its `samples` DataFrame ([#2440](https://github.com/scikit-bio/scikit-bio/pull/2440)).
 
+### Miscellaneous
+
+* Replaced SciPy sparse matrix constructors with sparse array constructors to align with SciPy's sparse array APIs. This affects `subsample_counts` (`csr_array`) (no public-facing effect) and `tree_basis` (`coo_array`).
+
 ## Version 0.7.2
 
 ### Features
--- a/skbio/stats/_subsample.py
+++ b/skbio/stats/_subsample.py
@@ -236,8 +236,8 @@ def subsample_counts(counts, n, replace=False, seed=None):
     if counts.ndim != 1:
         raise ValueError("Only 1-D vectors are supported.")
 
-    # csr_matrix will report ndim of 2 if vector
-    counts = sparse.csr_matrix(counts)
+    # csr_array will report ndim of 2 if vector
+    counts = sparse.csr_array(counts)
 
     counts_sum = counts.sum()
     if n > counts_sum and not replace:
--- a/skbio/stats/composition/tests/test_base.py
+++ b/skbio/stats/composition/tests/test_base.py
@@ -13,7 +13,7 @@ import numpy.testing as npt
 from numpy.exceptions import AxisError
 from numpy.random import rand, randint
 import pandas as pd
-from scipy.sparse import coo_matrix
+from scipy.sparse import coo_array
 
 from skbio import TreeNode
 from skbio.stats.distance import DistanceMatrixError
@@ -920,7 +920,7 @@ class TestTreeBasis(TestCase):
         tree = u"(a,b);"
         t = TreeNode.read([tree])
 
-        exp_basis = coo_matrix(
+        exp_basis = coo_array(
             np.array([[-np.sqrt(1. / 2),
                        np.sqrt(1. / 2)]]))
         exp_keys = [t.name]
@@ -938,7 +938,7 @@ class TestTreeBasis(TestCase):
     def test_tree_basis_unbalanced(self):
         tree = u"((a,b)c, d);"
         t = TreeNode.read([tree])
-        exp_basis = coo_matrix(np.array(
+        exp_basis = coo_array(np.array(
             [[-np.sqrt(1. / 6), -np.sqrt(1. / 6), np.sqrt(2. / 3)],
              [-np.sqrt(1. / 2), np.sqrt(1. / 2), 0]]
         ))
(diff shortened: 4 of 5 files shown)


skbio/stats/composition/_base.py
```diff
@@ -546,6 +567,124 @@ def _clr_inv(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
...
+    float_dtype = xp.float64
+    mat_float = xp.asarray(mat, dtype=float_dtype)
+
+    # Track which values were observed in the ORIGINAL input
+    observed_mask = (mat_float > 0) & ~xp.isnan(mat_float)
+
+    # Normalize to closure (using _nansum internally)
+    closed = _closure(xp, mat_float, axis)
+    closed_safe = xp.where(
+        closed > 0, closed, xp.asarray(float("nan"), dtype=float_dtype)
```
aj-pinto: It might be better to replace "closed > 0" with observed_mask, which has already been computed

skbio/stats/composition/_base.py
```diff
@@ -546,6 +567,124 @@ def _clr_inv(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
...
+def _rclr(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
+    """Perform rclr transform."""
+    float_dtype = xp.float64
+    mat_float = xp.asarray(mat, dtype=float_dtype)
+
+    # Track which values were observed in the ORIGINAL input
+    observed_mask = (mat_float > 0) & ~xp.isnan(mat_float)
+
+    # Normalize to closure (using _nansum internally)
+    closed = _closure(xp, mat_float, axis)
```
aj-pinto: Here, closure is performed before applying the observed_mask, so negative numbers might still be present in the array at this point.

This could reduce the row sums in the closure calculation (adding negative numbers).

If we are only considering positive values as "observed," (replacing negatives with NaN), should the mask be applied first?

aj-pinto: Actually, is normalizing to closure necessary? The result should not depend on the scale of the data, since clr(kx) = clr(x) for any constant k. So, the result will be the same regardless.

It is costlier to compute closure, but is there a tradeoff for stability, or some other reason?

mataton:
@qiyunzhu AJ and I discussed offline and feel this is ready to merge. What do you think? The git history is a bit messy but I'm not sure if that is a concern or not?

skbio/stats/composition/_base.py
```diff
@@ -60,8 +68,14 @@ def _check_composition(
     """
     if not xp.isdtype(mat.dtype, "numeric"):
         raise TypeError("Input matrix must have a numeric data type.")
-    if not xp.all(xp.isfinite(mat)):
-        raise ValueError("Input matrix cannot have infinite or NaN values.")
+    if allnum:
```
qiyunzhu, line 71: This looks like an elegant and efficient patch.

skbio/stats/composition/_base.py
```diff
@@ -546,6 +576,115 @@ def _clr_inv(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
...
+    """
+    xp, mat = ingest_array(mat)
+    if validate:
+        _check_composition(xp, mat, allnum=False)
+    return _rclr(xp, mat, axis)
+
+
+def _rclr(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
+    """Perform rclr transform."""
+    mat_array = xp.asarray(mat)
```
qiyunzhu: This seems unnecessary because `mat` is guaranteed to be an array in the `xp` namespace.

skbio/stats/composition/_base.py
```diff
@@ -546,6 +576,115 @@ def _clr_inv(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
...
+
+    # Track which values were observed in the original input
+    observed_mask = (mat_array > 0) & ~xp.isnan(mat_array)
```
qiyunzhu: This can be reduced into merely `~xp.isnan(mat_array)`, because `_check_composition` already guaranteed that all values are non-negative (or NaN).

skbio/stats/composition/_base.py
```diff
@@ -546,6 +576,115 @@ def _clr_inv(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
...
+
+    # Operate only on observed values (nonnegative)
+    mat_safe = xp.where(observed_mask, mat_array, xp.asarray(float("nan")))
```
qiyunzhu: Can `float("nan")` be replaced with `xp.nan`? "nan" is part of the array API standard. Also, why we need `xp.asarray` here? (This is a question before it becomes a suggestion. My understanding may not be accurate.)

skbio/stats/composition/_base.py
```diff
@@ -546,6 +576,115 @@ def _clr_inv(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
...
+
+    # Take log (will give -inf for zeros, NaN for NaN)
+    log_safe = xp.where(observed_mask, xp.log(mat_safe), xp.asarray(0.0))
```
qiyunzhu: Likewise, why we need `xp.asarray(0.0)` instead of just `0.0`? Same below.

qiyunzhu:
@aj-pinto Once you have addressed my comments, please remove "[WIP]" from the title, indicating that you think it is ready to be merged. And also fill the checklist. Thanks!

skbio/stats/composition/_base.py
aj-pinto: I think this also filters zero entries as well, which is not checked by check_composition for rclr. It could be replaced by "mat_array != 0", or maybe instead the comparison could be made with xp.isclose?

skbio/stats/composition/_base.py
aj-pinto: This change is possible, but I just realized using NaNs might be a problem if the matrix has a dtype int, in which case my understanding is that NaN cannot be represented? Should we restore the fixed float dtype for mat_array?

skbio/stats/composition/_base.py
aj-pinto: I think this was intended to copy the array so the operation would not be performed in-place on the input, but a new array is made immediately after anyway, so it can be removed.

skbio/stats/composition/_base.py
aj-pinto: Good point, I will make this change.

skbio/stats/composition/_base.py
aj-pinto: Actually, I just realized NaNs are only required if we choose to make an intermediate "mat_safe" matrix before taking the log of positive, non-NaN entries from the original matrix. However, this step seems unnecessary since we can simply use the observed_mask we took from the original matrix, so I will remove it.

skbio/stats/composition/_base.py
```diff
@@ -546,6 +576,115 @@ def _clr_inv(xp: ModuleType, mat: StdArray, axis: int) -> StdArray:
...
+
+    # Count observed values from original mask
+    n_observed = xp.sum(observed_mask, axis=axis, keepdims=True)
+
+    # Sum logs for observed values only
+    log_sum = xp.sum(log_safe, axis=axis, keepdims=True)
+
+    # Geometric mean
+    n_observed_safe = xp.where(n_observed > 0, n_observed, xp.asarray(1.0))
```
aj-pinto: I think this is unnecessary since check_composition already prevents compositions with all zeros?

aj-pinto: Do you agree @qiyunzhu?

qiyunzhu: @aj-pinto No. I checked the code of `_check_composition` and believe that it doesn't prevent compositions with all zeros. It only prevents individual cells from being zeros if `nozero=True` is set. If not set, all-zero compositions are still allowed.

aj-pinto: @qiyunzhu If nozero=False, it still falls back on another check regardless, right?

qiyunzhu: Oh! I see. The `_check_composition` function has the following logic. Therefore, you are right, that this code is unnecessary. @aj-pinto 

```python
if xp.any(~xp.any(mat, axis=axis)):
    raise ValueError("Input matrix cannot have compositions with all zeros.")
```

mataton:
@aj-pinto Have you pushed the latest changes yet?

aj-pinto:
@mataton I've just pushed the changes now, still waiting on some checks

qiyunzhu:
Everything passed. Should I merge? @aj-pinto

qiyunzhu:
@aj-pinto I have merged this PR. Good job! Note that the changelog hasn't been updated. Also we can discuss on implementing an `rclr_inv` function if needed. @mortonjt what do you think?

aj-pinto:
@qiyunzhu Great, thank you!

Issue #2375, opened 2026-02-06
Metadata module incompatibility with pandas 3.0


mataton:
Certain parts of scikit-bio's metadata module are incompatible with pandas 3.0. We're aware of the issue and are actively working on a fix for the next release. In the meantime, if you rely on SampleMetadata, CategoricalMetadataColumn, NumericMetadataColumn, or INSDC missing value schemes, we recommend pinning pandas<3.0.

The metadata module currently uses "NaN boxing"—storing missing value scheme information (like "missing", "not applicable", etc.) directly in the bit representation of NaN floats. This is clever and memory-efficient, but it relies on pandas preserving those exact bit patterns. Pandas 3.0 no longer guarantees this. Specifically, operations like .apply() now normalize NaN values to a canonical representation, which destroys the encoded metadata.

Rather than patching around each place pandas touches NaN bits (a game of whack-a-mole that's hard to maintain), I plan to refactor the approach entirely. The new design will store missing value labels in a parallel data structure instead of encoding them in NaN bits. This is slightly less memory-efficient, but significantly more maintainable and robust to future changes in pandas or numpy internals. Given that metadata objects are typically not large, I think this is the right tradeoff.

See #2371 for additional details.

PR #2387, opened 2026-02-17
Refactor metadata handling of missing values

mataton:
This PR addresses #2375. 

Replace NaN-payload bit-packing with a separate mask for tracking missing value provenance.

The old approach encoded missing value terms (e.g. INSDC vocabulary like "not applicable", "missing") into the mantissa bits of NaN floats via `_enan.py`. This was clever but fragile, and broke under Pandas 3.x which no longer preserves NaN bit patterns through Series operations.

This PR replaces that with a straightforward `pd.Series` mask that travels alongside the data, where non-null entries record the original missing term. The encode/extract roundtrip is preserved, `_enan.py` and its tests are removed, and all the `@unittest.skipIf(PANDAS_3, ...)` guards are dropped.

Key changes:
- `series_encode_missing` now returns `(series, mask)` instead of a single series
- `_missing_masks` is threaded through `SampleMetadata`, `MetadataColumn`, and `MetadataReader`
- `series_extract_missing` reconstructs original terms from the mask
- Tests updated for Pandas 3.x dtype handling (StringDtype vs object)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Bug Fixes
 
-* Fixed incompatibility issues between the `skbio.Metadata` module and Pandas >= 3.0 ([#2387](https://github.com/scikit-bio/scikit-bio/pull/2387))
+* Fixed incompatibility issues between the `skbio.metadata` module and Pandas >= 3.0 ([#2387](https://github.com/scikit-bio/scikit-bio/pull/2387))
 
 
 ## Version 0.7.2


qiyunzhu (approved):
Looks reasonable and elegant to me!

It appears that changes to the API are mainly in private functions, which make no impact to the public API. The only change to public API is `def __init__(self, series, missing_scheme=DEFAULT_MISSING, missing_mask=None)`, with the added parameter `missing_mask`. This adds flexibility but doesn't change the existing behavior.

The use of Pandas series to store missing masks may not be the most efficient, but the solution is simple in logic and easy to maintain. I tend to think this is the best solution.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Version 0.7.3-dev
 
+### Bug Fixes
+
+* Fixed incompatibility issues between the `skbio.Metadata` module and Pandas >= 3.0 ([#2387](https://github.com/scikit-bio/scikit-bio/pull/2387))
```
qiyunzhu:
```suggestion
* Fixed incompatibility issues between the `skbio.metadata` module and Pandas >= 3.0 ([#2387](https://github.com/scikit-bio/scikit-bio/pull/2387))
```

PR #2389, opened 2026-02-22
MAINT: add Self type hints across core modules (sequence, alignment, tree, stats.distance, metadata) (#2355)

### LINKED ISSUE #2355: MAI Adopt `typing.Self` since Python 3.11+
This type can simplify class method typing, especially for those decorated with `@classonlymethod`, which is a custom decorator and mypy cannot recognize it.

```python
class Foo:
    def return_self(self: Self) -> Self:
        ...
        return self
```

https://docs.python.org/3/library/typing.html#typing.Self

umi21:
This PR resolves #2355 by updating the type hints in the skbio.sequence, skbio.alignment, skbio.tree, skbio.stats.distance, and skbio.metadata modules. It adopts typing.Self for methods that return an instance of the class in which they are defined.

* [ ] This pull request includes code, documentation, or other content derived from external source(s).If this is the case, ensure the external source's license is compatible with scikit-bio's [license](https://github.com/scikit-bio/scikit-bio/blob/main/LICENSE.txt). Include the license in the licenses directory and add a comment in the code giving proper attribution. Ensure any other requirements set forth by the license and/or author are satisfied.

* It is your responsibility to disclose code, documentation, or other content derived from external source(s). If you have questions about whether something can be included in the project or how to give proper attribution, include those questions in your pull request and a reviewer will assist you.

### DIFF
--- a/skbio/tree/_tree.py
+++ b/skbio/tree/_tree.py
@@ -43,6 +43,7 @@ from ._compare import (
 if TYPE_CHECKING:  # pragma: no cover
     from collections.abc import Sequence, Iterable, Callable, Iterator
     from numpy.typing import ArrayLike
+    from typing import Self
 
 
 # ----------------------------------------------------------------------------
@@ -5785,7 +5786,7 @@ class TreeNode(SkbioObject):
             ("names", "id_list", "0.7.2", False),
         ]
     )
-    def from_linkage_matrix(cls, lnkmat: ArrayLike, names: Iterable[str]) -> TreeNode:
+    def from_linkage_matrix(cls, lnkmat: ArrayLike, names: Iterable[str]) -> Self:
         r"""Construct tree from a SciPy linkage matrix.
 
         Parameters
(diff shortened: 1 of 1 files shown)


mataton:
This looks like a good start @umi21. Do you want to add more to it or is it ready to merge?

mataton:
@qiyunzhu `Self` is only imported when performing type checking with `mypy` or some other static type checker. It won't run during testing or during normal use, so the test cases pass. The only time an import error would happen is really if a developer was doing work in an environment with Python 3.10 and decided to run `mypy`. So I think it's okay to keep these changes. EOL for Python 3.10 is in 7 months.

umi21:
> @mataton @umi21 I thought `Self` is supported since Python 3.11, and scikit-bio currently supports 3.10+. But the CI workflows passed on this PR? So it works for Python 3.10 too?

@qiyunzhu Issue #2355 says "... since 3.11+"; that's why I didn't bother about 3.10. But the `Self` type hint is supported in 3.10 through the `typing_extensions` module instead of the `typing`. If needed, we can still prevent any errors by adding a version check and importing `Self` from `typing_extensions` if the python version is 3.10. I'm not sure if that's important.

PR #2398, opened 2026-03-03
MNT: Support Array API in Correspondence Analysis (CA)

Sriniketh99:
I have implemented the xp wrapper for Correspondence Analysis to support hardware-agnostic execution (CPU/GPU) as part of the ongoing Array API migration for GSoC.

Ordination:

- [x] ca

- [ ] cca 

- [x] rda

Alpha diversity metrics (alpha):

- [ ] shannon

- [ ] simpson

- [ ] pielou_e

Multiplicative replacement (composition):

- [ ] multi_replace

- [ ] vlr

Differential abundance (DA) tests:

- [ ] ancom

- [ ] dirmult_lme

- [ ] dirmult_ttest

Augmentation techniques (table):

- [ ] phylomix

- [ ] compos_cutmix

- [ ] aitchison_mixup

- [ ] mixup

- [x] I have read the [contribution guidelines](https://github.com/scikit-bio/scikit-bio/blob/main/CONTRIBUTING.md).

- [x] All local tests passed via pytest skbio/stats/ordination/tests/test_correspondence_analysis.py.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -5,6 +5,8 @@ on:
     branches: [ main ]
   workflow_dispatch:
   workflow_call:
+  schedule:
+    - cron: "0 7 1,15 * *"
 
 env:
   latest_python: "3.14"
@@ -43,9 +45,9 @@ jobs:
         # This may need to be updated for new Python versions, double check before release.
         uses: pypa/cibuildwheel@v3.2.1
         env:
-        # Run full tests on release or manual trigger, otherwise (i.e. on PRs) do simple import/print test.
+        # Run full tests on release, manual trigger, and schedule (twice a month). Otherwise (i.e. on PRs) do simple import/print test.
           CIBW_TEST_COMMAND: >
-            ${{ ((github.event_name == 'push' && startsWith(github.ref, 'refs/tags')) || github.event_name == 'workflow_dispatch') &&
+            ${{ ((github.event_name == 'push' && startsWith(github.ref, 'refs/tags')) || github.event_name == 'workflow_dispatch' || github.event_name == 'schedule') &&
                 'python -m skbio.test' ||
                 'python -c "import skbio; print(skbio.__version__)"' }}
 


skbio/stats/ordination/_correspondence_analysis.py
```diff
@@ -113,17 +115,17 @@ def ca(X, scaling=1, sample_ids=None, feature_ids=None, output_format=None):
...
     # an scaled version of the contribution of each cell towards
     # Pearson chi-square statistic.
-    expected = np.outer(row_marginals, column_marginals)
-    Q_bar = (Q - expected) / np.sqrt(expected)  # Eq. 9.32
+    expected = xp.outer(row_marginals, column_marginals)
+    Q_bar = (Q - expected) / xp.sqrt(expected)  # Eq. 9.32
 
     # Step 2 (Singular Value Decomposition)
-    U_hat, W, Ut = svd(Q_bar, full_matrices=False)
+    U_hat, W, Ut = xp.linalg.svd(Q_bar, full_matrices=False)
```
qiyunzhu: It is interesting to see that `xp.linalg.svd` is now part of the Python array API standard. All libraries should support it, although I am not sure how many them actually do. Good job!

skbio/stats/ordination/_redundancy_analysis.py
```diff
@@ -113,6 +114,17 @@ def rda(
     Y, y_rows, y_cols = _ingest_table(y)
     X, x_rows, x_cols = _ingest_table(x)
 
+    # Correct unpacking: the array comes first, then the namespace
+    xp_y, Y = ingest_array(Y)
+    xp_x, X = ingest_array(X)
+
+    # Compare the namespace names
+    if xp_y.__name__ != xp_x.__name__:
```
qiyunzhu: This works but there is a better way: `xp, X, Y = ingest_array(X, Y)`. This helper function calls [`array_api_compat.array_namespace`](https://data-apis.org/array-api-compat/helper-functions.html#array_api_compat.array_namespace) under the hood to ensure that both arrays have the same namespace.

skbio/stats/ordination/_redundancy_analysis.py
```diff
@@ -125,10 +137,11 @@ def rda(
     feature_ids = y_cols
     # Centre response variables (they must be dimensionally
     # homogeneous)
-    Y = scale(Y, with_std=scale_Y)
+    Y = Y - xp.mean(Y, axis=0)
```
qiyunzhu: This works but there may be a cleaner solution: `Y -= xp.mean(Y, axis=0)`. This is because `Y -=` modifies `Y` in-place, whereas `Y = Y -` creates a new array. That being said, I am not sure if array libraries all support this behavior. Worth testing. Same applies to a few other locations.

skbio/stats/ordination/_correspondence_analysis.py
Sriniketh99: Thank You !!

skbio/stats/ordination/_redundancy_analysis.py
Sriniketh99: I Fixed it in the latest commit.
I've updated the code to use the unified xp, X, Y = ingest_array(X, Y)

skbio/stats/ordination/_redundancy_analysis.py
Sriniketh99: Updated in the latest commit.
Thanks for pointing it out !!

PR #2405, opened 2026-03-05
ENH: Support Array API in ANCOM 

Sriniketh99:
This PR implements the xp wrapper for ancom in the composition module to support the Array API standard.
Key Changes:

Namespace Integration: Added ingest_array to handle multiple backends (NumPy, CuPy, etc.).

Refactored Logic:

1)Replaced np.apply_along_axis with row-iteration and xp.stack.

2)Replaced np.fill_diagonal with identity-matrix masking and xp.where.

3)Replaced np.triu_indices with xp.triu and xp.nonzero.

Fixes:

1)Forced xp.float64 for the p-value matrix to prevent rounding errors on integer inputs.

2)Resolved TypeError for list_iterator inputs in percentiles.

3)Used np.asarray for percentile calculations to maintain pandas compatibility.

Verification

Passed 38 tests via pytest skbio/stats/composition/tests/test_ancom.py.

Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -57,3 +57,4 @@ web/_build
 # Cython files
 *.c
 !ssw.c
+skbio-venv/


skbio/stats/composition/_ancom.py
```diff
@@ -287,16 +288,21 @@ def ancom(
...
+   # validate percentiles
+    # validate percentiles
     if percentiles is None:
-        percentiles = np.arange(0, 125, 25.0)
+        percentiles = xp.asarray([0.0, 25.0, 50.0, 75.0, 100.0])
     else:
-        if not isinstance(percentiles, np.ndarray):
-            percentiles = np.fromiter(percentiles, dtype=float)
-        if (percentiles < 0.0).any() or (percentiles > 100.0).any():
+        # CRITICAL FIX: Convert iterator to a list before array ingestion
```
qiyunzhu, line 304: I think `ingest_array` can handle iterators. Can you test?

skbio/stats/composition/_ancom.py
```diff
@@ -323,24 +329,32 @@ def ancom(
     # compare log ratios
     pval_mat = _log_compare(matrix, labels, n_groups, test_f)
 
+    # correct for multiple testing problem
```
qiyunzhu: Why this comment shows up three times?

skbio/stats/composition/_ancom.py
```diff
@@ -361,17 +375,32 @@ def ancom(
     )
 
     # calculate percentiles
-    if percentiles.size == 0:
+    if int(percentiles.size) == 0:
```
qiyunzhu, line 388: Why is `int` needed here?

skbio/stats/composition/_ancom.py
```diff
@@ -403,24 +432,36 @@ def _log_compare(matrix, labels, n, test):
...
+    xp_labels = xp.asarray(labels)
+    
     # divide data by sample group
-    grouped = [log_mat[labels == i] for i in range(n)]
+    grouped = [log_mat[xp_labels == i] for i in range(n)]
 
     # determine all pairs of feature indices
     m = matrix.shape[1]
-    ii, jj = np.triu_indices(m, k=1)
+    upper_tri_mask = xp.triu(xp.ones((m, m), dtype=xp.bool), k=1)
```
qiyunzhu, line 461: Interesting solution!

skbio/stats/composition/_ancom.py
```diff
@@ -403,24 +432,36 @@ def _log_compare(matrix, labels, n, test):
...
+    ii, jj = xp.nonzero(upper_tri_mask)
 
     # calculate all log ratios (pairwise difference of log values)
     log_ratios = [x[:, ii] - x[:, jj] for x in grouped]
 
     # run statistical test on the 2-D arrays in a vectorized manner
     _, pvals = test(*log_ratios)
+# run statistical test on the 2-D arrays in a vectorized manner
```
qiyunzhu: Missing indent.

skbio/stats/composition/_ancom.py
```diff
@@ -403,24 +432,36 @@ def _log_compare(matrix, labels, n, test):
...
+    _, pvals = test(*log_ratios)
 
-    # populate p-value matrix
-    pval_mat = np.empty((m, m))
-    pval_mat[ii, jj] = pval_mat[jj, ii] = pvals
-    np.fill_diagonal(pval_mat, 0)
-
-    return pval_mat
+    # CRITICAL FIX: Force floats! If matrix was integers, this was rounding p-values to 0.
```
qiyunzhu: Why is this needed? I think an valid `test` won't return integer p-values.

skbio/stats/composition/_ancom.py
Sriniketh99: I tested ingest_array with a list_iterator locally. While it doesn't raise an error, it returns an array containing the iterator object itself (e.g., <numpy.ndarray> of type object) rather than converting the elements of the iterator into an array of floats. Because we need to perform boundary checks like xp.any(percentiles < 0.0), the explicit list() conversion is required to ensure the values are correctly ingested.

skbio/stats/composition/_ancom.py
Sriniketh99: You're right that in the NumPy backend, .size returns a standard int. I initially included the int() cast as a safeguard for Array API compatibility with backends like PyTorch, where .size can return a 0-d tensor.

skbio/stats/composition/_ancom.py
Sriniketh99: I confirmed that if the input matrix consists of integers (counts), pval_mat inherits that integer dtype if initialized via matrix.dtype. My tests show that float p-values (e.g., 0.045) are silently truncated to 0 during assignment in this scenario. Forcing xp.float64 is required to maintain precision for count-based data.

skbio/stats/composition/_ancom.py
Sriniketh99: Fixed it in the latest commit.
Thank you for catching those formatting errors!!

skbio/stats/composition/_ancom.py
Sriniketh99: I corrected the indentation for the _log_compare logic to ensure it is properly contained within the function block.

skbio/stats/composition/_ancom.py
Sriniketh99: Thank You !!

skbio/stats/composition/_ancom.py
qiyunzhu: Useful information. @mataton We probably should fix the behavior of `ingest_array` such that it can take iterators as input.

qiyunzhu (commented):
@Sriniketh99 I think this PR is approaching a good shape. Can you fix the workflow errors? Try to run it locally and you can catch the errors before updating the PR.

PR #2395, opened 2026-03-01
xp wrapper implementation

Aaraviitkgp:
I have made this pr for xp wrapper implementation and its checklist as well for gsoc:

Alpha diversity metrics (alpha):
- [x] shannon
- [x] simpson
- [x] pielou_e

Multiplicative replacement (composition):
- [x] multi_replace
- [x] vlr

Differential abundance (DA) tests:
- [ ] ancom
- [x] dirmult_lme
- [x] dirmult_ttest

Augmentation techniques (table):
- [ ] phylomix
- [ ] compos_cutmix
- [ ] aitchison_mixup
- [ ] mixup

Ordination:
- [x] ca
- [x] cca
- [x] rda

Please complete the following checklist:

* [x] I have read the [contribution guidelines](https://scikit.bio/contribute.html).

* [ ] I have documented all public-facing changes in the [changelog](https://github.com/scikit-bio/scikit-bio/blob/main/CHANGELOG.md).

* [ ] **This pull request includes code, documentation, or other content derived from external source(s).** If this is the case, ensure the external source's license is compatible with scikit-bio's [license](https://github.com/scikit-bio/scikit-bio/blob/main/LICENSE.txt). Include the license in the `licenses` directory and add a comment in the code giving proper attribution. Ensure any other requirements set forth by the license and/or author are satisfied.

  - **It is your responsibility to disclose** code, documentation, or other content derived from external source(s). If you have questions about whether something can be included in the project or how to give proper attribution, include those questions in your pull request and a reviewer will assist you.

* [x] This pull request does not include code, documentation, or other content derived from external source(s).

Note: [This document](https://scikit.bio/devdoc/review.html) may also be helpful to see some of the things code reviewers will be verifying when reviewing your pull request.

Issue #2379, opened 2026-02-10
`dirmult_lme` tests: investigate cross-platform numerical inconsistencies


mataton:
`test_dirmult.py` has accumulated some commented-out checks related to cross-platform numerical optimization differences. Would it make sense to dig into what's driving the inconsistencies? That way we might be able to find a fix rather than needing to keep commenting things out.

Some examples:
```python
    # def test_dirmult_lme_fail_some(self):
    #     # With the BFGS method, LME model fitting will not converge on two features.
    #     # Output will be NaN.
    #     msg = "LME fit failed for 2 features in all replicates, reporting NaNs."
    #     with self.assertWarns(UserWarning) as cm:
    #         res = dirmult_lme(table=self.table, metadata=self.metadata,
    #                           formula="Covar2 + Covar3", grouping="Covar1",
    #                           draws=1, seed=0, p_adjust="sidak",
    #                           fit_method="bfgs", fit_converge=True)
    #     self.assertEqual(str(cm.warning), msg)
    #     self.assertTrue(res.query(
    #         "FeatureID == ['feature1', 'feature3']")["Log2(FC)"].isnull().all())
    #     self.assertTrue(res.query(
    #         "FeatureID == ['feature2', 'feature4']")["Log2(FC)"].notnull().all())
```
```python
    def test_dirmult_lme_demo(self):
        # a regular analysis
        res = dirmult_lme(
            table=self.table, metadata=self.metadata, formula="Covar2 + Covar3",
            grouping="Covar1", draws=1, seed=0, p_adjust="sidak")
        exp = """
  FeatureID Covariate  Reps  Log2(FC)   CI(2.5)  CI(97.5)    pvalue    qvalue  Signif
0  feature1    Covar2     1  2.708376 -0.818699  6.235451  0.132319  0.247129   False
1  feature1    Covar3     1  1.696770 -1.224053  4.617594  0.254876  0.444790   False
2  feature2    Covar2     1  0.956017 -0.736692  2.648726  0.268312  0.464632   False
3  feature2    Covar3     1  0.325451 -0.657935  1.308836  0.516565  0.766291   False
4  feature3    Covar2     1 -1.990268 -4.362246  0.381711  0.100061  0.190110   False
5  feature3    Covar3     1 -0.812892 -2.910615  1.284830  0.447548  0.694797   False
6  feature4    Covar2     1 -1.674125 -3.885791  0.537540  0.137915  0.256810   False
[...]

PR #2407, opened 2026-03-06
xp implementation for dirmult_ttest and dirmult_lme

Aaraviitkgp:
#2395 , #2391 there is the test resolve code in it aswell with it issue to when that gets resolved all the test will pass on the dirmult_toy test fails which is handled in #2391 pr 😁.

### DIFF
--- a/skbio/stats/composition/_dirmult.py
+++ b/skbio/stats/composition/_dirmult.py
@@ -311,8 +311,8 @@ def dirmult_ttest(
         # The final lower and upper bounds are the minimum and maximum of all lower
         # and upper bounds seen during sampling, respectively.
         lower_, upper_ = cm.tconfint_diff(alpha=0.05, **cm_params)
-        xp.minimum(lower, lower_, out = lower)
-        xp.maximum(upper, upper_, out = upper)
+        lower = xp.minimum(lower, lower_)
+        upper = xp.maximum(upper, upper_)
 
     # Normalize metrics to averages over all replicates.
     delta /= draws
@@ -640,7 +640,7 @@ def dirmult_lme(
     upper = xp.full(shape, -xp.inf)  # 97.5% CI
 
     # number of replicates (draws) LME fitting is successful for each feature
-    fitted = xp.zeros(n_feats, dtype=int)
+    fitted = xp.zeros(n_feats, dtype=xp.int64)
 
     fit_fail_msg = "LME fit failed for feature {} in replicate {}, outputting NaNs."
 
--- a/skbio/table/_augment.py
+++ b/skbio/table/_augment.py
@@ -98,7 +98,7 @@ def _validate_labels(labels: NDArray, n: int) -> tuple[NDArray, NDArray]:
                 "Labels must be consecutive integers from 0 to n_classes - 1."
             )
 
-        labels_one_hot = xp.eye(n_classes, dtype=int)[labels]
+        labels_one_hot = xp.eye(n_classes, dtype=labels.dtype)[labels]
         return labels, labels_one_hot
 
     # input as one-hot encoded labels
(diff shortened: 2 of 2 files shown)


Aaraviitkgp:
@qiyunzhu  by the way are you guys using joblib for parallelizing  in scikit-bio ?

skbio/table/_augment.py
```diff
@@ -139,7 +148,8 @@ def _all_pairs(n: int) -> NDArray:
         Pairs of sample indices.
 
     """
-    return np.column_stack(np.triu_indices(n, k=1))
+    xp, _ = ingest_array()
+    return xp.stack(xp.triu_indices(n, k=1))
```
qiyunzhu: This task requires not only testing the code with specific library and device, but also checking if the functions are part of the [Python array API standard](https://data-apis.org/array-api/latest/). I checked and found that `triu_indices` is not part of the standard. Therefore, we cannot safely use it.

There is a potential that we develop a mechanism to support individual libraries without the standard, but it will add burden to maintenance. We can revisit this later.

skbio/table/_augment.py
```diff
@@ -76,33 +77,39 @@ def _validate_labels(labels: NDArray, n: int) -> tuple[NDArray, NDArray]:
     if labels.ndim == 1:
         # if labels is not integer-type, make sure it contains only whole numbers,
         # then convert to int
-        if not np.issubdtype(labels.dtype, np.integer):
-            if not (np.mod(labels, 1) == 0).all():
-                raise ValueError(f"Labels must only contain integer values.")
-            labels = labels.astype(int)
+        if not xp.isdtype(labels.dtype, "integral"):
+            labels_float = xp.astype(labels, xp.float64)
```
qiyunzhu, line 81: Why `float64` specifically? I can see a range of reasons. But I wish to know your rationale.

skbio/stats/composition/_dirmult.py
```diff
@@ -635,12 +670,13 @@ class from statsmodels.
...
-
-                # There are many ways model fitting may fail. Examples are LinAlgError,
-                # RuntimeError, OverflowError, and ZeroDivisionError. If any error
-                # occurs, the function will still proceed but the current run will be
-                # discarded.
+                    # Use fallbacks to ensure cross-platform stability.
+                    # Default to lbfgs if no method is provided.
+                    result = _fit_with_fallbacks(
+                        model,
+                        primary_method=fit_method if fit_method else 'lbfgs',
```
qiyunzhu, line 677: Is `lbfgs` better supported by various platforms? Please provide rationales.

skbio/stats/composition/_dirmult.py
```diff
@@ -21,8 +21,40 @@
...
 )
+from skbio.util._array import ingest_array
 
 
+
+def _fit_with_fallbacks(model, primary_method='lbfgs', fit_kwargs=None):
+    kwargs = fit_kwargs if fit_kwargs is not None else {}
+
+    # List of valid statsmodels methods for MixedLM
+    valid_methods = ['lbfgs', 'bfgs', 'powell', 'cg', 'ncg', 'nm']
```
qiyunzhu, line 32: Where is this list of methods from?

skbio/table/_augment.py
Aaraviitkgp: @qiyunzhu   I told you about that in the main pr , I mean either we have to hard code this (which can degrade performance) or we can keep it np.

skbio/stats/composition/_dirmult.py
Aaraviitkgp: yup from what I have seen lbfgs is more stable compared to bfgs for hessians.

Aaraviitkgp: I have given proper explanation in the issue discussion itself #2379

skbio/stats/composition/_dirmult.py
Aaraviitkgp: These are the allowed solver methods used by LikelihoodModel.fit in statsmodels have skipped newton raphson though. here is link to it https://www.statsmodels.org/stable/_modules/statsmodels/base/optimizer.html

skbio/table/_augment.py
Aaraviitkgp: @qiyunzhu,  I mean firstly the range float64 gives is huge (i think there is no dataset currently present that has that many labels ) so i covers it all, secoundly the precision (related to first point) the float32, float16 will truncate the value which can cause two dinstinct labels to collapse into  same integer  and lastly I mean under this we have  xp.int64 ,  (for example we cannot safely compare xp.int64 with xp.float16) , in numpy from what I remember they would first convert to single dtype and then compare and the dtype they convert to is totally depends on the backend.

skbio/stats/composition/_dirmult.py
Aaraviitkgp: @qiyunzhu Both scipy and statsmodel seems to prefer lbfgs over bfgs

skbio/table/_augment.py
Aaraviitkgp: @qiyunzhu even random.generator would have to be looked at , same issue as triu_indices.

PR #2408, opened 2026-03-06
parallelize geomedian_axis_one with prange

AhmedKhairy22:
The inner loops of `geomedian_axis_one` were sequential because `_dist_euclidean(X[:, i], y)` creates a temporary Python memoryview slice on every call, which requires the GIL and blocks `prange`.

To fix this, I added a `_dist_euclidean_col` cdef helper that accesses the column by index directly with no slicing, making it fully nogil-safe. This allowed switching the three independent inner loops, distance computation, weight normalization, and the weighted sum update, from sequential to `prange`.

A follow-up fix pre-transposes `X` into a C-contiguous `XT` before the main loop, so that `XT[i, :]` is a contiguous row read inside the helper, correcting strided access.

### Benchmark (Intel i7, 10 cores)

Input: `np.random.rand(1_000_000, 100)`, float64, `maxiters=100`
| Version  | Time     |
|----------|----------|
| Original | ~4.5 s |
| This PR  | ~1.8 s |

### DIFF
--- a/skbio/stats/distance/_cutils.pyx
+++ b/skbio/stats/distance/_cutils.pyx
@@ -500,7 +500,8 @@ def geomedian_axis_one(floating[:, :] X, floating eps=1e-7,
 
     cdef floating dist, Dinvs, total, r, rinv, tmp, Di
     cdef size_t nzeros = n
-    cdef size_t iteration, i, j
+    cdef size_t iteration
+    cdef Py_ssize_t i, j
 
     XT_np = np.ascontiguousarray(X.T, dtype=dtype)
     cdef floating[:, ::1] XT = XT_np
@@ -558,10 +559,11 @@ def geomedian_axis_one(floating[:, :] X, floating eps=1e-7,
             
     return y
 
-cdef floating _dist_euclidean_col(floating[:, ::1] XT, size_t row, floating[:] y, size_t p) nogil:
+cdef floating _dist_euclidean_col(floating[:, ::1] XT, Py_ssize_t row, 
+                                    floating[:] y, Py_ssize_t p) nogil:
     cdef float64_t d = 0.
     cdef float64_t tmp
-    cdef size_t j
+    cdef Py_ssize_t j
     for j in range(p):
         tmp = XT[row, j] - y[j]
         d += tmp * tmp


AhmedKhairy22:
I've pushed the Py_ssize_t changes for the prange loop variables as you suggested @qiyunzhu, hoping that this resolves the Windows build failure.

Regarding @sfiligoi's note on benchmarking with larger matrices, I ran a benchmark on a 20000x20000 float32 matrix (~1.6 GB) using memory_profiler to capture both time and real memory overhead. Here are the results:

| Version                   | Time     | Peak Memory | Overhead   |
|---------------------------|----------|-------------|------------|
| Original (no changes)     | 34.21 s  | 1697.36 MB  | 0.38 MB    |
| prange only (no transpose)| 23.59 s  | 1697.40 MB  | 0.25 MB    |
| prange + transpose        | 18.92 s  | 3223.09 MB  | 1526.10 MB |

A few observations from these numbers:

prange alone without the transpose already gives a meaningful speedup (~31% faster) with virtually zero memory overhead, only 0.25 MB on top of the base. This feels like the safest improvement.

Adding the transpose on top brings an additional speedup (~20% more) but at a cost of ~1526 MB of extra memory for a 20000x20000 float32 matrix. For float64, that overhead would double to ~3 GB, and for the 50k x 50k case, I guess it would be completely impractical.

So, what I think is to keep prange only as a more suitable approach, as the transpose trick I thought might be faster, but the memory cost makes it a poor fit here.

I'm happy to drop the transpose and keep only the prange changes if that aligns with your thinking. and i'm Curious to hear what you think

sfiligoi:
@AhmedKhairy22 Given that the speed improvement from the transpose is not drastic, I agree it is not worth the doubling of memory usage.

sfiligoi:
@AhmedKhairy22 Given the obvious benefits of using a transpose, have you checked where this function is called from, and if the input matrix is symmetrical already?
Most distance matrices are indeed symmetric, with X == X.T.
So, it may be worth exploiting that.

sfiligoi:
That said...
if the input matrices are really always tiny, it may not be worth optimizing to start with.
Plus, on top of a trivial starting runtime,
on a 10x10 matrix `prange` is likely (significantly) slower than sequential `range`, due to thread starting overhead.

Issue #2376, opened 2026-02-10
Paralellization of dirmult_lme


mataton:
### Discussed in https://github.com/scikit-bio/scikit-bio/discussions/2374

<div type='discussions-op-text'>

<sup>Originally posted by **The-Skov** February  4, 2026</sup>
I am using dirmult_lme as i believe it is the most appropriate for my use case, but as there is a very large amount if iterations that independent, this seems like a prime candidate for parallelization.
I have done it locally by subsetting the function calls to match the core count with great succes and obviously massive speed gains.
I have no idea how to implement this in a safe for everyone way and I do not have the proper experience to feel comfortable contributing myself so just wanted to throw this out there. :)</div>

Issue #2280, opened 2025-08-15
Use unifrac-binaries as implementation of unifrac functionality


sfiligoi:
The unifrac implementation in the unifrac-binaries library has been extensively used by several communities, but it is currently not directly accessible through the scikit-bio interface (relying instead on a separate python layer).

We should fix that, by adding the necessary library calls into scikit-bio itself.

ebolyen:
If this happens, I would recommend scikit-bio provide a thread parameter with a default of 1 or 2 as it's definitely bitten us over in QIIME 2. I realize the point is the multi-threading, but I think as a user of scikit-bio, defaulting to all-cores is somewhat unexpected and it gets really gnarly when you launch multiple processes (this took out one of our workshop servers once).

Here's how we accomplish that:

https://github.com/qiime2/q2-diversity-lib/blob/4565c49cd10620a58595555564c3d9fbe370ec44/q2_diversity_lib/beta.py#L203-L222

Although perhaps there's a cleaner way these days.

sfiligoi:
@ebolyen You do realize scipy, which is used under the hood in scikit-bio, is also multithreaded by default these days?
The only safe way for running multiple processes these days is to explicitly setting OMP_NUM_THREADS (or use taskset)... virtually all library respect this (and the batch system usually does this for you)

wasade:
NumPy also now can use OpenMP. I believe some of the underlying libraries it can optionally link against also make the same assumptions Igor notesOn Aug 15, 2025, at 12:01, Evan Bolyen ***@***.***> wrote:﻿ebolyen left a comment (scikit-bio/scikit-bio#2280)
No I had not realized that, although it is obvious in retrospect.
Maybe a quick note in the docs then? This only stood out to us because Galaxy does not appear to set that by default so we really thrashed the machine. It must be that any of the other functionality in QIIME 2 does not generally take long enough for this to matter. I'm a little surprised we hadn't encountered this with PCoA.

qiyunzhu:
I am prone to having a `threads` parameter at function level, as well as in the [global configuration](https://scikit.bio/docs/latest/config.html), which was added in skbio 0.7.0. I am hoping to implement it in skbio 0.7.1, due by Oct 2025 (after Python 3.14 release).

[...]

PR #2409, opened 2026-03-07
parallelizing dirmult_lme 

Aaraviitkgp:
#2395  #2376 

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,8 @@ dependencies = [
     "biom-format >= 2.1.16",
     "statsmodels >= 0.14.0",
     "patsy >= 0.5.0",
-    "array-api-compat >= 1.3"
+    "array-api-compat >= 1.3",
+    "joblib >= 1.3.0"
 ]
 
 [project.urls]


Aaraviitkgp:
@sfiligoi this is the final benchmarking script:

~~~python 
import time
import gc
import warnings
import multiprocessing
import sys
import os

import numpy as np
import pandas as pd

# Allow importing original implementation from o.py
sys.path.append(os.getcwd())

# New implementation
from skbio.stats.composition import dirmult_lme

# Original implementation
try:
    from o import dirmult_lme as dirmult_lme_original
    HAS_ORIGINAL = True
except ImportError:
    HAS_ORIGINAL = False


# DATA GENERATION

def generate_mock_data(n_samples=400, n_features=1500):

    rng = np.random.default_rng(42)

    counts = rng.integers(1, 1000, size=(n_samples, n_features))

    table = pd.DataFrame(
        counts,
        index=[f"s{i}" for i in range(n_samples)],
        columns=[f"f{i}" for i in range(n_features)],
    )

    group_size = 3
    n_groups = n_samples // group_size
    n_total = n_groups * group_size

    table = table.iloc[:n_total]

    treatment = np.tile([0, 1], n_total // 2 + 1)[:n_total]

    metadata = pd.DataFrame({
        "patient": np.repeat(np.arange(n_groups), group_size),
        "time": np.tile(np.arange(group_size), n_groups),
        "treatment": treatment
    }, index=table.index)

    return table, metadata


# MODEL RUNNERS
def run_model(table, metadata, n_jobs, draws):

    with warnings.catch_warnings():
        warnings.simplefilter("ignore")

        dirmult_lme(
            table,
            metadata,
            formula="time + treatment",
            grouping="patient",
            draws=draws,
            n_jobs=n_jobs,
            seed=42,
        )


def run_original(table, metadata, draws):

    if not HAS_ORIGINAL:
        raise RuntimeError("Original implementation not available")

    with warnings.catch_warnings():
        warnings.simplefilter("ignore")

        dirmult_lme_original(
            table,
            metadata,
            formula="time + treatment",
            grouping="patient",
            draws=draws,
            seed=42,
        )


# BENCHMARK

def benchmark(func, repeats=2):

    times = []

    for _ in range(repeats):

        gc.collect()

        start = time.perf_counter()
        func()
        end = time.perf_counter()

        times.append(end - start)

    return np.median(times)



# MAIN


def main():

    n_samples = 400
    n_features = 100
    draws = 128

    print("=" * 60)
    print("DIRICHLET-MULTINOMIAL LME BENCHMARK: ORIGINAL VS. PARALLEL")
    print("=" * 60)

    print("CPU cores:", multiprocessing.cpu_count())
    print(f"Dataset: {n_samples} samples, {n_features} features")
    print(f"Draws: {draws}")
    print()

    table, metadata = generate_mock_data(n_samples, n_features)

    results = []


    # WARMUP RUN


    print("Running warmup...")
    run_model(table.copy(), metadata.copy(), 1, draws)
    print("Warmup complete\n")


    # ORIGINAL IMPLEMENTATION


    if HAS_ORIGINAL:

        print("Benchmarking original serial implementation...")

        t_original = benchmark(
            lambda: run_original(table.copy(), metadata.copy(), draws)
        )

        print(f"Original Time: {t_original:.2f}s\n")

        results.append({
            "Mode": "Original (Serial)",
            "Time (s)": round(t_original, 2),
            "Speedup": 1.0
        })

    else:

        print("Original implementation not found in o.py\n")
        t_original = None

    # PARALLEL SCALING

    jobs_list = [1, 2, 4, 8, 12]

    t_serial = None

    for jobs in jobs_list:

        print(f"Benchmarking new implementation (n_jobs={jobs})...")

        t = benchmark(
            lambda: run_model(table.copy(), metadata.copy(), jobs, draws)
        )

        if jobs == 1:
            t_serial = t

        if t_original:
            speedup = t_original / t
        else:
            speedup = t_serial / t if t_serial else 1.0

        print(f"Time: {t:.2f}s | Speedup: {speedup:.2f}x\n")

        results.append({
            "Mode": f"New (n_jobs={jobs})",
            "Time (s)": round(t, 2),
            "Speedup": round(speedup, 2)
        })

    # SUMMARY

    print("-" * 60)
    print("Benchmark Summary\n")

    summary_df = pd.DataFrame(results)

    print(summary_df.to_string(index=False))

    print("=" * 60)


if __name__ == "__main__":
    main()
~~~

qiyunzhu:
@Aaraviitkgp I think this is a decent PR. Your code is a nice attempt to address a big performance bottleneck in scikit-bio (current `dirmult_ttest` is way too slow).

We also discussed this internally and had some plans (cc'ing @mortonjt). Basically, `MixedLM` is the source of slow-down. This class from statsmodels cannot vectorize over multiple features (unless SciPy's statistical tests, most of which can vectorize). To substantially improve the performance of `dirmult_ttest`, we will need to re-implement `MixedLM` from scratch, which isn't easy but might be worth doing, given high-dimensional data are so common in biology.

Regarding parallelization, I have an overall neural or positive feeling about `joblib`, which is also the parallelization engine of scikit-learn. However, our team had some discussions on how to do parallelization in scikit-bio. See #2280 for some of our discussions. Our current consensus is that scikit-bio should more closely follow SciPy instead of scikit-learn. In SciPy (and NumPy), threading is implicitly performed by the algebra engines. If one wants to control threading, one needs to do it in a low level such as exporting environment variables or using `threadpoolctl`. Doing multiprocessing could be tricky, because it is hard to control the number of threads each process actually uses, given a mixture of Python / Cython / NumPy / SciPy. I think (without support) that `joblib` can take care of this. But the implementation of a new parallelization engine will certainly require careful planning and discussion.

mortonjt:
Thanks @Aaraviitkgp for the implementation. I'm OK with the job lib dependency and am OK with enabling multi-threaded over samples now in order to improve usability of this methodology in the near future (which would be different from a new implementation, since it would open up more possibilities).

One other thing we probably want to do is have a test for correctness. I noticed that you did specify the random seeds in this new method, which is great, but we probably want to have a unit test just to confirm those random seeds. One scenario we don't want to have is to have the same random seed given to every worker or consistently given to the different workers, because in order to build established randomness, otherwise the p-values would be wrong.

PR #2414, opened 2026-03-10
Support sparse matrices in vectorize_counts_and_tree

### LINKED ISSUE #2268: vectorize_counts_and_tree is not memory efficient
Right now, vectorize_counts_and_tree requires a dense matrix as input.
We may want to investigate leverage sparse operations to aggregate internal nodes across the tree to handle large biom tables.
See below for a "vibe-coded" version of this using a taxonomic tree (extracted from metaphlan feature ids).

``` python

def vectorize_counts_and_taxonomy_sparse(table, delimiter='|', root_label='root'):
    """
    Memory‑efficient phylogenetic decomposition via sparse matrix multiplication.

    Returns a biom.Table of (nodes × samples) counts,
    aggregating every feature (ASV) up its taxonomy path.
    """
    # 1) Extract the sparse samples×features matrix
    #    biom_table.matrix_data is features×samples, so transpose:
    X = table.matrix_data.transpose().tocsr()
    sample_ids = table.ids(axis='sample')
    feature_ids = table.ids(axis='observation')

    # 2) Build a mapping from each feature-index → each node-index
    node_idx = {}   # node_label → column in P
    rows, cols, data = [], [], []
    for j, feat in enumerate(feature_ids):
        parts = [p.strip() or root_label for p in feat.split(delimiter)]
        for label in parts:
            idx = node_idx.setdefault(label, len(node_idx))
            rows.append(j)
            cols.append(idx)
            data.append(1.0)

    # 3) Construct the features×nodes sparse map P
    P = sp.csr_matrix(
        (data, (rows, cols)),
        shape=(len(feature_ids), len(node_idx))
    )

    # 4) Multiply: (samples×features) × (features×nodes) → (samples×nodes)
    T = X @ P    # still a sparse matrix

    # 5) Build the new BIOM table (obs×samples) = (nodes×samples)
    node_ids = [lbl for lbl, _ in sorted(node_idx.items(), key=lambda x: x[1])]
    # transpose T so that rows=nodes, cols=samples
    out_table = Table(T.transpose().tocsr(), node_ids, sample_ids)
    # make sure we didn't drop the original tip counts
    #out_table = table.merge(out_table)    

    return out_table

```

r0hansaxena:

---

Closes #2268

### Description
Adds support for sparse matrix inputs to vectorize_counts_and_tree, improving memory efficiency when working with large biological datasets. Internally routes to a new sparse-aware helper that uses scipy sparse operations instead of dense matrix expansion

### DIFF
--- a/benchmark_sparse.py
+++ b/benchmark_sparse.py
@@ -17,9 +17,7 @@ from skbio.diversity._util import vectorize_counts_and_tree
 
 def create_star_tree(n_tips):
     """Create a simple star tree with n_tips tips."""
-    newick = "({})root;".format(
-        ",".join(f"t{i}:1.0" for i in range(n_tips))
-    )
+    newick = "({})root;".format(",".join(f"t{i}:1.0" for i in range(n_tips)))
     return TreeNode.read([newick])
 
 
@@ -42,8 +40,10 @@ def run_scaling_benchmark():
 
     results = {
         "tips": [],
-        "dense_time": [], "sparse_time": [],
-        "dense_mem": [], "sparse_mem": [],
+        "dense_time": [],
+        "sparse_time": [],
+        "dense_mem": [],
+        "sparse_mem": [],
     }
 
     print("=== Scaling benchmark (sparsity=5%) ===")
@@ -79,8 +79,10 @@ def run_sparsity_benchmark():
 
     results = {
         "sparsity": [],
-        "dense_time": [], "sparse_time": [],
-        "dense_mem": [], "sparse_mem": [],
+        "dense_time": [],
+        "sparse_time": [],
+        "dense_mem": [],
+        "sparse_mem": [],
     }
 
     print("\n=== Sparsity benchmark (n_tips=2000) ===")
@@ -104,7 +106,10 @@ def run_sparsity_benchmark():
         results["sparse_mem"].append(sm)
 
         nnz_pct = (counts_sparse.nnz / (n_samples * n_tips)) * 100
-        print(f"dense={dt:.3f}s/{dm:.1f}MB  sparse={st:.3f}s/{sm:.1f}MB  nnz={nnz_pct:.1f}%")
+        print(
+            f"dense={dt:.3f}s/{dm:.1f}MB  "
+            f"sparse={st:.3f}s/{sm:.1f}MB  nnz={nnz_pct:.1f}%"
+        )
 
     return results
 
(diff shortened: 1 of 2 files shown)


mortonjt:
Thanks @r0hansaxena for your contribution.

This is nicely scoped. I would like to see if a small-scale benchmark of the runtime and the memory consumption of the sparse vs dense implementations. This will be important for users specifying this functionality to know the tradeoffs for sparse vs dense implementations (and when they should consider switching to sparse). If it could be contributed as a tutorial with plots and documentation, that would be the most informative.

Other issues to consider

1. **Performance concern with `tree.find()` and `tree.path()`**: The propagation matrix construction iterates over every taxon and calls `tree.find(tax)` + `tree.path(tip_node)` per taxon. `find()` does a full tree traversal (O(n_nodes)) unless there's a cached lookup. For large trees (tens of thousands of tips), this is O(n_taxa × n_nodes) just to build `P`. Consider building `P` directly from `tree_index['child_index']` instead — you already have the indexed tree, you don't need to walk it again via `find`/`path`.

2. **`tip_to_index` includes internal nodes, not just tips**: The dict is called `tip_to_index` but it's built from `tree_index['name']` which includes *all* nodes. The name is misleading — call it `node_to_index` or similar.

3. **The error message says "not found in tree tips" but checks all nodes**: Related to the above — the `ValueError` message references "tree tips" but `tip_to_index` maps all nodes. If a taxon name matches an internal node name but not a tip, this would silently succeed when it probably shouldn't. The dense path handles this differently (it filters to observed IDs that exist in nodes).

4. **Duplicate entries in the COO data**: If `tree.path()` returns overlapping nodes (e.g., the root appears in `[tree] + tree.path(...)` and is also `tree` itself), you'll get duplicate `(i, j)` entries. `csr_matrix` sums duplicates, so the propagation weights would be >1 for those entries, producing incorrect counts. I'd verify whether `[tree] + tree.path(tip_node) + [tip_node]` can produce duplicates — it likely does for the root and/or the tip itself.

5. **Return type inconsistency**: `_nodes_by_counts_sparse` calls `.toarray()` at the end, returning a dense ndarray. This is fine for the current call site, but it means the "sparse" benefit is limited to the matmul — if `counts` is sparse because it's huge, densifying the `(n_nodes, n_samples)` result may still blow up memory. Worth documenting this tradeoff.

6. **No dtype control**: The dense path casts to `COUNT_DTYPE` (int64). The sparse path uses float64 (from the `1.0` in `data`). The output types will differ between the two paths. The test uses `npt.assert_equal` which may mask this if values are numerically equal but types differ.

7. **Test coverage is thin**: Only one test case (the existing dense test converted to sparse). Would be good to add: a 1-D sparse vector input, a case with taxa that are a subset of tree tips, and an error case for missing taxa.

**Suggested refactor for the core of `_nodes_by_counts_sparse`:**

Instead of walking the tree via `find`/`path`, build the propagation matrix from `tree_index['child_index']` — iterate `child_index` in reverse (postorder) and set parent = sum of children, same logic as `_traverse_reduce` but as a sparse matrix construction. This would be both faster and consistent with the dense path's semantics.

r0hansaxena:
thanks for the review @mortonjt ! 
I'll get started on those performance and naming refactors and add the extra test cases.

r0hansaxena:
I'll also follow up with the benchmark results and tutorial content once I have the runtime/memory comparisons ready so users can see the tradeoffs.

PR #2425, opened 2026-03-18
BUG: Fix multi_replace output type for Pandas 3.0 apply compatibility

MAUK9086:
### Description
This addresses the `multi_replace` bug mentioned by @mataton in #2375.

#### The Issue
In Pandas 3.0, `df.apply(multi_replace, axis=1)` fails to expand the result into a DataFrame. Because the function was returning a raw `np.ndarray` via `mat.squeeze()`, Pandas 3.0 collapses the result into a single `object` dtype column instead of a 2D structure. This specifically affects workflows in the `05-marker-inference` tutorial.

#### The Fix
Added type-awareness to `multi_replace`. It now captures whether the input is a `pd.Series` or `pd.DataFrame`, preserves the metadata (index/columns), and restores the output to the corresponding Pandas object. This ensures `apply(axis=1)` correctly unpacks the results back into a DataFrame.

#### Technical Note
I verified the fix locally using Pandas 3.0.0. While `pytest` shows `DocTestFailures`, these are pre-existing issues related to NumPy 2.0 whitespace formatting and are unrelated to the changes in this PR.

---

### Checklist

* [x] All tests pass locally (excluding pre-existing NumPy 2.0 doctest formatting issues).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Version 0.7.3-dev
 
+### Bug Fixes
+
+* Fixed an issue where `composition.multi_replace` failed to expand into a DataFrame when used 
+
 
 ## Version 0.7.2
 
--- a/skbio/stats/composition/_base.py
+++ b/skbio/stats/composition/_base.py
@@ -168,6 +168,20 @@ def multi_replace(mat, delta=None):
            [ 0.0625,  0.4375,  0.4375,  0.0625]])
 
     """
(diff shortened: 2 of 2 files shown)


qiyunzhu:
Nice trial! Although I doubt if this is the right direction. Calling the `ingest_array` helper function to parse the input array-like data (dataframe, series, raw array or anything), like other functions do in `composition`, should be the way to go.

qiyunzhu:
Closing as this code isn't in the desired form.

PR #2429, opened 2026-03-19
ENH: Add Array API support to centralize and vlr

LarytheLord:
`centralize()` and `vlr()` are the last two composition stats functions that still use
hard-coded NumPy / SciPy calls. This brings them in line with `closure`, `clr`, `ilr`,
`alr`, and their inverses, which already go through `ingest_array`.

### What changed

**`centralize()`**

- Replaced the `scipy.stats.gmean` import with `exp(mean(log(x)))`, which only uses
  operations available in the Array API standard.
- Uses `ingest_array` → `xp` namespace like the rest of the module.
- Added a `validate` parameter (consistent with `closure`, `clr`, etc.).

**`vlr()` / `_vlr()`**

- The non-robust path now receives the `xp` namespace from `_closure_two` and computes
  log-ratio variance using `xp.log`, `xp.mean`, `xp.sum`.
- The `robust=True` path stays on NumPy (`np.ma` masked arrays have no Array API
  equivalent). The docstring notes this explicitly.

**Tests**

- Added `Tests_Centralize` and `Tests_VLR` classes in `test_ndarray.py` with NumPy,
  PyTorch, and JAX backend checks (same pattern as `Tests_Closure`, `Tests_CLR`, etc.).

### What I verified locally

```
$ python -m pytest skbio/stats/composition/tests/test_base.py \
                   skbio/stats/composition/tests/test_ndarray.py -q
40 passed, 40 skipped
```

(Torch/JAX/CUDA tests are skipped locally — they'll run in CI where those deps are available.)

### DIFF
--- a/skbio/stats/composition/_base.py
+++ b/skbio/stats/composition/_base.py
@@ -979,13 +979,20 @@ def _alr_inv(xp: ModuleType, mat: StdArray, ref_idx: int, axis: int) -> StdArray
     return _closure(xp, comp, axis)
 
 
-def centralize(mat: ArrayLike) -> StdArray:
+def centralize(mat: ArrayLike, validate: bool = True) -> StdArray:
     r"""Center data around its geometric average.
+
+    .. versionchanged:: 0.7.3
+        The function now supports the Array API standard and no longer depends on SciPy.
 
     Parameters
     ----------
     mat : array_like of shape (n_compositions, n_components)
         A matrix of proportions.
+    validate : bool, default True
+        Check if the compositions are legitimate.
+
+        .. versionadded:: 0.7.3
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


skbio/stats/composition/_base.py
```diff
@@ -998,26 +1005,31 @@ def centralize(mat: ArrayLike) -> StdArray:
...
-    array([[ 0.17445763,  0.30216948,  0.34891526,  0.17445763],
-           [ 0.32495488,  0.18761279,  0.16247744,  0.32495488]])
+    array([[0.17445763, 0.30216948, 0.34891526, 0.17445763],
+           [0.32495488, 0.18761279, 0.16247744, 0.32495488]])
 
     """
-    from scipy.stats import gmean
-
-    mat = closure(mat)
-    cen = gmean(mat, axis=0)
```
sfiligoi, line 1008: Does this have any impact on performance?
Both in terms of runtime and memory usage?

Especially when input is nparray.

skbio/stats/composition/_base.py
```diff
@@ -1109,22 +1127,14 @@ def vlr(x, y, ddof=1, robust=False):
...
-        "y": y,
-        "ddof": ddof,
-    }
-
-    # Run backend function
     if robust:
-        return _robust_vlr(**kwargs)
-    else:
-        return _vlr(**kwargs)
+        # robust path uses NumPy masked arrays; keep on numpy
```
sfiligoi, line 1131: Keeping the if... else.. pattern would be preferred.

sfiligoi:
Close as obsolete.

PR #2431, opened 2026-03-21
Fix permdisp mac ci

### LINKED ISSUE #2430: "EIGH: since no value for dimensions is specified ..." when running PERMDISP
I was running PERMDISP for some data I was running PERMANOVA on. I'm bombarded with these warnings:
```
_principal_coordinate_analysis.py:164: RuntimeWarning: EIGH: since no value for dimensions is specified, PCoA for all dimensions will be computed, which may result in long computation time if the original distance matrix is large.
```

But why is this warning there? The documentation for the dimensions parameter says:
> Dimensions to reduce the distance matrix to if using the fsvd method. Not used if the eigh method is being selected.

I found out, this warning is printed when a matrix is larger than 10×10 (btw, is 10x10 really "large"?):
https://github.com/scikit-bio/scikit-bio/blob/5bb487a350282509d8da16bccb41aa6955bda2c1/skbio/stats/ordination/_principal_coordinate_analysis.py#L163-L170

Actually, the dimension is deliberately set to zero by the permdisp function, thus triggering this warning all the time:
https://github.com/scikit-bio/scikit-bio/blob/5bb487a350282509d8da16bccb41aa6955bda2c1/skbio/stats/distance/_permdisp.py#L271-L274

Sriniketh99:

### Fixes Issue
Fixes #2430 

When running `permdisp` with the `eigh` method on a distance matrix larger than 10x10, users are receive a spurious with a `RuntimeWarning` from `pcoa` stating that no dimensions were specified. 

This happens because `permdisp` intentionally passes `dimensions=0` to `pcoa` (since `eigh` natively computes all dimensions). However, `pcoa` still triggers a warning for this, which is confusing since the `scikit-bio` documentation explicitly states that the `dimensions` parameter is not used for `eigh`.

**Changes made:**
To fix this without altering the core `pcoa` internal logic, I wrapped the `pcoa` call inside `skbio/stats/distance/_permdisp.py` using `warnings.catch_warnings()`. 

This safely filters out the specific `EIGH` dimension warning for `permdisp` users, but leaves legitimate mathematical warnings (e.g., negative eigenvalues) completely intact.

### Testing
- Ran the full test suite for `permdisp` and `pcoa` (`pytest`); all tests pass perfectly.
- Verified locally with a custom >10x10 distance matrix script that the target `EIGH` warning is completely suppressed when calling `permdisp`, while underlying math warnings remain functional.


.gitignore
```diff
@@ -57,3 +57,5 @@ web/_build
 # Cython files
 *.c
 !ssw.c
+skbio-venv/
```
sfiligoi, line 60: This definitely does not belong to the PR.

sfiligoi:
Obsoleted by #2456

PR #2432, opened 2026-03-23
ENH: support int8/int16/int32 dtypes with fixed-point scale in distance matrices

### LINKED ISSUE #2290: Add support for INT-based matrices, mimicking fixed point semantics
We currently only support fp64 and fp32 values in distance matrices.
However, many problems do not need neither that much precision nor that much dynamic range.
Indeed, the semantics of many distance matrices is conceptually fix point. 

Since there is no native support for fixed point representation in modern hardware,
I thus propose we simulate it with a combination of INT-based element values and a (single) matrix-wide multiplication factor.
With support for all of INT8, INT16, and INT32.

To clarify:
I am **not** proposing for this should be the default representation, just one option that users may optionally choose to use, e.g. to store the data, or when memory capacity is limited.

The various functions in skbio will obviously have to be adapted to deal with this new, optional format.

r0hansaxena:

---

Closes #2290

## Description

Adds optional fixed-point storage for distance matrices. Intended for cases where full floating-point precision is not required and memory footprint matters.

A matrix may now carry an optional scale factor, and integer dtypes (int8, int16, int32) are accepted as input. When a scale is set, stored integer values represent real distances multiplied by the scale. The scale is a per-matrix property. Integer data without a scale is rejected at construction, since it does not describe a meaningful distance.

The scale is preserved across filtering, copying, condensed/redundant conversion, and permutation, so downstream code does not need to rethread it. The on-disk formats persist it on write and restore it on read.

Analysis code consumes distance matrices through a single conversion point that applies the scale and returns floats. Consumers never see raw integer values, and dtype is preserved rather than widened by default. This covers every distance-matrix consumer in the library.

Opting in means constructing the matrix from integer data with a scale; analysis functions accept it transparently.


sfiligoi:
@r0hansaxena What happens if you pass a fixed-point DM to any non-trivial function?
E.g. pcoa?

I see you added the tests for the DM class itself, but I don't think any other functionality of skbio is automatically exercised right now, right?

skbio/io/format/lsmat.py
```diff
@@ -179,8 +187,6 @@ def _lsmat_to_matrix(cls, fh, delimiter, dtype):
     num_ids = len(ids)
     data = np.empty((num_ids, num_ids), dtype=dtype)
 
-    # The default separator of str.split (consecutive whitespaces) is equivalent to " "
```
sfiligoi, line 182: Why were most comments stripped out?

r0hansaxena:
> @r0hansaxena What happens if you pass a fixed-point DM to any non-trivial function? E.g. pcoa?
> 
> I see you added the tests for the DM class itself, but I don't think any other functionality of skbio is automatically exercised right now, right?

You're right, the tests currently only cover the DM class. I'll add the integration tests with downstream functions like `pcoa` to verify everything works.

skbio/io/format/lsmat.py
r0hansaxena: That was an oversight on my part, I trimmed them during a cleanup pass. I'll restore them.

skbio/stats/distance/_base.py
```diff
@@ -527,16 +552,19 @@ def filter(
                     pass
             ids = found_ids
 
-        # Note: Skip validation, since we assume self was already validated
```
sfiligoi, line 530: Another removed comment.

sfiligoi (changes requested):
I see you made changed to pcoa and added pcoa specific tests.

But it is just one of the many functions that use the DM objects;
and if it needed changes, many/most of the others will need it, too.

I am even more convinced the current PR is not complete, and cannot be merged as-is.

sfiligoi:
@qiyunzhu @mataton What is your take?

Should we allow to have a DM code that provides new capabilities, but cannot be ingested by some/most of the functions?
If so, what amount of "external use" testing would we want? (Beyond `pcoa`, which is already part of this PR)
We definitely want to make sure the new structures are usable, and performant, in practice.

And what guardrails should we use?  Should we just
a) Document it as a limitation?
b) Issue a runtime warning if any user tries to use the new features?
c) Just let it fail in functions that do not support the current functionality?

I see pros and cons to all of the above 4 options.

Note: The new code does not seem to break any existing functionality, but it is inevitable that some users will (eventually) try to use the new functionality, if available.

r0hansaxena:
> I see you made changed to pcoa and added pcoa specific tests.
> 
> But it is just one of the many functions that use the DM objects; and if it needed changes, many/most of the others will need it, too.
> 
> I am even more convinced the current PR is not complete, and cannot be merged as-is.

Thanks for the feedback! I initially did it for just `pcoa` but I'll be extending this PR with fixed-point integration tests for the main `DistanceMatrix` consumers in `stats.distance` (`permanova`, `anosim`, `permdisp`, `mantel`)

skbio/stats/distance/_base.py
```diff
@@ -1043,8 +1073,11 @@ def _validate_shape(self, data: NDArray) -> None:
             raise PairwiseMatrixError(
                 "Data must be square (i.e., have the same number of rows and columns)."
             )
-        if data.dtype not in (np.float32, np.float64):
-            raise PairwiseMatrixError("Data must contain only floating point values.")
+        if data.dtype not in _ALLOWED_DTYPES:
```
sfiligoi, line 1080: I think we need to ensure we have a scaling factor, if the values are integer.

skbio/stats/distance/_base.py
```diff
@@ -1382,8 +1420,11 @@ def _validate_shape(self, data: NDArray):
                 f"Found {data.ndim} dimensions."
             )
 
-        if data.dtype not in (np.float32, np.float64):
-            raise PairwiseMatrixError("Data must contain only floating point values.")
+        if data.dtype not in _ALLOWED_DTYPES:
```
sfiligoi, line 1431: Same as above, int only valid if we have a scaling factor.

skbio/stats/distance/_base.py
```diff
@@ -2295,11 +2349,43 @@ def _preprocess_input(
...
+        Matrix values ready for numeric computation. Integer buffers and/or
+        fixed-point matrices are converted to floating point with scale applied.
+
+    """
+    distances = distance_matrix.condensed_form() if condensed else distance_matrix.data
+
+    if distance_matrix.scale is None and distances.dtype in _FLOAT_DTYPES:
+        return distances
+
+    distances = distances.astype(np.float64, copy=False)
```
sfiligoi: This will force distances to be float64, even if the original was float32 and scale is float32, too.

skbio/stats/distance/_bioenv.py
```diff
@@ -14,6 +14,7 @@
 from scipy.stats import spearmanr
 
 from skbio.stats.distance import DistanceMatrix
+from skbio.stats.distance._base import _extract_distance_matrix_data
```
sfiligoi: Are we sure we want to export a "internal function" (starting with _)?
Wouldn't it be better if we promoted it to a trully public function, if we expect it to be generally usable?

qiyunzhu: This helper function may be generally useful across the codebase, but it probably isn't to the users. So I don't think it needs to be a public-facing function. We can delete the pretrailing underline (`_`) but we probably don't need to import it in `__init__.py` or add it to the documentation.

skbio/stats/distance/_base.py
r0hansaxena: I added a check so integer matrices require scale, otherwise we raise an error, also added the tests.

skbio/stats/distance/_base.py
r0hansaxena: Same fix applied here as well.

skbio/stats/distance/_bioenv.py
r0hansaxena: This helper is intended for internal reuse across modules, not as a user-facing API. I renamed it for internal clarity.

sfiligoi:
@r0hansaxena Can you expand the PR description to include a (semi-detailed) explanation of how this is implemented?
I.e. what new fields have you added, how they are used (i.e. their semantics), high level implications on the users of the new functionality and any backwards compatibility considerations.

Basically, I want to have a logical view of what this PR does, so anyone can understand the implication of this change without having to look at the details of the code changes.
(I will obviously review to make sure the code does what you claim it is doing, but nobody else should need to do that)

No need to describe the changes to the tests, just how the DMs have been changed (although you should mention any useful support functionality).

r0hansaxena:
@sfiligoi I've updated the PR description, let me know if any further changes are required!

skbio/stats/distance/_base.py
```diff
@@ -134,8 +138,16 @@ def __init__(
         data: NDArray | Sequence[float] | Sequence[Sequence[float]] | PairwiseMatrix,
         ids: Sequence[str] | None = None,
         validate: bool = True,
+        scale: float | None = None,
```
sfiligoi, line 141: You did not document `scale` in the Class comment.

skbio/stats/distance/_base.py
```diff
@@ -1043,8 +1077,11 @@ def _validate_shape(self, data: NDArray) -> None:
             raise PairwiseMatrixError(
                 "Data must be square (i.e., have the same number of rows and columns)."
             )
-        if data.dtype not in (np.float32, np.float64):
-            raise PairwiseMatrixError("Data must contain only floating point values.")
+        if data.dtype not in _ALLOWED_DTYPES:
```
sfiligoi, line 1080: `_validate_shape` should get a `scale` parameter, and check that data+scale make sense together.
I believe that's the intended semantics of this method.

sfiligoi: PS: We may have redundant checks in some other places, but this function in particular should be self-consistent.

skbio/stats/distance/_base.py
```diff
@@ -1382,8 +1428,11 @@ def _validate_shape(self, data: NDArray):
                 f"Found {data.ndim} dimensions."
             )
 
-        if data.dtype not in (np.float32, np.float64):
-            raise PairwiseMatrixError("Data must contain only floating point values.")
+        if data.dtype not in _ALLOWED_DTYPES:
```
sfiligoi, line 1431: As above

skbio/stats/distance/_base.py
```diff
@@ -1101,7 +1138,11 @@ def __init__(
         validate: bool = True,
         condensed: bool = False,
         diagonal: float | NDArray = None,
+        scale: float | None = None,
```
sfiligoi, line 1141: As above, document `scale`.

skbio/stats/distance/_base.py
```diff
@@ -1987,6 +2051,7 @@ def __init__(
         self._diagonal = 0.0
         self._data = self._init_data(data, condensed)
         self._flags = self._init_flags(condensed)
+        self._scale = scale
```
sfiligoi, line 2054: As above, document `scale`.

skbio/stats/distance/_base.py
```diff
@@ -2295,11 +2361,54 @@ def _preprocess_input(
...
+            data = distance_matrix.data
+            tri_idxs = np.triu_indices(data.shape[0], k=1)
+            distances = data[tri_idxs]
+    else:
+        distances = distance_matrix.data
+
+    if distance_matrix.scale is None and distances.dtype in _FLOAT_DTYPES:
+        return distances
+
+    if distance_matrix.scale is None:
```
sfiligoi, line 2402: I don't understand the logic behind this.
I thought having non-float data and no scale was illegal.

sfiligoi:
I was looking at `condensed_form`, which is one of the methods that returns a NDArray.
Right now, it will not take into account scale at all.

I don't think this is the right semantics.
The returned NDArray should be self-consistent.
At least the default should be the scaled FP variant...

But maybe we want to have the option of returning the original (unscaled) data.
the caller can then apply any scaling as a post-processing step.
But I don't believe it should not be the default.

PS: There may be other instances where we return NDArray, and the same logic would apply.

skbio/stats/distance/_bioenv.py
```diff
@@ -186,7 +187,7 @@ def bioenv(distmat, data_frame, columns=None):
     # columns within a tight loop and using a numpy array ends up being ~2x
     # faster.
     vars_array = _scale(vars_df).values
-    dm_flat = distmat.condensed_form()
+    dm_flat = extract_distance_matrix_data(distmat, condensed=True)
```
sfiligoi, line 190: I believe `condensed_form` should have already returned the right data, no need to explicitly call `extract_distance_matrix_data`.

skbio/io/format/binary_dm.py
```diff
@@ -181,6 +187,9 @@ def _skbio_mat_to_h5py_mat(obj, f):
     np_ids = np.array(b_ids)
     f.create_dataset("order", data=np_ids)
     f.create_dataset("matrix", data=obj.data)
+    scale = getattr(obj, "scale", None)
```
sfiligoi, line 190: At the very least, we must document the new element in the hdf5 data structure.

But that's not enough.
We are effectively changing the semantics of the on-disk format, so the `version` must be updated, too, when the "scale" element is added.

We likely want to keep the old format version for when `scale` is not present, to maximize backward compatibility.

skbio/io/format/lsmat.py
```diff
@@ -68,8 +68,13 @@
 
 dtype : str or dtype, optional
     The data type of the underlying matrix data. Only relevant when reading from a
-    file. Default is "float64", which maps to ``np.float64``. The only other available
-    option is "float32" (or ``np.float32``).
+    file. Default is "float64", which maps to ``np.float64``. Other available options
+    include "float32", "int8", "int16", and "int32".
```
sfiligoi, line 72: We should clarify that the int types are only valid when a scale is provided, too.

skbio/io/format/phylip_dm.py
```diff
@@ -369,18 +370,25 @@ def _matrix_to_phylip_dm(obj, fh, delimiter, layout):
             " two samples in the matrix."
         )
     ids = obj.ids
+    data = extract_distance_matrix_data(obj)
+    if data.ndim == 1:
```
sfiligoi, line 374: Where this conditional logic come from?
I don't see it in the original logic.

skbio/stats/distance/_permanova.py
```diff
@@ -205,14 +206,22 @@ def permanova(
     # if we got here, we could not use skbb
     # Calculate number of objects in each group.
     group_sizes = np.bincount(grouping)
-    s_T = (distmat.data**2).sum() / sample_size
-    if not distmat._flags["CONDENSED"]:
+    is_condensed = distmat._flags["CONDENSED"]
+    distance_data = extract_distance_matrix_data(distmat, condensed=is_condensed)
```
sfiligoi, line 210: This is too late.
distmat is already used 20 lines above.

sfiligoi:
Just for the record:
I went over all non-test code, so the above requests are pretty much a complete list at this commit point.

In summary, I think the direction is solid, but needs a bit of polishing.

skbio/io/format/lsmat.py
```diff
@@ -263,9 +271,28 @@ def _parse_data(fh, delimiter):
...
+    fh.seek(0)
+    return scale
+
+
 def _matrix_to_lsmat(obj, fh, delimiter):
     delimiter = "%s" % delimiter
     ids = obj.ids
+
+    if getattr(obj, 'scale', None) is not None:
+        fh.write("# scale: %s\n" % obj.scale)
```
sfiligoi, line 295: I am starting to think we **do not want** to do this,
as it is effectively changing the on-disk semantics, and would make the file unreadable by most existing tools.

I think it would be much better to always write out the scaled-float numbers, instead.
This format is optimized for ease of use, not compactness.

@qiyunzhu Please let me know if you disagree.

skbio/io/format/binary_dm.py
sfiligoi: I would recommend naming the new version `b"2026.04"`.

skbio/stats/distance/_base.py
r0hansaxena: I kept this in `__init__` since the int-requires-scale rule isn't really a property of the array itself, more of a cross-input check between `data` and `scale`. `_validate_shape` is just looking at the array.

r0hansaxena:
> I was looking at `condensed_form`, which is one of the methods that returns a NDArray. Right now, it will not take into account scale at all.
> 
> I don't think this is the right semantics. The returned NDArray should be self-consistent. At least the default should be the scaled FP variant...
> 
> But maybe we want to have the option of returning the original (unscaled) data. the caller can then apply any scaling as a post-processing step. But I don't believe it should not be the default.
> 
> PS: There may be other instances where we return NDArray, and the same logic would apply.

Yeah makes sense. Default's scaled floats now, `raw=True` gets  the underlying buffer if it's needed. Also cleaned up bioenv since the helper call was redundant.

skbio/io/format/binary_dm.py
r0hansaxena: Yes, I've documented the `scale` dataset, the version bumps to `2026.04` only when scale is present and the old files keep `2020.12`.

skbio/io/format/lsmat.py
r0hansaxena: Would it make more sense to always emit scaled floats and drop the int-write path? Compact int storage still works via binary_dm. Wanted to check before changing it.

skbio/io/format/phylip_dm.py
r0hansaxena: That was covering for the helper returning 1D when the matrix is stored condensed. Moved the squareform into the helper.

sfiligoi:
@r0hansaxena Some of the tests are failing.

sfiligoi:
@r0hansaxena Still interested in completing this, or should we close it?

sfiligoi:
Closing this due to lack of progress.

PR #2436, opened 2026-03-28
ENH: Add Array API support to center_distance_matrix

LarytheLord:
`center_distance_matrix()` currently only dispatches to the Cython implementation, which requires NumPy arrays via the CPython buffer protocol. This means GPU-backed arrays (CuPy, JAX, PyTorch) cannot pass through `center_distance_matrix()` or the PCoA pipeline that depends on it.

### What changed

- NumPy arrays continue using the optimized Cython path (zero regression)
- Non-NumPy arrays now go through a pure-xp fallback implementing the same Gower centering (Eqs. 9.20 and 9.21, Legendre & Legendre 1998) using `ingest_array` and `xp.mean`
- Array type and device are preserved for GPU backends
- `inplace=True` is documented as NumPy-only

### Tests

- Added `TestCenterDistanceMatrixArrayAPI` with NumPy, PyTorch, and JAX backends
- Validates type preservation, device preservation, and numerical correctness against `f_matrix(e_matrix())` reference
- All existing ordination tests pass (144 passed, 0 failures)

### Numerical agreement

Cython vs xp path max absolute difference at n=100: 8.88e-16

### Context

This follows the same `ingest_array` + `xp` namespace pattern established in the composition module (centralize, vlr, closure, clr, etc.) and extends Array API support into the ordination module for the first time.

Related: #2429

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * Added support to `TreeNode.from_taxonomy` for parsing taxonomic lineage strings into trees with an optional `extract_rank` parameter. ([#2406](https://github.com/scikit-bio/scikit-bio/pull/2406))
 * Added `mmvec` (Microbe-Metabolite Vectors) to `skbio.stats.ordination` for learning joint embeddings of two feature sets from co-occurrence patterns. Supports L-BFGS and Adam optimizers, cross-validation via Q² scores, and prediction of one modality from another. ([#2360](https://github.com/scikit-bio/scikit-bio/pull/2360))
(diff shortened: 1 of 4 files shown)


sfiligoi:
@LarytheLord Is this still relevant?

LarytheLord:
No, this can be closed. The array API support for center_distance_matrix is already in development_numba_gpu via #2403 (the ingest_array path in _principal_coordinate_analysis.py), and #2508 built the Numba engine on top of that. This PR added it a different way in _utils.py, so it's now superseded. We can close it.

PR #2440, opened 2026-04-01
BUG: Avoid mutating caller-owned OrdinationResults in permdisp

### LINKED ISSUE #2438: BUG: avoid mutating caller-owned `OrdinationResults` in `permdisp`
### Description
`permdisp` appears to modify the `samples` DataFrame of an `OrdinationResults` object in-place by adding a `"grouping"` column when the input is an `OrdinationResults` instance (e.g., from `pcoa`).

This introduces an unexpected side effect and makes the input object unsafe to reuse in downstream analyses.

### Steps to reproduce

```
from skbio import DistanceMatrix
from skbio.stats.ordination import pcoa
from skbio.stats.distance import permdisp

dm = DistanceMatrix(
    [
        [0.0, 1.0, 2.0, 2.0],
        [1.0, 0.0, 2.0, 2.0],
        [2.0, 2.0, 0.0, 1.0],
        [2.0, 2.0, 1.0, 0.0],
    ],
    ["s1", "s2", "s3", "s4"],
)

po = pcoa(dm, warn_neg_eigval=False)
print("before:", po.samples.columns.tolist())
permdisp(po, ["G1", "G1", "G2", "G2"], permutations=0, warn_neg_eigval=False)
print("after: ", po.samples.columns.tolist())
```

### Observed behavior

```
before: ['PC1', 'PC2', 'PC3', 'PC4']
after:  ['PC1', 'PC2', 'PC3', 'PC4', 'grouping']
```

### Expected behavior

`permdisp` should not modify the input `OrdinationResults` object.
The `samples` DataFrame should remain unchanged after the function call.

### Impact

This behavior can lead to:
- Hidden state mutation in analysis pipelines
- Reusing the same `OrdinationResults` object after permdisp exposes an extra `"grouping"` column in `samples`
- Confusion or bugs in downstream steps (e.g., plotting or metadata merging)

### Possible cause:

The implementation appears to take a direct reference to `ordination.samples` and then mutate it:

```
samples = ordination.samples
samples["grouping"] = grouping
```

### Workarounds:

Pass a deep copy of the ordination result:
```
import copy
permdisp(copy.deepcopy(po), grouping)
```

If the original `DistanceMatrix` is still available, call `permdisp` on that instead of the existing `OrdinationResults`:

```
permdisp(dm, grouping)
```
This avoids mutating `po`, but it recomputes the ordination internally and may not be a strict substitute for `permdisp(po, grouping)` in every workflow.

### Note

The mutation happens in _permdisp.py (line 290) and _permdisp.py (line 308).

r0hansaxena:

---

Closes #2438

### Description

`permdisp` was modifying the `samples` DataFrame of an `OrdinationResults` object in-place by adding a `"grouping"` column when the input was an `OrdinationResults` instance.

This PR fixes the bug by creating a shallow `.copy()` of `ordination.samples` alongside the initial setup so that the original object safely preserves its exact state without unexpected mutation side effects in downstream pipelines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 * Added support to `TreeNode.from_taxonomy` for parsing taxonomic lineage strings into trees with an optional `extract_rank` parameter. ([#2406](https://github.com/scikit-bio/scikit-bio/pull/2406))
 * Added `mmvec` (Microbe-Metabolite Vectors) to `skbio.stats.ordination` for learning joint embeddings of two feature sets from co-occurrence patterns. Supports L-BFGS and Adam optimizers, cross-validation via Q² scores, and prediction of one modality from another. ([#2360](https://github.com/scikit-bio/scikit-bio/pull/2360))
 
+### Bug Fixes
+
+* Fixed `permdisp` mutating the input `OrdinationResults` object by adding a `"grouping"` column to its `samples` DataFrame ([#2440](https://github.com/scikit-bio/scikit-bio/pull/2440)).
 
 ## Version 0.7.2
 
--- a/skbio/stats/distance/_permdisp.py
+++ b/skbio/stats/distance/_permdisp.py
@@ -331,5 +331,5 @@ def _config_med(x):
 
     Transpose the vector to be compatible with hd.geomedian.
     """
-    X = x.values
-    return pd.Series(np.array(geomedian_axis_one(X.T)), index=x.columns)
+    Xt = np.ascontiguousarray(x.to_numpy(copy=True).T)
+    return pd.Series(np.array(geomedian_axis_one(Xt)), index=x.columns)
--- a/skbio/stats/distance/_permdisp.py
+++ b/skbio/stats/distance/_permdisp.py
@@ -287,7 +287,7 @@ def permdisp(
     else:
         raise TypeError("Input must be a DistanceMatrix or OrdinationResults.")
 
-    samples = ordination.samples.copy()
+    samples = ordination.samples
 
     num_groups, grouping = _preprocess_input_sng(ids, sample_size, grouping, column)
 
@@ -305,17 +305,16 @@ def permdisp(
 def _compute_groups(samples, test_type, grouping):
     groups = []
 
-    samples["grouping"] = grouping
     if test_type == "centroid":
-        centroids = samples.groupby("grouping").aggregate("mean")
+        centroids = samples.groupby(grouping).aggregate("mean")
     else:  # median
         grouping_cols = samples.columns.to_list()
-        centroids = samples.groupby("grouping")[grouping_cols].apply(_config_med)
+        centroids = samples.groupby(grouping)[grouping_cols].apply(_config_med)
 
-    for label, df in samples.groupby("grouping"):
+    for label, df in samples.groupby(grouping):
         groups.append(
             cdist(
-                df.values[:, :-1].astype("float64"),
+                df.values.astype("float64"),
                 [centroids.loc[label].values],
                 metric="euclidean",
             )
@@ -328,10 +327,9 @@ def _compute_groups(samples, test_type, grouping):
 
 
 def _config_med(x):
-    """Slice and transpose the vector.
+    """Transpose the vector.
 
-    Slice the vector up to the last value to exclude grouping column
-    and transpose the vector to be compatible with hd.geomedian.
+    Transpose the vector to be compatible with hd.geomedian.
     """
-    X = x.values[:, :-1]
-    return pd.Series(np.array(geomedian_axis_one(X.T)), index=x.columns[:-1])
+    X = x.values
+    return pd.Series(np.array(geomedian_axis_one(X.T)), index=x.columns)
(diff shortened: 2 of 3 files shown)


qiyunzhu (commented):
@r0hansaxena Thank you for the timely and elegant fix! This looks to me should work well. A small consideration is that making a copy of the data will cost time and memory. However if it isn't a dominant step in the algorithm, then we don't have to worry about it. For example, if algorithm has to permute the samples 999 times, then adding a single copy won't bother at all, and the current solution is perfect. Otherwise, we may prefer removing the "grouping" column in the end, or (more cleanly) storing the grouping values in a standalone vector that is never added to the "samples" dataframe. What do you think?

r0hansaxena:
@qiyunzhu that does make sense regarding the memory footprint of `.copy()`, especially for larger arrays. We can instead bypass the `OrdinationResults` mutation without any copy overhead. Since` pandas.groupby` accepts an external array of matching length, I can just route the standalone vector directly into the aggregations in` _compute_groups`.

r0hansaxena:
I'll make the changes and push them

qiyunzhu (commented):
This looks like a cleaner and more performant fix, compared with the original code and your previous trial. The CI test failed at some point though. I think the reason is that `X = x.values` then `X.T` some times creates a view of the array that is read-only, and the Cython function `geomedian_axis_one` doesn't like this. Even the original code isn't robust in this, because NumPy will try to return a view instead of making a copy of the array.

A plausible fix is (note that `values` has been deprecated in favor of `to_numpy()`):

```python
X = x.to_numpy(copy=True)
```

If still not working, can do:

```python
Xt = np.ascontiguousarray(x.to_numpy(copy=True).T)
```

---

Additional note: The current `permdisp` function heavily relies on Pandas, which itself is an unnecessary overhead. A better (more performant, and less reliant on Pandas) way is to move the entire algorithm into NumPy (and potentially xp). You can refer to the [implementation](https://github.com/scikit-bio/scikit-bio/blob/3b445f3b138ffb82723abc366b63b0054e261b94/skbio/stats/composition/_dirmult.py#L52) of `dirmult_ttest` for an example. This potential optimization doesn't need to be done here in this PR. It can be postponed to a future PR.

r0hansaxena:
you're right, taking the transpose there would mess with the array formatting for Cython. I just pushed the `np.ascontiguousarray` fix.

r0hansaxena:
@qiyunzhu Regarding the note on Pandas dependency, moving to pure NumPy would be a good improvement. I can open a new issue for that migration so we can handle it separately.

qiyunzhu:
@r0hansaxena Thanks for the update! Have you tried the first solution (without `np.ascontiguousarray`) to see if it works?

`np.ascontiguousarray` creates another copy of the array if the original array is not C-contiguous. Therefore, you create up to two copies in this line of code. Often, `df.values` or `df.to_numpy` returns a row x column array, but it is in fact a transposed view of a column x row array (that's how data is stored in Pandas under the hood). When transposed, it returns the column x row array which is already C-contiguous. Therefore, `np.ascontiguousarray` won't copy the array. So this function call won't introduce any overhead. However, there might be exceptions.

If `np.ascontiguousarray` is desired, then we can do another simple optimization: Look at [here](https://github.com/scikit-bio/scikit-bio/blob/3b445f3b138ffb82723abc366b63b0054e261b94/skbio/stats/distance/_cutils.pyx#L472):

```cython
def geomedian_axis_one(floating[:, :] X, ...):
```

The definition of `X` allows arbitrary array formation -- not necessarily C-contiguous. Therefore, forcing the input array into C-contiguous by calling `np.ascontiguousarray` is not necessary.

However, if we can guarantee that the input array is C-contiguous, we can define it as `floating[:, ::1] X`. This will likely notably accelerate the algorithm. There is a trade-off between the gain of performance of this Cython function (which is a dominant step) and the cost of (potentially; meaning unlikely) creating another copy of the array. It seems to me that the gain is more obvious and desired.

These are my thoughts for further cleanup / optimization. I would encourage you to try the 2nd solution (replacing `[:, :]` with `[:, ::1]`) and benchmark to see if performance indeed increases. Thanks!

r0hansaxena:
@qiyunzhu Thanks! I did try a version without `np.ascontiguousarray` earlier, and I think I ran into the same read-only/view behavior you mentioned. I’ll revisit that path and also test the `[:, ::1]` option and then push the most stable version with benchmark results accordingly.

qiyunzhu:
@r0hansaxena Oh! Good to know that. There must be something I miss. Let's keep the `np.ascontiguousarray` version then!

r0hansaxena:
@qiyunzhu I tried both versions locally. The one without `np.ascontiguousarray` was a bit faster on my side, and the `permdisp` tests still passed.

r0hansaxena:
I'll push the changes, we can move forward from there

qiyunzhu (approved):
Looks like your latest commit (without `np.ascontiguousarray`) passed all tests. Let's merge then! Next, you are welcome to try to upgrade permdisp to a pure NumPy solution and we will highly appreciate it! No pressure though.

PR #2441, opened 2026-04-02
DOC: Polished pip installation and environment setup

qiyunzhu:
This PR extends @AhmedKhairy22 's #2433. Specifically:

1. Replaced the default virtual environment path from `~/virtualenvs/skbio-dev` to `.venv`. According to the [official documentation](https://docs.python.org/3/library/venv.html) and common practices, having a `.venv` directory in the project directory is often what people would do, and it is automatically recognized by IDEs like VS Code.
2. Wrapped the instructions inside a `note` box in the documentation webpage. This is to avoid the confusion that one or both commands are part of the main workflow instead of the alternative, such that users won't accidentally execute them.
3. Renamed `conda_requirements.txt` as `requirements.txt`, since we now recommend both conda and pip.

***


.gitignore
```diff
@@ -56,4 +56,13 @@ web/_build
 
 # Cython files
 *.c
-!ssw.c
+
+# Environments
```
qiyunzhu, line 60: This section is adopted from GitHub's recommendation ([link](https://github.com/github/gitignore/blob/main/Python.gitignore)). Such that `.venv` won't be synced in the repo, but only stays in the developer's local machine.

qiyunzhu:
@AhmedKhairy22 Can you please review my PR? Thanks!

AhmedKhairy22:
Thanks @qiyunzhu  for extending my PR and improving the docs!

Using `.venv` makes sense since IDEs detect it automatically. I initially suggested a home directory setup to prevent accidental commits, a pattern used in projects like [pandas](https://pandas.pydata.org/docs/development/contributing_environment.html). It also tends to keep my virtual environments in a central folder for easier management or when doing quick editing or testing via terminal or Vim. Since VS Code can easily be pointed to external paths, it stays functional either way.

Also, the official Python documentation mentions both approaches: environments can be placed in the project directory (commonly `.venv` or `venv`) **or under a container directory for many environments, such as `~/.virtualenvs`**.

So I was just wondering what happens if contributors don't follow the usual naming conventions. For example, if someone names their environment `skbio-dev`, it wouldn’t currently be ignored by `.gitignore`.

I’m not entirely sure what most contributors typically do here, just sharing my thoughts. Overall the changes look good to me.

qiyunzhu:
@AhmedKhairy22 Thank you for sharing your thoughts! This is very useful. Good to know pandas' recommendation. I have updated the instruction to make it more explicit on the choice of venv directory. Now it reads:

> `.venv` will become a hidden directory in the project directory, and it will be ignored by Git. Alternatively, you may replace `.venv` with a location outside the project directory, such as `~/.virtualenvs/skbio-dev`.

I think that the contribution guide document is primarily for beginners and causal contributors. Therefore the experience should be as simple as possible. Placing a `.venv` folder in the project will make things modular and won't leave residues in their system. So I still prefer this approach. But I added the `virtualenvs` approach to the text. In fact if I am to mainly rely on venv to develop scikit-bio, I would use the `virtualenvs` approach.

Issue #2442, opened 2026-04-03
ENH: Reduce Pandas overhead in permdisp


r0hansaxena:
I’d like to work on a follow-up optimization for `permdisp` by moving more internal computation from Pandas to NumPy.

This follows #2440, which fixed the mutation behavior and improved part of the median path. Next, I’ll rework parts of `_permdisp.py` to use NumPy more directly, especially where DataFrame/groupby handling is still driving overhead. The idea is to move those intermediate computations to NumPy arrays earlier in the flow, while keeping the statistical behavior and user-facing API exactly the same for both `test="centroid"` and `test="median"`.

I’ll validate parity with existing tests plus a few focused checks around the refactored path, and include small benchmarks to show the runtime impact.

r0hansaxena:
I've opened #2443 for the NumPy focused part of this issue, it covers the main internal shift from Pandas to NumPy in `permdisp`

I'll open a separate follow up PR for `xp`/Array API support

PR #2443, opened 2026-04-03
ENH: Move permdisp group computations from Pandas to NumPy

r0hansaxena:

---

### Description

This PR continues the `permdisp` optimization work by shifting the main internal computation path away from heavier Pandas operations and towards NumPy. In this pass, the core group-level distance calculations are handled directly with NumPy, and sample data is prepared once in array form before permutation runs. Behavior is kept the same for users, including both `test="centroid"` and `test="median"` cases.

This PR handles the NumPy scope for #2442.

### Testing

Ran focused tests:

`python -m pytest skbio/stats/distance/tests/test_permdisp.py -q`

Result:

`20 passed`


r0hansaxena:
@qiyunzhu for review

skbio/stats/distance/_permdisp.py
```diff
@@ -303,34 +302,51 @@ def permdisp(
 
 
 def _compute_groups(samples, test_type, grouping):
+    if hasattr(samples, "to_numpy"):
```
qiyunzhu: This is not necessary because `samples` already became a NumPy array in line 289.

skbio/stats/distance/_permdisp.py
```diff
@@ -303,34 +302,51 @@ def permdisp(
...
-        centroids = samples.groupby(grouping)[grouping_cols].apply(_config_med)
-
-    for label, df in samples.groupby(grouping):
-        groups.append(
-            cdist(
-                df.values.astype("float64"),
-                [centroids.loc[label].values],
-                metric="euclidean",
-            )
+    grouping_array = np.asarray(grouping)
```
qiyunzhu, line 308: Hmmm I think you are re-inventing the wheel there is already there in `skbio.stats.composition._utils._check_grouping`. I understand that permdisp is in `skbio.stats.distance`, which can't directly access that function. So a solution would be moving part of `skbio.stats.composition._utils` into `skbio.util`

skbio/stats/distance/_permdisp.py
```diff
@@ -303,34 +302,51 @@ def permdisp(
...
+def _encode_grouping_labels(grouping):
+    """Encode possibly mixed-type labels as integer group ids."""
+    labels = np.asarray(grouping, dtype=object)
+    codes = np.empty(labels.shape[0], dtype=np.intp)
+    mapping = {}
+    next_code = 0
 
-def _config_med(x):
-    """Transpose the vector.
+    for idx, label in enumerate(labels):
```
qiyunzhu: I feel that dictionary is a slow approach. A better approach is to get indices of each group using `np.unique`, then do indexing instead of dictionary lookup. Refer to `_check_grouping`.

skbio/stats/distance/_permdisp.py
r0hansaxena: you're right, I'll make the changes

skbio/stats/distance/_permdisp.py
r0hansaxena: You’re right about the duplication. I’ll switch this to the existing grouping/factorization flow instead of re-implementing it in `permdisp`.

skbio/stats/distance/_permdisp.py
r0hansaxena: hmm yes, I'll drop the dictionary based encoding and use the `np.unique` + indexing pattern for group handling here.

skbio/stats/distance/_permdisp.py
```diff
@@ -287,11 +286,11 @@ def permdisp(
     else:
         raise TypeError("Input must be a DistanceMatrix or OrdinationResults.")
 
-    samples = ordination.samples
+    sample_data = ordination.samples.to_numpy(dtype=np.float64, copy=False)
```
qiyunzhu: Data type isn't always float64. The downstream functions should be able to support both float64 and float32. So I think we shouldn't force type coercion here. Same below.

r0hansaxena: makes sense, I've updated this so `permdisp` preserves the input dtype, including the downstream path in `_compute_groups`

mataton:
@r0hansaxena I'm gonna push some simple edits and merge main onto your branch and then it  should be good to go.

PR #2447, opened 2026-04-11
Fix: Prevent infinite loops and array cycles during NNI swapping on unary roots

### LINKED ISSUE #2446: skbio.tree.nni crashes when the root has a single child
Calling `skbio.tree.nni` on a tree whose root has a single child seems to crash. I noticed this from the comment in skbio/tree/tests/test_me.py:629:
```python
# "(((b,c),((d,e),a)));",  # root has 1 child  # TODO This doesn't work.
```
and I also see the same behavior with that topology. 
`tree.is_bifurcating()` still returns `True`, but `nni` then fails with: `IndexError: index 7 is out of bounds for axis 0 with size 7`
From the traceback, it looks like the error starts in `nni`, goes through root-handling logic including the unary-root case, and then crashes in `skbio/tree/_me.py` while building internal data structures.
It might make sense to either handle this case explicitly, or raise a clearer `ValueError` earlier instead of ending in an `IndexError`.

AshutoshSingh9:
Fixes #2446 
## Summary

`nni` crashed with `IndexError` and entered infinite loops when the input tree had a **unary root** (a root node with a single child, e.g., `(((b,c),((d,e),a)));`). Even though `tree.is_bifurcating()` returns `True` for such a tree (the bifurcation check is not affected by the wrapper), the internal array-building code in `_root_from_treenode` produced a self-referential cycle in `tree[0]`, which caused downstream NNI traversals to loop forever.

### Root cause (two stages)

**Stage 1 -- unary root:** When the root has one child, `_root_from_treenode` would compute an empty siblings list and raise a misleading `ValueError`. The fix prunes the copy of the tree to collapse the unary wrapper before array conversion, without mutating the caller's object.

**Stage 2 -- direct-child circular reference (new in this PR):** After pruning `(((b,(c,(e,d))),a))` -> `((b,(c,(e,d))),a)`, `taxa[0]='a'` becomes a *direct child* of the (now rooted-binary) root. In `_root_from_treenode`, the `n_sibs==1, parent=None` branch ran with `pos=prev_pos=sib_pos=0`, causing `tree[0, 1] = 0` to overwrite the root's right-child pointer with `0` (itself). This circular self-reference made every NNI traversal loop infinitely. Fixed by guarding those four overwrite assignments with `if pos > 0:`.

### Changes

- `skbio/tree/_me.py` -- two additions to `_root_from_treenode`:
-   1. Unary-root pruning block at the top of the function.
-   2. `if pos > 0:` guard in the rooted-binary-root branch.
- - `skbio/tree/tests/test_me.py` -- three new tests:
-   - `test_root_from_treenode`: extended to verify unary root is handled (not `ValueError`).
-   - `test_unary_root_conversion`: array produced from a unary-root tree matches the equivalent normal tree.
-   - `test_unary_root_nni`: end-to-end NNI on a unary-root tree gives same topology as on the equivalent normal tree.
- - `CHANGELOG.md` -- bug-fix entry under Version 0.7.3-dev.
---

**PR checklist:**

* [x]  This pull request does not include code, documentation, or other content derived from external source(s).

### DIFF
--- a/skbio/tree/_me.py
+++ b/skbio/tree/_me.py
@@ -1060,7 +1060,18 @@ def _root_from_treenode(obj, taxa):
     always preordered.
 
     """
+    # If the root has a single child (unary root), promote the child to root
+    # to avoid generating invalid array mappings and infinite loops during NNI.
+    # This does not modify the caller's tree object.
+    if len(obj.children) == 1:
+        obj = obj.copy()
+        obj.prune()
+        while len(obj.children) == 1:
+            obj = obj.children[0]
+            obj.parent = None
+
     errmsg = "Tree is not strictly bifurcating."
+
     taxmap = {taxon: i for i, taxon in enumerate(taxa)}
 
     m = len(taxa)  # number of taxa
@@ -1112,6 +1123,7 @@ def _root_from_treenode(obj, taxa):
         if not siblings:  # singleton branch
             raise ValueError(errmsg)
         n_sibs = len(siblings)
+
         if n_sibs == 1:
             other = siblings[0]
 
@@ -1146,10 +1158,18 @@ def _root_from_treenode(obj, taxa):
             else:
                 _from_treenode(other, taxmap, tree, preodr, postodr, pos, depth)
 
-                tree[pos, 2] = prev_pos
-                tree[prev_pos, 1] = pos
-                tree[sib_pos, 3] = pos
-                tree[pos, 3] = sib_pos
+                # Guard: only link back when at least one internal node has
+                # been traversed (pos > 0). When pos == 0, taxa[0] is a
+                # direct child of a rooted-binary root; _from_treenode has
+                # already set tree[0] up correctly (parent=0, sibling=0), so
+                # the assignments below would overwrite the right-child index
+                # of tree[0] with 0 — creating a circular self-reference that
+                # causes infinite loops in the downstream NNI algorithms.
+                if pos > 0:
+                    tree[pos, 2] = prev_pos
+                    tree[prev_pos, 1] = pos
+                    tree[sib_pos, 3] = pos
+                    tree[pos, 3] = sib_pos
 
                 break
 
--- a/skbio/tree/tests/test_me.py
+++ b/skbio/tree/tests/test_me.py
@@ -621,17 +621,38 @@ class MeTests(TestCase):
             for o, e in zip(obs, exp):
                 npt.assert_array_equal(o, e)
 
-        # non-binary trees
+        # non-binary trees (should now be automatically bifurcated)
         taxa = list("abcde")
         nwks = [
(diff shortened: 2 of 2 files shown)


AshutoshSingh9:
@mataton @qiyunzhu @r0hansaxena check this commit and let me know if this works

AshutoshSingh9 (commented):
kindly let me know if this fixes the issue :)

r0hansaxena:
> @mataton @qiyunzhu @r0hansaxena check this commit and let me know if this works

I confirmed this addresses the specific crash from #2446 (nni on "(((b,c),((d,e),a)));" no longer throws IndexError).

I do have a scope concern tho:

This PR changes behavior for non-bifurcating trees (auto-bifurcating instead of raising), which seems broader than #2446 and the current API contract.
It also includes unrelated changes in skbio/io/* and skbio/metadata/_metadata.py.
Could we split this to just the unary-root nni fix and related tests in skbio/tree/_me.py and skbio/tree/tests/test_me.py?

r0hansaxena:
@AshutoshSingh9 Also, it would be good practice moving forward to use the PR template provided by the org and link the issue the PR resolves. :)

AshutoshSingh9:
@r0hansaxena Thank you for the thorough review! I've addressed both points you raised:

1. **Scope** -- The PR now touches *only* `skbio/tree/_me.py`, `skbio/tree/tests/test_me.py`, and `CHANGELOG.md`. All unrelated changes to `skbio/io/*`, `skbio/metadata/_metadata.py`, `pyproject.toml`, and `.gitignore` have been reverted to `upstream/main`.
2. **Root cause fixed properly** -- The original fix (pruning the unary wrapper) exposed a second bug: after pruning `(((b,(c,(e,d))),a))` -> `((b,(c,(e,d))),a)`, `taxa[0]='a'` becomes a direct child of the rooted root. The `_root_from_treenode` code then wrote `tree[0,1]=0` (right-child = itself), creating a circular reference that caused `nni` to loop forever. This is now fixed with a `if pos > 0:` guard. All 38 tests in `test_me.py` pass in under 2 seconds with no hangs.
3. **Issue linked** -- This PR now closes #2446.
Please let me know if anything else needs attention.

AshutoshSingh9:
@mataton @qiyunzhu please review

AshutoshSingh9:
@mortonjt @aj-pinto @r0hansaxena

qiyunzhu:
@AshutoshSingh9 Thank you for making an attempt to resolve this issue. After careful review, I think that the solution you proposed may not be desired, because it alters the default behavior of the program. Instead of manipulating trees to make them compatible, I think the program should raise an error to let the user know they are not. The additional check you introduced in the algorithm and the in-line comment associated with it are also not most convenient from an aspect of maintenance. Therefore, I decided to write a simple patch myself (#2474). Thank you for your trial though!

PR #2448, opened 2026-04-13
ENH: Improving MMvec interface, documentation and algorithm.

qiyunzhu:
This PR polishes and improves @mortonjt 's porting of MMvec in #2360 .

## Most important change:

- Significantly accelerated the algorithm (up to 10x speedup). The optimizations were multifaceted. But the major algorithmic improvements were: 1) Skipped the reference part in the augmented matrices. 2) Avoided repeated array space allocations and data transfers. The improved algorithm is mathematically equivalent to the original one, although not completely numerically identical. This is expected. Tests have shown consistency between results.

Bencmarks (sec) on Ryzen 9 9950X with 8 cores.

| Dataset | Optimizer | Before | After | Speedup |
| --- | --- | --: | --: | --: |
| soils | L-BFGS | 2.554 | 0.244 | 10.47 |
| cf | L-BFGS | 45.440 | 4.662 | 9.75 |
| soils | Adam | 26.664 | 10.526 | 2.53 |
| cf | Adam | 311.439 | 46.959 | 6.63 |

image

## Major changes (and notes):

- Per our discussion, I have renamed "microbes" and "metabolites" into "X" and "Y" to emphasize the generalizability of MMvec to various omics data types.
- Added a detailed tutorial on how to analyze paired microbiome-metabolome data using MMvec.
- Per our discussion, I refactored mmvec into a [scikit-learn-compatible estimator](https://scikit-learn.org/stable/developers/develop.html) interface: `MMvec`. Its `fit` - `predict` interface works perfectly, and can be integrated into scikit-learn workflows (such as hyperparameter tuning with cross validation). However, there are still some design decisions we need to make before deciding to switch scikit-bio to scikit-learn style. Therefore, I temporarily hid this `MMvec` estimator class from the documentation.
- Instead, a function `mmvec` is the only public-facing function available. It returns an `MMvecResult` object, which provides downstream methods such as `score` and `predict`. I realized that SciPy is adopting a similar design: transitioning named tuples into classes with downstream methods. For example, `ttest_ind` now returns a [`TtestResult`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats._result_classes.TtestResult.html) object, which has a [`confidence_interval`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats._result_classes.TtestResult.confidence_interval.html) method. Given this finding, whether we should keep our current design or switch to scikit-learn style remains a question.
- The output of model fitting now only stores `x_embeddings` and `y_embeddings` in an eager manner, whereas `ranks` and `probs` are lazy and will only be calculated at the first call. This is because they are xdim by ydim and can be very large with high-dimensional multiomics data.

## Minor changes:

- Slightly optimized both L-BFGS and Adam optimizers (some 5-10% performance gain).
- Polished the mathematical equations in the documentation.
- Organized unit tests. Increased unit test coverage.
- Renamed the `probabilities` method as the `probs` attribute.
- Fixed a bug with the `ranks` attribute, which re-computes ranks all over due to namespace contamination.
- Replaced `coo_matrix` with the newer `coo_array` class.
- Adopted scikit-bio's `_ingest_table` helper function and `output_format` mechanism to handle various formats of tabular data.
- Adopted scikit-bio's `seed` interface for random number generation. If we choose to follow the scikit-learn style, we may want to rename it as `random_state`. But scikit-learn itself has an older `random_state` interface that needs to be migrated to `rng`. Upstream packages like SciPy are adopting an `rng` interface. So we don't have to take action too soon.
- Renamed embedding column names from "PC0", "PC1",... into "PC1", "PC2",... to follow scikit-bio's convention.
- Renamed `n_components` into `dimensions` in the function interface (`mmvec`) to follow scikit-bio's convention (that in the estimator interface `MMvec` stays the same).
- Added citations to the test datasets.
- Replaced the non-ASCII character "²".
- Some minor parameter name tweaks.

***


qiyunzhu:
Will self-merge after lots of local works and some discussions.

PR #2456, opened 2026-04-22
FIX: suppress spurious EIGH RuntimeWarning in `permdisp` (fixes #2430)

### LINKED ISSUE #2430: "EIGH: since no value for dimensions is specified ..." when running PERMDISP

jissen706:
## Summary

- `permdisp(dm, grouping)` with the default `method="eigh"` has been emitting a `RuntimeWarning: EIGH: since no value for dimensions is specified...` from `pcoa` on every call where the distance matrix has more than 10 samples.
- Root cause: `permdisp` intentionally overrides `dimensions = 0` before calling `pcoa`, which trips `pcoa`'s warning branch. The advice in that warning ("specify dimensions to avoid long computation") does not apply to `eigh`, which always computes all eigenvectors regardless of `dimensions` (see `pcoa`'s own docstring at `_principal_coordinate_analysis.py` lines 115–118).
- Fix: pass `dimensions = distmat.shape[0]` instead of `0`. Inside `pcoa`, `dimensions == 0` gets replaced by `shape[0]` anyway, so behavior is unchanged, but the warning's guard `dimensions == 0` is no longer satisfied.

Minimal change, scoped to `_permdisp.py`. No change to `pcoa` itself.

## Test plan

- [x] Added regression test `test_no_spurious_eigh_warning_large_matrix` that calls `permdisp` on a 12-sample matrix (>10, the warning threshold) and asserts no `RuntimeWarning` containing `"EIGH"` is captured. Fails on `main`, passes with the fix.
- [x] Full `skbio/stats/distance/tests/test_permdisp.py` suite: 21/21 pass.
- [x] Full `skbio/stats/distance/` + `skbio/stats/ordination/` suites: pass (the 5 remaining failures are pre-existing `SymmetricMatrix` doctest diffs from the NumPy 2.x array repr change — confirmed on unpatched `main`, unrelated to this PR).
- [x] `ruff check --output-format=github .` exits clean.

Fixes #2430.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 ### Bug Fixes
 
 * Fixed `permdisp` mutating the input `OrdinationResults` object by adding a `"grouping"` column to its `samples` DataFrame ([#2440](https://github.com/scikit-bio/scikit-bio/pull/2440)).
+* Fixed `permdisp` emitting pcoa's `"EIGH: since no value for dimensions..."` `RuntimeWarning` on every call with a distance matrix larger than 10 samples. `permdisp` was internally forcing `dimensions=0` before calling `pcoa`, which both tripped the warning and discarded the `eigh` speedup introduced in [#2285](https://github.com/scikit-bio/scikit-bio/pull/2285). The `dimensions` argument is now passed through to `pcoa` unchanged, matching the `fsvd` path; callers who deliberately pass `dimensions=0` on a large matrix still see the warning ([#2430](https://github.com/scikit-bio/scikit-bio/issues/2430)).
 
 ### Miscellaneous
 
--- a/skbio/stats/distance/_permdisp.py
+++ b/skbio/stats/distance/_permdisp.py
@@ -176,7 +176,8 @@ def permdisp(
     the output deterministic. You may skip it if that's not necessary.
 
     >>> from skbio.stats.distance import permdisp
-    >>> permdisp(dm, grouping, permutations=99, seed=42) # doctest: +ELLIPSIS
+    >>> permdisp(dm, grouping, permutations=99, seed=42,
+    ...          dimensions=dm.shape[0]) # doctest: +ELLIPSIS
     method name               PERMDISP
     test statistic name        F-value
     sample size                      6
@@ -192,7 +193,7 @@ def permdisp(
     To suppress calculation of the p-value and only obtain the F statistic, specify
     zero permutations:
 
-    >>> permdisp(dm, grouping, permutations=0)
+    >>> permdisp(dm, grouping, permutations=0, dimensions=dm.shape[0])
     method name               PERMDISP
     test statistic name        F-value
     sample size                      6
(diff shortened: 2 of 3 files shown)


qiyunzhu:
@jissen706 Thanks for contributing!

@sfiligoi I recall that one recent patch you made to PCoA would also theoretically accelerate PERMDISP. Is this PR coincidentally addressing the same issue using the method we discussed?

sfiligoi:
Right... `pcoa` now support the  `dimensions` parameter for `eigh`, too.

The correct implementation thus is to pass-through the value, just like we do for `fsvd`.
I completely forgot to check if any user of `pcoa` needed changing, too.

@jissen706 Appreciate you finding the issue, but the solution needs to be changed.

sfiligoi:
PS: The warning in `pcoa` when passing 0 as dimensions was intentional, to point out places where pcoa was used in "the wrong way".
Masking it and using "the wrong way" should not be the solution.

I.e., going forward, if a user uses `permdisp` with dimensions=0, we should still display the warning, for the very same reason.
(As a reminder, with the new `pcoa` implementation, using the right dimensions reduces the runtime by orders of magnitude on large DMs)

jissen706:
Thanks @sfiligoi, good catch, I didn't realize #2285 made eigh actually honor dimensions. Pushed a new commit that just passes dimensions through, same as the fsvd path.

- **Removed the `dimensions = 0` override.** `permdisp` now passes `dimensions` straight through to `pcoa`, matching the `fsvd` path.
- **Preserved the warning for `dimensions=0`.** Users who explicitly opt in still see the pcoa `EIGH` warning on large matrices — I'm no longer masking it.
- **Updated the existing eigh tests (and the three docstring examples)** to pass `dimensions=dm.shape[0]` explicitly. Without an explicit value, the default `dimensions=10` was exceeding the 6- and 9-sample matrices those tests use, mirroring how the existing fsvd tests already pass `dimensions=3`.
- **Strengthened the regression test** to assert both directions:
  - normal usage (positive `dimensions`) does not leak the `EIGH` warning;
  - explicit `dimensions=0` on a 12-sample matrix still surfaces it.

Full `test_permdisp.py` suite: 21 pass; doctest also passes. Happy to iterate if any of the test updates feel off.

jissen706:
Hey @sfiligoi, gentle bump — anything else you'd like me to change here?

sfiligoi (approved):
Sorry, did not realize you were waiting on me.
Looks great to me.

qiyunzhu:
Cool! Given @sfiligoi 's approval, I am going to merge this PR once tests pass.

qiyunzhu:
Merged. @jissen706 Thanks for the contribution!

PR #2457, opened 2026-04-25
fix gff3 attribute decoding

### LINKED ISSUE #2454: BUG: GFF3 reader does not percent-decode attribute values
Ran into something weird with the GFF3 reader. If you give it a file with percent-encoded characters in the attributes column, the parser leaves them as-is. Percent-encoding of reserved characters like semicolons and commas (%3B, %2C) is standard for GFF3. So a Note field that should say "important; useful" comes back as "important%3B useful" in the metadata. The writer does escape those characters, so if you read a file, write it back, and read it again, the round-trip doesn't give you the same value. There's no error or warning, the value just ends up different from the input.

I think the decode step was never added on the read side. I'll make a PR for this.

r0hansaxena:

- [ ] I have documented all public-facing changes in the [changelog](https://github.com/scikit-bio/scikit-bio/blob/main/CHANGELOG.md).

- [ ] **This pull request includes code, documentation, or other content derived from external source(s).** If this is the case, ensure the external source's license is compatible with scikit-bio's [license](https://github.com/scikit-bio/scikit-bio/blob/main/LICENSE.txt). Include the license in the `licenses` directory and add a comment in the code giving proper attribution. Ensure any other requirements set forth by the license and/or author are satisfied.

Closes #2454

## Description

The GFF3 reader left percent-encoded characters (`%3B`, `%2C`, `%3D`, `%26`) literal in attribute values, so write/read round-trips didn't return the original value. Added `unquote` in `_parse_attr` to decode keys and values on read, splitting multi-value attributes on `,` first so escaped commas are preserved. Tests cover the unescape path and a write/read round-trip.

### DIFF
--- a/skbio/io/format/gff3.py
+++ b/skbio/io/format/gff3.py
@@ -214,6 +214,7 @@ Specifications/blob/master/gff3.md
 
 import re
 from collections.abc import Iterable
+from urllib.parse import unquote
 
 from skbio.sequence import DNA, Sequence
 from skbio.io import create_format, GFF3FormatError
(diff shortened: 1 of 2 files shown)


r0hansaxena:
@qiyunzhu @mataton for review

PR #2459, opened 2026-04-28
WIP:  Use new gpu runner for array-api Actions workflow

mataton:
## Add GPU/Array API CI workflow

Adds a new GitHub Actions workflow (`array-api.yml`) that runs array-API-decorated tests on a T4 GPU runner against PyTorch, JAX, and CuPy with CUDA. Tests are filtered via a new `array_api` pytest marker so only relevant tests run.

### Changes

- **New workflow** (`array-api.yml`): runs weekly (Mondays 06:00 UTC), on manual dispatch, or when a PR is labeled `gpu-ci`. Verifies each backend sees the GPU at install time, then runs `pytest -m array_api`.
- **`@array_backends` decorator** now applies a `pytest.mark.array_api` marker so CI can filter on it. Marker registered in `pyproject.toml`.
- **Skip guards tightened** in `test_array.py`: replaced "is the module importable" checks with usability probes that actually exercise the backend. Fixes spurious errors when a library installs but can't allocate (e.g., CuPy without a GPU).
- **`_normalize_device` replaced** with `_device_specs_match` and `_on_device`. The old function blurred two concerns; the new split distinguishes "do these spec strings refer to the same hardware" from "is this array actually on that hardware."
- **`assert_type_preserved`** now uses `_on_device` to inspect real array device placement instead of string-comparing device reprs.
- **Developer docs**: new `devdoc/array_api.rst` covering how to write array-API library code and tests, run them locally, and use the `gpu-ci` label.

### Notes

- Dask was considered but excluded. Its support in scikit-bio is dev-only and not fully built out; adding CI commitment now would be premature.
- Workflow is intentionally not a required check — GPU runners can be flaky, and most PRs don't touch array code. Reviewers apply the `gpu-ci` label when validation is needed.
- Only 16 tests are currently decorated (in `composition` and `ordination`). Coverage will grow as more of the package adopts the array API.

### DIFF
--- a/.github/workflows/array-api.yml
+++ b/.github/workflows/array-api.yml
@@ -18,6 +18,7 @@ jobs:
       matrix:
         include:
           - backend: torch
+            device: cuda
           # - backend: jax
           # - backend: cupy
     steps:
@@ -35,6 +36,6 @@ jobs:
           pip install -r ci/requirements.txt -r ci/requirements.test.txt
       - name: Run tests
         env:
-          SKBIO_ARRAY_BACKEND: torch # ${{ matrix.backend }}
-          SKBIO_DEVICE: gpu
+          SKBIO_ARRAY_BACKEND: ${{ matrix.backend }}
+          SKBIO_DEVICE: ${{ matrix.device }}
         run: pytest skbio/stats/composition/tests/test_base.py
--- a/.github/workflows/array-api.yml
+++ b/.github/workflows/array-api.yml
@@ -72,10 +72,10 @@ jobs:
               print("cupy:", cupy.__version__, "devices:", n)
           PY
 
-      - name: Check for compiled extensions
-        run: |
-          find . -name "_intersection*" -type f -not -path "*/\.*"
-          python -c "import skbio.metadata._intersection; print('OK:', skbio.metadata._intersection.__file__)"
+      # - name: Check for compiled extensions
+      #   run: |
+      #     find . -name "_intersection*" -type f -not -path "*/\.*"
+      #     python -c "import skbio.metadata._intersection; print('OK:', skbio.metadata._intersection.__file__)"
 
       - name: Run tests
         env:
--- a/.github/workflows/array-api.yml
+++ b/.github/workflows/array-api.yml
@@ -26,8 +26,10 @@ jobs:
           python-version: ${{ env.latest_python }}
       - name: Install scikit-bio + backend
         run: |
-          pip install -e .
+          python -m ensurepip
+          pip install --no-deps .
           pip install ${{ matrix.backend }}
+          pip install -r ci/requirements.txt -r ci/requirements.test.txt
       - name: Run tests
         env:
           SKBIO_ARRAY_BACKEND: torch # ${{ matrix.backend }}
--- a/.github/workflows/array-api.yml
+++ b/.github/workflows/array-api.yml
@@ -11,17 +11,14 @@ env:
 jobs:
   test:
     name: Array API (${{ matrix.backend }})
-    runs-on: ${{ matrix.runs-on }}
+    runs-on: gpu-t4-4-core
     strategy:
       fail-fast: false
       matrix:
         include:
           - backend: torch
-            runs-on: ubuntu-latest
           - backend: jax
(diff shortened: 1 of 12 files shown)


qiyunzhu:
Looks like we need to patch the unit test for `_on_device` with some dummy namespaces in order to bring up the coverage, or mark it to skip coverage calculation. @mataton

mataton:
@qiyunzhu Still need to add a similar workflow to test things on CPU, but other than that this should be ready for review.

skbio/util/tests/test_array.py
```diff
@@ -25,15 +25,18 @@
...
-dask = get_package("dask", raise_error=False)
-if dask is not None:
-    da = get_package("dask.array")
+da = get_package("dask.array", raise_error=False)
+
+CUPY_OK = _is_usable(cp, lambda: cp.arange(1))
+TORCH_OK = _is_usable(torch, lambda: torch.arange(1))
+JAX_OK = _is_usable(jax, lambda: jax.numpy.arange(1))
+DASK_OK = _is_usable(da, lambda: da.arange(1, chunks=1).compute())
+XARRAY_OK = _is_usable(xr, lambda: xr.DataArray([1]))
```
qiyunzhu, line 37: Do we have official Xarray support? If not, I am not sure if we should include Xarray tests. I also think that Xarray is more similar to dataframes instead of arrays. I have an intension to add Xarray to the "table-like" support list. But Xarray can handle more than two dimensions, and we haven't explicited dealt with this logic.

.github/workflows/array-api.yml
```diff
@@ -1,37 +1,85 @@
...
+      - name: Check CUDA driver
+        run: nvidia-smi
+
       - name: Install scikit-bio + backend
         run: |
-          pip install -e .
-          pip install ${{ matrix.backend }}
+          python -m ensurepip
+          pip install -v .
+          case "${{ matrix.backend }}" in
```
qiyunzhu, line 47: Looks like a specific CUDA version is pinned. It will be good to add a comment on how to update CUDA version in the future.

.github/workflows/array-api.yml
```diff
@@ -1,37 +1,85 @@
...
+  latest_python: "3.13"   # torch has no 3.14 wheels yet
 
 jobs:
   test:
+    if: |
+      github.event_name != 'pull_request' ||
+      contains(github.event.pull_request.labels.*.name, 'gpu-ci')
     name: Array API (${{ matrix.backend }})
-    runs-on: ${{ matrix.runs-on }}
+    runs-on: gpu-t4-4-core
```
qiyunzhu, line 21: Can add a comment here that `gpu-t4-4-core` is a specific runner configured in the scikit-bio organization.

.github/workflows/array-api.yml
```diff
@@ -1,37 +1,85 @@
...
 
 env:
-  latest_python: "3.14"
+  latest_python: "3.13"   # torch has no 3.14 wheels yet
 
 jobs:
   test:
+    if: |
+      github.event_name != 'pull_request' ||
+      contains(github.event.pull_request.labels.*.name, 'gpu-ci')
```
qiyunzhu, line 18: Can add a comment here that the workflow will be triggered when the `gpu-ci` tag is added to the PR.

web/devdoc/array_api.rst
```diff
@@ -0,0 +1,217 @@
+Array API and GPU support
+=========================
+
+Scikit-bio is gradually adopting the `Python array API standard
+<https://data-apis.org/array-api/latest/>`_ to allow functions to work
+across multiple array libraries (NumPy, PyTorch, JAX, CuPy) and run on
```
qiyunzhu: Shall we add "Dask" to this list?

skbio/util/tests/test_array.py
mataton: The Xarray tests were from the original `test_array.py`, I didn't add them so I didn't want to remove them. But if you think we can remove them then I think that's okay. The only test it's used in is for `ingest_array` specifically.

qiyunzhu:
Matt and I went through the code and discussed offline. We agreed that it can be merged.

PR #2460, opened 2026-05-02
ENH: Added gamma corrections for sequence distance models: JC69, F81, K2P, and TN93

### LINKED ISSUE #2458: Gamma-corrected distance models
I would like to add support for gamma-distributed rate variation to distance models. For several substitution models, gamma-corrected distance estimators have analytical closed-form expressions, allowing incorporation of among-site rate heterogeneity directly into pairwise distance computation. See, for example, Li & Nei (1990, formula A4) and Tamura & Nei (1993, formula 15).

I see two approaches to insert this into the current interface:

- add an optional alpha parameter to existing models, e.g. jc69(..., alpha=None);
- or introduce separate functions such as jc69_gamma(...).

In the first case, gamma correction would only be applied when gamma shape parameter is provided. Both approaches should retain backward compatibility with the current interface.

I would appreciate guidance on which approach would better align with the current API design.

keder:
Resolves  #2458

### DIFF
--- a/skbio/sequence/distance.py
+++ b/skbio/sequence/distance.py
@@ -1312,36 +1312,38 @@ def tn93(seq1, seq2, freqs=None, gamma=None):
     as:
 
     .. math::
-        \begin{aligned}
-        D = &-2\frac{\pi_A\pi_G}{\pi_R}
-            ln(1-\frac{\pi_R}{2\pi_A\pi_G}P_1-\frac{1}{2\pi_R}Q) \\
-            &-2\frac{\pi_C\pi_T}{\pi_Y}
-            ln(1-\frac{\pi_Y}{2\pi_C\pi_T}P_2-\frac{1}{2\pi_Y}Q) \\
-            &-2(\pi_R\pi_Y-\frac{\pi_A\pi_G\pi_Y}{\pi_R}-\frac{\pi_C\pi_T\pi_R}{\pi_Y})
-            ln(1-\frac{1}{2\pi_R\pi_Y}Q)
-        \end{aligned}
+        D = -2A\, ln\left(1-\frac{P_1}{2A}-\frac{Q}{2\pi_R}\right)
+        -2B\, ln\left(1-\frac{P_2}{2B}-\frac{Q}{2\pi_Y}\right)
+        -2C\, ln\left(1-\frac{Q}{2\pi_R\pi_Y}\right)
 
     Where :math:`P_1` and :math:`P_2` are the proportions of purine and pyrimidine
-    transitions, respectively. :math:`Q` is the proportion of transversions.
+    transitions, respectively. :math:`Q` is the proportion of transversions. And:
+
+    .. math::
+        \begin{aligned}
+        &A = \frac{\pi_A\pi_G}{\pi_R} \\
+        &B = \frac{\pi_C\pi_T}{\pi_Y} \\
+        &C = \pi_R\pi_Y - A\pi_Y - B\pi_R
+        \end{aligned}
+
 
     The TN93 model can be corrected for site-rate heterogeneity by assuming that
     evolutionary rates follow a gamma distribution:
 
     .. math::
         \begin{aligned}
-        D = &\alpha\left[2\frac{\pi_A\pi_G}{\pi_R}
-            \left(1-\frac{\pi_R}{2\pi_A\pi_G}P_1-\frac{1}{2\pi_R}Q\right)
-            ^{-\frac{1}{\alpha}} \right. \\
-            &\quad +2\frac{\pi_C\pi_T}{\pi_Y}
-            \left(1-\frac{\pi_Y}{2\pi_C\pi_T}P_2-\frac{1}{2\pi_Y}Q\right)
-            ^{-\frac{1}{\alpha}} \\
-            &\quad +2\left(\pi_R\pi_Y-\frac{\pi_A\pi_G\pi_Y}{\pi_R}-
-            \frac{\pi_C\pi_T\pi_R}{\pi_Y}\right)
-            \left(1-\frac{1}{2\pi_R\pi_Y}Q\right)^{-\frac{1}{\alpha}} \\
-            &\quad \left. -2(\pi_A\pi_G+\pi_T\pi_C+\pi_R\pi_Y)\right]
+        D = 2\alpha\Bigg[
+            &A\left(1-\frac{P_1}{2A}-\frac{Q}{2\pi_R}\right)^{-\frac{1}{\alpha}}
+            +B\left(1-\frac{P_2}{2B}-\frac{Q}{2\pi_Y}\right)^{-\frac{1}{\alpha}} \\
+            &\quad +\; C\left(1-\frac{Q}{2\pi_R\pi_Y}\right)^{-\frac{1}{\alpha}}
+            - E\Bigg]
         \end{aligned}
 
-    Where :math:`\alpha > 0` is the shape parameter of the gamma distribution.
+
+    Where :math:`\alpha > 0` is the shape parameter of the gamma distribution. And:
+
+    .. math::
+        E = \pi_A\pi_G + \pi_T\pi_C + \pi_R\pi_Y
 
     Parameters
     ----------
--- a/skbio/sequence/distance.py
+++ b/skbio/sequence/distance.py
@@ -634,17 +634,16 @@ def jc69(seq1, seq2, gamma=None):
 
     Where :math:`\alpha > 0` is the shape parameter of the gamma distribution.
 
-
     Parameters
     ----------
     seq1, seq2 : {DNA, RNA}
         Sequences to compute the JC69 distance between.
     gamma : float, optional
-        Shape parameter (:math:`\alpha`) of the gamma distribution for among-site
-        rate heterogeneity. If not provided, no gamma correction is applied.
-        If a positive number, JC69 distance is computed under a gamma model with
-        shape parameter :math:`\alpha = \text{gamma}`.
+        Shape parameter (:math:`\alpha`) of the gamma distribution for among-site rate
+        heterogeneity. Must be a positive number. If not provided, no gamma correction
+        will be applied.
 
+        .. versionadded:: 0.7.3
 
     Returns
     -------
@@ -667,7 +666,7 @@ def jc69(seq1, seq2, gamma=None):
 
     Site heterogeneity in evolutionary rates can be incorporated by extending the JC69
     framework to allow rate variation across sites. Under the assumption that
-    substitution rates vary among sites and follow a Gamma distribution [2]_,
+    substitution rates vary among sites and follow a gamma distribution [2]_,
     the observed genetic distance can be corrected by integrating over this rate
     heterogeneity, yielding a closed-form expression for the evolutionary distance.
 
@@ -890,7 +889,7 @@ def f81(seq1, seq2, freqs=None, gamma=None):
     .. math::
         D = b\, \alpha \left[\left(1 - \frac{p}{b}\right)^{-\frac{1}{\alpha}} - 1\right]
 
-    Where :math:`\alpha > 0` is a shape parameter of a Gamma distribution.
+    Where :math:`\alpha > 0` is the shape parameter of the gamma distribution.
 
     Parameters
     ----------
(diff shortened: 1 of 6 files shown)


skbio/sequence/distance.py
```diff
@@ -1009,8 +1133,16 @@ def func(seqs, i, out):
...
         a2[a2 <= 0] = np.nan
 
-        # the formula (Eq. 10 of Kimura, 1980)
-        out[:] = -0.5 * np.log(a1) - 0.25 * np.log(a2)
+        if gamma is None:
+            # the formula (Eq. 10 of Kimura, 1980)
+            out[:] = -0.5 * np.log(a1) - 0.25 * np.log(a2)
+        else:
+            _check_gamma(gamma)
+            out[:] = (
```
qiyunzhu: Micro-optimization: `-1 / gamma` just needs to be calculated once before this equation and reused in it. Also, I think `np.power` can be written as `**` since we don't need `out=` here. Same below for TN93.

skbio/sequence/distance.py
```diff
@@ -1198,6 +1330,22 @@ def tn93(seq1, seq2, freqs=None):
     Where :math:`P_1` and :math:`P_2` are the proportions of purine and pyrimidine
     transitions, respectively. :math:`Q` is the proportion of transversions.
 
+    The TN93 model can be corrected for site-rate heterogeneity by assuming that
+    evolutionary rates follow a gamma distribution:
+
+    .. math::
```
qiyunzhu, line 1323: Have you tested if this can be correctly rendered?

skbio/stats/distance/_base.py
```diff
@@ -219,6 +219,7 @@ def from_iterable(
         metric: Callable,
         key: Any | None = None,
         keys: Iterable[Any] | None = None,
+        **kwargs: Any,
```
qiyunzhu: Let's not do this for now, and limit the scope of this PR to `skbio.sequence.distance` only. It is generally a good idea to improve `from_iterable` but we should do one thing at a time.

skbio/sequence/distance.py
```diff
@@ -235,6 +236,28 @@ def _check_seqtype(name, this, valid=None):
     )
 
 
+def _check_gamma(gamma):
```
qiyunzhu, line 239: I think this function is good but a bit overkill. Since the function signature already says `gamma` is a float, the user is responsible for supplying a float, which must be a real number when it is >0. You can reduce the code to merely `if not gamma > 0:` Also note that NaN is a real number but we don't want NaN here. `gamma > 0` comparison can simultaneously fence against this case.

qiyunzhu (changes requested):
@keder Thank you for the contribution! Your code looks quite elegant. Other than the literature source (Nei & Kumar, 2000) you cited, I checked the MEGA manual and confirmed that your equation for JC69 and K2P are consistent with MEGA's. I didn't check TN93's because I suspect there is a Latex error in your code. For F81, you didn't supply a source, and I don't see the gamma correction in the McGuire (1999) paper. Do you have a source? I understand that the equation is very straightforward derivation from the J69 equation. So it is perhaps fine not citing a source. But in case you have that will be nicer. Finally, have you checked your results against ape's output and found they are consistent?

skbio/sequence/distance.py
keder: Original formula (the one above) is also not rendering for me properly. Since it's basically a LaTeX formula, I inserted it in the overleaf document and it was rendering appropriately. Please let me know if there is more appropriate way to check it.

skbio/stats/distance/_base.py
keder: I will revert commit here.

skbio/sequence/distance.py
keder: This is reasonable, I will put positivity check instead.

skbio/sequence/distance.py
keder: Does formula above rendering correctly for you (the one for TN93 without gamma correction)? If yes, I will try something else to check rendering.

keder:
> For F81, you didn't supply a source, and I don't see the gamma correction in the McGuire (1999) paper. Do you have a source?

To be fair, I cannot find source. Since JC69 and F81 only differs in constant multipliers, I think integration should yield the same result (with different constant multipliers, of course). My understanding is that ape's developers followed the same line of thought for their F81 code.

> Finally, have you checked your results against ape's output and found they are consistent?

I checked ape against the current code and it was consistent up to 5th significant digit. This was visual comparison though, if needed I can cook some automatic validation.

skbio/sequence/distance.py
qiyunzhu: No it doesn't. You can type `make doc` to render the documentation locally and check it. I made some edits to your PR to make the rendering correct.

skbio/sequence/distance.py
```diff
@@ -50,6 +50,7 @@
 from typing import TYPE_CHECKING
 from inspect import signature
 import functools
+import numbers
```
qiyunzhu: This import can be deleted since it isn't used anymore.

qiyunzhu (changes requested):
Thanks for the update. I updated your code to make the rendering correct (screenshot below). Please git pull before you do further edits. I think you can keep the `_check_gamma` helper function, because in that way the error message "Parameter 'gamma' must be a positive number." doesn't have to be typed multiple times (for sake of code reusability). Also, I think `if gamma <= 0.0:` and `if not gamma > 0.0:` are not the same thing. If you test it with NaN you will know what I mean.

Image

qiyunzhu:
Please check whether the equation I rendered is correct.

skbio/sequence/distance.py
keder: This is what I did and it was not rendering. I assumed that there is a difference between how formulas rendered from static HTML and when put through the website engine.

keder:
> Thanks for the update. I updated your code to make the rendering correct (screenshot below). 

Thanks for the help!

> I think you can keep the `_check_gamma` helper function, because in that way the error message "Parameter 'gamma' must be a positive number." doesn't have to be typed multiple times (for sake of code reusability). 

Reverted.

> Also, I think `if gamma <= 0.0:` and `if not gamma > 0.0:` are not the same thing. If you test it with NaN you will know what I mean.

You right, I forgot about nans.

> image

I see, there should be "aligned" instead of "align".  I should not have trusted the code in formula above blindly. It looked too similar to LaTeX syntax.

keder:
> Please check whether the equation I rendered is correct.

Now I it renders in static html, so I checked it again on my machine. It checks with MEGA docs. I personally would prefer to split it into series of formulas (formulas are way too congested for a regular reader), but I think consistency with previous formula is more important.

skbio/sequence/distance.py
keder: Removed

qiyunzhu:
@keder Thanks for the update. Splitting it into series of formulas is actually a great idea to me. Given that the uncorrected and corrected formulae share multiple parts, it will be more concise and readable to split both of them and reuse the parts. See the documentation of [F84](https://scikit.bio/docs/latest/generated/skbio.sequence.distance.f84.html) for an example. I will let you decide whether you want to make another commit to split, or you would prefer to leave as-is (and I will merge).

qiyunzhu (approved):

keder:
> @keder Thanks for the update. Splitting it into series of formulas is actually a great idea to me. Given that the uncorrected and corrected formulae share multiple parts, it will be more concise and readable to split both of them and reuse the parts. See the documentation of [F84](https://scikit.bio/docs/latest/generated/skbio.sequence.distance.f84.html) for an example. I will let you decide whether you want to make another commit to split, or you would prefer to leave as-is (and I will merge).

@qiyunzhu I want to edit formulas then. Please put merge on hold for now.

keder:
@qiyunzhu Could you, please, check if formulas look pretty to you. I tried to align them neatly, but I need second opinion.

skbio/sequence/distance.py
```diff
@@ -1315,8 +1479,16 @@ def func(seqs, i, out):
         a3 = 1.0 - Q / piRxY
         a3[a3 <= 0] = np.nan
 
-        # the formula (Eq. 3 of Tamura & Nei, 1993)
-        out[:] = c1 * np.log(a1) + c2 * np.log(a2) + c3 * np.log(a3)
+        if gamma is None:
+            # the formula (Eq. 3 of Tamura & Nei, 1993)
+            out[:] = c1 * np.log(a1) + c2 * np.log(a2) + c3 * np.log(a3)
+        else:
+            _check_gamma(gamma)
```
qiyunzhu: I think parameter validation should go to the top of the public-facing function, instead of in the middle of calculation. This design is good because it separates algorithmic core from user interface. Same for other places.

skbio/sequence/distance.py
```diff
@@ -648,6 +681,12 @@ def jc69(seq1, seq2):
     evolution as a continuous-time Markov chain, and corrects the observed distance
     (*p*-distance) for repeated substitutions to estimate the true distance.
 
+    Site heterogeneity in evolutionary rates can be incorporated by extending the JC69
```
qiyunzhu: I noticed that this paragraph has been repeated multiple times. This might be fine, but it could reduce readability (to human readers) when the rest of the Notes is short. Given that you already defined the Gamma distribution equation above Parameters, having a long paragraph below may not be necessary. I think you may either remove these paragraphs (and citation) from functions other than `jc69`, or replace them with a sentence like: "Refer to :func:`jc69` for gamma correction for rate variation across sites." to link the readers to the full description.

skbio/sequence/distance.py
```diff
@@ -656,12 +695,15 @@ def jc69(seq1, seq2):
     ----------
     .. [1] Jukes, T. H., & Cantor, C. R. (1969). Evolution of protein molecules.
        Mammalian Protein Metabolism, 3(21), 132.
+    .. [2] Nei, M., & Kumar, S. (2000). Molecular evolution and phylogenetics.
```
qiyunzhu: This literature source is a full book, and it may not be easy for a reader to find the specific chapters and pages where the gamma correction was proposed or defined. I manually checked this book and found that JC69 gamma correction is equation 2.12 of page 22, chapter 2. But the relevant paragraph of the book also noted two previous papers (Nei 1976a; Ota and Nei 1994b). I haven't checked them though. Also, I haven't checked if the book contains gamma correction equations of other metrics. Ideally, we should identify and note the most original literature where each equation was proposed. I understand that this may not always be feasible, given that phylogenetics is such an old field and many classical methods are not documented to today's standard. I will leave this up to you to decide.

qiyunzhu (approved):
Thanks for the update. I rendered the equations locally and they look beautiful! I appreciate that you identified an elegant way to denote those equations. Well done! I have a few additional comments.

skbio/sequence/distance.py
qiyunzhu: PS: If we do want to cite a book, I think chapter, book is enough. No need to cite page or equation number (but you can note in the code if this information is available). It is acceptable to only cite a book without chapter if necessary.

skbio/sequence/distance.py
keder: I was hesitant where to put it originally. Thought that this way would simplify and centralize checking a bit. I will move checks up then.

skbio/sequence/distance.py
keder: I think it would be nice to have citation that will allow reader to read more on topic. I will remove the paragraph, but leave citation. I will cite textbook near the formula.

skbio/sequence/distance.py
keder: I have tried to find original source for a while, but only found a couple of text books so far. I will try to search once more and add chapters if no citation will be available.

qiyunzhu: Thanks! If it's indeed hard to find the original source, you can leave the current citation as-is.

keder: Ok, I managed to find original papers.

skbio/sequence/distance.py
keder: Now I am getting the following issue. Tests use align_dists() which in turn calls models with underscore (_jc69, e.g.). So when gamma check moved to the jc69(), tests skip gamma check. Which of course causes test fails. Not sure how to approach this issue.

qiyunzhu: Good point! I think the distance metrics and `align_dists` are bit engineered that it isn't straightforward to alter their behaviors. How about we temporarily merge your current code, then I will find a time to work on it later?

As far as I think, a solution is to modify [`_metric_specs`](https://github.com/scikit-bio/scikit-bio/blob/101d76c6a3fa3cd66b3ff1783b024af9947bdb4c/skbio/sequence/distance.py#L91) to add a `_has_gamma` flag to the public-facing metric function (like `jc69`), then to add `_check_gamma` into `align_dists` like this [example](https://github.com/scikit-bio/scikit-bio/blob/101d76c6a3fa3cd66b3ff1783b024af9947bdb4c/skbio/alignment/_distance.py#L137). This will do the job. However, it looks bit complicated. So feel free to leave it to me.

keder: Ok, I will do that. I have not pushed all changes yet because of unit test fails, so I will push everything once I fix them.

keder: @qiyunzhu I pushed all changes. Could you, please, check if I implemented changes you suggested right?

qiyunzhu:
@keder Thanks for the update! I think the code is perfect now. Once the CI tests pass, I will merge. We appreciate your contribution to scikit-bio and the scientific community!

keder:
@qiyunzhu Hope you do not mind if I contribute more.

qiyunzhu:
You are always welcome!

PR #2464, opened 2026-05-16
ENH: add Numba CPU permuted Pearson helpers for Mantel

LarytheLord:
Add optional Numba CPU implementations for the Mantel Pearson permutation
helpers. Numba is detected at import time; if unavailable the existing Cython
path is used as fallback.

## Implementation

This PR is a direct Cython -> Numba conversion of the Mantel Pearson
permutation helpers.

Two helpers added:
- `_mantel_perm_pearsonr_numba` — full matrix
- `_mantel_perm_pearsonr_condensed_numba` — condensed matrix

Both use `@njit(parallel=True)` with `prange` over independent permutations.
Each permutation writes to its own `permuted_stats[p]` slot, so there is no
shared reduction. The implementation follows the same loop structure and
numerical operations as the Cython helpers.

Any Numba-specific optimization work should be handled in a follow-up PR.

## Testing

```
$ python -m pytest skbio/stats/distance/tests/test_mantel.py -x -q
61 passed in 11.13s
```

## Benchmark Correction

The earlier local benchmark that showed a large Numba CPU speedup used a local
macOS editable build where the Cython extension had been built with Apple Clang
and was not OpenMP-enabled. That comparison was not a fair optimized-Cython
baseline.

I reran the benchmark with corrected setups:
- local macOS rebuild using a conda OpenMP-capable toolchain, verifying that
  `_cutils` links against `libomp`
- fork-only GitHub Actions Linux runs using conda compilers and explicit
  `OMP_NUM_THREADS` / `NUMBA_NUM_THREADS` controls

## Corrected Controlled Results

Linux fork-only GitHub Actions results (`ubuntu-latest`, 4 CPU / 16 GB RAM):

| Form | n | Perms | `OMP_NUM_THREADS` | `NUMBA_NUM_THREADS` | Cython | Numba warm | Speedup | Max diff |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| full | 5000 | 999 | 8 | 8 | 14.878774s | 19.712725s | 0.75x | 0.000e+00 |
| full | 5000 | 9999 | 8 | 8 | 143.913849s | 188.065629s | 0.77x | 0.000e+00 |
| condensed | 5000 | 9999 | 8 | 8 | 1092.922673s | 648.562209s | 1.69x | 0.000e+00 |
| condensed | 25000 | 999 | 8 | 8 | 2956.519255s | 1802.022791s | 1.64x | 0.000e+00 |

Additional control runs confirmed:
- OpenMP helps the Cython path on the corrected Linux baseline.
- `NUMBA_NUM_THREADS` does control Numba thread count as expected.
- The current performance picture depends on matrix form: full-form Mantel is
  slower under this direct Numba port, while condensed-form Mantel can be
  faster at larger scales.

Experimental `condensed`, `n=25000`, `permutations=9999` did not finish on the
public GitHub runner within about 6 hours.

### DIFF
--- a/skbio/stats/distance/_mantel.py
+++ b/skbio/stats/distance/_mantel.py
@@ -21,11 +21,138 @@ from ._cutils import mantel_perm_pearsonr_cy, mantel_perm_pearsonr_condensed_cy
 from skbio.stats.distance import DistanceMatrix
 from skbio.util import get_rng
 
+try:
+    from numba import njit, prange
+    NUMBA_AVAILABLE = True
+except ImportError:
+    NUMBA_AVAILABLE = False
+
 if TYPE_CHECKING:  # pragma: no cover
     from collections.abc import Sequence
     from ._base import DistanceMatrix
     from numpy.typing import ArrayLike
     from skbio.util._typing import SeedLike
+
+
+if NUMBA_AVAILABLE:
+
+    @njit(parallel=True)
+    def _mantel_perm_pearsonr_numba(
+        x_data, perm_order, xmean, normxm, ym_normalized, permuted_stats
+    ):
+        """Fused permute, normalize, and Pearson correlation for Mantel test.
+
+        Replaces the following Python code::
+
+            def _mantel_perm_pearsonr_one(x_flat, xmean, normxm, ym_normalized):
+                xm_normalized = (x_flat - xmean) / normxm
+                one_stat = np.dot(xm_normalized, ym_normalized)
+                one_stat = max(min(one_stat, 1.0), -1.0)
+                return one_stat
+
+            permuted_stats = np.array([
+                _mantel_perm_pearsonr_one(
+                    distmat_reorder_condensed(x._data, perm_order[p]), ...
+                ) for p in range(permutations)
+            ])
+
+        Takes a pre-computed permutation order matrix and normalizes values on
+        the fly using mul and add (derived from xmean and normxm) instead of
+        materializing the full permuted matrix. Each permutation is independent
+        and computed in parallel via prange.
+
+        Parameters
+        ----------
+        x_data : 2D array_like
+            Full distance matrix (n x n).
+        perm_order : 2D array_like
+            Permutation index array (n_perm x n). Each row is a permutation
+            of range(n) specifying the reordering of rows and columns.
+        xmean : float
+            Mean of the condensed form of x.
+        normxm : float
+            Norm of (x_flat - xmean).
+        ym_normalized : 1D array_like
+            (y_flat - ymean) / normym, pre-normalized.
+        permuted_stats : 1D array_like
+            Output array (n_perm,) to write correlation coefficients into.
+
+        """
+        n = x_data.shape[0]
+        n_perm = perm_order.shape[0]
+        mul = 1.0 / normxm
+        add = -xmean / normxm
+
+        for p in prange(n_perm):
+            my_ps = 0.0
+            for row in range(n - 1):
+                vrow = perm_order[p, row]
+                row_start = row * (n - 1) - ((row - 1) * row) // 2
+                for icol in range(n - row - 1):
+                    col = icol + row + 1
+                    yval = ym_normalized[row_start + icol]
+                    xval = x_data[vrow, perm_order[p, col]] * mul + add
+                    my_ps = yval * xval + my_ps
+
+            if my_ps > 1.0:
+                my_ps = 1.0
+            elif my_ps < -1.0:
+                my_ps = -1.0
+            permuted_stats[p] = my_ps
+
+    @njit(parallel=True)
+    def _mantel_perm_pearsonr_condensed_numba(
+        x_data, perm_order, xmean, normxm, ym_normalized, permuted_stats
+    ):
+        """Fused permute, normalize, and Pearson for condensed Mantel test.
+
+        Same as _mantel_perm_pearsonr_numba but accepts x_data in condensed
+        form (1D array of length n*(n-1)/2) instead of the full 2D matrix.
+        Uses condensed_index formula to map (vrow, vcol) pairs from the
+        permuted ordering back to the 1D condensed array.
+
+        Parameters
+        ----------
+        x_data : 1D array_like
+            Condensed distance matrix (lower triangle only).
+        perm_order : 2D array_like
+            Permutation index array (n_perm x n).
+        xmean : float
+            Mean of the condensed form of x.
+        normxm : float
+            Norm of (x_flat - xmean).
+        ym_normalized : 1D array_like
+            (y_flat - ymean) / normym, pre-normalized.
+        permuted_stats : 1D array_like
+            Output array (n_perm,) to write correlation coefficients into.
+
+        """
+        n_perm = perm_order.shape[0]
+        n = perm_order.shape[1]
+        mul = 1.0 / normxm
+        add = -xmean / normxm
+
+        for p in prange(n_perm):
+            my_ps = 0.0
+            for row in range(n - 1):
+                vrow = perm_order[p, row]
+                row_start = row * (n - 1) - ((row - 1) * row) // 2
+                for icol in range(n - row - 1):
+                    col = icol + row + 1
+                    vcol = perm_order[p, col]
+                    if vrow < vcol:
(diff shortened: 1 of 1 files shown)


skbio/stats/distance/_mantel.py
```diff
@@ -21,13 +21,140 @@
...
+        permuted_stats : 1D array_like
+            Output array (n_perm,) to write correlation coefficients into.
+
+        """
+        n = x_data.shape[0]
+        n_perm = perm_order.shape[0]
+        mul = 1.0 / normxm
+        add = -xmean / normxm
+
+        for p in prange(n_perm):
```
qiyunzhu, line 86: This triple-loop structure looks quite heavy... Wondering if there is a simple NumPy vectorized solution... (Out of the scope of this PR; no need to address.)

skbio/stats/distance/_mantel.py
```diff
@@ -21,13 +21,140 @@
...
+            my_ps = 0.0
+            for row in range(n - 1):
```
qiyunzhu, line 88: `n - 1` can be pre-calculated such that you won't need to calculate it many times during iterations. Modern compilers may be able to optimize it automatically, so you may not notice much performance change. But making it explicit can improve stability.

skbio/stats/distance/_mantel.py
```diff
@@ -21,13 +21,140 @@
...
+                vrow = perm_order[p, row]
+                row_start = row * (n - 1) - ((row - 1) * row) // 2
+                for icol in range(n - row - 1):
+                    col = icol + row + 1
+                    yval = ym_normalized[row_start + icol]
+                    xval = x_data[vrow, perm_order[p, col]] * mul + add
```
qiyunzhu, line 94: Here is a trick which works for Cython, but I am not sure if it does for Numba: pre-assign row pointers `x_data_vrow = x_data[vrow]` and `perm_order_p = perm_order[p]` in the outer loop, then do `xval = x_data_vrow[perm_order_p[col]] * mul + add` in the inner loop.

skbio/stats/distance/_mantel.py
```diff
@@ -21,13 +21,140 @@
...
+        add = -xmean / normxm
+
+        for p in prange(n_perm):
+            my_ps = 0.0
+            for row in range(n - 1):
+                vrow = perm_order[p, row]
+                row_start = row * (n - 1) - ((row - 1) * row) // 2
+                for icol in range(n - row - 1):
+                    col = icol + row + 1
+                    yval = ym_normalized[row_start + icol]
```
qiyunzhu, line 93: Likewise, the trick here is `ym_normalized_row_start = ym_normalized[row_start:]` in the outer loop, then you can do `yval = ym_normalized_row_start[icol]` in the inner loop.

sfiligoi: Any smart compiler should automatically optimize that out.
Adding additional pointer variables may just add to the register pressure. (Although any good compiler should optimize that out, too)

skbio/stats/distance/_mantel.py
sfiligoi: As you said, any good compiler will do the right thing.
n is constant, so it is trivial for the compiler to reason on it.

The question is more:
Which way it is more readable? Understandable?

skbio/stats/distance/_mantel.py
sfiligoi: This is the straight conversion from the Cython version:
As a reminder: The outer parallel loop is optimized for L2 cache reuse, since ym_normalized accesses do not depend on p.

skbio/stats/distance/_mantel.py
sfiligoi: A reasonable suggestion, but I am not sure this PR is the right place to address this.
Personally, I would rather have this PR to match as closely as possible the Cython version, and address these kind of changes (if beneficial) in a separate PR that modifies both the Cython and the Numba versions at the same time, since we plan to keep both in the long run.

sfiligoi:
As a reminder:
Mantel is completely memory-bound (for any non-trivial DM size).
Virtually all time is spent for data to be loaded in the registers... the compute is basically free.

Multi-threading helps, a different threads can reuse the same data, from the chip-wide caches.

sfiligoi:
Hi @LarytheLord 
My benchmark numbers on a 6-core i5-12400F (32GB RAM) Linux system show your Numba implementation is noticeably slower than the Cython version (g++ 15.2, numba numba-0.65.1)
(BTW, frirst time, JIT penalty is about 0.3s)

I ran mantel on uw_emp_5k.h5 vs wn_emp_5k.h5 (5116^2)
nperm | Cython | Numba
-- | -- | --
99 | 0.33 | 0.41
999 | 2.9 | 3.7
9999 | 27.9 | 36.8

For completeness, here are the numbers when using a single thread (`taskset -c 4`)
nperm | Cython | Numba
-- | -- | --
99 | 1.1 | 1.7
999 | 10.5 | 16.4
9999 | 104.7 | 146

sfiligoi:
Same 6-core i5-12400F (32GB RAM) Linux system  (g++ 15.2, numba numba-0.65.1),

but benchmarking the whole EMP Unifrac DM (UW vs WN, 25145^2):
nperm | Cython | Numba
-- | -- | --
99 | 9.4 | 11.2
999 | 83.3 | 98.1

sfiligoi:
The difference is not order-of-magnitude worse, but it definitely is not faster.
Not ready to replace the Cython version on performance merits alone.

LarytheLord:
@sfiligoi I tracked down the earlier benchmark discrepancy and updated the PR description with corrected results.

The original local macOS comparison used an Apple Clang Cython build that was not OpenMP-enabled, so that baseline was misleading. I reran the benchmark with a corrected local OpenMP-capable build and with fork-only Linux GitHub Actions runs using conda compilers plus explicit OMP_NUM_THREADS / NUMBA_NUM_THREADS controls.

With the corrected setup, the direct Cython -> Numba port is not uniformly faster on CPU: full-form Mantel is slower in the controlled runs, while condensed-form Mantel can be faster at larger scales. I kept this PR as a direct conversion and treated any Numba-specific optimization work as follow-up work, as discussed.

I also rebased this branch onto the latest development_numba and reran the Mantel test suite successfully (61 passed).

LarytheLord:
For completeness, here are corrected local numbers using the real EMP UniFrac pair on an OpenMP-enabled local build.

This supersedes my earlier local EMP comment. The previous one-shot timings were noisier than expected, so I reran the real-data benchmark more narrowly.

Same `uw_emp_5k.h5` vs `wn_emp_5k.h5` pair (`5116^2`) on an Apple M2 Mac (8 logical cores), using a local editable build rebuilt with conda clang and `libomp` linked.

Multi-threaded reruns (`OMP_NUM_THREADS=8`, `NUMBA_NUM_THREADS=8`):

nperm | Cython | Numba
-- | -- | --
99 | 0.493 | 0.733
999 | 4.044 | 3.974
9999 | 41.409 | 40.271

Single-thread rerun (`OMP_NUM_THREADS=1`, `NUMBA_NUM_THREADS=1`):

nperm | Cython | Numba
-- | -- | --
999 | 18.248 | 15.640

So on this corrected local real-data benchmark, the multi-threaded full-matrix Mantel results are close to parity, with Cython slightly ahead at low `nperm`, while the single-thread `999` rerun favors Numba locally.

skbio/stats/distance/_mantel.py
qiyunzhu: I agree with you that compilers can handle it already (I guess), so add this trivial optimization may not have a profound effect. It is neither more readable nor understandable, but it relieves the burden of optimization from compiler, which is system-specific and may not always produces the same machine code (remember that numba is JIT). I think my suggestion is overall trivial, unless benchmarks show there is a visible performance difference.

skbio/stats/distance/_mantel.py
qiyunzhu: Fully agreed with the development procedures you proposed!

sfiligoi:
The Numba variant is still not tested by the CI, but will address that in a separate PR.

qiyunzhu:
A trivial suggestion: How about we use suffixes `_nb` (Numba) and `_cy` (Cython) to indicate different engines of the code?

sfiligoi:
Good idea, I think we should do that in a follow-up PR.

PR #2468, opened 2026-05-20
TEST: add Numba-aware Mantel tests

LarytheLord:
## Summary
- Add a `numba_code` pytest marker and test decorator for optional Numba-specific tests.
- Add Mantel tests that explicitly exercise the full, condensed, and dispatch Numba paths.
- Add a dedicated CPU Numba workflow that installs Numba and runs only marked tests.

## Tests
- `python -m pytest skbio/stats/distance/tests/test_mantel.py -m numba_code -q`
- `python -m pytest skbio/stats/distance/tests/test_mantel.py -q`
- `cd ci && python -m pytest -m numba_code --pyargs skbio -q`
- `ruff check skbio/util/_testing.py skbio/util/__init__.py`

### DIFF
--- a/.github/workflows/numba.yml
+++ b/.github/workflows/numba.yml
@@ -7,12 +7,22 @@ on:
   workflow_dispatch:
 
 env:
-  latest_python: "3.13"
+  latest_python: "3.14"
 
 jobs:
   test:
-    name: Numba CPU
-    runs-on: ubuntu-latest
+    name: Numba (${{ env.latest_python }}, ${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: bash
+    strategy:
+      fail-fast: false
+      matrix:
+        # macos-15-intel is omitted: Numba and llvmlite no longer ship macOS
+        # x86_64 wheels, so `pip install numba` would fall back to a source
+        # build requiring an LLVM toolchain.
+        os: [ubuntu-latest, macos-latest, windows-latest, ubuntu-24.04-arm]
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-python@v6
--- a/.github/workflows/numba.yml
+++ b/.github/workflows/numba.yml
@@ -1,0 +1,31 @@
+name: Numba
+
+on:
+  pull_request:
+    branches: [ development_numba ]
+    paths-ignore: ["web/**", "**.md", "README.rst"]
+  workflow_dispatch:
+
+env:
+  latest_python: "3.13"
+
+jobs:
+  test:
+    name: Numba CPU
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-python@v6
+        with:
+          python-version: ${{ env.latest_python }}
+
+      - name: Install scikit-bio + Numba
+        run: |
+          python -m ensurepip
+          pip install --prefer-binary -r ci/requirements.txt -r ci/requirements.test.txt
+          pip install numba
+          pip install --no-deps .
+          pip list
+
+      - name: Run Numba tests
+        run: cd ci && pytest -m numba_code --pyargs skbio
--- a/.github/workflows/numba.yml
+++ b/.github/workflows/numba.yml
@@ -27,5 +27,8 @@ jobs:
           pip install --no-deps .
           pip list
 
-      - name: Run Numba tests
+      - name: Run Numba-marked tests
         run: cd ci && pytest -m numba_code --pyargs skbio
+
+      - name: Run full test suite with Numba installed
+        run: make test
--- a/.github/workflows/numba.yml
+++ b/.github/workflows/numba.yml
@@ -10,8 +10,22 @@ env:
   latest_python: "3.14"
 
 jobs:
+  conf:
+    # This job is needed to route the global environment variables into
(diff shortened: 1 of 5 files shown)


skbio/stats/distance/tests/test_mantel.py
```diff
@@ -77,6 +79,90 @@ def _compute_perf_one(self, x_data, order, xmean, normxm, ym_normalized):
         one_stat = max(min(one_stat, 1.0), -1.0)
         return one_stat
 
+    def _numba_test_inputs(self):
```
sfiligoi: Why do we have different test inputs for Numba vs the Cython version?

I believe it would be better if we used identical tests for the two code paths, as we expect the same results for both.

skbio/stats/distance/tests/test_mantel.py
```diff
@@ -77,6 +79,90 @@ def _compute_perf_one(self, x_data, order, xmean, normxm, ym_normalized):
...
+        xm = x_flat - xmean
+        normxm = np.linalg.norm(xm)
+        ymean = y_flat.mean()
+        ym = y_flat - ymean
+        ym_normalized = ym / np.linalg.norm(ym)
+
+        return x_data, x_flat, perm_order, xmean, normxm, ym_normalized
+
+    @numba_code
+    def test_perm_pearsonr_numba_full(self):
```
sfiligoi: Is this the Numba equivalent of `test_perm_pearsonr3`?
If so, we should keep a similar test naming, e.g. `test_perm_pearsonr3_numba`.

May be also worth renaming the Cython version to expicility include the cy name, e.g. `test_perm_pearsonr3_cy`.

Sam comment for all the other functions that explicity target a specific backend.

.github/workflows/numba.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Install scikit-bio + Numba
+        run: |
+          python -m ensurepip
+          pip install --prefer-binary -r ci/requirements.txt -r ci/requirements.test.txt
+          pip install numba
+          pip install --no-deps .
+          pip list
+
+      - name: Run Numba tests
+        run: cd ci && pytest -m numba_code --pyargs skbio
```
sfiligoi, line 55: It is nice to have a numba-only test task, to make sure the numba subroutines work as expected.

But we should also have a full test suite run, to make sure having numba installed does not break anything (even if we may expect no numba code should be invoked, eventually... it actually should trigger the numba code-path in the current, intermediate mantel implementation).

skbio/stats/distance/tests/test_mantel.py
```diff
@@ -77,6 +79,90 @@ def _compute_perf_one(self, x_data, order, xmean, normxm, ym_normalized):
...
+                     ym_normalized, permuted_stats):
+            calls.append("full")
+            permuted_stats[:] = 0.0
+
+        def condensed_spy(x_data, perm_order, xmean, normxm,
+                          ym_normalized, permuted_stats):
+            calls.append("condensed")
+            permuted_stats[:] = 0.0
+
+        with patch.object(mantel_mod, "_mantel_perm_pearsonr_numba",
```
sfiligoi: I don't understand what this is doing.
Can you please explain?

Ideally, add a comment in the code itself.

skbio/stats/distance/tests/test_mantel.py
```diff
@@ -77,8 +80,16 @@ def _compute_perf_one(self, x_data, order, xmean, normxm, ym_normalized):
         one_stat = max(min(one_stat, 1.0), -1.0)
         return one_stat
 
-    def test_perm_pearsonr3(self):
-        # data pre-computed using released code
```
sfiligoi, line 81: You should not delete the comment

skbio/stats/distance/tests/test_mantel.py
```diff
@@ -90,17 +101,9 @@ def test_perm_pearsonr3(self):
...
-        permuted_stats = np.empty(len(perm_order), dtype=x_data.dtype)
-        mantel_perm_pearsonr_cy(x_data, perm_order, xmean, normxm,
-                                ym_normalized, permuted_stats)
-        for i in range(len(perm_order)):
-            exp_res = self._compute_perf_one(x_data, perm_order[i, :],
-                                             xmean, normxm, ym_normalized)
-            self.assertAlmostEqual(permuted_stats[i], exp_res)
-
-    def test_perm_pearsonr6(self):
-        # data pre-computed using released code
```
sfiligoi, line 103: Same here... preserve the comments

skbio/stats/distance/tests/test_mantel.py
```diff
@@ -174,15 +170,75 @@ def test_perm_pearsonr_full(self):
...
+    @numba_code
+    def test_pearson_flat_numba_matches_cython_fallback(self):
+        y_flat = self.miny_dm.condensed_form()
+
+        for x in (self.minx_dm, self.minx_cond):
+            obs = mantel_mod._mantel_stats_pearson_flat(x, y_flat, 3, seed=0)
+
+            # Force the fallback path so the Numba dispatch result is compared
+            # with Cython using the same input matrix and permutation seed.
+            with patch.object(mantel_mod, "NUMBA_AVAILABLE", False):
```
sfiligoi, line 237: This is OK for now.
In the long run numba vs cython should be user selectable, without any patching.

PR #2470, opened 2026-05-26
ENH: add Numba CPU s_W helpers for PERMANOVA

LarytheLord:
Builds on #2464 and #2468.

### Description

Add optional Numba CPU implementations for the PERMANOVA s_W helpers. Numba is detected at import time; if unavailable the existing Cython path is used as fallback.

This PR is a direct Cython -> Numba conversion of `permanova_f_stat_sW_cy` and `permanova_f_stat_sW_condensed_cy`.

Two helpers added in `skbio/stats/distance/_permanova.py`:

- `_permanova_f_stat_sW_numba` — full matrix
- `_permanova_f_stat_sW_condensed_numba` — condensed matrix

Both use `@njit(parallel=True)` with `prange` over half the rows, pairing each `row_idx` with its mirror `n - row_idx - 2` to preserve the Cython implementation's load-balancing pattern.

The helpers write per-row contributions to a partials array and sum sequentially at the end, instead of doing the natural scalar reduction inside `prange`. This works around a Numba 0.65.x parfor cycle bug that fires when a `prange` scalar reduction reads from an array parameter used inside the inner loop. The docstrings include a `Replaces::` block showing the natural Numba code that would be used once the bug is fixed.

`_compute_f_stat` dispatches to Numba when available, Cython otherwise. The Cython path is untouched.

Any Numba-specific optimization work (e.g. the block-permutation strategy from `scikit-bio-binaries`) should be handled in a follow-up PR.

### Testing

New `PERMANOVANumbaTests` class with:

- Paired `_cy` / `_numba` helper tests for full and condensed matrices, asserting both produce the same s_W as a pure-Python reference.
- A dispatch fallback test using `patch.object(NUMBA_AVAILABLE, False)` to verify the Cython path under the same input matrix and seed.

The new tests use the `@numba_code` decorator from #2468 so they are skipped when Numba is not installed. The `numba.yml` workflow from #2468 picks them up automatically across `ubuntu-latest`, `macos-latest`, `windows-latest`, and `ubuntu-24.04-arm`.

```
$ python -m pytest skbio/stats/distance/tests/test_permanova.py -q
25 passed
$ python -m pytest skbio/stats/distance/tests/test_permanova.py -m numba_code -q
3 passed, 22 deselected
```

### DIFF
--- a/skbio/stats/distance/_permanova.py
+++ b/skbio/stats/distance/_permanova.py
@@ -113,10 +113,7 @@ if NUMBA_AVAILABLE:
 
             partials[row_idx] = group_sum
 
-        s_W = 0.0
-        for i in range(n_half):
-            s_W += partials[i]
-        return s_W
+        return partials.sum()
 
     @njit(parallel=True)
     def _permanova_f_stat_sW_condensed_numba(condensed_matrix, group_sizes, grouping):
@@ -181,10 +178,7 @@ if NUMBA_AVAILABLE:
 
             partials[row_idx] = group_sum
 
-        s_W = 0.0
-        for i in range(n_half):
-            s_W += partials[i]
-        return s_W
+        return partials.sum()
 
 
 @params_aliased([("distmat", "distance_matrix", "0.7.0", False)])
--- a/skbio/stats/distance/_permanova.py
+++ b/skbio/stats/distance/_permanova.py
@@ -156,7 +156,10 @@ if NUMBA_AVAILABLE:
             local_sum = 0.0
             for col_idx in range(row_idx + 1, n):
                 if grouping[col_idx] == group_idx:
-                    condensed_idx = row_idx * n + col_idx - ((row_idx + 2) * (row_idx + 1)) // 2
+                    condensed_idx = (
+                        row_idx * n + col_idx
+                        - ((row_idx + 2) * (row_idx + 1)) // 2
+                    )
                     val = condensed_matrix[condensed_idx]
                     local_sum += val * val
 
@@ -168,7 +171,10 @@ if NUMBA_AVAILABLE:
                 local_sum = 0.0
                 for col_idx in range(mirror_row + 1, n):
                     if grouping[col_idx] == group_idx:
-                        condensed_idx = mirror_row * n + col_idx - ((mirror_row + 2) * (mirror_row + 1)) // 2
+                        condensed_idx = (
+                            mirror_row * n + col_idx
+                            - ((mirror_row + 2) * (mirror_row + 1)) // 2
+                        )
                         val = condensed_matrix[condensed_idx]
                         local_sum += val * val
                 group_sum += local_sum / group_sizes[group_idx]
--- a/skbio/stats/distance/_permanova.py
+++ b/skbio/stats/distance/_permanova.py
@@ -8,6 +8,7 @@
 
 from __future__ import annotations
 
+import math
 from functools import partial
 from warnings import warn
 from typing import TYPE_CHECKING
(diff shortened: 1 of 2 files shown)


skbio/stats/distance/tests/test_permanova.py
```diff
@@ -408,5 +412,69 @@ def test_condensed_data_shape(self):
         self.assertEqual(self.dm_ties_redundant.data.shape, (n, n))
 
 
+class PERMANOVANumbaTests(TestCase):
```
sfiligoi: Why are you creating a new class, instead of just adding tests to the existing class?

skbio/stats/distance/_permanova.py
```diff
@@ -27,12 +28,165 @@
...
 
 
+if NUMBA_AVAILABLE:
+
+    @njit(parallel=True)
+    def _permanova_f_stat_sW_numba(distance_matrix, group_sizes, grouping):
+        """Compute s_W for full (non-condensed) distance matrix using Numba.
+
+        Replaces the following natural Numba code, which triggers a parfor
+        cycle bug in Numba 0.65.x when the scalar reduction variable is read
```
sfiligoi, line 50: Honestly, I find it had to believe that such a basic operation is broken in Numba.
It is pretty much a textbook example, shown in the official docs.

That said, the documentation does state it is a CPU-only feature, so moving to this alternative implementation will help with GPU porting.

sfiligoi:
@LarytheLord Do you have any benchmark numbers comparing Cython and Numba performance?

skbio/stats/distance/_permanova.py
```diff
@@ -27,12 +28,165 @@
...
+                local_sum = 0.0
+                for col_idx in range(mirror_row + 1, n):
+                    if grouping[col_idx] == group_idx:
+                        val = distance_matrix[mirror_row, col_idx]
+                        local_sum += val * val
+                group_sum += local_sum / group_sizes[group_idx]
+
+            partials[row_idx] = group_sum
+
+        s_W = 0.0
```
sfiligoi: Can you use
s_w = partials.sum()
?

LarytheLord:
The natural reduction code on Numba 0.65.1. It crashes at compile time:

File ".../numba/parfors/parfor.py", line 3118, in run
  p.redvars, p.reddict = get_parfor_reductions(self.func_ir, ...)
File ".../numba/parfors/parfor.py", line 3758, in get_parfor_reductions
  reduce_nodes = get_reduce_nodes(param, param_nodes[param], func_ir)
File ".../numba/parfors/parfor.py", line 3898, in get_reduce_nodes
  in_vars = set(noncyclic_lookup(v, True).name ...)
File ".../numba/parfors/parfor.py", line 3886, in noncyclic_lookup
  raise AssertionError("unexpected cycle in lookup()")
AssertionError: unexpected cycle in lookup()

Tried a few variants to narrow it down. A prange with only a scalar reduction compiles fine. Adding an inner loop reading a parameter array unconditionally is also fine. Adding an if conditional on a parameter array inside that inner loop is what breaks it. Removing the cross-iteration scalar reduction instead (writing to a per-iter slot) also compiles. So both pieces are needed together: the outer scalar reduction AND the conditional inner read on a parameter.

Our workaround removes the scalar reduction (per-iter slot writes plus partials.sum() after the prange exits) and keeps the conditional.

Agree on the GPU side too. parallel=True reductions are CPU-only in Numba, so this partials shape ports to @cuda.jit more directly later.

Reproducer:
import numpy as np
from numba import njit, prange

@njit(parallel=True)
def natural_sW(mat, group_sizes, grouping):
    n = mat.shape[0]
    s_W = 0.0
    for row in prange(n - 1):
        group = grouping[row]
        local = 0.0
        for col in range(row + 1, n):
            if grouping[col] == group:
                val = mat[row, col]
                local += val * val
        s_W += local / group_sizes[group]
    return s_W

mat = np.random.rand(6, 6); mat = (mat + mat.T) / 2; np.fill_diagonal(mat, 0)
grouping = np.array([0, 1, 2, 1, 0, 0], dtype=np.intp)
group_sizes = np.bincount(grouping).astype(np.intp)
natural_sW(mat, group_sizes, grouping)

skbio/stats/distance/tests/test_permanova.py
```diff
@@ -408,5 +385,68 @@ def test_condensed_data_shape(self):
...
+        self.dm_full = np.ascontiguousarray(self.dm_unequal.data)
+        self.dm_condensed = squareform(
+            self.dm_full, force='tovector', checks=False
+        )
+        self.grouping = np.asarray([0, 1, 2, 1, 0, 0], dtype=np.intp)
+        self.group_sizes = np.bincount(self.grouping).astype(np.intp)
+        self.ids = self.dm_unequal.ids
+        self.grouping_labels = self.grouping_unequal
+
+    def _expected_sW(self):
```
sfiligoi: Why do we need to re-implement permanova here?
Don't we already have the expected values to compare against?

I don't believe the test should have its own implementation.
We are just validating that the tested alogrithm returns the expected values, which were computed using a trusted implementation offline once.

sfiligoi (approved):
Looks good

LarytheLord:
**Linux benchmarks**  on AMD Ryzen 9 7940HS, 8c/8t, Ubuntu 22.04, GCC 11.4  
Python 3.12.13 | skbio 0.7.3-dev | numba 0.65.1 | numpy 2.4.6  
OMP_NUM_THREADS=8, NUMBA_NUM_THREADS=8, cython_openmp=YES, cpu_governor=powersave

**5k² EMP UniFrac DMs** (June 9)

| matrix | nperm | Cython (s) | Numba (s) | speedup |
|---|---:|---:|---:|---:|
| uw_emp_5k | 999 | 1.31 | 1.22 | 1.07× |
| uw_emp_5k | 9999 | 12.52 | 12.08 | 1.04× |
| uw_emp_5k | 99999 | 127.74 | 121.24 | 1.05× |
| wn_emp_5k | 999 | 1.33 | 1.23 | 1.08× |
| wn_emp_5k | 9999 | 12.79 | 12.14 | 1.05× |
| wn_emp_5k | 99999 | 129.12 | 121.60 | 1.06× |

**25k² EMP UniFrac DMs** (June 12)

| matrix | nperm | Cython (s) | Numba (s) | speedup |
|---|---:|---:|---:|---:|
| uw_emp_25k | 999 | 42.50 | 23.65 | 1.80× |
| uw_emp_25k | 9999 | 428.95 | 235.69 | 1.82× |
| wn_emp_25k | 999 | 43.81 | 23.92 | 1.83× |
| wn_emp_25k | 9999 | 436.33 | 236.50 | 1.84× |

F-stats match Cython exactly on all rows. Speedup increases with matrix size at 25k² the matrix exceeds all cache levels and Numba's LLVM generates more efficient code for float32 at that scale.

PR #2471, opened 2026-05-26
Fix documentation generation on windows

mataton:
# Fix invalid RST in Array API compatibility note

## Summary

The Array API compatibility table injected by `array_api_doc` was being placed
as body content inside a docstring's `Notes` section. numpydoc renders a `Notes`
header as a `.. rubric:: Notes` directive, which permits **no body content**, so
the injected intro sentence and table produced:

```
ERROR: Error in "rubric" directive: no content permitted.
```

This broke clean documentation builds (`make doc`, which runs `clean html` with
`-W`). Affected functions include `skbio.stats.composition.alr`, `closure`, and
`clr`.

## Fix

In `skbio/util/_docstring.py`, wrap the injected content in a self-contained
`.. note::` admonition (which accepts body content) instead of placing it
directly under the `Notes` rubric. The note block is indented to match the
surrounding docstring so the directive line and its body keep their relative
nesting.

## Why it surfaced inconsistently

The failure only appears on a **clean** build. Incremental builds
(`make -C doc html` without `clean`) reuse cached doctrees and never re-render
the affected docstrings, masking the error. This is also why it initially
appeared platform-specific — it is not; source and tool versions were identical.

## Testing

Verified that `make doc` (clean build, `-W`) succeeds on both Windows and Linux.
Rendered output confirmed: the compatibility table now displays correctly inside
a Note admonition.


qiyunzhu (approved):
Thanks for catching this!

PR #2483, opened 2026-06-06
ENH: add engine= selector to mantel and permanova

LarytheLord:
Builds on #2464, #2468, and #2470.

### Description

This makes the Numba implementations opt-in rather than automatically selected when Numba is installed, via an `engine=` selector on `mantel()` and `permanova()`.

- `engine={"cython", "numba"}` chooses the compute backend per call. It defaults to the global setting, which is configurable through `skbio._config` (`set_config("engine", ...)`) and is `"cython"` by default. `engine="numba"` requires the optional Numba dependency and raises `ImportError` if it is not installed.
- The internal Numba helper functions are renamed from the `_numba` suffix to `_nb`, matching the existing `_cy` (Cython) suffix.

The `engine=` value is currently a single string (`"cython"` or `"numba"`); the design leaves room to add a preference list and further backends later without breaking the API. The selector covers Mantel and PERMANOVA, the two functions that currently have Numba implementations.

### Testing

- New tests for the `engine` config option and the `_resolve_engine` helper (default resolution, unsupported values, and `engine="numba"` when Numba is absent).
- The existing Mantel and PERMANOVA dispatch tests now run both engines via `engine=` and assert they agree, replacing the previous `patch.object(..., NUMBA_AVAILABLE, False)` approach.
- `mantel` and `permanova` produce identical results with `engine="cython"` and `engine="numba"` on the EMP 5k UniFrac matrices (uw/wn), for both the F-statistic / correlation and the p-value.

### DIFF
--- a/skbio/stats/distance/_mantel.py
+++ b/skbio/stats/distance/_mantel.py
@@ -20,6 +20,7 @@ from scipy.stats import kendalltau, ConstantInputWarning, NearConstantInputWarni
 from ._cutils import mantel_perm_pearsonr_cy, mantel_perm_pearsonr_condensed_cy
 from skbio.stats.distance import DistanceMatrix
 from skbio.util import get_rng
+from skbio._config import _resolve_engine
 
 try:
     from numba import njit, prange
@@ -37,7 +38,7 @@ if TYPE_CHECKING:  # pragma: no cover
 if NUMBA_AVAILABLE:
 
     @njit(parallel=True)
-    def _mantel_perm_pearsonr_numba(
+    def _mantel_perm_pearsonr_nb(
         x_data, perm_order, xmean, normxm, ym_normalized, permuted_stats
     ):
         """Fused permute, normalize, and Pearson correlation for Mantel test.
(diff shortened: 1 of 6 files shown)


skbio/stats/distance/_mantel.py
```diff
@@ -472,7 +482,7 @@ def mantel(
     return orig_stat, p_value, n
 
 
-def _mantel_stats_pearson_flat(x, y_flat, permutations, seed=None):
```
sfiligoi, line 475: Wouldn't it be better if the default was None?
This way the default can change at a global level in the future, and no code changes will be needed.
(Same for all the places where engine is passed in functions)

LarytheLord: Agreed, switched to engine=None

sfiligoi (approved):
Looks good to me.

qiyunzhu (approved):
Hi @LarytheLord Good job! I only have two overarching questions regarding stochasticity: Is the Numba engine numerically reproducible given the same seed? Do Cython and Numba engines produce the same numerical result given the same seed? If the former is no, I would invite you to examine the code again. If the latter is no, that is fine but you may want to note it in the documentation. Otherwise the PR is good to go.

sfiligoi:
@qiyunzhu These is no stochasticity in the current Numba implementation;
the Numba implementations for both Mantel and Permanova are completely deterministic.
All stochasticity is handled in "plain Python" code right now.

So I think we are good for this PR.

That said, the next step is to bring stochasticity inside the Numba code, like scikit-bio-binaries did.
It's the only way to get full performance.
So, thanks for the reminder.

PR #2484, opened 2026-06-15
PERF: parallelize PERMANOVA Numba engine across the permutation axis

LarytheLord:
cc @sfiligoi

### What

The full-matrix `engine="numba"` path for `permanova()` previously called the
single-permutation s_W kernel once per permutation via the Python Monte Carlo
loop. This PR adds `_permanova_f_stat_sW_parallel_nb`, which computes s_W for
all `permutations + 1` groupings in a single Numba kernel that parallelises
across the permutation axis (`prange` over permutations). This is the
"running many f_stats in parallel" approach.

Condensed matrices and the Cython engine are unchanged — they still go through
the existing Monte Carlo path. (Parallelising the condensed numba path is a
natural follow-up; condensed already produces correct numba results today.)

Memory is bounded: permutations are processed in chunks (CHUNK x n buffer
reused), so peak memory does not grow with the permutation count.

### Benchmarks

Igor's node: AMD Ryzen 9 7940HS (8 cores / 8 threads used), Ubuntu 22.04,
GCC 11.4, `OMP_NUM_THREADS=8 NUMBA_NUM_THREADS=8`. Cython is OpenMP-enabled
(`libgomp`). EMP UniFrac DMs, **float32**. Min of repeated runs.

**5k²:**

| matrix | nperm | Cython (s) | Numba (s) | speedup |
|---|---:|---:|---:|---:|
| uw_emp_5k | 999 | 1.26 | 0.86 | 1.47× |
| uw_emp_5k | 9999 | 12.58 | 8.79 | 1.43× |
| wn_emp_5k | 999 | 1.28 | 0.88 | 1.45× |
| wn_emp_5k | 9999 | 12.85 | 8.86 | 1.45× |

**25k² — 3-way same-session run:**

| matrix | nperm | Binaries (s) | Cython (s) | Numba (s) | vs Cython |
|---|---:|---:|---:|---:|---:|
| uw_emp_25k | 999 | 2.23 | 42.03 | 18.77 | **2.24×** |
| uw_emp_25k | 9999 | 21.45 | 427.41 | 189.84 | **2.25×** |
| wn_emp_25k | 999 | 2.25 | 43.36 | 19.97 | **2.17×** |
| wn_emp_25k | 9999 | 22.37 | 436.93 | 194.35 | **2.25×** |

Speedup grows with problem size (better load balance across the permutation axis
amortises overhead at scale). scikit-bio-binaries remains ~8.7× faster via
hand-tuned AVX-512; that gap is expected for CPU Numba and is the motivation for
the GPU `@cuda.jit` path in follow-up work.

### Correctness

- **float64 input:** numba and cython are byte-identical — same statistic and
  p-value for the same seed (kernel consumes permutations in the same RNG order
  as `_run_monte_carlo_stats`). Verified across 72 configurations (sizes × group
  counts × permutation counts).
- **float32 input:** the numba kernel accumulates in float64 (promoting distances
  before squaring), whereas the cython kernel returns a float32 statistic.
  Because pseudo-F is built from `s_A = s_T - s_W` (small difference of large
  sums), that precision gap amplifies into the statistic's ~4th–5th digit
  (F differs ~1e-4). The numba value is the more precise of the two. p-values
  match except where float32 rounding flips a borderline permutation at large n
  — observed ≤0.002 at 25k², well below MC standard error (~0.016 at 999 perms).
- New tests: float32 agreement vs Cython, `permutations=0` → nan, negative
  permutations → `ValueError`, condensed falls back to the single-perm path,
  and a >CHUNK permutation count to verify RNG continuity across batches.
  Full PERMANOVA suite: 31/31 pass.

### Notes

- This is an engine optimisation, not a fresh conversion; the single-perm
  kernels remain for the condensed and Cython-fallback paths.

### DIFF
--- a/skbio/stats/distance/_permanova.py
+++ b/skbio/stats/distance/_permanova.py
@@ -26,11 +26,12 @@ from skbio.binaries import (
     permanova_available as _skbb_permanova_available,
     permanova as _skbb_permanova,
 )
+from skbio.util import get_rng
 from skbio.util._decorator import params_aliased
 from skbio._config import _resolve_engine
 
 try:
-    from numba import njit, prange
+    from numba import njit, prange, get_num_threads
     NUMBA_AVAILABLE = True
 except ImportError:
     NUMBA_AVAILABLE = False
@@ -97,7 +98,9 @@ if NUMBA_AVAILABLE:
             local_sum = 0.0
             for col_idx in range(row_idx + 1, n):
                 if grouping[col_idx] == group_idx:
-                    val = distance_matrix[row_idx, col_idx]
+                    # promote to float64 before squaring to match the Cython
+                    # ``double`` accumulator (matters for float32 input)
+                    val = np.float64(distance_matrix[row_idx, col_idx])
                     local_sum += val * val
 
             group_sum = local_sum / group_sizes[group_idx]
@@ -108,7 +111,7 @@ if NUMBA_AVAILABLE:
                 local_sum = 0.0
                 for col_idx in range(mirror_row + 1, n):
                     if grouping[col_idx] == group_idx:
-                        val = distance_matrix[mirror_row, col_idx]
+                        val = np.float64(distance_matrix[mirror_row, col_idx])
                         local_sum += val * val
                 group_sum += local_sum / group_sizes[group_idx]
 
@@ -158,7 +161,7 @@ if NUMBA_AVAILABLE:
                         row_idx * n + col_idx
                         - ((row_idx + 2) * (row_idx + 1)) // 2
                     )
-                    val = condensed_matrix[condensed_idx]
+                    val = np.float64(condensed_matrix[condensed_idx])
                     local_sum += val * val
 
             group_sum = local_sum / group_sizes[group_idx]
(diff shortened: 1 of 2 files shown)


skbio/stats/distance/_permanova.py
```diff
@@ -181,6 +184,122 @@ def _permanova_f_stat_sW_condensed_nb(condensed_matrix, group_sizes, grouping):
...
+
+        rng = get_rng(seed)
+        n_total = permutations + 1
+        out_sW = np.empty(n_total, np.float64)
+
+        # Process permutations in fixed-size chunks so peak memory is bounded
+        # by the chunk (CHUNK x n) rather than by the total permutation count
+        # (n_total x n). Permutations are generated in the same RNG order as
+        # the sequential Monte Carlo path, so p-values are unchanged.
+        CHUNK = 512
```
sfiligoi: This should be dynamic.
I.e. must scale with the number of threads/parallel compute.

I think skbio-binaries uses 32, but may need to be slightly different for Numba.

Reminder: The reason we use a small chunk is to fit the active part of buf (groupings_2d) in L1 cache (and you have an independent L1 cache on each core).
(Will be important once you implement block processing)

skbio/stats/distance/_permanova.py
```diff
@@ -368,12 +487,22 @@ def permanova(
...
 
-    test_stat_function = partial(
-        _compute_f_stat, sample_size, num_groups, distmat, group_sizes, s_T, engine
-    )
-    stat, p_value = _run_monte_carlo_stats(
-        test_stat_function, grouping, permutations, seed
-    )
+    if (
+        engine == "numba"
+        and NUMBA_AVAILABLE
```
sfiligoi: You don't need to check for NUMBA_AVAILABLE.
Just like you did not before.

(It has already been taken care of in _resolve_engine)

skbio/stats/distance/_permanova.py
```diff
@@ -181,6 +184,122 @@ def _permanova_f_stat_sW_condensed_nb(condensed_matrix, group_sizes, grouping):
...
+        # the sequential Monte Carlo path, so p-values are unchanged.
+        CHUNK = 512
+        buf = np.empty((min(CHUNK, n_total), sample_size), dtype=np.intp)
+        for start in range(0, n_total, CHUNK):
+            end = min(start + CHUNK, n_total)
+            for i in range(start, end):
+                if i == 0:
+                    buf[0] = grouping
+                else:
+                    buf[i - start] = rng.permutation(grouping)
```
sfiligoi, line 293: Just for clarification:
This uses the same RND scheme as the cython version, so it should result in identical stochastic behavior.
Which is GREAT.

skbio/stats/distance/_permanova.py
```diff
@@ -43,7 +44,7 @@
 
 if NUMBA_AVAILABLE:
 
-    @njit(parallel=True)
+    @njit(parallel=True, cache=True)
```
sfiligoi: Let's not force cache at this point of the development cycle.
May be appropriate later, but I think it is safer to keep the default behavior for now.

sfiligoi (approved):
Loos good to me.

PR #2488, opened 2026-06-21
PERF: single-pass row-tile PERMANOVA Numba kernel

LarytheLord:
This replaces the per-permutation Numba engine added in #2484 with a single-pass row-tile kernel, following up on the chunking/cache discussion there.

### Approach

The per-permutation kernel parallelises across permutations, so every worker streams the whole distance matrix — it is memory-bound on large matrices, which is why it plateaued around 2x. This kernel instead parallelises over **distance-matrix rows** and processes a whole chunk of permutations against each matrix element while it is loaded. Each row is owned by exactly one worker, so the matrix is read from memory once per chunk rather than once per worker. This is the cross-thread cache reuse the scikit-bio-binaries CPU path relies on (`pmn_f_stat_sW_cpu` in `permanova_dyn_impl.hpp`).

Details:
- The permutation loop is **branchless** (`rsum[p] += val2 * (groupings_T[col, p] == groupings_T[row, p])`) so Numba auto-vectorises it. The branchy form used in the C++ does not vectorise under Numba — I confirmed in the generated LLVM IR that the faithful port (and a branchless `NBLOCK=16` block variant, with and without `fastmath`) stays scalar, while this long branchless loop vectorises to AVX-512 on the test machine.
- Groupings are passed **sample-major `int32`** so the loop reads contiguous, low-bandwidth windows (int32 measured ~2.9x faster than int64 at 25k²).
- Rows are **mirror-paired** (`row` with `n - row - 2`) for load balance, matching `_permanova_f_stat_sW_nb`.
- `CHUNK = 32 * get_num_threads()` (same as before); larger chunks were measurably slower, consistent with keeping each worker's active grouping window in L1.
- Permutations are drawn in the same RNG order as before, so p-values are unchanged for float64 input.

### Benchmarks

EMP UniFrac DMs, AMD Ryzen 9 7940HS (8 threads), GCC/OpenMP, 999 / 9999 permutations. skbio-binaries (`libskbb.so`, x86-64-v4/AVX-512) shown for reference:

| matrix | perms | cython (s) | numba row-tile (s) | skbio-binaries (s) | numba vs cython | numba vs binaries |
|---|---:|---:|---:|---:|---:|---:|
| uw_emp_5k | 999 | 1.35 | 0.23 | 0.14 | 5.9x | 1.6x slower |
| uw_emp_5k | 9999 | 12.51 | 2.31 | 1.35 | 5.4x | 1.7x slower |
| wn_emp_5k | 999 | 1.31 | 0.24 | 0.14 | 5.5x | 1.7x slower |
| wn_emp_5k | 9999 | 12.75 | 2.23 | 1.35 | 5.7x | 1.7x slower |
| uw_emp_25k | 999 | 42.63 | 5.50 | 2.18 | 7.8x | 2.5x slower |
| uw_emp_25k | 9999 | 429.36 | 53.26 | 21.53 | 8.1x | 2.5x slower |
| wn_emp_25k | 999 | 42.96 | 5.51 | 2.21 | 7.8x | 2.5x slower |
| wn_emp_25k | 9999 | 434.86 | 52.19 | 21.61 | 8.3x | 2.4x slower |

At 25k²/9999 this is ~3.6x faster than the per-permutation engine (~191s) and ~2.5x off skbio-binaries; on the cache-resident 5k matrices binaries' lead narrows to ~1.6-1.7x. All three agree on the F-statistic (binaries also float32). Closing the remaining gap is the GPU `@cuda.jit` follow-up.

### Numerical agreement

Statistic and p-value are identical to the cython engine for float64 input. For float32 input the statistic differs at the ~1e-4 level because the Numba kernels accumulate in float64 (the cython path does not), so the p-value can differ by Monte-Carlo noise; this matches the existing per-permutation engine's behaviour.

### Notes

- Removes the now-unused `_permanova_f_stat_sW_parallel_nb` / `_run_permanova_parallel_nb`; the corresponding tests were renamed and still cover permutations=0, negative-raises, condensed fallback, and chunk-boundary behaviour.
- 31 tests pass with the JIT enabled and with `NUMBA_DISABLE_JIT=1`.

@sfiligoi

### DIFF
--- a/skbio/stats/distance/_permanova.py
+++ b/skbio/stats/distance/_permanova.py
@@ -31,7 +31,7 @@ from skbio.util._decorator import params_aliased
 from skbio._config import _resolve_engine
 
 try:
-    from numba import njit, prange, get_num_threads
+    from numba import njit, prange
     NUMBA_AVAILABLE = True
 except ImportError:
     NUMBA_AVAILABLE = False
@@ -192,15 +192,18 @@ if NUMBA_AVAILABLE:
 
         Parallelises over distance-matrix rows (each paired with its mirror row
         ``n - row - 2`` for load balance, as in ``_permanova_f_stat_sW_nb``)
-        rather than over permutations. Each row is owned by one worker, so the
-        matrix is read from memory only once per chunk, whereas the
-        per-permutation kernel has every worker stream the whole matrix. This
-        follows the cross-thread cache reuse of scikit-bio-binaries'
-        ``pmn_f_stat_sW_cpu`` in ``permanova_dyn_impl.hpp``.
+        rather than over permutations. Each row is owned by one worker, so each
+        distance is read from memory once per chunk and reused across all
+        permutations of the chunk, instead of streaming the whole matrix once
+        per permutation. The kernel is memory-bound, so this reuse -- not the
+        arithmetic -- is what determines its speed; it mirrors the cache reuse
+        of scikit-bio-binaries' ``pmn_f_stat_sW_cpu`` in
+        ``permanova_dyn_impl.hpp``.
 
-        The permutation loop is branchless so Numba can vectorise it, and
-        ``groupings_T`` is sample-major ``int32`` so the loop reads contiguous,
-        low-bandwidth windows.
+        The permutation loop is kept branchless (a uniform read-and-accumulate
+        over the grouping windows, with no data-dependent control flow), and
+        ``groupings_T`` is sample-major ``int32`` so each permutation's labels
+        are read as a contiguous, low-bandwidth window.
 
         Parameters
         ----------
@@ -211,20 +214,25 @@ if NUMBA_AVAILABLE:
         inv_group_sizes : np.ndarray, shape (num_groups,)
             Reciprocal group sizes (``1.0 / group_sizes``), precomputed so the
             hot loop multiplies instead of dividing.
-        partials : np.ndarray, shape (n // 2, n_perm)
-            Output, zeroed by the caller; ``partials[i]`` receives the s_W
-            contribution of the mirror-pair handled by iteration ``i``. The
-            caller sums over axis 0 to obtain s_W per permutation.
+        partials : np.ndarray, shape (n // 2, C) with C >= n_perm
+            Reused scratch (allocated once by the caller). Each iteration zeros
+            and fills the first ``n_perm`` entries of its own row ``partials[i]``
+            with the s_W contribution of the mirror-pair it handles. The caller
+            sums the first ``n_perm`` columns over axis 0 to obtain s_W per
+            permutation.
 
         """
         n = distance_matrix.shape[0]
         n_perm = groupings_T.shape[1]
         n_half = n // 2
 
-        # Each worker owns one row (and its mirror for load balance), so the
-        # distance matrix is read from memory only once per chunk.
+        # Each worker owns one row (and its mirror for load balance), so each
+        # distance is read once per chunk and reused across all permutations.
         for row_idx in prange(n_half):
+            # partials is reused across chunks, so zero this row's accumulator.
             local_sW = partials[row_idx]
+            for p in range(n_perm):
+                local_sW[p] = 0.0
             rsum = np.empty(n_perm, np.float64)
 
             for p in range(n_perm):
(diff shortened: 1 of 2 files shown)


skbio/stats/distance/_permanova.py
```diff
@@ -274,26 +313,36 @@ def _run_permanova_parallel_nb(
...
-        # resident in its per-core cache (scikit-bio-binaries uses 32
-        # permutations per worker). Chunking also bounds peak memory to
-        # (CHUNK x n) rather than the full permutation count (n_total x n).
-        # Permutations are generated in the same RNG order as the sequential
-        # Monte Carlo path, so p-values are unchanged.
+        # Same chunk size as the per-permutation path (32 per worker), which on
+        # this kernel keeps each worker's active grouping window in L1; chunks
+        # larger than this were measurably slower. Chunking also bounds peak
+        # memory to (CHUNK x n) regardless of the permutation count.
         CHUNK = 32 * get_num_threads()
```
sfiligoi: I don't think you are parallelizing over CHUNK anymore, right?
Does it still make sense to scale it with the number of threads?

skbio/stats/distance/_permanova.py
```diff
@@ -274,26 +313,36 @@ def _run_permanova_parallel_nb(
...
                     buf[0] = grouping
                 else:
                     buf[i - start] = rng.permutation(grouping)
-            _permanova_f_stat_sW_parallel_nb(
-                distmat.data, group_sizes, buf[: end - start], out_sW[start:end]
+
+            groupings_T = np.ascontiguousarray(
+                buf[:chunk_size].T, dtype=np.int32
+            )
+            partials = np.zeros((n_half, chunk_size), np.float64)
```
sfiligoi: This may be creating memory allocations inside the loop, which tends to be slow.
Why not create it outside the loop, and reuse the buffer?
You know the max size ahead of time.

skbio/stats/distance/_permanova.py
```diff
@@ -274,26 +313,36 @@ def _run_permanova_parallel_nb(
...
+            chunk_size = end - start
             for i in range(start, end):
                 if i == 0:
                     buf[0] = grouping
                 else:
                     buf[i - start] = rng.permutation(grouping)
-            _permanova_f_stat_sW_parallel_nb(
-                distmat.data, group_sizes, buf[: end - start], out_sW[start:end]
+
+            groupings_T = np.ascontiguousarray(
```
sfiligoi: I don't understand why you need ascontiguousarray here.
The buffer is already linear.

LarytheLord: The kernel reads each permutation's labels for a fixed sample as a contiguous window (groupings_T[row, :] / [col, :]), so it needs sample-major layout. buf is permutation-major (rng.permutation fills a row) and buf.T is a non-contiguous view, so this materializes the C-contiguous copy and with buf now int32 it's an int32→int32 transpose. Generating sample-major directly would need strided column writes, which I measured as slower, but happy to change it if you see a cleaner approach.

skbio/stats/distance/_permanova.py
LarytheLord: moved it out of the loop and reuse a single buffer (max size (n_half, CHUNK), known up front); the kernel now zeros its own row each iteration.

skbio/stats/distance/_permanova.py
LarytheLord: fixed it to a constant sized for L1 instead of scaling with thread count. A 64/128/256 sweep at 25k² was flat for int32 (256 ≈ 64, ~5.2s), so I kept 256.

skbio/stats/distance/_permanova.py
sfiligoi: Ah, yes, I see it now.
`buf` is really 2D, even though you use a single index most of the time.

Any reason why you could not create the (and fill) the buffer using the alternate/transposed memory layout to start with?

LarytheLord: I tried that filling the buffer transposed means writing each permutation down a column of the (n, n_perm) array one strided write per sample which thrashes cache during generation.
an earlier version that filled it column by column like this was measurably slower. Filling it permutation major instead 'each rng.permutation lands in a contiguous row, so generation stays sequential' and transposing once per chunk is faster overall the transpose is a single bulk pass over the small grouping buffer, a small fraction of the run at 25k². 
Together with this try I also did other trials with little different combinations and all of them were slower.

sfiligoi (approved):
Looks good to me.

PR #2490, opened 2026-06-24
ENH: Add optional Numba support

sfiligoi:
This is a cumulative PR that brings optional Numba support for Permanova and Mantel.

It is composed of the following partial PRs:

- https://github.com/scikit-bio/scikit-bio/pull/2464
- https://github.com/scikit-bio/scikit-bio/pull/2466
- https://github.com/scikit-bio/scikit-bio/pull/2468
- https://github.com/scikit-bio/scikit-bio/pull/2470
- https://github.com/scikit-bio/scikit-bio/pull/2483
- https://github.com/scikit-bio/scikit-bio/pull/2484
- https://github.com/scikit-bio/scikit-bio/pull/2486
- https://github.com/scikit-bio/scikit-bio/pull/2488

### DIFF
--- a/.github/workflows/numba.yml
+++ b/.github/workflows/numba.yml
@@ -2,7 +2,7 @@ name: Numba
 
 on:
   pull_request:
-    branches: [ development_numba ]
+    branches: [ main ]
     paths-ignore: ["web/**", "**.md", "README.rst"]
   workflow_dispatch:
 
--- a/.github/workflows/numba.yml
+++ b/.github/workflows/numba.yml
@@ -56,3 +56,23 @@ jobs:
 
       - name: Run full test suite with Numba installed
         run: make test
+
+      # coverage.py cannot trace JIT-compiled kernels, so re-run the suite
+      # with the JIT disabled (kernels run as plain Python). Linux x86_64 only,
+      # matching ci.yml; Codecov merges this report with the main one.
+      - name: Run unit tests with coverage (Numba JIT disabled)
+        if: runner.os == 'Linux' && runner.arch == 'X64'
+        env:
+          WITH_COVERAGE: "TRUE"
+          NUMBA_DISABLE_JIT: "1"
+        run: make test
+      - name: Generate coverage reports
+        if: runner.os == 'Linux' && runner.arch == 'X64'
+        run: |
+          cd ci && coverage lcov --rcfile ../.coveragerc
+      - name: Upload coverage reports to Codecov
+        if: runner.os == 'Linux' && runner.arch == 'X64'
+        uses: codecov/codecov-action@v5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          slug: scikit-bio/scikit-bio
--- a/.github/workflows/numba.yml
+++ b/.github/workflows/numba.yml
@@ -1,0 +1,58 @@
+name: Numba
+
+on:
+  pull_request:
+    branches: [ development_numba ]
+    paths-ignore: ["web/**", "**.md", "README.rst"]
+  workflow_dispatch:
+
+env:
+  latest_python: "3.14"
+
+jobs:
+  conf:
+    # This job is needed to route the global environment variables into
+    # a context that's available for matrix (and name, but that's unimportant)
+    name: Prepare Numba Test Plan
+    runs-on: ubuntu-latest
+    outputs:
+      latest_python: ${{ steps.set-vars.outputs.latest_python }}
+    steps:
+      - name: Report Plan
+        id: set-vars
+        run: |
+          echo "latest_python=$latest_python" >> $GITHUB_OUTPUT
+
+  test:
+    name: Numba (${{ needs.conf.outputs.latest_python }}, ${{ matrix.os }})
+    needs: conf
+    runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: bash
+    strategy:
+      fail-fast: false
+      matrix:
+        # macos-15-intel is omitted: Numba and llvmlite no longer ship macOS
+        # x86_64 wheels, so `pip install numba` would fall back to a source
+        # build requiring an LLVM toolchain.
+        os: [ubuntu-latest, macos-latest, windows-latest, ubuntu-24.04-arm]
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-python@v6
+        with:
+          python-version: ${{ needs.conf.outputs.latest_python }}
+
+      - name: Install scikit-bio + Numba
+        run: |
+          python -m ensurepip
+          pip install --prefer-binary -r ci/requirements.txt -r ci/requirements.test.txt
+          pip install numba
+          pip install --no-deps .
+          pip list
+
+      - name: Run Numba-marked tests
+        run: cd ci && pytest -m numba_code --pyargs skbio
+
+      - name: Run full test suite with Numba installed
+        run: make test
--- a/.github/workflows/numba.yml
+++ b/.github/workflows/numba.yml
@@ -1,6 +1,9 @@
 name: Numba
 
 on:
+  push:
+    branches: [ main ]
+    paths-ignore: ["web/**", "**.md", "README.rst"]
   pull_request:
     branches: [ main ]
     paths-ignore: ["web/**", "**.md", "README.rst"]
@@ -33,9 +36,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # macos-15-intel is omitted: Numba and llvmlite no longer ship macOS
-        # x86_64 wheels, so `pip install numba` would fall back to a source
-        # build requiring an LLVM toolchain.
+        # macos-15-intel and windows-11-arm are omitted: Numba and llvmlite
+        # do not ship wheels for those targets (macOS x86_64 or Windows ARM64),
+        # so `pip install numba` would fall back to a source build requiring an
+        # LLVM toolchain.
         os: [ubuntu-latest, macos-latest, windows-latest, ubuntu-24.04-arm]
     steps:
       - uses: actions/checkout@v6
@@ -52,7 +56,7 @@ jobs:
           pip list
 
       - name: Run Numba-marked tests
-        run: cd ci && pytest -m numba_code --pyargs skbio
+        run: cd ci && pytest -m numba --pyargs skbio
 
       - name: Run full test suite with Numba installed
         run: make test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
     branches: [ main ]
     paths-ignore: ["web/**", "**.md", "README.rst"]
   pull_request:
-    branches: [ main, development_numba ]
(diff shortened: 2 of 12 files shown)


LarytheLord:
Looked into both the security alerts and the failing check neither is specific to the Numba code.

Security alerts CodeQL "Workflow does not contain permissions"

These aren't specific to numba.yml. None of the six workflows (ci, array-api, release, website, wheels, numba) declare a permissions: block, so GITHUB_TOKEN defaults to write-all (visible in the run logs: Contents: write, PullRequests: write, …). numba.yml is flagged only because it's the new file in this PR's diff; the same gap is in the others, just not in this diff. So it's a repo-wide hardening item rather than something new here.

The fix is GitHub's suggested changeset a least-privilege block:

permissions:
  contents: read
We can add it to numba.yml, and to the other workflows in one pass if you'd like.

Failing check Build Documentation

This is a transient network failure, not a content/doc error. The build runs sphinx-build -W (warnings = errors), and intersphinx timed out fetching scipy's inventory:

intersphinx inventory 'https://docs.scipy.org/doc/scipy/objects.inv' not fetchable ... ConnectTimeout
build finished with problems, 1 warning (with warnings treated as errors).
That single fetch warning, under -W, fails the build unrelated to this PR's changes; it would hit any PR whenever scipy's docs server is briefly unreachable. Re running the job should clear it. (Reproduced locally: an unreachable intersphinx inventory under -W gives the exact same failure, and passes once the inventory is reachable or -W is dropped.)

If it recurs, durable options are to drop -W or cache/vendor the intersphinx inventories. suppress_warnings and a shorter intersphinx_timeout don't help the former doesn't catch the fetch-failure warning, the latter just fails faster.

.github/workflows/numba.yml
```diff
@@ -0,0 +1,78 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      latest_python: ${{ steps.set-vars.outputs.latest_python }}
+    steps:
+      - name: Report Plan
+        id: set-vars
+        run: |
+          echo "latest_python=$latest_python" >> $GITHUB_OUTPUT
+
+  test:
```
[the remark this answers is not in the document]
sfiligoi: Given that all other CI scripts are consistent with this, I am dismissing this alert.
But it is probably worth considering changing all CI workflows along the lines, so I opened issue #2492

.github/workflows/numba.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Generate coverage reports
+        if: runner.os == 'Linux' && runner.arch == 'X64'
+        run: |
+          cd ci && coverage lcov --rcfile ../.coveragerc
+      - name: Upload coverage reports to Codecov
+        if: runner.os == 'Linux' && runner.arch == 'X64'
+        uses: codecov/codecov-action@v5
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          slug: scikit-bio/scikit-bio
```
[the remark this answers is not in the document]
sfiligoi: As above, see #2492

qiyunzhu:
@sfiligoi @LarytheLord I think it is a good idea to add `permissions: contents: read` to all workflows. To my understanding, this command only limits writing to the same repo, whereas none of the workflows attempt to do so. But we won't know unless we test. I think this can be a separate PR.

intersphinx inventory not fetchable... This is common, due to stochastic network issues. Just re-run the workflow and it will likely work.

.github/workflows/numba.yml
```diff
@@ -0,0 +1,78 @@
+name: Numba
+
+on:
+  pull_request:
+    branches: [ main ]
+    paths-ignore: ["web/**", "**.md", "README.rst"]
+  workflow_dispatch:
```
qiyunzhu, line 10: This command means the workflow can be manually triggered. In what scenarios will this happen?

.github/workflows/numba.yml
```diff
@@ -0,0 +1,78 @@
...
+    defaults:
+      run:
+        shell: bash
+    strategy:
+      fail-fast: false
+      matrix:
+        # macos-15-intel is omitted: Numba and llvmlite no longer ship macOS
+        # x86_64 wheels, so `pip install numba` would fall back to a source
+        # build requiring an LLVM toolchain.
+        os: [ubuntu-latest, macos-latest, windows-latest, ubuntu-24.04-arm]
```
qiyunzhu, line 43: Windows ARM is not here either. There will still be some time until Numba becomes a generally adopted component of modern computers.

.github/workflows/numba.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Install scikit-bio + Numba
+        run: |
+          python -m ensurepip
+          pip install --prefer-binary -r ci/requirements.txt -r ci/requirements.test.txt
+          pip install numba
+          pip install --no-deps .
+          pip list
+
+      - name: Run Numba-marked tests
+        run: cd ci && pytest -m numba_code --pyargs skbio
```
qiyunzhu: I would call it `numba` instead of `numba_code` for conciseness.

.github/workflows/numba.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Run full test suite with Numba installed
+        run: make test
+
+      # coverage.py cannot trace JIT-compiled kernels, so re-run the suite
```
qiyunzhu, line 64: Some questions here: 1) Can Numba code not be counted toward coverage at all? 3) If so, will our Codecov rate drop significantly? 2) If so (to 1), I am not sure if we need to upload reports to Codecov in this workflow, since `ci.yml` already takes care of it, in the absence of Numba.

.github/workflows/numba.yml
LarytheLord: workflow_dispatch adds a manual "Run workflow" trigger that runs the workflow against a chosen branch, separate from any PR. Since this workflow only has a pull_request trigger (no push:), manual dispatch is a way to run the Numba suite on a branch directly without opening a PR e.g. against development_numba/main itself, or to re check the numba paths after a new numba/llvmlite release without a code change. array-api.yml and wheels.yml also include it. It's a convenience rather than a necessity, we can drop it also and rely on the PR trigger if you'd prefer to keep the surface minimal.

.github/workflows/numba.yml
LarytheLord: Yes windows-11-arm is omitted for the same reason as macos-15-intel: Numba/llvmlite don't ship Windows ARM64 wheels, so pip install numba there would fail (or fall back to an unsupported source build needing an LLVM toolchain). The comment only spells out the macOS Intel case I'll extend it to mention Windows ARM too so the omission is documented. And agreed on the broader point: Numba's platform coverage is still incomplete (Windows ARM64, Intel macOS), the matrix only covers the platforms where it actually installs.

.github/workflows/numba.yml
LarytheLord: Will do, I'll rename the marker to numba. 
Heads up that numba_code is also the exported test decorator (@numba_code), so to avoid from skbio.util import numba shadowing the package I'd rename that one to requires_numba or to something similar rather than numba, would that work.

.github/workflows/numba.yml
LarytheLord: 1) Not while it's JIT compiled. @njit kernels compile to native code via LLVM and never go through CPython's line tracer, so coverage.py can't see inside them I checked on a minimal kernel: JIT on → body lines missing, NUMBA_DISABLE_JIT=1 → fully covered. That re-run is the way to count them.

2) Total/project coverage barely moves, but the patch check would fail: the kernels are ~360 new lines (most of this PR's diff), and .codecov.yml sets an 80% patch target uncovered, they'd pull the patch well under it.

3) ci.yml can't reach the kernels: it runs without Numba, so NUMBA_AVAILABLE is False, the whole guarded block never executes, and the marked tests skip those lines are missing in its report. The JIT disabled run here is the only place they're traced. 
Neither upload uses flags, so Codecov just combines both reports on the same commit they're complementary, not duplicate. Your point that ci.yml covers the non-Numba path is exactly why this one is needed for the Numba path.

If you'd rather centralize uploads in ci.yml, the alternative is to exclude the kernels from coverage (# pragma: no cover / a .coveragerc omit) and accept no signal on them but I'd lean toward keeping the upload since these are the new performance critical paths which are developed continuously.

qiyunzhu: Let me try to understand, by saying:

> Codecov just combines both reports on the same commit they're complementary, not duplicate.

Do you mean that Numba code can still be executed with JIT off, and they are still counted by `coverage.py` in that situation. Therefore, despite that a fraction of the codebase is Numba, we can still reach near-100% coverage because of this. `ci.yml` currently doesn't run Numba code, with JIT on or off. `numba.yml` does. Therefore, the coverage reported by `ci.yml` and `numba.yml` are complementary. Is my understanding correct?

.github/workflows/numba.yml
qiyunzhu: Makes sense to me. I wonder if we should add `push` to `numba.yml` as well, just like `ci.yml`. Will there be any concern of that?

.github/workflows/numba.yml
qiyunzhu: Also I don't know how complementary Codecov reports work. If `ci.yml` reports 95% coverage because it doesn't test Numba code, and `numba.yml` reports 5% coverage because it only runs Numba code, will the combined report be 100%?

.github/workflows/numba.yml
LarytheLord: Yes we can scoped to branches: [main] like ci.yml, no real concern. 
It actually helps in two ways, it validates the Numba paths on main after a merge (numba.yml only runs on PRs today) and it keeps main's coverage accurate on push to main, ci.yml uploads without Numba, so without a numba.yml push the kernels would show uncovered on main.

The only thing to avoid is an unscoped push: that would run the 4-OS Numba matrix on every topic branch push and double up with the PR runs. Keeping it to branches: [main] (matching ci.yml) sidesteps that.

.github/workflows/numba.yml
LarytheLord: Yes, our understanding is correct. With NUMBA_DISABLE_JIT=1 the @njit kernels run as plain Python, so coverage.py traces them, ci.yml never runs the Numba code (Numba isn't installed there), numba.yml does so the reports are complementary.

On the 95% / 5% question: Codecov merges at the line level, not by averaging percentages. A line is covered if any report hits it, so disjoint 95% + 5% does combine to ~100% (overlapping lines just aren't double counted).

qiyunzhu: Very good. Thank you!

.github/workflows/numba.yml
sfiligoi: This was addressed

.github/workflows/numba.yml
sfiligoi: This was addresssed.

sfiligoi:
All the comments have been answered, all requests for change addressed and the updated code is passing all the tests, so I am merging this PR.

PR #2491, opened 2026-06-24
Cdm numba cpu clean

fyash1010:

* [X] I have read the [contribution guidelines](https://scikit.bio/contribute.html).

### DIFF
--- a/skbio/stats/ordination/_center_distance_matrix_numba.py
+++ b/skbio/stats/ordination/_center_distance_matrix_numba.py
@@ -1,0 +1,75 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2013--, scikit-bio development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+"""Optional CPU Numba implementation of distance-matrix double-centering."""
+
+import numpy as np
+
+try:
+    from numba import njit, prange
+
+    NUMBA_AVAILABLE = True
+except ImportError:
+    NUMBA_AVAILABLE = False
+
+
+if NUMBA_AVAILABLE:
+
+    @njit(parallel=True)
+    def e_matrix_means_nb(mat, centered, row_means):
+        """Apply E-matrix transform and collect row/global means in one pass."""
+        n = mat.shape[0]
+        global_sum = np.float64(0.0)
+
+        for row in prange(n):
+            row_sum = np.float64(0.0)
+
+            for col in range(n):
+                val = np.float64(mat[row, col])
+                el = np.float64(-0.5) * val * val
+                centered[row, col] = el
+                row_sum += el
+
+            row_means[row] = row_sum / np.float64(n)
+            global_sum += row_sum
+
+        return (global_sum / np.float64(n)) / np.float64(n)
+
+    @njit(parallel=True)
+    def f_matrix_inplace_nb(row_means, global_mean, centered):
+        """Double-center E-matrix in-place."""
+        n = centered.shape[0]
+
+        n_blocks = (n + 23) // 24
+        for brow in prange(n_blocks):
+            trow = brow * 24
+            trow_max = min(trow + 24, n)
+
+            for tcol in range(0, n, 24):
+                tcol_max = min(tcol + 24, n)
+
+                for row in range(trow, trow_max):
+                    gr_mean = global_mean - row_means[row]
+
+                    for col in range(tcol, tcol_max):
+                        centered[row, col] += gr_mean - row_means[col]
+
+    def center_distance_matrix_nb(mat, centered):
+        """Drop-in style replacement for ``center_distance_matrix_cy``.
+
+        Parameters
+        ----------
+        mat : ndarray (n, n), float32 or float64, C-contiguous
+            Input distance matrix.
+        centered : ndarray (n, n), same dtype as ``mat``
+            Pre-allocated output array. May alias ``mat`` for in-place use.
+        """
+        n = mat.shape[0]
+        row_means = np.zeros(n, dtype=mat.dtype)
+        global_mean = e_matrix_means_nb(mat, centered, row_means)
+        f_matrix_inplace_nb(row_means, np.float64(global_mean), centered)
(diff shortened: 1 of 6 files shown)


fyash1010:
@sfiligoi @LarytheLord

sfiligoi:
@fyash1010 Please rebase from `main` and make sure only the relevant commits show in the PR.

skbio/stats/ordination/_utils.py
```diff
@@ -6,11 +6,23 @@
...
 # ----------------------------------------------------------------------------
 
+import os
+
 import numpy as np
 
 from ._cutils import center_distance_matrix_cy
 
 
+def _get_center_backend():
```
sfiligoi: I don't think we want to use an environment variable.
You should use the `engine` argument, like everywhere else.

skbio/stats/ordination/_center_distance_matrix_numba.py
```diff
@@ -0,0 +1,81 @@
...
+                tcol_max = min(tcol + 24, n)
+
+                for row in range(trow, trow_max):
+                    gr_mean = global_mean - row_means[row]
+
+                    for col in range(tcol, tcol_max):
+                        centered[row, col] += gr_mean - row_means[col]
+
+
+def center_distance_matrix_nb(mat, centered):
```
sfiligoi: You should define this function in the `if` block, not have the `if` block inside the function.

skbio/stats/ordination/tests/test_util.py
```diff
@@ -117,11 +149,152 @@ def test_center_distance_matrix_single(self):
...
+            center_distance_matrix_cy(mat, cy_centered)
+            center_distance_matrix_nb(mat, nb_centered)
+
+            npt.assert_allclose(nb_centered, cy_centered, rtol=rtol, atol=atol)
+
+    @numba_code
+    def test_center_distance_matrix_numba_matches_cpu_dispatch(self):
+        matrix_copy = copy.deepcopy(self.dist_mat)
+
+        with patch.dict("os.environ", {"SKBIO_PCOA_CENTER_BACKEND": "numba"}):
```
sfiligoi: We should not use the environment variables.

skbio/stats/ordination/tests/test_util.py
```diff
@@ -27,13 +38,27 @@ def setUp(self):
...
 
         self.small_mat = np.array([[7, 5, 5], [4, 4, 9], [7, 5, 3]])
-        self.dist_mat = np.asarray([[0., 7., 5., 5.], [7., 0., 4., 9.],
-                                    [5., 4., 0., 3.], [5., 9., 3., 0.]],
-                                   dtype=np.float64)
-        self.dist_mat_fp32 = np.asarray([[0., 7., 5., 5.], [7., 0., 4., 9.],
-                                         [5., 4., 0., 3.], [5., 9., 3., 0.]],
-                                        dtype=np.float32)
-
+        self.dist_mat = np.asarray(
```
sfiligoi: You should not change the formatting of the code that has no functional changes.

skbio/stats/ordination/_center_distance_matrix_numba.py
```diff
@@ -0,0 +1,75 @@
...
+    from numba import njit, prange
+
+    NUMBA_AVAILABLE = True
+except ImportError:
+    NUMBA_AVAILABLE = False
+
+
+if NUMBA_AVAILABLE:
+
+    @njit(parallel=True, cache=True)
```
sfiligoi: Please remove the cache directive, to be consistent with other Numba code we have. (e.g. permanova)

sfiligoi:
The Numba implementation of center_matrix seems reasonable to me.

But I have a hard time understanding the updated tests.
@LarytheLord Could you please have a look?

LarytheLord:
Looked through it. The Numba kernel matches what we did for permanova, so no concerns on the implementation.

On the tests they're correct, but they're hard to follow because most of them check that engine= routes to the right helper (via patch / assert_called_once) rather than checking the result, and the same numba vs cython equivalence is repeated a few times at different levels. A couple also test the Cython path, which isn't what this PR changes.

@fyash1010 the permanova tests in test_permanova.py are the closest template they just compare engine="numba" against engine="cython" outputs directly, no mocking. Matching that shape would cut this down and make the intent obvious.

skbio/stats/ordination/tests/test_util.py
```diff
@@ -122,6 +126,145 @@ def test_center_distance_matrix_single(self):
...
+        with (
+            patch(
+                "skbio.stats.ordination._center_distance_matrix_numba."
+                "center_distance_matrix_nb",
+                wraps=cdm_numba.center_distance_matrix_nb,
+            ) as center_distance_matrix_nb,
+        ):
+            dm_centered = center_distance_matrix(matrix_copy, engine="numba")
+
+        center_distance_matrix_nb.assert_called_once()
```
LarytheLord: The patch(wraps=...) + assert_called_once() verifies that engine= routed to the numba helper. The permanova tests don't assert routing this way. They test the numba helper directly (test_center_distance_matrix_float64_numba already does that here) and then compare engine="numba" against engine="cython" outputs. Matching that pattern would drop the mock while keeping the same coverage.

fyash1010: Thanks, updated the CDM tests to match the PERMANOVA-style pattern. I removed the mock/wraps routing assertions and the Cython-only helper checks, kept the direct Numba helper coverage, and now compare engine="numba" against engine="cython" outputs directly for both out-of-place and inplace centering.

LarytheLord:
Thanks, the tests read much better now, the mock/wraps routing checks are gone which was my main concern. Ran them locally and they pass, including under -m numba and with the JIT disabled, so the coverage path is fine. Implementation looks good to me.

One optional thing: the numba tests all use the 4x4 matrix, which stays inside a single tile in f_matrix_inplace_nb. A random ~100x100 case compared numba vs cython would exercise the block loop. Not a blocker.

On CI, the red Test (3.14, ...) jobs are a conda install flake (sqlite3.OperationalError: database is locked), not a test failure, the tests were skipped and fail-fast cancelled the rest. A re run should clear it.

sfiligoi:
I am OK delaying improved tests (100x100) to later.
@qiyunzhu Is the current PR with you?

sfiligoi:
Superseded by #2508 .
Closing without merging.

PR #2495, opened 2026-06-28
ENH: Add inplace parameter to TreeNode.prune and bifurcate

### LINKED ISSUE #1571: Inconsistency in manupulating existing tree or creating new tree
Multiple scikit-bio TreeNode methods that manipulates the entire tree, behaves inconsistently whether to create a new tree or to edit the existing tree. For examples, `shear` and `root_at_midpoint` return a new tree, while `prune` and `bifurcate` operates on the same tree. There is no `inplace` argument for any of these methods.

jissen706:
## Summary

Closes #1571.

Several `TreeNode` methods that manipulate the whole tree behave inconsistently: `shear`, `root`, `root_at_midpoint`, and `root_by_outgroup` accept an `inplace` argument and **return** the resulting tree, whereas `prune` and `bifurcate` only ever modify the tree in place and return `None`.

This PR closes that gap by giving `prune` and `bifurcate` the same `inplace` interface as the other whole-tree methods:

* Add an `inplace: bool` parameter to both methods.
* Both methods now **return the resulting tree** (previously they returned `None`).
* With `inplace=False`, the original tree is left untouched and the operation is performed on a copy (mirroring `shear`).

### Backward compatibility

`inplace` defaults to **`True`**, which preserves the existing in-place behavior. Existing calls like `tree.prune()` and `tree.bifurcate()` keep mutating the tree exactly as before; they simply now also return it (callers that ignore the return value are unaffected). No existing call sites in the codebase rely on the old `None` return.

I chose `inplace=True` as the default specifically to avoid a silent breaking change. The other whole-tree methods default to `inplace=False`, so this leaves a small residual difference in defaults — happy to flip `prune`/`bifurcate` to `inplace=False` instead if maintainers prefer full consistency and are comfortable with the (breaking) behavior change, in which case I'd add a deprecation path.

## Testing

* Added `test_prune_inplace` and `test_bifurcate_inplace` covering: the default in-place behavior returns `self`; explicit `inplace=True` returns `self`; and `inplace=False` leaves the original unchanged and returns a modified copy.
* Full `skbio/tree` test suite passes locally (165 passed).
* `prune`/`bifurcate` doctests updated (they now capture the returned tree) and pass.
* `ruff check` is clean on the modified module.

## Checklist

- [x] This pull request includes a `CHANGELOG.md` entry (under 0.7.4-dev, Features).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 * Add optional support for the Numba backend [#2483](https://github.com/scikit-bio/scikit-bio/pull/2483), with Permanova and Mantel currently using it [#2488](https://github.com/scikit-bio/scikit-bio/pull/2488)and [#2464](https://github.com/scikit-bio/scikit-bio/pull/2464).
+* `TreeNode.prune` and `TreeNode.bifurcate` now accept an `inplace` parameter (default `True`, preserving the previous in-place behavior) and return the resulting tree. This makes them consistent with other whole-tree methods such as `shear` and `root_at_midpoint`. Set `inplace=False` to leave the original tree unchanged and operate on a copy ([#2495](https://github.com/scikit-bio/scikit-bio/pull/2495)).
 
 ### Performance enhancements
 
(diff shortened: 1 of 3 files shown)


qiyunzhu (approved):
This looks good. Thanks for contributing!

PR #2496, opened 2026-06-30
EHN: Transition probability matrices

### LINKED ISSUE #2463: Transitional probability models
I am interested in working on adding log-likelihood (LL) computation for trees. As part of this, I would need to implement transition probability matrices for substitution models.

Before starting implementation, I wanted to ask about the preferred project architecture/design for this functionality.

My main question is where substitution models and transition probability computations should live within the project structure. Conceptually, they seem closely related to the `tree` submodule, but I am not sure whether that is the best place for them.

I also have questions about the preferred interface for substitution models. One possibility would be standalone functions, for example:

```python
def k2p_transition_probabilities(d, kappa):
    ...
```

However, this would require likelihood computation code to pass both the function and all model parameters separately.

Another possibility is to represent models as closures/factories:

```python
def get_k2p_model(kappa):
    def k2p(d):
        <use kappa here>
        ...
    return k2p
```

This would encapsulate model parameters cleanly and simplify the LL interface. However it fells a bit less intuitive to me.

A third option would be using model objects/classes, where each model stores its parameters and exposes a method for computing transition probabilities.

At the moment, I am leaning toward the object-oriented approach because it seems easier to extend to additional models and likelihood-related functionality later, but I would appreciate guidance on which design style would best fit the project.

keder:
Resolves #2463


skbio/sequence/tpm.py
```diff
@@ -0,0 +1,696 @@
...
+    GrammaredSequence,
+    DNA,
+    RNA,
+    Protein,
+    SubstitutionMatrix,
+)
+import skbio.sequence as sk_seqtype
+
+if TYPE_CHECKING:  # pragma: no cover
+    from numpy.typing import ArrayLike
```
qiyunzhu: I think `NDArray` can be imported here, rather than used as `npt.NDArray`.

skbio/sequence/tpm.py
```diff
@@ -0,0 +1,696 @@
...
+        molecules. Mammalian Protein Metabolism, 3(21), 132.
+    """
+
+    return _wrap_vector_tpm_function(
+        _jc69, d, seqtype=seqtype, allowed_seqtypes=("DNA", "RNA")
+    )
+
+
+def _jc69(
+    d: npt.NDArray[np.float64],
```
qiyunzhu: I think it can just be `NDArray` without fixing to float64. The algorithm should support float32 and other float formats as well I guess.

skbio/sequence/tpm.py
```diff
@@ -0,0 +1,696 @@
...
+
+
+def jc69(
+    d: float,
+    seqtype: str = "DNA",
+) -> SubstitutionMatrix:
+    r"""
+    Calculate the JC69 transition probability matrix for a given distance.
+
+    .. versionadded:: 0.7.3
```
qiyunzhu: The version number should be 0.7.4 now.

skbio/sequence/transition.py
```diff
@@ -0,0 +1,696 @@
...
+        String that holds type of the sequence. "DNA" (default) and "RNA"
+        are valid options. Needed to assign proper nucleotide letters to
+        SubstitutionMatrix dimensions.
+
+    Returns
+    -------
+    SubstitutionMatrix
+        Transition probability matrix. Rows are ancestral nucleotides,
+        columns are descendant nucleotides.
+
```
qiyunzhu, line 271: Redundant empty line here.

qiyunzhu (commented):
@keder Thanks for the efforts! Your code looks great already. There is no particularly major technical issue I can find. I have a few overarching thoughts to share:

1. I tend to rename the `tpm` submodule as `transition`. This is just a personal flavor thing, but I am trying to keep the language style consistent across the package. What do you think of the word "transition"? Will it cause confusions or interfere with any future functionalities you intend to have?

2. I thought about where should the `transition` submodule go? A technically intuitive approach is to make it a method under `SubstitutionMatrix`, since all tpm functions don't need a `Sequence` instance as input and they produce a `SubstitutionMatrix`. But then I think that this approach will make these nice functions too deep in the module tree, which isn't good for their exposure.

3. An intermediate approach may be to have a dedicated function `tpm` which takes a string parameter as the model name, and a few `kwargs` for that model. But in this way there is no place to write the nice equations.

4. So I finally think that having a `skbio.sequence.transition` module as-is is the best idea. This will introduce some namespace redundancy though (e.g., both `distance` and `transition` have `jc69`), but this isn't a big issue.

5. Since there are further features to add, we can live with the current design for now, and it can be subject to updates in the future.

keder:
> @keder Thanks for the efforts! Your code looks great already. There is no particularly major technical issue I can find. I have a few overarching thoughts to share:
> 
>     1. I tend to rename the `tpm` submodule as `transition`. This is just a personal flavor thing, but I am trying to keep the language style consistent across the package. What do you think of the word "transition"? Will it cause confusions or interfere with any future functionalities you intend to have?
> 
>     2. I thought about where should the `transition` submodule go? A technically intuitive approach is to make it a method under `SubstitutionMatrix`, since all tpm functions don't need a `Sequence` instance as input and they produce a `SubstitutionMatrix`. But then I think that this approach will make these nice functions too deep in the module tree, which isn't good for their exposure.
> 
>     3. An intermediate approach may be to have a dedicated function `tpm` which takes a string parameter as the model name, and a few `kwargs` for that model. But in this way there is no place to write the nice equations.
> 
>     4. So I finally think that having a `skbio.sequence.transition` module as-is is the best idea. This will introduce some namespace redundancy though (e.g., both `distance` and `transition` have `jc69`), but this isn't a big issue.
> 
>     5. Since there are further features to add, we can live with the current design for now, and it can be subject to updates in the future.

1. No problem, I will rename it.
2. I was not sure where it exactly should go either, those are also suitable places to put it. I will trust your judgment for this. I can move it to another location whenever necessary.
4. Unfortunately, each of those models (like JC96) are in fact a collection of models that computes different variables using same assumptions, so I think some degree of redundancy is expected. The only think we need to make sure is that users will not be confused by that.
5. Of course, I will make all necessary changes, if needed.

keder:
@qiyunzhu I have pushed the fixes.

qiyunzhu:
@keder Thank you for the patience! This looks good and I will merge. Then I will do some clean-up stuff from my end. Thanks for your contribution!

PR #2497, opened 2026-06-30
BUG: Fix TreeNode.copy of attributes referencing other nodes

### LINKED ISSUE #2084: TreeNode recursive copy of internal references
The copy mechanism of `TreeNode` has been improved in #2073 . However, there is still an issue: if an attribute of a node is a reference to another node, the copy operation will fall in infinite recursion. For example: `node1.partner = node2`. The previous code avoided this problem by pre-defining a list of cache attributes that should not be copied. However, this doesn't work if a user decides to create new attributes, for example ecological connections or horizontal gen transfer.

A possible solution is:
1. Assign node IDs throughout the tree.
2. Copy the tree. When the value of an attribute is a node with an ID, it will be replaced by the ID.
3. Traverse the copied tree and replace the IDs back to nodes, but this time the nodes are copied nodes in the new tree.

jissen706:
### Fixes Issue

Fixes #2084

### Summary

When a `TreeNode` carried a custom attribute that referenced another node in the same tree (e.g. `node1.partner = node2`), `TreeNode.copy` — and the `copy.copy` / `copy.deepcopy` protocols it backs — handled the reference incorrectly:

- **Deep or shallow copy of a non-cyclic reference** produced a *detached duplicate* of the referenced node rather than pointing at the corresponding node in the copied tree. The copy's internal reference topology was silently wrong.
- **A cyclic reference** (e.g. `node1.partner = node2; node2.partner = node1`) caused `RecursionError` / infinite recursion.

The previous implementation only avoided this for a hard-coded set of cache attributes, so any user-defined node-referencing attribute (ecological links, horizontal-gene-transfer edges, pairing relationships, etc.) hit the bug.

### Fix

`_copy` is now performed in two passes:

1. **Structure pass** — build the new tree (built-in attributes only) and record an `{id(old_node): new_node}` mapping. Custom attributes are deferred so that *every* node exists before any reference between nodes is resolved.
2. **Attribute pass** — copy custom attributes, redirecting intra-tree node references to the corresponding nodes in the copy. For deep copies this is done by seeding the `deepcopy` memo with the node mapping, which also transparently handles references nested inside containers (lists, dicts, …) and cycles. For shallow copies, direct node references are substituted via the mapping.

References to nodes *outside* the tree being copied retain the prior behavior.

### Tests

Added four focused tests in `skbio/tree/tests/test_tree.py` covering: deep-copy remapping, shallow-copy remapping, cyclic references (no infinite recursion), and node references nested inside a container. All existing tree tests continue to pass.

### Checklist

* [ ] **This pull request includes code, documentation, or other content derived from external source(s).**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
 
 * When using the Numba backend, Permanova is up to 8x faster [#2488](https://github.com/scikit-bio/scikit-bio/pull/2488).
 
+### Bug Fixes
+
+* Fixed `TreeNode.copy` (and `copy.copy`/`copy.deepcopy`) producing detached duplicate nodes, or recursing infinitely on cyclic references, when a node carried a custom attribute referencing another node in the same tree. Such references are now redirected to the corresponding nodes in the copy ([#2497](https://github.com/scikit-bio/scikit-bio/pull/2497)).
+
 ## Version 0.7.3
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,8 @@
 ### Performance enhancements
 
 * When using the Numba backend, Permanova is up to 8x faster [#2488](https://github.com/scikit-bio/scikit-bio/pull/2488).
+* Improved `TreeNode.copy` such that it can handle node cross-references correctly: If a node attribute refers to another node in the tree, the copied node attribute will be redirected to the corresponding node in the new tree ([#2497](https://github.com/scikit-bio/scikit-bio/pull/2497)).
 
-### Bug Fixes
-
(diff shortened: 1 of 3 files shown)


qiyunzhu:
@jissen706 This seems to be a very smart solution. When I was humanly engineering the `copy` function I wasn't able to think of it. I have quickly glanced your code and think it should work as anticipated. But I will need some more time to carefully think about it.

qiyunzhu:
I have revised this PR to: 1) Improved the performance of shallow copy, because the code before this commit maps all nodes regardless if there are node references or node, which regresses the performance. This commit fixes it. 2) Improved the documentation to make the behavior of the `copy` method explicit to the user.

PR #2498, opened 2026-07-05
ENH: add the balanced parentheses (BP) succinct tree backend

monkeytim19:
## Summary
Adds `BPTree`, a succinct balanced-parentheses representation for data with a tree structure, ported from improved-octo-waddle (https://github.com/biocore/improved-octo-waddle). Topology is stored in a compact bit array (~2 bits/node) for memory-efficient handling of very large trees, with fast topological queries (parent, children, siblings, LCA, depth/height, subtree size, pre/post-order, shear, collapse, …). Targets the `bp_tree` feature integration branch.

## What's included
- New `skbio.tree.bp` package: the `BPTree` Cython class, the vendored BitArray C library (`_bitarray/`), and supporting modules.
- `BPTree` is exposed via `skbio.tree`; method names/docstrings are aligned with `TreeNode` (numpydoc). Newick/jplace IO, TreeNode conversion, and jplace placement remain in the backend but are intentionally not part of the public API yet (importable only from`skbio.tree.bp`).
- Tests covering the full public `BPTree` API (49 tests), including the `.npz` read/write round-trip.
- `CHANGELOG.md` entry; BitArray license recorded under `licenses/`.

## Licensing
- Derived from improved-octo-waddle (https://github.com/biocore/improved-octo-waddle), originally by Daniel McDonald — Modified BSD (same as scikit-bio); attributed in each ported file's header.
- Vendors the BitArray C library (https://github.com/noporpoise/BitArray/, Isaac Turner) — Public Domain; recorded in `licenses/bitarray.txt`.

## Deferred
- **scikit-bio IO registry integration.** `read`/`write` and the Newick/jplace parsing are currently standalone. To align with the rest of scikit-bio they should move behind the `skbio.io` registry

- **Dunder-method consistency with `TreeNode`.** BPTree's dunder methods (`__repr__`, `__eq__`, `__len__`, `__contains__`, etc.) should be harmonized with `TreeNode`'s so the two tree types present a consistent interface.

- **Further Cython structure / optimization.** Modifying `BPTree` into a public-facing Python class over an under-the-hood Cython core
 

---

* [x] **This pull request includes code, documentation, or other content derived from external source(s).** BitArray (Public Domain; `licenses/bitarray.txt` + in-source attribution) and improved-octo-waddle (Modified BSD; attributed in each ported file's header).


skbio/tree/bp/_bp_insert.pyx
```diff
@@ -0,0 +1,129 @@
...
+
+    if insert_type == 'fully_resolved':
+        placements = placements.assign(
+            parent=placements.apply(_preallocate_empty, axis=1))
+
+    return placements, sktree, node_lookup
+
+
+# pd.DataFrame is not a resolved type so we cannot use it here for cython
+def insert_fully_resolved(object placements, BPTree bptree):
```
mortonjt: I think this is more of an API question - its not clear to me if we want to have a skbio.TreeNode returned, or a BPtree. My gut feeling is that it would be better to return a BPtree, and have conversion methods downstream handle BPtree -> TreeNode

skbio/tree/bp/_bp_conv.pyx
```diff
@@ -0,0 +1,153 @@
...
+# ----------------------------------------------------------------------------
+
+import skbio
+import numpy as np
+cimport numpy as cnp
+
+from ._bp cimport BPTree
+
+
+def to_skbio_treenode(BPTree bp):
```
mortonjt: I think this is a good opportunity to lock in the naming scheme.

Perhaps `to_treenode` is better here (unless we want to be more creative with the io-registry). Thoughts @qiyunzhu ?

skbio/tree/bp/_bp_conv.pyx
```diff
@@ -0,0 +1,153 @@
...
+            parent = bp.preorder_rank(bp.parent(node_idx)) - 1
+            # uncache=False avoids repeated cache clearing during bulk construction
+            nodes[parent].append(nodes[i], uncache=False)
+
+    root.length = None
+
+    return root
+
+
+def from_skbio_treenode(tree):
```
mortonjt: Same here, I think we now have better naming. I think `from_treenode` is better here.

skbio/tree/bp/_bp_conv.pyx
```diff
@@ -0,0 +1,153 @@
...
+            if n.is_tip():
+                ptr += 1
+
+            seen.add(n)
+
+        ptr += 1
+    return BPTree(topo, names=names, lengths=lengths, edges=edges)
+
+
+def to_skbio_treearray(BPTree bp):
```
mortonjt: Thoughts on just `to_array`?

mortonjt: Does it warrant a `from_array` method as well? Or should we defer to later?

skbio/tree/bp/_bp_io.pyx
```diff
@@ -0,0 +1,443 @@
...
+        edge = np.int32(token.strip()[1:end - 1])
+    else:
+        name = token.replace("'", "").replace('"', "").strip()
+
+    names[ptr] = name
+    lengths[ptr] = length
+    edges[ptr] = edge
+
+
+def write_newick(BPTree tree, object output, bint include_edge):
```
mortonjt, line 95: This is a good start, but this isn't quite what we meant regarding the file registry.
See here https://github.com/scikit-bio/scikit-bio/blob/main/skbio/io/format/newick.py for examples on how the existing newick is formatted. We will want to use the same io registry functionality for all read/write newick methods here (with the option to overload TreeNode with BPTree).

Same with the jplace format, but I'm OK with deferring the jplace to a separate PR.

skbio/tree/bp/_bp_io.pyx
```diff
@@ -0,0 +1,443 @@
...
+            if tree.next_sibling(open_paren_stack.pop()) == 0:
+                if idx != root_close:
+                    output.write(')')
+            else:
+                output.write(',')
+
+    output.write(';')
+
+
+cpdef parse_newick(unicode data):
```
mortonjt, line 168: Same here, we will want to leverage the io registry.

skbio/tree/bp/_bp_io.pyx
```diff
@@ -0,0 +1,443 @@
...
+        if _ccheck(c):
+            if idx == start:
+                return idx + 1
+            else:
+                return idx
+
+    return idx + 1
+
+
+def parse_jplace(object data):
```
mortonjt, line 366: Best to register jplace as its own format (this should be explicitly defined as a new file jplace.py under `skbio/io/format/`)

mortonjt: Probably should also define a jplace writer method as well? OK with deferring to a separate PR.

skbio/tree/bp/_bp_io.pyx
mortonjt: Specifically for newick, we will want to expand upon the `skbio/io/format/newick.py` file to append BPTree parsers and write functionality, so that newick can return both `TreeNode` and `BPTree` objects

mortonjt (commented):
I think this is off to a good start -- the docstrings could use some more work. 

More importantly, the io registry remains outstanding. Given that this already a huge PR, I'm OK with deferring, but we will want to get all of the io registry functionality in place before the next release.

skbio/tree/bp/_bitarray/bit_array.c
```diff
@@ -0,0 +1,3181 @@
+/*
+ bit_array.c
+ project: bit array C library
+ url: https://github.com/noporpoise/BitArray/
+ maintainer: Isaac Turner <turner.isaac@gmail.com>
+ license: Public Domain, no warranty
+ date: Aug 2014
```
wasade, line 7: I'm assuming this is the same commit as the `iow` project, are there any meaningful changes to this project that should be accounted for such as important bugs?

skbio/tree/bp/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+# ----------------------------------------------------------------------------
+
+from ._bp import BPTree
+from ._bp_io import parse_newick, write_newick, parse_jplace
+from ._bp_conv import to_skbio_treenode, from_skbio_treenode, to_skbio_treearray
+from ._bp_insert import insert_fully_resolved
+
+__all__ = [
+    "BPTree",
+    "parse_newick",
```
wasade, line 25: We already have a newick parser and writer in scikit-bio. I worry about having multiple parsers as there is potential for drift between them leading to subtle inconsistencies.

skbio/tree/bp/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+    "write_newick",
+    "parse_jplace",
+    "to_skbio_treenode",
```
wasade: The to/from methods are likely better within scikit-bio as class methods off the respective objects

skbio/tree/bp/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+from ._bp import BPTree
+from ._bp_io import parse_newick, write_newick, parse_jplace
+from ._bp_conv import to_skbio_treenode, from_skbio_treenode, to_skbio_treearray
+from ._bp_insert import insert_fully_resolved
+
+__all__ = [
+    "BPTree",
+    "parse_newick",
+    "write_newick",
+    "parse_jplace",
```
wasade, line 27: Can jplace parsing go into the `io` module of skbio?

skbio/tree/bp/_bp.pyx
```diff
@@ -0,0 +1,1365 @@
...
+            self._edge_lookup = None
+
+        # precursor for select index cache
+        _r_index_0 = np.cumsum((1 - B), dtype=SIZE)
+        _r_index_1 = np.cumsum(B, dtype=SIZE)
+
+        # construct a select index. These operations are performed frequently,
+        # and easy to cache at a relatively minor memory expense. It cannot be
+        # assumed that open and close will be same length so can't stack
+        _k_index_0 = np.unique(_r_index_0,
```
wasade, line 234: I was really hoping that here and with the `_e_index` that these could be derived at runtime rather than cached consistent with the original publication and the `TODO` statements. Has that been investigated?

skbio/tree/bp/_bp.pyx
```diff
@@ -0,0 +1,1365 @@
...
+        self._k_index_1 = _k_index_1
+
+        # construct an excess index. These operations are performed a lot, and
+        # similarly can to rank and select, can be cached at a minimal expense.
+        _e_index = np.empty(B.size, dtype=SIZE)
+        for i in range(B.size):
+            _e_index[i] = self._excess(i)
+        self._e_index = _e_index
+
+    def write(self, object fname):
```
wasade, line 248: `read` and `write` I think would normally be decorated by the `io` registry?

skbio/tree/bp/_bp.pyx
```diff
@@ -0,0 +1,1365 @@
...
+        max_v = self.excess(i)  # a value larger than what will be tested
+        for k in range(i, j + 1):
+            obs_v = self.excess(k)
+            if obs_v > max_v:
+                max_k = k
+                max_v = obs_v
+
+        return max_k
+
+    def __len__(self):
```
wasade, line 583: It would be beneficial for the dunder methods to be reasonably consistent with `TreeNode`

skbio/tree/bp/_bp.pyx
```diff
@@ -0,0 +1,1365 @@
...
+            pos = self.previous_sibling(self.open(i))
+
+        if pos < 0:
+            return 0
+        elif self._b_ptr[pos]:
+            return pos
+        else:
+            return 0
+
+    cpdef SIZE_t preorder_rank(self, SIZE_t i) nogil:
```
wasade, line 823: There likely is an opportunity to harmonize the namespace, perhaps not the function signatures, but at least the namespace with TreeNode.

skbio/tree/bp/_bp_conv.pyx
wasade: These probably should be class methods on the respective objects though

skbio/tree/bp/_bp_insert.pyx
wasade: Remember that BPTree is immutable in its topology, and this is a mutable operation. skbio as far as I know does not have support for placement resolution. It doesn't have to be off of BPTree or even in Cython, the code here should likely be moved. Note there is also support for multifurcation, which was what Siavash had recommended for GG2. However, the derivation of the algorithm was off of GPL code so that resolution method lives in the `iow-gpl` project 

All that said, placement resolution is _much faster_ with duckdb-miint. Adopting it though changes the underlying model of a phylogeny considerably.

skbio/tree/bp/_bp_io.pyx
wasade: Agree

wasade (changes requested):

skbio/tree/bp/_bitarray/bit_array.c
monkeytim19: Not that I am aware of — the `BitArray` commit being used the submodule to `iow` being ported right now should be the latest version.

doc/source/_templates/BP.rst
```diff
@@ -0,0 +1,61 @@
...
+
+   .. rubric:: Tree analysis
+
+   .. autosummary::
+      :toctree:
+
+      ~{{ name }}.is_tip
+      ~{{ name }}.depth
+      ~{{ name }}.height
+      ~{{ name }}.subtree_size
```
qiyunzhu: `subtree_size` and `ntips` are equivalent to the [`count`](https://scikit.bio/docs/latest/generated/skbio.tree.TreeNode.count.html) method of `TreeNode`. I would suggest making the API consistent.

skbio/tree/bp/__init__.py
qiyunzhu: Good idea. We should explicitly add jplace to `io` as a new format. Your code in #2452 and the current PR can follow that. Note that I think jplace should be registered a standalone format, rather than part of Newick, but the underlying parser can be combined if that works well.

skbio/tree/bp/__init__.py
qiyunzhu: Agreed.

skbio/tree/tests/bp/data/200/placement.jplace
```diff
@@ -0,0 +1,156 @@
+{
```
qiyunzhu, line 1: These test datasets are good for now. But later (after adding `jplace` format) we should refine them.

skbio/tree/bp/_bp_conv.pyx
qiyunzhu: Agreed. Since both data structures live in skbio, there is no need to prefix with `skbio`.

skbio/tree/bp/_bp_conv.pyx
qiyunzhu: Agreed with @wasade . Also, I would suggest adding `from_bptree` to `TreeNode`, rather than adding `to_treenode` to `BPTree`. This is because the latter forces the output class to be `TreeNode`, whereas the former also support subclasses. Also because `to_xx` assumes the existence and consistence of each other, which makes maintenance less convenient.

skbio/tree/bp/_bp_conv.pyx
```diff
@@ -0,0 +1,153 @@
+# ----------------------------------------------------------------------------
```
qiyunzhu: I think there are several design pattern issues with these Cython files (*.pyx).

1. Some variables are not defined before use in the Cython functions. For example, in line 77 you have `for n in tree.pre_and_postorder(include_self=True):`. This will leave the variable `n` in the Python namespace, and slows down the execution of the Cython function. The correct approach is to `cdef Py_ssize_t n` before the loop. See [here](https://github.com/scikit-bio/scikit-bio/blob/main/skbio/alignment/_cutils.pyx#L99) for an example. Note that `Py_ssize_t` is preferred over `int` because the former is platform-independent (equivalent to `np.intp`). Therefore, your line 32 can also be updated to adapt this.

2. Cython files should only house compute-intense algorithms that runs *under-the-hood*, whereas the interface should be kept in Python files. Additionally, allocation of arrays can live in Python whereas calculations (particularly loops) can go the Cython. For example, the method `to_skbio_treenode` should stay in Python, while `for i in range(bp.data.sum()):...` can live in Cython.

3. I doubt how much performance gain can be obtained by Cython in these functions. They all involve lots of Python interface like set and attribute.

skbio/tree/bp/_bp_io.pyx
```diff
@@ -0,0 +1,443 @@
+# encoding: utf-8
+# cython: profile=False, boundscheck=False, wraparound=False
```
qiyunzhu: See [here](https://github.com/scikit-bio/scikit-bio/blob/25107a5dae11f1d2c8514bceb77e4b224931d7b2/skbio/alignment/_cutils.pyx#L9) for even better performance if applicable.

skbio/tree/bp/_bp.pyx
```diff
@@ -0,0 +1,1365 @@
...
+                else:
+                    self.mM[node, self.m_idx] = min(self.mM[lchild, self.m_idx],
+                                                    self.mM[rchild, self.m_idx])
+                    self.mM[node, self.M_idx] = max(self.mM[lchild, self.M_idx],
+                                                    self.mM[rchild, self.M_idx])
+
+                self.r[node] = self.r[lchild]
+
+
+@cython.final
```
qiyunzhu, line 147: I am not sure why `@cython.final` is necessary here.

skbio/tree/bp/_ba.pyx
```diff
@@ -0,0 +1,81 @@
...
+# originally authored by Daniel McDonald, distributed under the Modified BSD License.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+cimport libc.stdlib
+from cpython cimport Py_INCREF, Py_DECREF, bool
+cimport numpy as cnp
```
qiyunzhu: This pyx file doesn't create NumPy arrays. So this import is unnecessary.

skbio/tree/bp/_bp.pyx
```diff
@@ -0,0 +1,1365 @@
...
+cdef class BPTree:
```
qiyunzhu, line 148: I understand that declaring a class directly in Cython has speed advantages due to [early binding](https://cython.readthedocs.io/en/latest/src/userguide/early_binding_for_speed.html). But I am also concerned that it will complicate debugging. I think a better design pattern is to have a public-facing class in Python and an under-the-hood class in Cython. The latter only provide methods that are used elsewhere in Cython for speed. The former has a full set of methods.

wasade:
On jplace, I can't seem to reply to the comment -- a single parser does not work. Jplace has a field that contains newick, but jplace itself is not newick. #2452 adds support for edge IDs, which is valid "newick" depending on the exact grammar you accept and not specific to jplace.

monkeytim19:
### Added Changes
- **Conversions** reworked into class constructors — `TreeNode.from_bptree` / `BPTree.from_treenode` (subclass-preserving), dropping the standalone functions and the `to_*` mirror.
- **`count(i=0, tips=False)`** replaces `subtree_size`/`ntips`, mirroring `TreeNode.count`.
- **Placement/insertion (jplace)** deferred to a dedicated follow-up.
- **Robustness/correctness:** unbalanced topology now raises `ValueError` (was `assert`); fixed BitArray extern decls (`uint64_t` width, `bit_array_not` signature).
- **Cython hygiene:** standardized file headers/directives, tidied imports, removed stray trailing commas.
- **Docs/tests:** documented `read()`'s `allow_pickle` risk; Cython-level tests now run under the test runner instead of at import.

skbio/tree/bp/_bitarray/bit_macros.h
```diff
@@ -0,0 +1,205 @@
...
+
+// trailing_zeros is number of least significant zeros
+// leading_zeros is number of most significant zeros
+#if defined(_WIN32)
+  #define trailing_zeros(x) ({ __typeof(x) _r; _BitScanReverse64(&_r, x); _r; })
+  #define leading_zeros(x) ({ __typeof(x) _r; _BitScanForward64(&_r, x); _r; })
+#else
+  #define trailing_zeros(x) ((x) ? (__typeof(x))__builtin_ctzll(x) : (__typeof(x))sizeof(x)*8)
+  #define leading_zeros(x) ((x) ? (__typeof(x))__builtin_clzll(x) : (__typeof(x))sizeof(x)*8)
+#endif
```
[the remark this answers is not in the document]
qiyunzhu: Note: scikit-bio does not support 32-bit operating systems.

qiyunzhu (approved):
@monkeytim19 Thanks for the update! I took a quick look and didn't notice any major issue. How about we just merge and move on?

monkeytim19:
Happy to merge it for now and proceed from there. @mortonjt @wasade any further comments?

skbio/tree/bp/_bp.pyx
```diff
@@ -284,6 +284,114 @@ cdef class BPTree:
...
+        Returns
+        -------
+        BPTree
+            The tree represented in balanced-parentheses form.
+
+        See Also
+        --------
+        skbio.tree.TreeNode.from_bptree
+
+        """
```
wasade, line 316: Could the variables be typed? It's one of the primary performance gains

skbio/tree/bp/_bp.pyx
```diff
@@ -284,6 +284,114 @@ cdef class BPTree:
...
+        -------
+        dict
+            Dictionary with keys ``'child_index'``, ``'length'``,
+            ``'id_index'`` and ``'name'``.
+
+        See Also
+        --------
+        skbio.tree.TreeNode.to_array
+
+        """
```
wasade, line 364: Same here, there are many untyped variables

monkeytim19:
Addressed @wasade's feedback on untyped variables in the conversion helpers (d6f2c9ef) with variables typed for `BPTree.from_treenode` and `BPTree.to_array`.

qiyunzhu:
Thank! @monkeytim19 Let's merge.

PR #2499, opened 2026-07-06
ENH: support array-API (GPU-resident) buffers in DistanceMatrix

LarytheLord:
This is a first step toward letting the distance functions operate on GPU-resident distance matrices. It allows `DistanceMatrix` to hold a non-NumPy, array-API-compatible buffer (e.g. a cupy/torch/jax array) instead of coercing everything to NumPy at construction.

### What changed
- `DistanceMatrix` (and the `SymmetricMatrix` machinery it inherits) now preserves a non-NumPy array-API buffer at construction instead of forcing `np.asarray`.
- Validation (symmetry, hollowness, dtype), `permute`, `copy` and `filter` dispatch to an array-API (`xp`) path when the buffer is non-NumPy.
- The NumPy path (Cython kernels and existing validation) is untouched, so behavior for NumPy inputs is unchanged.
- Adds array-API tests for `DistanceMatrix`.

### Approach and scope
Dispatch keys off whether `.data` is a NumPy array, following the existing `ingest_array`/`xp` pattern. This is deliberately a focused first step on the class itself:
- IO and plotting on a non-NumPy buffer are left for a follow-up.
- `PairwiseMatrix` is kept NumPy-only; `DistanceMatrix` and `SymmetricMatrix` are the array-API-capable ones.
- The duplicated `_normalize_input` across the classes is kept as-is rather than refactored.

### Testing
- The array-API tests are parametrized over numpy/jax/torch/cupy (picked up by the array-api CI). I verified them locally on NumPy and PyTorch (CPU).
- The `DistanceMatrix` array-API tests were also run on a real AMD Instinct MI210 via torch-ROCm and pass: construction, validation, `permute`, `copy` and `filter` all behave correctly with a GPU-resident buffer.
- Existing distance / ordination / diversity suites remain green.

### Follow-ups
- `permanova`/`mantel` array-API support, consuming a GPU-backed `DistanceMatrix`.
- Docstring and changelog updates.

cc @sfiligoi

### DIFF
--- a/skbio/stats/distance/_base.py
+++ b/skbio/stats/distance/_base.py
@@ -26,7 +26,7 @@ from skbio.io.descriptors import Read, Write
 
 from ._utils import is_symmetric_and_hollow, is_symmetric
 from ._utils import distmat_reorder, distmat_reorder_condensed
-from skbio.util._array import ingest_array
+from skbio.util._array import ingest_array, copy_array
 
 import array_api_compat as _aac
 
@@ -537,16 +537,7 @@ class PairwiseMatrix(SkbioObject, PlottableMixin):
             self._validate_ids(filtered_data, ids)
             return self.__class__(filtered_data, ids, validate=False, condensed=True)
         else:
-            if _aac.is_numpy_array(self._data):
-                filtered_data = distmat_reorder(self.redundant_form(), idxs)
-            else:
-                # non-NumPy (e.g. GPU) buffer: reorder rows/cols via an xp gather
-                xp = _aac.array_namespace(self._data)
-                idx = xp.asarray(
-                    np.asarray(idxs, dtype=int),
-                    device=getattr(self._data, "device", None),
-                )
-                filtered_data = self._data[idx][:, idx]
+            filtered_data = distmat_reorder(self.redundant_form(), idxs)
             self._validate_ids(filtered_data, ids)
             return self.__class__(filtered_data, ids, validate=False)
 
@@ -1201,10 +1192,14 @@ class SymmetricMatrix(PairwiseMatrix):
         # array or (b) the data are not a single or double precision numpy
         # data type.
         if _aac.is_array_api_obj(data) and not _aac.is_numpy_array(data):
-            # Preserve a non-NumPy (e.g. GPU-resident) array-API buffer as-is so
-            # the DistanceMatrix can hold GPU data. NumPy / list / tuple inputs
-            # are still normalized to a NumPy array below.
+            # Preserve a non-NumPy (e.g. GPU-resident) array-API buffer so the
+            # DistanceMatrix can hold GPU data. NumPy / list / tuple inputs are
+            # still normalized to a NumPy array below. A non-floating buffer is
+            # cast to float64, mirroring the NumPy path (which keeps float32/64
+            # and casts everything else to float).
             _xp, data = ingest_array(data)
+            if not _xp.isdtype(data.dtype, (_xp.float32, _xp.float64)):
+                data = _xp.astype(data, _xp.float64)
         else:
             _issue_copy = True
             if isinstance(data, np.ndarray):
@@ -1299,6 +1294,18 @@ class SymmetricMatrix(PairwiseMatrix):
             1-D or 2-D array containing the values of the matrix.
 
         """
+        if _aac.is_array_api_obj(data) and not _aac.is_numpy_array(data):
+            # Array-API (e.g. GPU-resident) buffers are stored in redundant (2-D)
+            # form only. Condensed storage relies on scipy.squareform, which is
+            # NumPy-only, so a condensed or 1-D non-NumPy buffer could not be
+            # reordered/converted without a silent copy back to the host.
+            if condensed or data.ndim != 2:
+                raise SymmetricMatrixError(
+                    "Condensed form is not supported for non-NumPy (array-API) "
+                    "arrays. Provide a 2-D array to store the matrix on a "
+                    "non-NumPy device."
+                )
+            return data
         if condensed:
             # case where input is 1d and stays 1d
             if data.ndim == 1:
@@ -1327,15 +1334,6 @@ class SymmetricMatrix(PairwiseMatrix):
 
         """
         if data.ndim == 1:
-            return
-        if _aac.is_array_api_obj(data) and not _aac.is_numpy_array(data):
-            # xp-aware symmetry check for a non-NumPy buffer (a NaN makes the
-            # inequality True, so this also rejects NaNs, matching is_symmetric).
-            xp = _aac.array_namespace(data)
-            if bool(xp.any(xp.matrix_transpose(data) != data)):
-                raise DistanceMatrixError(
-                    "Data must be symmetric and cannot contain NaNs."
-                )
             return
         if not is_symmetric(data):
             raise DistanceMatrixError("Data must be symmetric and cannot contain NaNs.")
@@ -1411,13 +1409,11 @@ class SymmetricMatrix(PairwiseMatrix):
                 f"Found {data.ndim} dimensions."
             )
 
-        if _aac.is_array_api_obj(data) and not _aac.is_numpy_array(data):
-            # array-API (e.g. GPU) buffer: check the float dtype via the namespace
-            if not _aac.array_namespace(data).isdtype(data.dtype, "real floating"):
-                raise PairwiseMatrixError(
-                    "Data must contain only floating point values."
-                )
-        elif data.dtype not in (np.float32, np.float64):
+        # isdtype works for NumPy and non-NumPy (e.g. GPU) buffers alike, via each
+        # array's own namespace. Restrict to single/double precision (as the legacy
+        # NumPy check did) so both paths accept exactly the same dtypes.
+        xp = _aac.array_namespace(data)
+        if not xp.isdtype(data.dtype, (xp.float32, xp.float64)):
             raise PairwiseMatrixError("Data must contain only floating point values.")
 
     @property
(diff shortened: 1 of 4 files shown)


skbio/stats/distance/_base.py
```diff
@@ -1188,18 +1200,24 @@ def _normalize_input(self, data, ids, diagonal):
...
-
-        # Make data_ explicitly an ndarray to help with type checking.
-        # At this point in the code we can be certain that data is an
-        # ndarray.
-        assert isinstance(data, np.ndarray)
+        if _aac.is_array_api_obj(data) and not _aac.is_numpy_array(data):
+            # Preserve a non-NumPy (e.g. GPU-resident) array-API buffer as-is so
+            # the DistanceMatrix can hold GPU data. NumPy / list / tuple inputs
+            # are still normalized to a NumPy array below.
+            _xp, data = ingest_array(data)
```
sfiligoi, line 1200: I don't think this is correct.
If a non-numpy array is not float, it should be converted to float.
But you don't do that here.

skbio/stats/distance/_base.py
```diff
@@ -1308,7 +1326,18 @@ def _validate_data(self, data: NDArray) -> None:
         be symmetric.
 
         """
-        if (data.ndim != 1) and (not is_symmetric(data)):
+        if data.ndim == 1:
+            return
+        if _aac.is_array_api_obj(data) and not _aac.is_numpy_array(data):
```
sfiligoi: I think it would be better to update `is_symmetric` to accept non-numpy data.

skbio/stats/distance/_base.py
```diff
@@ -1382,7 +1411,13 @@ def _validate_shape(self, data: NDArray):
                 f"Found {data.ndim} dimensions."
             )
 
-        if data.dtype not in (np.float32, np.float64):
+        if _aac.is_array_api_obj(data) and not _aac.is_numpy_array(data):
+            # array-API (e.g. GPU) buffer: check the float dtype via the namespace
+            if not _aac.array_namespace(data).isdtype(data.dtype, "real floating"):
```
sfiligoi: Does't array_namespace handle `numpy` data, too?
What is the downside of always using the `isdtype`, also for (default) numpy?

skbio/stats/distance/_base.py
```diff
@@ -1751,6 +1786,20 @@ def permute(
         rng = get_rng(seed)
         order = rng.permutation(self.shape[0])
 
+        if not _aac.is_numpy_array(self._data):
```
sfiligoi: Wouldn't it be better to put this login in `distmat_reorder_condensed` (and `distmat_reorder_condensed_py`) instead?

If it really cannot be fit into existing functions, create a similar "aac" function in those same files, and call the new function.

skbio/stats/distance/_base.py
```diff
@@ -1802,7 +1851,11 @@ def _copy(
 
         """
         # adding for backward compatibility
-        data = self._data.copy()
+        if _aac.is_numpy_array(self._data):
+            data = self._data.copy()
+        else:
+            # non-NumPy (e.g. GPU) buffer: copy via the array-API namespace
+            data = _aac.array_namespace(self._data).asarray(self._data, copy=True)
```
sfiligoi: I think this is used often enough that we should have a helper function for it, similar to `ingest_array`.

I would suggest you create a `copy_array` in `utility/_array.py`, then call that function here.

skbio/stats/distance/_base.py
```diff
@@ -2026,18 +2079,24 @@ def _normalize_input(self, data, ids):
...
-
-        # Make data_ explicitly an ndarray to help with type checking.
-        # At this point in the code we can be certain that data is an
-        # ndarray.
-        assert isinstance(data, np.ndarray)
+        if _aac.is_array_api_obj(data) and not _aac.is_numpy_array(data):
+            # Preserve a non-NumPy (e.g. GPU-resident) array-API buffer as-is so
+            # the DistanceMatrix can hold GPU data. NumPy / list / tuple inputs
+            # are still normalized to a NumPy array below.
+            _xp, data = ingest_array(data)
```
sfiligoi, line 2064: As above.

skbio/stats/distance/_base.py
```diff
@@ -2048,16 +2107,24 @@ def _validate_data(self, data: NDArray) -> None:
 
         """
         # if the input data is 1D, we don't need to check for hollowness or symmetry
-        if data.ndim == 2:
+        if data.ndim != 2:
+            return
+        if _aac.is_array_api_obj(data) and not _aac.is_numpy_array(data):
```
sfiligoi: I think it would be better to make ` is_symmetric_and_hollow` array API friendly, instead of implementing inline.

skbio/stats/distance/_base.py
```diff
@@ -2075,7 +2142,11 @@ def _copy(self, transpose: bool = False, condensed: bool = False) -> DistanceMat
 
         """
         # adding for backward compatibility
-        data = self._data.copy()
+        if _aac.is_numpy_array(self._data):
+            data = self._data.copy()
+        else:
+            # non-NumPy (e.g. GPU) buffer: copy via the array-API namespace
+            data = _aac.array_namespace(self._data).asarray(self._data, copy=True)
```
sfiligoi: As above.

skbio/stats/distance/_base.py
```diff
@@ -534,7 +537,16 @@ def filter(
             self._validate_ids(filtered_data, ids)
             return self.__class__(filtered_data, ids, validate=False, condensed=True)
         else:
-            filtered_data = distmat_reorder(self.redundant_form(), idxs)
+            if _aac.is_numpy_array(self._data):
+                filtered_data = distmat_reorder(self.redundant_form(), idxs)
+            else:
+                # non-NumPy (e.g. GPU) buffer: reorder rows/cols via an xp gather
+                xp = _aac.array_namespace(self._data)
```
sfiligoi: I think it would be better to make `distmat_reorder` capable of accepting non-numpy data.

LarytheLord:
Pushed a commit addressing the review comments. The symmetry/hollowness, reordering and copy handling for non-NumPy (array-API) buffers now lives in the helpers instead of inline in the class methods:

- `is_symmetric_and_hollow`, `distmat_reorder` and `distmat_reorder_condensed` handle a non-NumPy buffer directly, so `_validate_data`, `permute` and `filter` just call them.
- Added a `copy_array` helper in `skbio.util._array` (next to `ingest_array`) for the device-preserving copy used by `_copy`.
- `_validate_shape` is now a single `isdtype` check for both paths.

Two fixes to keep the non-NumPy path consistent with NumPy:
- A non-floating buffer is now cast to float64 (float32/float64 are kept), matching the NumPy path — previously the non-NumPy path skipped this cast, so its dtype handling didn't line up with NumPy (e.g. an integer buffer, which NumPy casts and accepts).
- Non-NumPy buffers are stored in redundant (2-D) form only; a condensed or 1-D non-NumPy buffer is now rejected rather than silently copied back to the host through scipy.squareform, which is NumPy-only.

Tested on NumPy and PyTorch, and on a AMD MI210 (ROCm): the DistanceMatrix array-API tests pass on-device, and the buffer stays on the GPU through construction, permute, filter and copy.

skbio/stats/distance/_utils.py
```diff
@@ -31,11 +32,23 @@ def is_symmetric_and_hollow(mat):
...
     See the :install:`parallelization guide <#parallelization>` for information on
     controlling the number of threads used.
 
     """
+    if _aac.is_array_api_obj(mat) and not _aac.is_numpy_array(mat):
+        # Non-NumPy (e.g. GPU-resident) buffer: check symmetry and hollowness via
+        # the array-API namespace. A NaN makes the inequality True, so this also
+        # rejects NaNs, matching the Cython path.
+        xp = _aac.array_namespace(mat)
+        is_symmetric = not bool(xp.any(xp.matrix_transpose(mat) != mat))
```
sfiligoi, line 48: This is OK in this PR.
But should be improved in the future, using a similar approach as the one Cython uses, as the current naive approach can be quite slow for large matrices.

skbio/stats/distance/_base.py
```diff
@@ -2048,16 +2089,15 @@ def _validate_data(self, data: NDArray) -> None:
...
-            data_sym, data_hol = is_symmetric_and_hollow(data)
-            if not data_sym:
-                raise DistanceMatrixError(
-                    "Data must be symmetric and cannot contain NaNs."
-                )
-            if not data_hol:
-                raise DistanceMatrixError(
-                    "Data must  be hollow (i.e., the diagonal can only contain zeros)."
-                )
+        if data.ndim != 2:
```
sfiligoi: Any reason why we cannot preserve the original logic in here?
Looks only a cosmetic change to me.

skbio/stats/distance/_base.py
```diff
@@ -1281,6 +1294,18 @@ def _init_data(self, data: NDArray, condensed: bool = False) -> NDArray:
             1-D or 2-D array containing the values of the matrix.
 
         """
+        if _aac.is_array_api_obj(data) and not _aac.is_numpy_array(data):
+            # Array-API (e.g. GPU-resident) buffers are stored in redundant (2-D)
+            # form only. Condensed storage relies on scipy.squareform, which is
+            # NumPy-only, so a condensed or 1-D non-NumPy buffer could not be
+            # reordered/converted without a silent copy back to the host.
+            if condensed or data.ndim != 2:
```
sfiligoi, line 1302: This is OK for this PR, but should be "fixed" in a follow-up PR.
We do want to support condensed form also on the GPU.

skbio/stats/distance/_base.py
```diff
@@ -1308,7 +1333,9 @@ def _validate_data(self, data: NDArray) -> None:
         be symmetric.
 
         """
-        if (data.ndim != 1) and (not is_symmetric(data)):
+        if data.ndim == 1:
```
sfiligoi: This is just a cosmetic change, right?
Please keep the original, unless there is a semantic change involved that I don't see.

LarytheLord:
Reverted the two cosmetic changes `_validate_data` in both `SymmetricMatrix` and `DistanceMatrix` is back to its original control flow. The array-API handling lives in the `is_symmetric` / `is_symmetric_and_hollow` helpers now, so the original form works unchanged.

sfiligoi (approved):
Looks good to me.

PR #2503, opened 2026-07-11
ENH: array-API (GPU-resident) support for permanova and mantel

LarytheLord:
### Description
Follow-up to #2499. Adds first-pass array-API / GPU support to `permanova` and `mantel`, dispatching on where the `DistanceMatrix` data lives: a NumPy-backed `DistanceMatrix` keeps the existing cython/numba `engine=` path unchanged, while a non-NumPy (GPU-resident cupy/torch/jax) buffer is computed through the array-API (`xp`) namespace so the matrix stays on its device.

`mantel` covers pearson and spearman on the xp path; kendalltau has no xp path and is materialized on the host (SciPy's kendalltau is host-only). IDs are aligned exactly like the NumPy path (reorder / `strict` / `lookup`) via `_order_dms`, which became array-API aware in #2499. Both functions keep DistanceMatrix-only semantics (no raw-array input). The permutation p-values match the NumPy path to floating-point precision, using the same RNG order.

### Scope / follow-ups
- First-pass GPU support (roadmap step 2). The performant numba-GPU (`@cuda.jit` / `@hip.jit`) kernel backend and explicit `engine=` GPU selection are planned as later performance work (step 4).
- Permutations rebuild the index array on the host each iteration, and mantel spearman does a one-time host assembly of the ranked matrix; keeping these fully on-device is a performance follow-up.

### Testing
- array-API tests added for both functions, run on numpy and PyTorch (including id-reordering and spearman tie handling).
- Verified on a real AMD Instinct MI210 (torch-ROCm): the xp path runs on-device and matches the numpy result to floating-point precision (permanova pseudo-F, mantel pearson/spearman r and p-values), including id-reordering happening on the GPU buffer.
- Existing permanova/mantel test suites pass unchanged.

cc @sfiligoi @qiyunzhu

### DIFF
--- a/skbio/stats/distance/_mantel.py
+++ b/skbio/stats/distance/_mantel.py
@@ -20,7 +20,7 @@ from scipy.stats import kendalltau, ConstantInputWarning, NearConstantInputWarni
 from ._cutils import mantel_perm_pearsonr_cy, mantel_perm_pearsonr_condensed_cy
 from skbio.stats.distance import DistanceMatrix
 from skbio.util import get_rng
-from skbio.util._array import ingest_array, _to_numpy
+from skbio.util._array import ingest_array, _get_array
 from skbio._config import _resolve_engine
 
 import array_api_compat as _aac
@@ -467,9 +467,9 @@ def mantel(
     # (the fallback above). A non-NumPy DistanceMatrix is materialized on the host
     # here (SciPy's kendalltau is host-only, and _order_dms runs on host).
     if x_is_xp:
-        x = DistanceMatrix(_to_numpy(x.data), x.ids)
+        x = DistanceMatrix(_get_array(x.data, to_numpy=True), x.ids)
     if y_is_xp:
-        y = DistanceMatrix(_to_numpy(y.data), y.ids)
+        y = DistanceMatrix(_get_array(y.data, to_numpy=True), y.ids)
 
     x, y = _order_dms(x, y, strict=strict, lookup=lookup)
 
@@ -581,8 +581,8 @@ def _mantel_stats_pearson_xp(x, y, permutations, rng, alternative, spearman=Fals
         y_flat = _xp_rank_average(xp, y_flat)
         # rebuild the ranked symmetric matrix so permuted pairs can be gathered
         # (one-time host assembly; on-device gather is a possible follow-up).
-        xr = _to_numpy(x_flat)
-        i0, j0 = _to_numpy(iu0), _to_numpy(iu1)
+        xr = _get_array(x_flat, to_numpy=True)
+        i0, j0 = _get_array(iu0, to_numpy=True), _get_array(iu1, to_numpy=True)
         full = np.zeros((n, n), dtype=np.float64)
         full[i0, j0] = xr
         full[j0, i0] = xr
--- a/skbio/stats/distance/_mantel.py
+++ b/skbio/stats/distance/_mantel.py
@@ -20,7 +20,10 @@ from scipy.stats import kendalltau, ConstantInputWarning, NearConstantInputWarni
 from ._cutils import mantel_perm_pearsonr_cy, mantel_perm_pearsonr_condensed_cy
 from skbio.stats.distance import DistanceMatrix
 from skbio.util import get_rng
+from skbio.util._array import ingest_array, _to_numpy
 from skbio._config import _resolve_engine
+
+import array_api_compat as _aac
 
 try:
     from numba import njit, prange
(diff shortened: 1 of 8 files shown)


skbio/stats/distance/_mantel.py
```diff
@@ -431,6 +434,37 @@ def mantel(
...
+    # A DistanceMatrix backed by a non-NumPy (e.g. GPU-resident) buffer is
+    # dispatched to the backend-agnostic xp path for pearson/spearman so the
+    # matrices stay on their device. kendalltau has no xp path and falls through
+    # to the host/scipy route below, where _order_dms materializes on host.
+    x_is_xp = isinstance(x, DistanceMatrix) and not _aac.is_numpy_array(x.data)
+    y_is_xp = isinstance(y, DistanceMatrix) and not _aac.is_numpy_array(y.data)
+
+    if special and (x_is_xp or y_is_xp):
+        # both inputs must be non-NumPy DistanceMatrix objects on the same backend
+        if not (x_is_xp and y_is_xp) or (
```
sfiligoi, line 445: We should probably just do the numpy path here.
I don't think a hard-fail is appropriate.

Maybe use the warning functionality.

skbio/stats/distance/_mantel.py
```diff
@@ -431,6 +434,37 @@ def mantel(
...
+            )
+        # align ids exactly like the NumPy path; the reorder/filter stays on
+        # the input's device (distmat_reorder is array-API aware).
+        x, y = _order_dms(x, y, strict=strict, lookup=lookup)
+        return _mantel_stats_pearson_xp(
+            x.data, y.data, permutations, rng, alternative,
+            spearman=(method == "spearman"),
+        )
+
+    # kendalltau has no xp path; materialize a non-NumPy DistanceMatrix on the
```
sfiligoi: Minor request:
Clarify in the comment that this is the default numpy path, because not all the inputs were xp.

skbio/stats/distance/_mantel.py
```diff
@@ -482,6 +516,128 @@ def mantel(
...
+    full (non-condensed) array-API matrices. Matches the RNG usage of the NumPy
+    path (identity permutation first, then ``permutations`` calls to
+    ``rng.permutation(n)``) so p-values are reproducible and identical.
+
+    Returns ``(orig_stat, p_value, n)``, matching ``mantel``'s return.
+    """
+    xp, X, Y = ingest_array(x, y)
+    if X.shape != Y.shape:
+        raise ValueError("Distance matrices must have the same shape.")
+    if X.ndim != 2 or X.shape[0] != X.shape[1]:
```
sfiligoi: Minor request:
Please use parenthesis to clarify what the grouping of operations is.

sfiligoi:
I know you expect a follow up performance improvement, but can you still provide how fast/slow the current implementation in on the GPU, compared to the CPU-only version?

sfiligoi:
@LarytheLord 
The code coverage report is a bit low.
Is it just a matter of the CI not counting the GPU-specific tests? Or something else?

Any chance you can raise the coverage fraction a bit more?

LarytheLord:
Addressed the inline comments: mixed or mismatched backends now warn and fall back to the NumPy path instead of raising, updated that comment to describe the default NumPy path taken when the inputs aren't all on the same backend and added the parentheses.

On the GPU vs CPU perf of the xp path: will post the numbers once I'm back on the AMD node.

LarytheLord:
On coverage the uncovered lines are the array API compute helpers, which only run when the DistanceMatrix is on a non NumPy backend and the development branch CI is NumPy only (array-api.yml only triggers on main).
I raised it by adding tests that call those helpers directly with NumPy arrays, which is array API compatible, so they run in ordinary CI and cover the added lines, including the shape and constant input branches. 
One question: do we want array-api.yml extended to also run on the development branch or leave it for later.

sfiligoi:
> On coverage the uncovered lines are the array API compute helpers, which only run when the DistanceMatrix is on a non NumPy backend and the development branch CI is NumPy only (array-api.yml only triggers on main). I raised it by adding tests that call those helpers directly with NumPy arrays, which is array API compatible, so they run in ordinary CI and cover the added lines, including the shape and constant input branches. One question: do we want array-api.yml extended to also run on the development branch or leave it for later.

@LarytheLord Yes, please enable the correct coverage test for PRs onto the development branch, too.

LarytheLord:
Added development_numba_gpu to array-api.yml's pull_request trigger. Because GitHub reads the pull_request trigger from the base branch, this won't run on the current PR, but once it merges the Array API CI will run on future PRs onto the development branch that carry the gpu-ci label.

LarytheLord:
@sfiligoi Extending array-api.yml to the dev branch made the Array API job run here, and jax is failing (torch is fine). Both test_kendalltau_backends and the mixed-backend fallback end up calling DistanceMatrix(_to_numpy(dm.data), dm.ids) on the host. jax gives back a read-only numpy array, and is_symmetric_and_hollow_cy needs a writable buffer so it errors with "buffer source array is read-only". is_symmetric_and_hollow only copies when the array isn't contiguous so a read-only contiguous one goes straight through.

Also _to_numpy says in its docstring that it's meant for test assertions and not library functions so calling it inside mantel might be the wrong thing to do in the first place.

How do we want to handle it? I can copy read-only inputs inside is_symmetric_and_hollow or make the cython take a const memoryview or stop using _to_numpy in mantel and materialize on the host some other way.

sfiligoi:
@LarytheLord Yeah, `_to_numpy` may not be the best choice.
See if there is any other function that does what you need... if not, create one.

As for "read-only buffers"...
Is there as standard way to test for read-only-ness?
If yes, checking that anywhere you expect in-place changes would be a good idea.
If not, use exception handling to deal with it, if possible.

In both cases, you will need to create a new array, if you cannot change the original.
I would expect the semantics is OK with most functions.
And try to just create the new array in a single pass, and not force a "copy to mutable-array then update in-place" logic, if at all possible.
And keep it on the same device, if at all possible, of course.

sfiligoi:
BTW, feel free to rebase, or merge, from the latest content of the dev branch, anytime, if it helps.

qiyunzhu:
I see there is a `development_numba_gpu` branch, where the GPU code is being developed on. I am fine with this. Just noting that I see it.

LarytheLord:
moved mantel's host path off _to_numpy and onto _get_array, the library-facing conversion. For the read-only part, my first cut copied inside is_symmetric_and_hollow, but that only fixed validation, so permute still failed later in distmat_reorder_condensed_cy. 
Fixed it at construction instead DistanceMatrix makes one writable copy when the input is a read-only or non-contiguous numpy array so validation, permutation and reordering all work. Writable inputs are still stored without a copy, and non-numpy buffers stay on their device.

Added a numpy-only test that builds a DistanceMatrix from a read-only array and permutes it, so it's covered in the normal CI and not just the array-api job. Also merged the latest development branch, the array-api jax and torch jobs are green now.

LarytheLord:
On perf: the current xp path is correctness first, not speed first. Each permutation copies the grouping to the device and runs a handful of small array API kernels so on the GPU it comes out slower than the CPU at the sizes I tested, and the gap narrows as n grows. So the current GPU path isn't a speedup yet.

The speedup is the fused numba GPU kernel (the follow up). It's validated on hardware now: permanova at 25k samples / 9999 permutations runs in 6.07s on an MI300X (7.47s float64) and 19.71s on an MI210, versus about 21.7s for scikit-bio-binaries, with identical p-values. That comes as a separate PR with engine=numba picking the GPU kernel when the matrix is already on a device.

sfiligoi (approved):
Looks good to me.

PR #2504, opened 2026-07-12
fix: handle atol=None in pair_align

steps-re:
## Summary

`pair_align`'s docstring says setting `atol` to 0 or `None` is valid and "will slightly increase performance." In practice, `atol=None` raises `TypeError` as soon as the code path that uses it runs:

- For `mode="local"`, the check `abs(score) <= atol` raises `TypeError: '<=' not supported between instances of 'float' and 'NoneType'`.
- For `mode="global"` (and for local, once past that check), `atol` is forwarded as `eps` into the Cython traceback routines (`_trace_one_linear`, `_trace_one_affine`, `_all_stops`, etc.), which declare it as a typed `floating eps` and raise `TypeError: must be real number, not NoneType`.

So today `atol=None` fails for every mode and every `max_paths` value, not just local alignment.

## Fix

Normalize `atol=None` to `0.0` at the top of `pair_align`, before it reaches any of the downstream comparisons. This makes `atol=None` behave identically to `atol=0`, matching the documented contract.

## Testing

Added `test_pair_align_atol_none`, which checks `atol=None` against `atol=0` for both `global`/`local` modes and `max_paths=1`/`max_paths=5` (exercising both the single-path and multi-path traceback code), plus a no-similarity local case that should return an empty path list without raising.

```
$ pytest skbio/alignment/tests/test_pair.py -q
35 passed
```

Confirmed the new test fails with the pre-fix code (`TypeError: must be real number, not NoneType`) and passes after the one-line fix.

## Checklist

- [x] I have documented this change in the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 
 ### Bug Fixes
 
-* Fixed `pair_align` raising `TypeError` when called with `atol=None`, a documented valid value equivalent to `atol=0` ([#2497](https://github.com/scikit-bio/scikit-bio/pull/2497)).
+* Fixed `pair_align` raising `TypeError` when called with `atol=None`, a documented valid value equivalent to `atol=0` ([#2504](https://github.com/scikit-bio/scikit-bio/pull/2504)).
 
 ## Version 0.7.3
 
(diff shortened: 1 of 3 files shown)


qiyunzhu (commented):
Thanks for noting this issue and fixing it. The code looks good but a bit too verbose. Let me refine and merge.

qiyunzhu (approved):

qiyunzhu:
After careful examination, there is actually a subtle problem with this `atol`. I will fix in a separate PR.

PR #2505, opened 2026-07-14
ENH: support condensed form for GPU-resident distance matrices

LarytheLord:
### Description
Follow-up to #2499, continuing the GPU-resident support for the distance-matrix classes. This makes the condensed-form conversions and the host-side export helpers work when the matrix data is a non-NumPy (GPU-resident cupy/torch/jax) buffer, so a distance matrix that lives on a device behaves the same as a NumPy-backed one for these operations.

`condensed_form()` on a non-NumPy 2-D buffer now extracts the row-major upper triangle on-device through the existing `distmat_reorder_condensed` helper (array-aware since #2499) rather than calling `scipy.spatial.distance.squareform`, which is NumPy-only. A non-NumPy buffer can also be constructed directly in condensed form, and a condensed non-NumPy buffer can be expanded back to redundant form. Expanding a condensed vector into a symmetric matrix goes through one host round-trip, because scattering values into a matrix needs item-assignment, which some backends (e.g. JAX) do not support. `plot()` and `to_data_frame()` materialize the matrix to the host before handing it to matplotlib and pandas, which are host-only.

NumPy inputs are unaffected: the form-conversion changes are guarded on a non-NumPy buffer, and the `plot`/`to_data_frame` exports route through a helper that returns a NumPy array unchanged. This is independent of #2503 (permanova/mantel) and touches only the matrix classes in `_base.py`.

### Scope / follow-ups
- File IO for a non-NumPy matrix (`lsmat` write) is not covered here and is left as a follow-up, as noted on #2499. Reading a matrix from a file always produces a NumPy array, so only writing is affected.
- The array-API symmetry check in `is_symmetric_and_hollow` compares the full transpose; a faster upper-triangle version is a possible performance follow-up (it needs to be measured, since an index-gather version I tried was slower on the backends I benchmarked).

### Testing
- array-API tests added for `DistanceMatrix` and `SymmetricMatrix`, run on numpy and PyTorch: condensed round-trip in both directions, `condensed_form`/`redundant_form` agreement with the NumPy path, `plot`/`to_data_frame` on a non-NumPy buffer, and a non-zero diagonal preserved through a condensed round-trip.
- Verified on a real AMD Instinct MI210 (torch-ROCm): the condensed conversions and exports run on-device and match the NumPy result exactly, including the non-zero-diagonal round-trip.
- The two tests that asserted condensed form was rejected for a non-NumPy buffer are replaced, since that restriction is now lifted.
- Existing distance test suites pass unchanged.

cc @sfiligoi @qiyunzhu

### DIFF
--- a/skbio/stats/distance/_base.py
+++ b/skbio/stats/distance/_base.py
@@ -26,7 +26,7 @@ from skbio.io.descriptors import Read, Write
 
 from ._utils import is_symmetric_and_hollow, is_symmetric
 from ._utils import distmat_reorder, distmat_reorder_condensed
-from skbio.util._array import ingest_array, copy_array
+from skbio.util._array import ingest_array, copy_array, _to_numpy
 
 import array_api_compat as _aac
 
@@ -776,7 +776,9 @@ class PairwiseMatrix(SkbioObject, PlottableMixin):
         fig, ax = self.plt.subplots()
 
         # use pcolormesh instead of pcolor for performance
-        heatmap = ax.pcolormesh(self.redundant_form(), cmap=cmap)
+        # matplotlib needs a host array; a non-NumPy (e.g. GPU-resident)
+        # buffer is materialized here since a heatmap is inherently host-side.
+        heatmap = ax.pcolormesh(_to_numpy(self.redundant_form()), cmap=cmap)
         fig.colorbar(heatmap)
 
         # center labels within each cell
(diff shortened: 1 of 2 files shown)


sfiligoi:
@LarytheLord 
Re: Expanding a condensed vector into a symmetric matrix goes through one host round-trip, because scattering values into a matrix needs item-assignment, which some backends (e.g. JAX) do not support.

Can you have a different path for JAX vs more-numpy-like GPU backends, like cuPy?
I don't think it makes sense penalizing all GPU implementations due to limitations of one (which likely is not the most commonly used one either).

LarytheLord:
Changed it so the mutable backends (PyTorch, CuPy) scatter the condensed values into the matrix on device via item-assignment and only JAX (which doesn't support item assignment) keeps the host round trip, so the common GPU backends aren't penalized anymore. The same host assembly pattern shows up in the Mantel Spearman path on the other PR, I'll give that the same treatment in a follow up.

sfiligoi (approved):
Looks good to me.

PR #2507, opened 2026-07-23
Improving RPCA code

qiyunzhu:
I did some minor tweaks of the code. A few questions for @aj-pinto :

1. `rpca` currently has a `max_iter` of 10000. Will it be too large? It might take a long time on real-world data.
2. `X = np.asarray(X, dtype=np.float64)` we probably shouldn't fix the float type to float64.
3. `rclr`'s `log_safe = xp.where(observed_mask, xp.log(mat), 0.0)` will raise a warning: "RuntimeWarning: divide by zero encountered in log". This is not ideal. We should improve this code.

***

### DIFF
--- a/skbio/stats/ordination/_optspace.py
+++ b/skbio/stats/ordination/_optspace.py
@@ -24,12 +24,12 @@ of singular values, and :math:`P_\Omega` projects onto the observed entries.
 
 References
 ----------
-.. [1] Keshavan RH, Montanari A, Oh S. 2010. Matrix Completion from a
-   Few Entries. IEEE Transactions on Information Theory 56(6):2980-2998.
+.. [1] Keshavan, R. H., Montanari, A., & Oh, S. (2010). Matrix completion from a
+   few entries. IEEE transactions on information theory, 56(6), 2980-2998.
 
-.. [2] Martino C, Morton JT, Marotz CA, Thompson LR, Tripathi A,
-   Knight R, Zengler K. 2019. A Novel Sparse Compositional Technique
-   Reveals Microbial Perturbations. mSystems 4:e00016-19.
+.. [2] Martino, C., Morton, J. T., Marotz, C. A., Thompson, L. R., Tripathi, A.,
+   Knight, R., & Zengler, K. (2019). A novel sparse compositional technique reveals
+   microbial perturbations. mSystems, 4(1), 10-1128.
 
 """
 
@@ -289,7 +289,7 @@ def solve_gauss_newton_step(U, V, S, observed_mask, R, tol, damp):
 
 
 def line_search(U, V, dU, dV, obj0, obj_fn, alpha0, tau, c):
-    """Backtracking line search"""
+    """Backtracking line search."""
 
     # Maximum depth of line search (no greater than 1e-16)
     max_ls = 16
@@ -304,7 +304,7 @@ def line_search(U, V, dU, dV, obj0, obj_fn, alpha0, tau, c):
     best = (U, V, obj0, alpha0)
     converged = False
 
-    for i in range(max_ls):
+    for _ in range(max_ls):
         # Retract step to Grassmann manifold
         U_try = retract_grassmann(U, alpha * dU)
         V_try = retract_grassmann(V, alpha * dV)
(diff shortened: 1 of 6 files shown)


skbio/stats/ordination/_optspace.py
```diff
@@ -343,24 +342,27 @@ def retract_grassmann(X, dX):
 def optspace(X, dimensions=3, max_iter=10000, tol=1e-5, method="GD"):
     r"""Matrix completion using the OptSpace algorithm.
 
-    OptSpace is an algorithm for recovering a low-rank matrix from a
-    subset of observed entries. It uses gradient descent on the
-    Grassmann manifold to find the optimal low-rank approximation.
+    OptSpace is an algorithm for recovering a low-rank matrix from a subset of observed
+    entries. It uses gradient descent on the Grassmann manifold to find the optimal
```
aj-pinto: Small wording change I overlooked:
```suggestion
    entries. It uses optimization on the Grassmann manifold to find the optimal
```

skbio/stats/ordination/_optspace.py
```diff
@@ -343,24 +342,27 @@ def retract_grassmann(X, dX):
...
+    X : array_like of shape (n_samples, n_features)
+        A matrix with observed values and NaN for missing entries.
     dimensions : int, optional
         The rank of the matrix to recover. Default is 3.
     max_iter : int, optional
-        Maximum number of iterations. Default is 20.
+        Maximum number of iterations. Default is 10000.
     tol : float, optional
         Convergence tolerance. Default is 1e-5.
+    method : {'GD', 'GN'}, optional
```
aj-pinto, line 359: If this exposed to the user, we should develop it more in the future with better documentation on the differences between the methods.

skbio/stats/ordination/_optspace.py
```diff
@@ -381,42 +383,49 @@ def optspace(X, dimensions=3, max_iter=10000, tol=1e-5, method="GD"):
...
     The algorithm proceeds as follows:
 
-    1. Initialize U, V using trimmed SVD of the observed matrix
+    1. Initialize `U`, `V` using trimmed SVD of the observed matrix.
     2. Iteratively:
-       a. Compute optimal S given current U, V
-       b. Update U, V with the Gauss-Newton step dU, dV
-       c. Project U, V back to Grassmann manifold
+       a. Compute optimal `S` given current `U`, `V`.
+       b. Update `U`, `V` with the Gauss-Newton step `dU`, `dV`.
```
aj-pinto:
```suggestion
       b. Update `U`, `V` with the manifold optimization step `dU`, `dV`.
```

aj-pinto:
@qiyunzhu 

Great, I have left some comments. To address your concerns:

> 1. `rpca` currently has a `max_iter` of 10000. Will it be too large? It might take a long time on real-world data.

I set max_iter artificially high during testing to accommodate differences in Gauss-Newton and gradient descent.

Gauss-Newton iterations are individually slower, but at most about 100 iterations are needed for any given problem. Gradient descent iterations are much faster, but many more of them are typically needed (sometimes in the thousands). 

For large, slow problems, the max iteration count will safely never be reached for either method. For smaller or more ill-conditioned problems, I think a large max_iter is safer, but this does risk unnecessarily long wait times. More testing would be necessary to tune this properly. 

> 2. `X = np.asarray(X, dtype=np.float64)` we probably shouldn't fix the float type to float64.

I agree, we can change this to allow other float types. I'm not sure if types other than float would be allowed since my understanding is that some data types do not allow NaNs. Currently, the algorithm depends on NaN values to detect the observed data.

> 3. `rclr`'s `log_safe = xp.where(observed_mask, xp.log(mat), 0.0)` will raise a warning: "RuntimeWarning: divide by zero encountered in log". This is not ideal. We should improve this code.

Is this due to float values very close to zero? The observed mask should eliminate any true zeros: `observed_mask = (mat != 0) & ~xp.isnan(mat)`. Maybe the `!=` should be swapped for `~xp.isclose`, but with a very small tolerance so that no true proportions are discarded. I had discussed this with Matt previously.

qiyunzhu:
Hello @aj-pinto Thanks for reviewing my code and providing comments! I have merged all suggestions you made. They are reasonable to me. With regard to your comments:

https://numpy.org/doc/1.23/reference/generated/numpy.asfarray.html

> > 1. `rpca` currently has a `max_iter` of 10000. Will it be too large? It might take a long time on real-world data.
> 
> I set max_iter artificially high during testing to accommodate differences in Gauss-Newton and gradient descent.
> 
> Gauss-Newton iterations are individually slower, but at most about 100 iterations are needed for any given problem. Gradient descent iterations are much faster, but many more of them are typically needed (sometimes in the thousands).
> 
> For large, slow problems, the max iteration count will safely never be reached for either method. For smaller or more ill-conditioned problems, I think a large max_iter is safer, but this does risk unnecessarily long wait times. More testing would be necessary to tune this properly.

I see. I agree with you that more testing will be needed. For now I think we should 1) set a small number by default, and 2) throw a warning if not converged.


> > 2. `X = np.asarray(X, dtype=np.float64)` we probably shouldn't fix the float type to float64.
> 
> I agree, we can change this to allow other float types. I'm not sure if types other than float would be allowed since my understanding is that some data types do not allow NaNs. Currently, the algorithm depends on NaN values to detect the observed data.

We can do this:

```python
X = np.asfarray(X)
```

This will guarantee that the output array is float, but it can be any float depending on the input. But the input is not float (e.g., int64, which does not support NaN), then it will be converted into float64.


> > 3. `rclr`'s `log_safe = xp.where(observed_mask, xp.log(mat), 0.0)` will raise a warning: "RuntimeWarning: divide by zero encountered in log". This is not ideal. We should improve this code.
> 
> Is this due to float values very close to zero? The observed mask should eliminate any true zeros: `observed_mask = (mat != 0) & ~xp.isnan(mat)`. Maybe the `!=` should be swapped for `~xp.isclose`, but with a very small tolerance so that no true proportions are discarded. I had discussed this with Matt previously.

I am not sure. Worth investigation.

PR #2508, opened 2026-07-24
ENH: add optional Numba engine for center_distance_matrix

LarytheLord:
## Summary

Adds an optional Numba implementation of distance-matrix double-centering, selectable via `engine="numba"` on `center_distance_matrix` and `pcoa`, alongside the existing Cython path. This extends the `engine=` selector already used by `permanova` and `mantel` to the ordination centering step.

This carries forward the Numba centering work started in #2491 by @fyash1010, rebased on the current `development_numba_gpu`, with the engine dispatch refactored to share the in-place/out-of-place logic, docstrings aligned to the `permanova`/`mantel` convention, and tests added. It is opened as a fresh PR since #2491 has been inactive.

The Numba kernels mirror the Cython implementation in `_cutils.pyx` (`e_matrix_means_cy`, `f_matrix_inplace_cy`), including the 24-wide block tiling, so the two engines are numerically interchangeable (agreement below 1e-15 for float64). Numba stays an optional dependency and Cython remains the default. The array-API (GPU-resident) centering path is already present on this branch, so this PR only adds the CPU Numba engine.

## Tests

- Numba-vs-Cython agreement for float64 and float32, in-place and out-of-place, including a size that is larger than the 24-wide tile and not a multiple of it to exercise the partial final block.
- The public `pcoa(engine="numba")` path against the Cython result.
- Invalid-engine handling.

@sfiligoi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 * Add optional support for the Numba backend [#2483](https://github.com/scikit-bio/scikit-bio/pull/2483), with Permanova and Mantel currently using it [#2488](https://github.com/scikit-bio/scikit-bio/pull/2488)and [#2464](https://github.com/scikit-bio/scikit-bio/pull/2464).
+* `pcoa` and `center_distance_matrix` can now use the Numba backend for distance-matrix centering via `engine="numba"` [#2508](https://github.com/scikit-bio/scikit-bio/pull/2508).
 
 ### Performance enhancements
 
--- a/skbio/stats/ordination/_center_distance_matrix_numba.py
+++ b/skbio/stats/ordination/_center_distance_matrix_numba.py
@@ -1,0 +1,91 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2013--, scikit-bio development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+"""Optional CPU Numba implementation of distance-matrix double-centering.
+
+Mirrors the Cython implementation in ``_cutils.pyx`` (``e_matrix_means_cy``,
+``f_matrix_inplace_cy`` and ``center_distance_matrix_cy``) so the two engines
+are numerically interchangeable.
+"""
+
+import numpy as np
+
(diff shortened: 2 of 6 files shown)


sfiligoi:
@LarytheLord 
I think you should add the engine check to the 
`_skbb_pcoa_fsvd_available` test, too.
That one is the equivalent of requesting cython, so when "numba" is requested, that code path should not be entered.

sfiligoi:
BTW: I just found out that JAX does have a version of `eigh` that supports `subset_by_index`, like host-based scipy:
https://docs.jax.dev/en/latest/_autosummary/jax.lax.linalg.eigh.html#jax.lax.linalg.eigh
used here
https://github.com/LarytheLord/scikit-bio/blob/ce00dc185bec60c8e697d77953850bf462241014/skbio/stats/ordination/_principal_coordinate_analysis.py#L289
https://github.com/LarytheLord/scikit-bio/blob/ce00dc185bec60c8e697d77953850bf462241014/skbio/stats/ordination/_principal_coordinate_analysis.py#L80

Not sure if this is the right PR to put it in, but should be at least tested.
(looks like neither cuPy nor PyTorch `eigh` supports it, so the host path is appropriate, there)

LarytheLord:
Thanks,
fsvd binaries path was a good catch, I'll gate _skbb_pcoa_fsvd_available on the engine so engine="numba" skips the binaries path, with the default (cython) behavior unchanged, plus a test that pcoa(method="fsvd", engine="numba") doesn't enter it.
Related permanova has the same pattern, where _skbb_permanova_available is checked before the engine is considered, so for a NumPy matrix engine="numba" there also ends up on the binaries path. Want me to gate it the same way, in this PR or a small separate follow-up?

JAX eigh: I tested it. jax.lax.linalg.eigh with subset_by_index raises NotImplementedError: subset_by_index not supported on CPU and GPU (jax 0.11.0), so it can't replace the host round-trip on the platforms we run on (the message implies it's TPU-only). Since neither cupy nor torch support subset_by_index either (as you noted), the scipy host path stays appropriate for all three. 
We can do a revisit a TPU only branch later if it's ever worth it.

sfiligoi:
Let's defer fixing of `_skbb_pcoa_fsvd_available` test to a follow-up PR, so it can covert permanova, too.

PR #2509, opened 2026-07-25
BUG: accumulate permanova s_T in float64 for float32 distance matrices

LarytheLord:
## Summary

PERMANOVA's total sum of squares `s_T` was accumulated in the distance matrix's dtype. For a **float32** matrix this gives an inaccurate pseudo-F, because `s_T` and the within-group sum of squares `s_W` nearly cancel in `s_A = s_T - s_W`, so a small float32 rounding error in `s_T` is amplified in the F ratio.

On real float32 UniFrac matrices the error grows with size — about 0.2% at n=5k and **~5% at n=25k** (the EMP matrices are float32). `s_W` is already accumulated in float64; this accumulates `s_T` in float64 too, in both the `DistanceMatrix` (Cython/Numba) path and the array-API path. Float64 matrices are unaffected. Found while benchmarking on the EMP UniFrac matrices.

## Change

`s_T` is computed with a float64 accumulator (`.sum(dtype=np.float64)` / `xp.sum(..., dtype=xp.float64)`). The per-element float32 square is negligible; it was the accumulation that mattered.

## Test

`test_float32_matches_float64`: a float32 matrix with weak grouping (where the cancellation is strongest) must give the same pseudo-F as its float64 equivalent. Fails on the old code (about 5e-5 at n=2000), passes with the fix.

@sfiligoi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 * When using the Numba backend, Permanova is up to 8x faster [#2488](https://github.com/scikit-bio/scikit-bio/pull/2488).
 
+### Bug fixes
+
+* Fixed `permanova` returning an inaccurate pseudo-F for float32 distance matrices, caused by accumulating the total sum of squares in float32 rather than float64 [#2509](https://github.com/scikit-bio/scikit-bio/pull/2509).
+
 ## Version 0.7.3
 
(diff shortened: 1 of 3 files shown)


sfiligoi (approved):
Looks good

PR #2510, opened 2026-07-27
MAINT: skip the scikit-bio-binaries path when engine="numba" in permanova and pcoa

LarytheLord:
## Summary

Follow-up to #2508 (as per review: "Let's defer fixing of `_skbb_pcoa_fsvd_available` test to a follow-up PR, so it can cover permanova, too").

When `engine="numba"` is requested, `permanova` and `pcoa` (`method="fsvd"`) no longer take the scikit-bio-binaries path. That path is the Cython-equivalent acceleration, so entering it contradicts a numba request. The binaries availability is now gated on the resolved engine: skipped for `"numba"`, kept for `"cython"` and the default.

## Changes

- `permanova`: gate `_skbb_permanova_available` on the already-resolved engine.
- `pcoa`: resolve the engine once at the top (like `permanova`) and gate `_skbb_pcoa_fsvd_available` on it. A consequence is that `pcoa(engine="numba")` now requires Numba to be installed even for a GPU-resident matrix, matching `permanova`.
- Tests: each asserts the binaries availability is not checked for `engine="numba"`, and is for `"cython"`.
- Docstrings note that `"numba"` skips the binaries acceleration.

@sfiligoi

### DIFF
--- a/skbio/stats/distance/tests/test_permanova.py
+++ b/skbio/stats/distance/tests/test_permanova.py
@@ -8,7 +8,7 @@
 
 import io
 from functools import partial
-from unittest import TestCase, main
+from unittest import TestCase, main, mock
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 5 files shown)


skbio/stats/distance/tests/test_permanova.py
```diff
@@ -123,6 +123,18 @@ def test_call_no_permutations(self):
         obs = permanova(self.dm_no_ties, self.grouping_equal, permutations=0)
         self.assert_series_equal(obs, exp)
 
+    @numba_code
+    def test_engine_numba_skips_binaries(self):
+        # A numba request must not enter the scikit-bio-binaries path (the
+        # Cython-equivalent), so its availability is not even checked; a cython
+        # request still consults it.
+        with mock.patch.object(permanova_mod, "_skbb_permanova_available",
```
sfiligoi: I appreciate you checking this, but I don't think we need it in every CI run.
I'd rather avoid using mock, unless it is absolutely needed.

skbio/stats/ordination/tests/test_principal_coordinate_analysis.py
```diff
@@ -72,6 +72,21 @@ def test_engine_numba_matches_cython(self):
...
 
+    @numba_code
+    def test_fsvd_engine_numba_skips_binaries(self):
+        # A numba request must not enter the scikit-bio-binaries fsvd path (the
+        # Cython-equivalent), so its availability is not even checked; a cython
+        # request still consults it.
+        from skbio.stats.ordination import (
+            _principal_coordinate_analysis as pcoa_mod,
+        )
+        with mock.patch.object(pcoa_mod, "_skbb_pcoa_fsvd_available",
```
sfiligoi: As above.

sfiligoi (approved):
Looks great

PR #2511, opened 2026-07-27
ENH: add a Numba GPU backend for permanova and mantel

LarytheLord:
## Summary

I added a fused, single-source Numba GPU backend for `permanova` and `mantel`. The same kernel source compiles through `numba.cuda` on NVIDIA and `numba.hip` on AMD. It runs when `engine="numba"` and the `DistanceMatrix` is already resident on the GPU: the kernel reads the device matrix in place with no host round-trip, and I dispatch on where the matrix lives instead of adding a new engine value. If the kernel cannot build or run on the current stack (for example the ROCm init-order case on some torch-ROCm builds), the call warns once and falls back to the array-API GPU path, which is correct on any array-API backend. I draw the permutations on the host in the same RNG order as the cython, numba, and xp engines, so the p-values match.

This builds on the array-API GPU support (#2503) and the float64 `s_T` fix (#2509), and it sits on top of the binaries engine gate (#2510).

## Correctness

I validated the kernels against the cython and double-precision reference:

- **permanova**: exact p-value parity on the AMD MI210 and MI300X at 1k, 5k, and 25k, including 25k with 9999 permutations, with the pseudo-F matching to about 1e-13. On an NVIDIA RTX 4060 the pseudo-F is within 8.6e-8 of the cython float64 reference at 25k. The kernel reads a float32 matrix but accumulates in float64, so its single-precision result stays close to double precision.
- **mantel**: matches the cython engine and the double-precision reference at 5k, to about 4e-7. I have not run a 25k mantel correctness check yet, so the 25k mantel rows below are timing only.

## Benchmarks

I ran two benchmarks, each on a single machine. The first, on an AMD MI300X, compares the fused kernel to the CPU engines; its cython baseline is built with OpenMP. The second, on an NVIDIA RTX 4060, is the direct same-card comparison against scikit-bio-binaries, the existing GPU accelerator. The two are on different GPUs, so compare within a table, not across.

### AMD Instinct MI300X: the fused kernel vs the CPU engines (seconds)

`permanova`

| n | perms | cython | numba (CPU) | numba-hip (kernel) | xp (array-API) |
|---|---|---|---|---|---|
| 5000 | 999 | 1.28 | 2.05 | **0.045** | 1.58 |
| 5000 | 9999 | 9.81 | 3.16 | **0.22** | 2.24 |
| 25000 | 999 | 43.8 | 5.77 | **1.84** | 3.61 |
| 25000 | 9999 | 426 | 30.0 | **7.29** | 12.5 |

`mantel`

| n | perms | cython | numba (CPU) | numba-hip (kernel) | xp (array-API) |
|---|---|---|---|---|---|
| 5000 | 999 | 3.91 | 5.60 | **0.37** | 1.96 |
| 5000 | 9999 | 36.6 | 44.5 | **5.96** | 8.13 |
| 25000 | 999 | 120 | 138 | **11.2** | 12.98 |
| 25000 | 9999 | 1407 | 1547 | **157** | 161.6 |

The kernel is a large win over the CPU engines. Against the array-API GPU fallback its advantage is largest at small and medium sizes; for mantel it narrows at the largest `n`, where the kernel and the xp path run about even (157 vs 162 s at 25k / 9999).

### NVIDIA RTX 4060: every engine on one card (`permanova`, 999 permutations, seconds)

Here I put all the engines on the same card so they line up directly, including scikit-bio-binaries, the existing GPU accelerator. SKBB accelerates `permanova` and `pcoa_fsvd`, not `mantel`, so there is no SKBB mantel row. I confirmed SKBB's GPU build auto-detects and uses the card, and that its result is correct: the pseudo-F is within 5e-12 of the cython float64 reference at 25k in double precision. cython here is an OpenMP build; both it and numba (CPU) run in float64, while xp, SKBB (fp32), and the kernel run in float32 (the kernel accumulates in float64).

| n | cython | numba (CPU) | xp (array-API) | SKBB GPU (fp32) | SKBB GPU (fp64) | numba-cuda kernel |
|---|---|---|---|---|---|---|
| 1000 | 0.20 | 0.026 | 0.73 | **0.01** | 0.17 | 0.027 |
| 5000 | 3.86 | 0.41 | 4.74 | **0.14** | 0.43 | 0.43 |
| 25000 | 116.8 | 10.37 | OOM (8 GB) | **3.19** | 6.23 | 8.57 |

On this 8 GB consumer card the array-API path runs out of memory at 25k: its `xp` statistic (`_permanova_sW_xp`) materializes several full n-by-n intermediates, two of them (`mask` and `dm*dm`) about 2.5 GB each in float32 at 25k, which together exceed the card. The fused kernel walks the matrix in place with no n-by-n temporary, so it (and SKBB) fit. For permanova here SKBB's C++ is the fastest GPU option (3.2 s fp32, 6.2 s fp64 at 25k), ahead of my kernel (8.6 s), and on this entry-level GPU the kernel leads the CPU engines only modestly. The kernel earns its place elsewhere: it also does `mantel`, which SKBB does not; it is much faster than the CPU engines on the datacenter MI300X above; it is one Python source across NVIDIA and AMD inside the `engine=` dispatch with no separate compiled dependency; and its float64 accumulation keeps the single-precision result within 8.6e-8 of the reference at 25k, where the SKBB fp32 path is at 4.9e-4.

## Tests

Host logic (permutation order, statistic assembly), the dispatch (backend routing, warn-once fallback), and the kernel math against the cython engine. The kernel bodies run on hardware and are gated to the GPU CI.

@sfiligoi

### DIFF
--- a/skbio/stats/distance/_gpu.py
+++ b/skbio/stats/distance/_gpu.py
@@ -40,8 +40,8 @@ def _numba_gpu_module_for(arr):
     unavailable backend, when Numba GPU support is not installed, or when this
     backend's fused kernel has already failed once in this process (see
     :func:`_mark_gpu_unavailable`); the caller then takes the array-API path. A
-    CuPy (CUDA) array is never routed to ``numba.hip`` (mixing the two crashes on
-    the array handoff).
+    CuPy array is routed to ``numba.cuda``, never ``numba.hip`` (the two target
+    different GPU vendors).
 
     Parameters
     ----------
--- a/skbio/stats/distance/_gpu.py
+++ b/skbio/stats/distance/_gpu.py
@@ -11,9 +11,9 @@
 The permutation-test statistics (PERMANOVA, Mantel) each ship a fused kernel that
 compiles through ``numba.cuda`` on NVIDIA and ``numba.hip`` on AMD. This module
 holds the piece they share: choosing the Numba GPU module for a given
-device-resident array (``numba.cuda`` for CuPy / CUDA-PyTorch, ``numba.hip`` for
-ROCm-PyTorch), and a correctness-first fallback to the array-API path whenever the
-fused kernel is unavailable or fails to build on the running stack.
+device-resident array (``numba.cuda`` for CUDA CuPy / PyTorch, ``numba.hip`` for
+ROCm CuPy / PyTorch), and a correctness-first fallback to the array-API path
+whenever the fused kernel is unavailable or fails to build on the running stack.
 """
 
 from warnings import warn
@@ -35,13 +35,12 @@ _unavailable = set()
 def _numba_gpu_module_for(arr):
     """Return the Numba GPU module for ``arr``'s device, or None.
 
-    CuPy and CUDA-built PyTorch map to ``numba.cuda``; ROCm-built PyTorch maps to
-    ``numba.hip``. Returns None for any other namespace (e.g. JAX, Dask), an
-    unavailable backend, when Numba GPU support is not installed, or when this
-    backend's fused kernel has already failed once in this process (see
-    :func:`_mark_gpu_unavailable`); the caller then takes the array-API path. A
-    CuPy array is routed to ``numba.cuda``, never ``numba.hip`` (the two target
-    different GPU vendors).
+    CUDA-built CuPy and PyTorch map to ``numba.cuda``; ROCm-built CuPy and PyTorch
+    map to ``numba.hip`` (both report the same array-API namespace, so the build's
+    own flag disambiguates them). Returns None for any other namespace (e.g. JAX,
+    Dask), an unavailable backend, when Numba GPU support is not installed, or when
+    this backend's fused kernel has already failed once in this process (see
+    :func:`_mark_gpu_unavailable`); the caller then takes the array-API path.
 
     Parameters
     ----------
@@ -57,7 +56,11 @@ def _numba_gpu_module_for(arr):
     if name in _unavailable:
         return None
     if name == "cupy":
-        want = "cuda"
+        import cupy
+
+        # cuPy reports the same namespace for CUDA and ROCm builds; the build's
+        # own is_hip flag disambiguates them (ROCm -> numba.hip).
+        want = "hip" if getattr(cupy.cuda.runtime, "is_hip", False) else "cuda"
     elif name == "torch":
         import torch
 
--- a/skbio/stats/distance/_gpu.py
+++ b/skbio/stats/distance/_gpu.py
@@ -1,0 +1,105 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2013--, scikit-bio development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+"""Shared helpers for the single-source Numba CUDA/HIP GPU backends.
+
+The permutation-test statistics (PERMANOVA, Mantel) each ship a fused kernel that
+compiles through ``numba.cuda`` on NVIDIA and ``numba.hip`` on AMD. This module
+holds the piece they share: choosing the Numba GPU module for a given
+device-resident array (``numba.cuda`` for CuPy / CUDA-PyTorch, ``numba.hip`` for
+ROCm-PyTorch), and a correctness-first fallback to the array-API path whenever the
+fused kernel is unavailable or fails to build on the running stack.
+"""
+
+from warnings import warn
+
+import array_api_compat as _aac
+
+from skbio.util._array import _get_backend_name
+
+# One thread block per permutation; _TPB threads walk the columns (the tile width
+# used by scikit-bio-binaries).
+_TPB = 128
+
+# Backend names whose fused Numba GPU kernel failed to build or run in this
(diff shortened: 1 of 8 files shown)


sfiligoi:
> The RTX 4060 is a consumer card without ECC memory and showed non-reproducible memory corruption at 25k

This does not make sense.

ECC should not be needed.
Memory corruption should be extremely rare... definitely not more than once a day.
If you see repeated errors, it is due to algorithm problems, not lack of ECC.
(Unless the GPU is broken... possible, but unlikely)

skbio/stats/distance/_gpu.py
```diff
@@ -0,0 +1,105 @@
...
+    Returns
+    -------
+    module or None
+        The Numba GPU module usable on this array's device, else None.
+    """
+    name = _get_backend_name(_aac.array_namespace(arr))
+    if name in _unavailable:
+        return None
+    if name == "cupy":
+        want = "cuda"
```
sfiligoi: Does this work on AMD GPUs?
e.g. cuPy still uses "cuda" even on AMD HW?

skbio/stats/distance/_permanova_gpu.py
```diff
@@ -0,0 +1,177 @@
...
+
+                local = 0.0
+                row = trow
+                while row < max_row:
+                    min_col = max(tcol, row + 1)
+                    gidx = row_grouping[row - trow]
+                    col = min_col + icol      # thread -> column: coalesced read
+                    if col < max_col:
+                        if groupings[gel, col] == gidx:
+                            v = nb_f64(mat[row, col])
```
sfiligoi: Why are you upgrading v to fp64, instead of only keeping the sum in higher precision?

skbio/stats/distance/_mantel_gpu.py
```diff
@@ -0,0 +1,202 @@
...
+        row = t
+        while row < n_dims - 1:
+            vrow = perm_order[p, row]
+            row_start = row * (n_dims - 1) - ((row - 1) * row) // 2
+            icol = 0
+            col = row + 1
+            while col < n_dims:
+                vcol = perm_order[p, col]
+                y = ym_norm[row_start + icol]
+                x = nb_f64(mat[vrow, vcol]) * mul + add
```
sfiligoi: Do we need to keep the variables in FP64?
Wouldn't be enough to keep only `acc` at higher precision?

skbio/stats/distance/_permanova_gpu.py
LarytheLord: fixed, the per element product now stays in the matrix dtype and only the accumulator (s_W) is float64. 
It doesn't cost accuracy: at 25k the observed s_W still matches the cython float64 reference to 2.5e-8, the same as with the cast, so the upcast wasn't buying anything. On the MI300X it's speed-neutral, as expected with its strong fp64. 

On the RTX 4060, where fp64 is weak, the kernel-only permanova time at 25k/999 drops from about 10.8 s to 5.7 s, roughly 1.9x. I'll refresh the full table with the new numbers.

skbio/stats/distance/_mantel_gpu.py
LarytheLord: Same fix here: only acc stays float64, the per-element math is in the matrix dtype. Verified at 25k, the observed r still matches cython to 1.4e-11.

skbio/stats/distance/_gpu.py
LarytheLord: A CuPy array maps to numba.cuda here. On AMD, numba.cuda reports unavailable, so this returns None and the call takes the array-API path, which is correct, it just does not use the fused kernel. The kernel's AMD path is torch-ROCm, detected via torch.version.hip. You are right that CuPy reports the "cuda" namespace even for a ROCm build, so the namespace alone can't distinguish CuPy-on-AMD. CuPy can be built for ROCm, so if we want CuPy-on-AMD to use the kernel I can detect the ROCm build and route it to numba.hip; otherwise the current behavior is the safe array-API fallback. Happy to do either.

skbio/stats/distance/tests/test_mantel.py
```diff
@@ -6,7 +6,7 @@
 # The full license is in the file LICENSE.txt, distributed with this software.
 # ----------------------------------------------------------------------------
 
-from unittest import TestCase, main
+from unittest import TestCase, main, mock
```
sfiligoi: Please avoid using mock.

skbio/stats/distance/tests/test_permanova.py
```diff
@@ -8,7 +8,7 @@
 
 import io
 from functools import partial
-from unittest import TestCase, main
+from unittest import TestCase, main, mock
```
sfiligoi: Please avoid using mock.

sfiligoi:
Main code looks great, thanks.

@LarytheLord  But can you avoid using mock in the tests?

LarytheLord:
Done, mock is out of the tests. The warn once test now uses a real numpy array. I dropped the two fallback tests, they hit the kernel raises path, which can't be triggered without a GPU present, so they needed mock to fake a GPU failure. The kernel can't run in CI anyway, so I validate it directly on the GPU (MI300X and the 4060). If you want the graceful degradation path covered, I can add it back with a mock, but I left it out to keep mock out of the suite.

sfiligoi (approved):
Looks great.

PR #2515, opened 2026-08-04
ENH: make the DM class family fully device-resident

LarytheLord:
## Summary

Igor asked whether the DM class family (DistanceMatrix, SymmetricMatrix, and the asymmetric PairwiseMatrix/DissimilarityMatrix) is fully GPU-aware and optimized, as a first step ahead of the functions that build DMs. I went through the class and found two spots that still forced a non-NumPy (e.g. GPU-resident) buffer to the host, both fixed here.

## What changed

I found that `distmat_reorder_condensed_py`, used by `filter()` and `permute()` when a matrix is stored condensed, always converted its input to a NumPy array. I fixed it so the reorder indices are computed on the host (cheap) but only the actual gather runs on the input's device, matching the pattern already used by `distmat_reorder` and `distmat_reorder_condensed`.

I also found that the asymmetric `PairwiseMatrix` (the `DissimilarityMatrix` base) could not hold a non-NumPy buffer at all. Its input normalization asserted a NumPy array, its shape validation compared dtypes against `np.float32`/`np.float64` directly (which never matches a non-NumPy dtype object), and its copy path called `data.copy()`, which not every backend implements. I fixed all three to mirror the patterns already used by `SymmetricMatrix`: `ingest_array` to preserve the buffer and its device, `xp.isdtype` for the dtype check, and `copy_array` for a backend-safe copy. I also handled a 1D condensed vector on a non-NumPy backend, which can't be symmetric-expanded by scipy's `squareform`: it now goes through a single host round-trip and gets placed back on the input's device instead of raising.

## Correctness

I verified this on both an AMD MI300X (torch-ROCm) and an NVIDIA RTX 4060 (CuPy): a condensed-stored DistanceMatrix's `filter()` and `permute()` stay on-device and match the NumPy reference, and a DissimilarityMatrix built from a device buffer stays on-device through construction, copy, and transpose. Same result on both vendors.

## Tests

I added array-backend tests (numpy, jax, torch, cupy) covering both fixes: `filter()` and `permute()` on a condensed-stored DistanceMatrix, and a new `PairwiseMatrixArrayAPITests` class covering construction, dtype casting, copy, transpose, and the 1D condensed input case for the asymmetric matrix. These run on CPU torch and jax in CI, not just on a GPU, since the array-API branch itself is backend-generic.

## Note

While I was checking `filter()`/`permute()`'s condensed branch I found an unrelated pre-existing issue: on a `SymmetricMatrix` with a non-zero diagonal, both methods silently reset the diagonal to 0 when reconstructing the result, since the reconstruction never passes `diagonal=`. It's identical on the NumPy path and unaffected by this change (`DistanceMatrix` never hits it, since its diagonal is always 0). I didn't fix it here, flagging it separately.

### DIFF
--- a/skbio/stats/distance/_base.py
+++ b/skbio/stats/distance/_base.py
@@ -142,7 +142,17 @@ class PairwiseMatrix(SkbioObject, PlottableMixin):
         # convert data to redundant if 1D input.
         # should do this for PairwiseMatrix only.
         if data.ndim == 1:
-            data = squareform(data, force="tomatrix", checks=False)
+            if _aac.is_array_api_obj(data) and not _aac.is_numpy_array(data):
+                # A condensed vector is inherently a symmetric-matrix
+                # representation; scipy's squareform has no array-API path, so
+                # expand it via a single host round-trip and place the result
+                # back on the input's device.
+                xp = _aac.array_namespace(data)
+                dev = getattr(data, "device", None)
+                mat_np = squareform(_to_numpy(data), force="tomatrix", checks=False)
+                data = xp.asarray(mat_np, device=dev)
+            else:
+                data = squareform(data, force="tomatrix", checks=False)
 
         if ids is None:
             ids = self._generate_ids(data)
@@ -184,6 +194,18 @@ class PairwiseMatrix(SkbioObject, PlottableMixin):
         # np.asarray to situations where the data are (a) not already a numpy
         # array or (b) the data are not a single or double precision numpy
         # data type.
+        if _aac.is_array_api_obj(data) and not _aac.is_numpy_array(data):
+            # Preserve a non-NumPy (e.g. GPU-resident) array-API buffer so the
+            # PairwiseMatrix can hold GPU data. NumPy / list / tuple inputs are
+            # still normalized to a NumPy array below. A non-floating buffer is
+            # cast to float64, mirroring the NumPy path (which keeps float32/64
+            # and casts everything else to float).
+            _xp, data = ingest_array(data)
+            if not _xp.isdtype(data.dtype, (_xp.float32, _xp.float64)):
+                data = _xp.astype(data, _xp.float64)
+            data_: NDArray = data
+            return (data_, ids, validate_shape, validate_ids)
+
         _issue_copy = True
         if isinstance(data, np.ndarray):
             if data.dtype in (np.float32, np.float64):
@@ -433,7 +455,7 @@ class PairwiseMatrix(SkbioObject, PlottableMixin):
             Deep copy of the matrix. Will be the same type as ``self``.
 
         """
-        data = self._data.copy()
+        data = copy_array(self._data)
         if transpose:
             data = data.T
         return self.__class__(data, deepcopy(self.ids), validate=False)
@@ -1050,7 +1072,11 @@ class PairwiseMatrix(SkbioObject, PlottableMixin):
             raise PairwiseMatrixError(
                 "Data must be square (i.e., have the same number of rows and columns)."
             )
-        if data.dtype not in (np.float32, np.float64):
+        # isdtype works for NumPy and non-NumPy (e.g. GPU) buffers alike, via each
+        # array's own namespace. Restrict to single/double precision (as the
+        # legacy NumPy check did) so both paths accept exactly the same dtypes.
+        xp = _aac.array_namespace(data)
+        if not xp.isdtype(data.dtype, (xp.float32, xp.float64)):
             raise PairwiseMatrixError("Data must contain only floating point values.")
 
     def _index_list(self, list_: Sequence[str]) -> dict:
@@ -2574,7 +2600,28 @@ def distmat_reorder_condensed_py(in_mat, reorder_vec):
     np.ndarray
         Condensed matrix.
 
+    Notes
+    -----
+    A non-NumPy array-API buffer (e.g. GPU-resident) is gathered on its own
+    device: the reorder indices are cheap index arithmetic computed on the
+    host, then only the actual data gather runs on the input's device.
+
     """
+    if _aac.is_array_api_obj(in_mat) and not _aac.is_numpy_array(in_mat):
+        n_original = _vec_to_shape(in_mat)
+        reorder_np = np.asarray(reorder_vec)
+        n_filtered = len(reorder_np)
+        i_indices, j_indices = np.triu_indices(n_filtered, k=1)
+        old_indices = _condensed_index(
+            reorder_np[i_indices], reorder_np[j_indices], n_original
+        )
+        xp = _aac.array_namespace(in_mat)
+        idx = xp.asarray(
+            np.asarray(old_indices, dtype=np.int64),
+            device=getattr(in_mat, "device", None),
+        )
+        return in_mat[idx]
+
     reorder_vec = np.asarray(reorder_vec)
     in_mat = np.asarray(in_mat)
     n_original = _vec_to_shape(in_mat)
--- a/skbio/stats/distance/_base.py
+++ b/skbio/stats/distance/_base.py
@@ -71,6 +71,76 @@ class MissingIDError(PairwiseMatrixError):
         self.args = ("The ID '%s' is not in the matrix." % missing_id,)
 
 
+def _expand_condensed(condensed: NDArray, diagonal=None) -> NDArray:
+    """Expand a 1-D condensed vector to a redundant 2-D matrix.
+
+    Works for a NumPy array or any non-NumPy array-API buffer (e.g.
+    GPU-resident), staying on the input's device for the latter. Only one
+    (n, n) buffer is allocated on the input's device, scattered into from
+    both triangles; no boolean-mask buffer or full-matrix transpose is
+    needed. Mutable backends (PyTorch, CuPy) scatter via item-assignment;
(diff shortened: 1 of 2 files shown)


skbio/stats/distance/_base.py
```diff
@@ -142,7 +142,17 @@ def __init__(
...
         if data.ndim == 1:
-            data = squareform(data, force="tomatrix", checks=False)
+            if _aac.is_array_api_obj(data) and not _aac.is_numpy_array(data):
+                # A condensed vector is inherently a symmetric-matrix
+                # representation; scipy's squareform has no array-API path, so
+                # expand it via a single host round-trip and place the result
+                # back on the input's device.
+                xp = _aac.array_namespace(data)
+                dev = getattr(data, "device", None)
+                mat_np = squareform(_to_numpy(data), force="tomatrix", checks=False)
```
sfiligoi: Any reason why we could not implement this in Numba, instead?

sfiligoi:
@LarytheLord Regarding the `SymmetricMatrix` problem you found, can you open a GitHub Issue?

skbio/stats/distance/_base.py
sfiligoi: Note: Using the native mechanisms may be even better (e.g. JAX using a JAX implementation, torch using a torc implementation, etc.), but that may be too complicated.
Having just a numba implementation (if numba-gpu installed) would be good enough at this stage (with CPU roundtrip if Numba not present). 
But adding a comment about the desire of having native implementations would be nice.

skbio/stats/distance/_base.py
```diff
@@ -142,7 +142,31 @@ def __init__(
...
+                # values into the matrix on-device via item-assignment, no host
+                # round-trip. Immutable backends (JAX), which do not support
+                # item-assignment, go through scipy's squareform on the host
+                # instead, mirroring SymmetricMatrix._condensed_to_redundant.
+                # A native per-backend implementation (or a Numba/GPU kernel,
+                # if available) could replace this for JAX too; this generic
+                # array-API scatter is the simpler option for now.
+                xp = _aac.array_namespace(data)
+                dev = getattr(data, "device", None)
+                if _get_backend_name(xp) in ("torch", "cupy"):
```
sfiligoi: Any reason this is limited to cupy and torch?
And not e.g. JAX?

sfiligoi:
One comment about manual testing:
Could you check all of cupy, pytorch and JAX on both AMD and NVIDIA GPUs?
(i.e. all 6 combinations)

skbio/stats/distance/_base.py
LarytheLord: The array-API scatter already avoids the host round-trip for torch and cupy without needing Numba installed at all and this is a rare, non-performance-critical construction-time path rather than a hot loop so I didn't add a kernel for it. Added a comment noting a native per-backend or Numba/GPU implementation could replace this later.

skbio/stats/distance/_base.py
```diff
@@ -142,7 +142,31 @@ def __init__(
...
+                    k = data.shape[0]
+                    n = int((1.0 + (1.0 + 8.0 * k) ** 0.5) / 2.0)
+                    idx = xp.arange(n, device=dev)
+                    upper = idx[:, None] < idx[None, :]
+                    mat = xp.zeros((n, n), dtype=data.dtype, device=dev)
+                    mat[upper] = data
+                    data = mat + xp.matrix_transpose(mat)  # diagonal stays zero
```
sfiligoi: This seems expensive, both in terms of compute and memory need.
Have you checked what happens on very large matrices, e.g. > 5k^2?

skbio/stats/distance/_base.py
LarytheLord: noted on JAX, it doesn't need the host round-trip either. It can scatter on device via its functional update API instead of the item assignment I used for torch/cupy. 
Verified locally (CPU JAX, no JAX-GPU installed on either node yet to test the literal GPU case, but the operation is device-preserving by construction, the same guarantee the torch/cupy scatter relies on). 
I mirrored the existing _condensed_to_redundant precedent for the JAX exception, which has the same limitation for the same reason. 
Can extend both to cover JAX on-device and fix the older method too for consistency, if you want.

sfiligoi: Yes, fixing `_condensed_to_redundant` would be great.

One question though:
Why are you not using it here, too?

I know it is currently defined later, but is there any reason not to move it up?

BTW: Could we change the semantics of `_condensed_to_redundant` to handle both numpy and xp arrays, so that we do not have to have an if statement every time it is invoked?

LarytheLord: Makes sense, agreed on all three. Plan is to pull the expansion logic out into a shared function that takes the condensed data and an optional diagonal (so both PairwiseMatrix, which has none and SymmetricMatrix, which does, can use it), handles numpy and array-API backends in one place instead of branching at each call site and lives early enough in the file for PairwiseMatrix's constructor to call it directly. 
Haven't made this change yet, it's next up, will push by tomorrow.

skbio/stats/distance/_base.py
LarytheLord: Hadn't checked at scale, fair to ask. At n=5000 the original version peaks around 425MB (bool mask plus two full n by n arrays). A leaner version with paired index arrays only gets to about 400MB, not the real win I expected. A more aggressive in place version would nearly halve it, but PyTorch rejects it outright with a RuntimeError (aliasing), so that's not viable. Given these numbers, a real kernel is probably the right call here, not just an array-API scatter. Will fold this into the _condensed_to_redundant consolidation.

LarytheLord:
Pushed (d657b24a). _expand_condensed is defined ahead of PairwiseMatrix, both PairwiseMatrix.__init__ and SymmetricMatrix._init_data/redundant_form call it now instead of branching separately and it reuses the existing _vec_to_shape helper rather than re-deriving n from k inline. _condensed_to_redundant is gone, no remaining callers anywhere in the codebase. Tests pass on numpy, torch, and jax (185 each), ruff clean.

Also finished the 6 combination hardware check you asked for. torch and cupy confirmed correct on both AMD and NVIDIA. JAX confirmed correct on NVIDIA. JAX on AMD is blocked, not by this PR: jax-rocm7-plugin installed per AMD's documented recipe (targets ROCm 7.0.2) against this node's ROCm 7.1/7.2 and jax.devices() falls back to CPU with AttributeError: module 'jax._src.hardware_utils' has no attribute 'num_available_amd_gpus' in the log, a real version mismatch in the plugin itself, not something wrong in the PR. 5 of 6 confirmed on real hardware, can open an upstream issue for the sixth if useful.

sfiligoi (approved):
Looks good to me.

Some large scale benchmarks would have been appreciated, but not essential.

LarytheLord:
What I'm measuring is `_expand_condensed(condensed, diagonal=None)` directly, the function this PR added, not a full `DistanceMatrix` construction or `filter()`/`permute()` call. For a given n, `condensed` is a random 1-D vector of length n(n-1)/2 (the upper-triangle count), and I'm timing the call that expands it to the redundant n×n matrix:

​```python
k = n * (n - 1) // 2
condensed = np.random.default_rng(0).random(k).astype(np.float64)
out = _expand_condensed(condensed)  # move condensed to torch/cupy/jax first for the device paths
​```

Ran this at scale on GPU, each case in its own process to avoid cross-backend memory contamination.

**CPU baseline (numpy, host-only, AMD EPYC 9684X, single-threaded)**

| n | numpy |
|---|---|
| 5,000 | 45 ms, 200 MB |
| 10,000 | 253 ms, 800 MB |
| 15,000 | 661 ms, 1.8 GB |
| 20,000 | 1.2 s, 3.2 GB |
| 25,000 | 2.0 s, 5.0 GB |

**NVIDIA node (RTX 4060, 8 GB)**

| n | torch | cupy | jax |
|---|---|---|---|
| 5,000 | 76 ms, 502 MB | 100 ms, 500 MB | 304 ms, 700 MB |
| 10,000 | 255 ms, 2.0 GB | 359 ms, 2.0 GB | 561 ms, 2.6 GB |
| 15,000 | 555 ms, 4.5 GB | 776 ms, 4.5 GB | OOM |
| 20,000 | OOM | OOM | OOM |
| 25,000 | OOM | OOM | OOM |

**AMD node (MI300X, 192 GB)**

| n | torch | cupy |
|---|---|---|
| 5,000 | 214 ms, 502 MB | 472 ms, 500 MB |
| 10,000 | 369 ms, 2.0 GB | 504 ms, 2.0 GB |
| 15,000 | 636 ms, 4.5 GB | 759 ms, 4.5 GB |
| 20,000 | 1.0 s, 8.0 GB | 1.5 s, 8.0 GB |
| 25,000 | 1.5 s, 12.5 GB | 1.6 s, 12.5 GB |

On the NVIDIA node, torch and cupy hold up to n=15000 and OOM at n=20000; jax OOMs a step earlier, at n=15000, because its two functional `.at[].set()` updates each allocate a new buffer instead of mutating in place. On the AMD node none of that matters, n=25000 runs in under 2 seconds either way. The peak at every size matches the arithmetic exactly: two O(n²) index arrays from `np.triu_indices` plus the O(n²) output plus the O(n²) input, e.g. at n=25000 that's 5+5+2.5=12.5GB, which is exactly what both backends measured.

PR #2519, opened 2026-08-06
ANCOM-BC2 reimplementation

iiiime:


skbio/stats/composition/_ancombc2.py
```diff
@@ -0,0 +1,1863 @@
...
+        -----
+        Results are cached: repeated calls return the same object without
+        recomputation.
+        """
+        if self._structural_zeros is not None:
+            return self._structural_zeros
+        if self._group is None:
+            raise ValueError(
+                "structural_zeros requires a group variable; "
+                "pass group=... to ancombc2()."
```
qiyunzhu, line 210: Is `structural_zeros` specific to `ancombc2`, or does it apply to `ancombc` as well?

skbio/stats/composition/_ancombc2.py
```diff
@@ -0,0 +1,1863 @@
...
+            )
+        _struc_zero_fn = globals()["struc_zero"]
```
qiyunzhu, line 212: Looking up a function from the module's namespace using `globals` may not be the desired behavior. Suggest directly calling `struc_zero` if it's within the same module (I guess yes), or call it through `skbio.stats.composition.struc_zero`.

skbio/stats/composition/_ancombc2.py
```diff
@@ -0,0 +1,1863 @@
...
+
+        Raises
+        ------
+        ValueError
+            If no group variable was specified (``group`` was None when
+            :func:`ancombc2` was called).
+
+        Notes
+        -----
+        Results are cached: repeated calls return the same object without
```
qiyunzhu, line 202: This is a nice design.

skbio/stats/composition/_ancombc2.py
```diff
@@ -0,0 +1,1863 @@
...
+            _dmat=dmat,
+            _beta_hat=beta_hat,
+            _vcov_hat=vcov_hat,
+            _alpha=alpha,
+            _p_adjust=p_adjust,
+        )
+
+    else:
+        # Data preprocessing (no filtering. user pre-filters)
+        O1 = pd.DataFrame(matrix, index=samples, columns=features)
```
qiyunzhu, line 1555: I think the variable names `O1` and `O2` are too vague. I'd suggest using `data` and `agg_data` instead.

skbio/stats/composition/_ancombc2.py
```diff
@@ -0,0 +1,1863 @@
...
+        beta_hat = res_main["beta_hat"]
+        se_hat = res_main["se_hat"]
+        W = res_main["W"]
+        p_hat = res_main["p_hat"]
+        q_hat = res_main["q_hat"]
+        diff_abn = res_main["diff_abn"]
+
+        lfc = beta_hat
+
+        res_dict = {
```
qiyunzhu, line 1598: I thought part of this code overlaps with `ancombc`, and should be re-used.

skbio/stats/composition/_ancombc2.py
```diff
@@ -0,0 +1,1863 @@
...
+    using stored intermediate data
+    Returned by both :func:`ancombc` and :func:`ancombc2`.
+
+    Attributes
+    ----------
+    res : pd.DataFrame
+        Primary results with (FeatureID, Covariate) multi-index.
+        For ANCOM-BC (reestimate=False): columns Log2(FC), SE, W, pvalue,
+        qvalue, Signif.
+        For ANCOM-BC2 (reestimate=True): columns lfc, SE, W, pvalue, qvalue,
```
qiyunzhu, line 73: Are `Log2(FC)` and `lfc` two different things?

skbio/stats/composition/_ancombc2.py
```diff
@@ -0,0 +1,1863 @@
...
+        Signif.
+    delta_em : ndarray
```
qiyunzhu, line 75: I think intermediate variables like `delta_em` don't need to be exposed to the end user.

skbio/stats/composition/_ancombc2.py
```diff
@@ -0,0 +1,1863 @@
...
+        EM-estimated biases per covariate.
+    delta_wls : ndarray
+        WLS-estimated biases per covariate.
+    samp_frac : ndarray
+        Estimated sampling fractions per sample.
+    feature_table : pd.DataFrame or None
+        Preprocessed feature table. None for ANCOM-BC.
+    reestimate : bool
```
qiyunzhu, line 83: I think the Boolean parameter `reestimate` should be renamed as a string parameter `method` with possible values `{'ANCOM-BC' and 'ANCOM-BC2'}`.

skbio/stats/composition/_ancombc2.py
```diff
@@ -0,0 +1,1863 @@
...
+    pseudo=0,
+    pseudo_sens=True,
+    s0_perc=0.05,
+    group=None,
+    struc_zero=False,
+    neg_lb=False,
+    alpha=0.05,
+    max_iter=100,
+    tol=1e-5,
+    mdfdr_fwer_ctrl_method="holm",
```
qiyunzhu, line 1447: Does this parameter need to be different from `p_adjust`? If not, we can omit it.

skbio/stats/composition/_ancombc2.py
```diff
@@ -0,0 +1,1863 @@
...
+    mdfdr_B=100,
+    trend_contrast=None,
```
qiyunzhu, line 1746: It looks to me that these additional parameters are not used in the core algorithm, but directly passed to the result. Then, downstream methods like `trend_test` will use them. If that's the case, I would suggest moving those parameters from the main function to the downstream methods (and make the results of downstream methods non-cached because they can be customized).

skbio/stats/composition/_ancombc2.py
```diff
@@ -0,0 +1,1863 @@
...
+        alpha=alpha,
+        p_adjust=p_adjust,
+    )
+
+
+def ancombc2(
+    table,
+    metadata,
+    formula,
+    aggregate_data=None,
```
qiyunzhu, line 1733: I think we have discussed and agreed that `aggregate_data` should become `aggregator`. But this can be postponed to a later PR.

skbio/stats/composition/_ancombc2.py
```diff
@@ -0,0 +1,1863 @@
...
+        if self._trend_test is not None:
+            return self._trend_test
+        if not self.reestimate:
+            raise ValueError("trend_test() is only available for ANCOM-BC2 results.")
+        if self._group is None:
+            raise ValueError(
+                "trend_test requires a group variable; pass group=... to ancombc2()."
+            )
+
+        raw = _ancombc_trend(
```
qiyunzhu, line 430: No need to prefix every function with `_ancombc` because this entire script is already called `_ancombc.py`.

qiyunzhu (commented):
Hi @iiiime This is a good start. I left some comments. Two overarching comments are:

1. CI failed on MacOS with PyPI due to numerical issues. This is often expected. I would suggest relaxing the precision of the comparison, if that works, or simply remove such comparisons.
2. The content of `_ancombc2.py` and `_ancombc2_tests.py` should be merged into `_ancombc.py`.

skbio/stats/composition/_ancombc2.py
iiiime: The initial estimation in ancombc2 does overlaps with ancombc, but separate calculations might be needed due to the difference in data preprocessing. Creating a common function would work

PR #2520, opened 2026-08-07
BUG: fix center_distance_matrix public wiring and two torch pcoa crashes

LarytheLord:
## Summary

You asked back on August 2nd to add versionadded tags and the array-API compatibility grid to permanova, mantel, pcoa, and center_distance_matrix, and confirmed the backend list should be numpy, jax, torch, and cupy, not dask. While putting that together I found the documentation would have been claiming support that didn't actually work, so this PR fixes three real bugs first and adds the documentation on top of working code.

## What changed

The public skbio.stats.ordination.center_distance_matrix was wired to the wrong function. __init__.py imported it from _utils.py, a NumPy-only implementation used internally by pcoa() as its Cython/Numba-accelerated engine, instead of from _principal_coordinate_analysis.py, where the actual array-API-aware version already lives and which pcoa() itself correctly calls. A user calling the public name on a torch, jax, or cupy buffer would hit the NumPy-only path and crash. I fixed the import to point at the right function. This changes nothing for NumPy input, since the array-API-aware version still calls the same Cython/Numba engine for a real np.ndarray.

pcoa() crashed on torch for both the eigh and fsvd methods. xp.isclose(eigvals, 0) fails because the array-API standard requires both arguments to be arrays, and torch's backend enforces that strictly while NumPy and JAX tolerate a bare Python int. I fixed it with xp.zeros_like(eigvals). Fixing that surfaced a second crash in the same block: eigvals.argsort()[::-1] also fails on torch, which doesn't support negative-step slicing. I fixed that with xp.flip(eigvals.argsort(), axis=0).

I then added the versionadded 0.7.4 tags and the array-API compatibility grid to all four functions, with the backend list you confirmed.

## Correctness

I verified both pcoa fixes against a NumPy reference on torch and jax, both eigh and fsvd, agreement under 4e-16. I also checked cupy specifically, since it isn't installed in my local environment: ran the two exact fixed expressions on AMD's cupy-ROCm build and confirmed neither crashes and both produce correct values.

## Tests

I added TestPCoABackends, covering both eigh and fsvd against a NumPy reference across numpy, jax, torch, and cupy, and a direct identity test guarding the wiring fix specifically, since the existing TestCenterDistanceMatrixBackends imports center_distance_matrix directly from _principal_coordinate_analysis and would never have caught the public-import bug. Neither of the two bugs had any test coverage before this.

@sfiligoi

### DIFF
--- a/skbio/stats/distance/_mantel.py
+++ b/skbio/stats/distance/_mantel.py
@@ -23,6 +23,7 @@ from ._mantel_gpu import _run_mantel_gpu
 from skbio.stats.distance import DistanceMatrix
 from skbio.util import get_rng
 from skbio.util._array import ingest_array, _get_array
+from skbio.util._decorator import array_api_doc
 from skbio._config import _resolve_engine
 
 import array_api_compat as _aac
@@ -161,6 +162,7 @@ if NUMBA_AVAILABLE:
             permuted_stats[p] = my_ps
 
 
+@array_api_doc(backends=["numpy", "jax", "torch", "cupy"])
 def mantel(
     x: DistanceMatrix | ArrayLike,
     y: DistanceMatrix | ArrayLike,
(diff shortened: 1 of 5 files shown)


sfiligoi (approved):
Looks good to me.

Issue #2349, opened 2025-12-16
Is it expected to replace underscores from tip names in the tree?


antgonza:
I wasn't expecting to have the `_` be replaced by ' ', for example:
```
In [1]: [n.name for n in skbio.tree.TreeNode.read(["((a_a,b_b)c,(d,e)f);"]).tips()]
Out[1]: ['a a', 'b b', 'd', 'e']
```

Naty89:
Hi, I'd like to work on this for my GSoC 2026 application. I'll implement the change to disable underscore-to-space    conversion by default (with an opt-in parameter to preserve legacy behavior) and open a PR. Is this the right approach @qiyunzhu @wasade ?

PR #2528, opened 2026-08-12
ENH: read and write BPTree via the newick format through skbio.io

monkeytim19:
## Summary

Wire `BPTree`'s newick I/O into the `skbio.io` registry so it behaves like `TreeNode`: `BPTree.read()` / `BPTree.write()` delegate to the registry, with `newick` as the default write format. Reading and writing never materialize a `TreeNode` (or any per-node object) — the parenthesis, name, length, and edge arrays are built and serialized directly — and both paths are backed by the existing compiled (Cython) parser, so they stay fast on very large trees.

Addresses the review feedback on #2498 asking that `BPTree` read/write newick through the shared registry rather than as bespoke methods.

## What's in this PR

- `BPTree.read` / `BPTree.write` now route through `skbio.io` (`default_write_format = "newick"`), mirroring `TreeNode`.
- A `newick` reader and writer registered for `BPTree` in `skbio/io/format/newick.py`, implemented as thin adapters over the compiled `parse_newick` / `write_newick`.
- The former npz serialization is **kept, not removed** — it moves to dedicated `BPTree.to_npz` / `BPTree.from_npz` methods (logic unchanged) now that `read`/`write` carry the registry meaning. `.npz` stores topology, names, and branch lengths (not edge numbers).

## Design note (grammar)

The `BPTree` reader uses the compiled newick scanner rather than the shared pure-Python tokenizer, so `convert_underscores` does not apply to it (underscores in unquoted `BPTree` labels are preserved literally) and `[N]` is read as an edge number rather than a comment. This is documented in the format docstring.

## Scope / follow-ups

Deliberately scoped to **newick only**. Planned as separate follow-up PRs:

- **jplace** as its own `skbio.io` format (per @qiyunzhu's suggestion).

## Testing

- New `TestNewickBPTree`: exact BP arrays, `{N}` edge numbers, write round-trips, registry read + sniffing, and error handling.
- `BPTree` newick round-trips via `read`/`write`, plus a `to_npz`/`from_npz` round-trip test.
- Full `skbio/io` and `skbio/tree` suites pass, plus the newick doctest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 
 * Added `BPTree`, a succinct balanced-parentheses representation of trees for memory-efficient storage and fast topological queries on very large trees. Ported from [improved-octo-waddle](https://github.com/biocore/improved-octo-waddle) ([#2498](https://github.com/scikit-bio/scikit-bio/pull/2498)).
+* `BPTree` reads and writes the `newick` format through the `skbio.io` registry, mirroring `TreeNode`: `BPTree.read` and `BPTree.write` delegate to the registry with `newick` as the default format. The reader and writer translate directly to and from `BPTree`'s parenthesis/name/length/edge arrays without constructing any intermediate per-node objects, and are backed by a compiled (Cython) implementation so that reading and writing stay fast on trees with millions to billions of nodes. ([#2528](https://github.com/scikit-bio/scikit-bio/pull/2528)).
 
 ## Version 0.7.3
 
--- a/doc/source/_templates/BP.rst
+++ b/doc/source/_templates/BP.rst
@@ -11,6 +11,8 @@
 
       ~{{ name }}.read
       ~{{ name }}.write
+      ~{{ name }}.to_npz
+      ~{{ name }}.from_npz
 
    .. rubric:: Tree navigation
 
--- a/skbio/io/format/newick.py
+++ b/skbio/io/format/newick.py
@@ -176,17 +176,10 @@ In addition to :class:`~skbio.tree.TreeNode`, this format reads and writes
 :class:`~skbio.tree.BPTree`, the succinct balanced-parentheses representation
 intended for very large trees. Reading and writing a ``BPTree`` never
 materializes a ``TreeNode`` (or any other per-node object): the parenthesis,
-name, length, and edge-number arrays are built directly, preserving
-``BPTree``'s memory scalability. For throughput on trees with millions to
-billions of nodes, the ``BPTree`` reader and writer are backed by a compiled
-(Cython) implementation in :mod:`skbio.tree.bp`.
-
-Edge numbers -- integer identifiers assigned to edges (as used by phylogenetic
-placement) -- are written in the canonical ``{N}`` form immediately after a
-branch length (e.g. ``A:0.01{4}``). On read they are parsed into the tree's
-edge array when present (both ``{N}`` and ``[N]`` are accepted). The ``BPTree``
-writer accepts an `include_edge_nums` parameter (``False`` by default); when
-``True`` it emits the ``{N}`` suffix for every node.
+name, and length arrays are built directly, preserving ``BPTree``'s memory
+scalability. For throughput on trees with millions to billions of nodes, the
+``BPTree`` reader and writer are backed by a compiled (Cython) implementation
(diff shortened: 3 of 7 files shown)


monkeytim19:
@qiyunzhu @mortonjt Would be good to have review on this before submitting a separate pull request on the jplace integration into IO work that builds on this.

qiyunzhu:
@monkeytim19 I have read your code. This is a good attempt and it looks overall sound to me. There are some issues related to design decision, which I can discuss with you here. But they may be deferred to a later PR.

First of all, `BPTree` needs to be added to `skbio/tree/__init__.py`. `from_npz` and `to_npz` need to be added to `BP.rst`. Otherwise they won't show up in the rendered documentation.

I think `newick.py` is not the ideal place to elaborate on BP tree. This file is dedicated to defining the Newick format, which itself has nothing to do with BP tree. You may notice that this file doesn't even discuss TreeNode much. I think that file format and internal data structure should be separated and modular. But I see that BP tree is related to edge number, which in turn is related to how the output Newick file looks like. So I kind-of understand your current design.

Edge number is not part of Newick definition. Is it related to jplace, by the way? If so, it should be deferred to the jplace PR.

AI flagged quote and underscore treatment. They were some pre-historic design decisions which I think should be re-visited. See #2349 for example. But this needs to be dealt with in a separate PR.

***

FYR, here is my local AI's opinion.

> - **P1**: The BPTree reader does not support Newick comments and can misparse them as topology or edge numbers. `_bp_io.pyx:247` only tracks quotes, so parentheses inside `[comments]` alter the balanced-parentheses structure. Then `_bp_io.pyx:82` treats any token containing `[` as an edge annotation and attempts integer conversion. This breaks valid existing inputs such as `a[comment]` and quoted labels containing brackets, despite Newick’s documented comment support.
> 
> - **P1**: Quoted labels lose apostrophes on read and cannot be safely written. `_bp_io.pyx:81` strips outer quotes, but its no-length branch removes every single quote via `replace("'", "")` at `_bp_io.pyx:88`. Thus `'b_a'''` becomes `b_a` instead of `b_a'`. Conversely, `_bp_io.pyx:147` emits apostrophes unchanged and never doubles them, yielding invalid Newick for names containing `'`. The established serializer explicitly doubles apostrophes at `newick.py:387`.
> 
> - **P2**: BPTree serialization does not quote all labels requiring escaping. `_bp_io.pyx:149` only checks `; , ( ) : _`, omitting whitespace and `[`/`]`. A name such as `e; foo` happens to be quoted due to `;`, but labels like `de d` are emitted with unescaped whitespace and bracket-containing names are emitted as comments. This diverges from the existing writer’s `,:_;()[]` rules at `newick.py:366`.
> 
> The current round-trip test battery includes some quoted labels but misses apostrophe and bracket fidelity because it tests BPTree-to-BPTree state after the reader has already discarded the metadata (`test_newick.py:431`).

monkeytim19:
# Changes Made
- **Default write format**: `BPTree.write()`** now uses `self.default_write_format` instead of a hardcoded "newick", so the default stays consistent if it ever changes or a subclass overrides it.
- **`to_npz` docstring**
- **Docs rendering**: Added to_npz / from_npz under the Tree IO rubric in BP.rst so they show up in the generated documentation.
- **Edge numbers deferred from newick**: agreed edge numbers aren't part of the Newick spec; removed the edge-number documentation, the include_edge_nums writer parameter, and the corresponding tests from the newick format layer. The shared compiled grammar still parses edge annotations, which the follow-up jplace format uses directly.

# Deferred
- Edge-number support feature — the upcoming jplace PR, where edge numbers are actually motivated
- Quote / apostrophe / comment ([...]) correctness and underscore/quote handling revisit

monkeytim19:
@qiyunzhu Following our discussion earlier, I attempted to rebuild the documentation locally to inspect the changes for `BPTree.to_npz()` and `BPTree.from_npz()`. The build was successful and the documentation display properly (_see attached_). If everything is as intended, then it should be good to be merged. 
 
Screenshot 2026-08-24 at 8 19 47 PM
Screenshot 2026-08-24 at 8 19 13 PM
Screenshot 2026-08-24 at 8 19 20 PM

qiyunzhu:
@monkeytim19 Thanks! It looks about right. Let me merge.

PR #2536, opened 2026-08-18
BUG: preserve non-zero diagonal in SymmetricMatrix.filter/permute on condensed matrices

### LINKED ISSUE #2516: SymmetricMatrix.filter()/permute() drop a non-zero diagonal on condensed matrices
`SymmetricMatrix.filter()` and `.permute()` silently reset a non-zero diagonal to 0 when the matrix is stored condensed.

Both methods reconstruct the result via `self.__class__(new_data, ids, condensed=True)`, but neither passes `diagonal=`, so the constructor falls back to its default (zero).

Repro:

```python
import numpy as np
from skbio.stats.distance import SymmetricMatrix

a = np.array([[5., 1, 2], [1, 7, 3], [2, 3, 9]])
cond = np.array([1., 2, 3])
sm = SymmetricMatrix(cond, ids=['a', 'b', 'c'], condensed=True, diagonal=np.diag(a).copy())
sm.diagonal        # [5. 7. 9.]

f = sm.filter(['a', 'c'])
f.diagonal         # 0.0, expected something like [5. 9.]
```

`permute()`'s `condensed=False` branch has the same reconstruction pattern and hits the same issue.

`DistanceMatrix` never triggers this, since its diagonal is always exactly 0.

Found while working on #2515 (unrelated GPU-completeness fix); this is identical on the NumPy path, not specific to that PR.

youdie006:
Fixes #2516.

## Root cause
`SymmetricMatrix.filter()` and `.permute()` reconstruct their result from the condensed form via `self.__class__(..., condensed=True)` without passing `diagonal=`. The condensed representation stores only off-diagonal values, so the constructor defaults the diagonal to 0. A `SymmetricMatrix` stored condensed with a non-zero diagonal therefore has it silently reset to 0 by filter/permute. `DistanceMatrix` never surfaces this because it is always hollow. The failure is silent because `__eq__` compares only the off-diagonal data, not the diagonal.

## Fix
Introduce an overridable helper `_diagonal_kwargs(idxs=None)`:
- `PairwiseMatrix` returns `{}` (no explicit diagonal).
- `SymmetricMatrix` returns `{"diagonal": ...}`, subsetting/reordering an array diagonal by `idxs` and passing a scalar (uniform) diagonal through unchanged (full diagonal when `idxs is None`).
- `DistanceMatrix` overrides back to `{}` (it stays hollow, and its constructor does not accept `diagonal=`). Note `DistanceMatrix` subclasses `SymmetricMatrix`, so this explicit override is required to avoid inheriting the diagonal-threading version.

The helper is threaded through the three condensed reconstruction sites (filter same-ids fast path, filter subset, permute), so `SymmetricMatrix` preserves its diagonal while `DistanceMatrix` is unaffected.

## Tests
Added regression tests: SymmetricMatrix filter (subset, reorder, same-ids, scalar-uniform) and permute preserve the diagonal; DistanceMatrix filter/permute remain hollow. Verified red (assertions fail with diagonal == 0 before the fix) and green (full `test_base.py`: 169 passed) in a clean `python:3.12` container.

## Changelog
Added a Bug Fixes entry referencing #2516.

Credit to @LarytheLord for reporting and diagnosing the root cause.

---
This contribution was prepared with AI assistance and reviewed by me before submission.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 * Fixed a subtle floating-point arithmetic issue in `pair_align` under a linear gap penalty. Previously it could be less tolerant than expected when `atol` was set smaller than the default (1e-5) and scores involved decimal numbers ([#2513](https://github.com/scikit-bio/scikit-bio/pull/2513)).
 * Fixed a bug in `pair_align` which raised a `TypeError` when called with `atol=None`. This should be equivalent to `atol=0` ([#2504](https://github.com/scikit-bio/scikit-bio/pull/2504)).
 * Patched `rclr` such that it won't raise a zero division warning ([#2526](https://github.com/scikit-bio/scikit-bio/pull/2526)).
+* Fixed `SymmetricMatrix.filter` and `SymmetricMatrix.permute` silently resetting a non-zero diagonal to `0` when the matrix was stored in condensed form. Both methods reconstructed the result from its condensed representation, which only carries off-diagonal values, without passing the diagonal through. The diagonal is now subset and reordered along with the rows and columns. `DistanceMatrix` is unaffected because it is always hollow ([#2516](https://github.com/scikit-bio/scikit-bio/issues/2516)). Thank @LarytheLord for reporting and diagnosing the root cause.
 
 
 ## Version 0.7.3
--- a/skbio/stats/distance/_base.py
+++ b/skbio/stats/distance/_base.py
@@ -471,6 +471,28 @@ class PairwiseMatrix(SkbioObject, PlottableMixin):
             new_ids = tuple(mapper(x) for x in self.ids)
         self.ids = new_ids
 
+    def _diagonal_kwargs(self, idxs=None):
+        """Return constructor keyword arguments that preserve the diagonal.
+
+        Reconstructing a matrix from its condensed form loses the diagonal,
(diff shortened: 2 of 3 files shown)


LarytheLord:
I reported this one, so for what it is worth, here is a check of the implementation against the original repro.

It is correct. I verified on this branch:

- the exact repro from #2516: `filter(['a', 'c']).diagonal` returns `[5. 9.]`
- condensed `permute` returns a correct permutation of the diagonal
- a scalar (uniform) diagonal passes through unchanged
- a diagonal given as a plain list works
- `DistanceMatrix` stays hollow, so the subclass override does the right thing
- the full `test_base.py` suite passes, 169 tests

I also went looking for reconstruction sites that might have been missed. `filter()` and `permute()` are the only ones that rebuild through the constructor with `condensed=True`. The other two paths go through `_copy()`, which preserves the diagonal already.

One correction to my own issue text, since it otherwise overstates the scope. I wrote there that `permute()`'s `condensed=False` branch was also affected. That was imprecise: `permute(condensed=True)` returns a raw ndarray, and a matrix stored non-condensed keeps its diagonal in `_data`, so both are fine. Only condensed storage was ever broken, which is exactly what this PR covers.

sfiligoi:
Closes #2516

PR #2538, opened 2026-08-18
MAINT: move the coverage configuration into pyproject.toml

LarytheLord:
Mirrors #2533 on this branch.

### Description

Main removed `.coveragerc` in #2533 and moved the configuration into `pyproject.toml`, and its Makefile, ci.yml and numba.yml now pass `--rcfile ../pyproject.toml`. `development_numba_gpu` still has `.coveragerc` and still points at it in those same three places.

That divergence matters for the merge into main. Coverage does not fall back to defaults when the rcfile is missing, it exits with `Couldn't read '../.coveragerc' as a config file`, so any call site here that still names the old file fails once main's deletion lands alongside it.

This copies main's `[tool.coverage.run]` and `[tool.coverage.report]` sections into `pyproject.toml` verbatim, removes `.coveragerc`, drops its now-dead `check-manifest` ignore entry, and repoints the three call sites. `numba.yml` needed no change to its test step, since it runs `make test` and picks the command up from the Makefile.

One side effect worth noting: this also brings the `**/tests/**` omit pattern from #2531, so the branch stops using the `*/tests*` form that was not matching reliably.

If you rebase this branch onto main, it will pick the same change up from #2533 directly and this becomes unnecessary. Happy to close it in that case.

### Testing

Run the way CI invokes it, from `ci/`:

```
coverage run --rcfile ../pyproject.toml -m pytest ../skbio/stats/distance/tests/test_base.py   # 185 passed
coverage lcov --rcfile ../pyproject.toml                                                       # 132 files, 0 test files
```

The generated report contains no test files, confirming the new omit patterns apply. The configuration block is byte-for-byte identical to main's, and `# pragma: no cover` still takes effect, because `exclude_also` adds to coverage's defaults rather than replacing them.


qiyunzhu:
I guess you did the changes manually rather than merging the upstream. Anyway that is not important. Let me merge.

PR #2541, opened 2026-08-19
PERF: Optimization to BPTree range-excess queries and widen its search kernel to 64-bit

monkeytim19:
## Summary

Two focused, **output-preserving** improvements to the balanced-parentheses `BPTree` backend introduced in #2498. Both target the `bp_tree` feature branch and leave the public API, the stored arrays, and every query result bit-for-bit identical.

1. **PERF** — answer range-minimum/maximum *excess* queries in **O(log n)** instead of O(span), speeding up `rmq`, `rMq`, and everything built on them (`lca`, `height`, `deepest_node`).
2. **MAINT** — widen the rmM-tree construction and the `fwdsearch`/`bwdsearch` search kernel from 32-bit C `int` to 64-bit indices, removing a silent overflow ceiling at ~537 M tips.

## 1. O(log n) range min/max excess queries

`rmq(i, j)` / `rMq(i, j)` find the position of the minimum / maximum *excess* in the parenthesis range `[i, j]`. The original port answered them by **linearly scanning the range**, so cost grew with the array distance `j − i`. Because `lca`, `height`, and `deepest_node` are all thin wrappers over these, a lowest-common-ancestor query between two distant nodes, or a whole-subtree height, degraded to an O(n) scan.

This rewrite instead **descends the rmM (range min-max) tree** — the block-summary structure already built at construction — to obtain the min/max excess over the range in O(log n), scanning only the two partial end-blocks. The attaining position is then recovered with a single existing `fwdsearch`, so no new "leftmost-qualifying-block” descent logic is introduced (`i` if `excess(i)` already equals the target, else `fwdsearch(i, target − excess(i))`). Segment-tree bounds are clamped so a fully-covered node never reads into the partially-filled last level of the heap.

**This is a general asymptotic win, not a deep-tree special case.** For far-separated node pairs the array distance is ~O(n) on *every* topology, so the improvement scales with tree size across balanced, caterpillar, and random shapes alike (see benchmarks). Queries between *near* nodes (small range) were already cheap and are unchanged.

## 2. 64-bit search kernel

The rmM-tree construction and the `fwdsearch`/`bwdsearch` kernel used C `int` for parenthesis positions, block sizes, block counts, and excess deltas — inherited verbatim from the [improved-octo-waddle](https://github.com/biocore/improved-octo-waddle) port. Above `2³¹` parentheses (~537 M tips) these **silently overflowed**, corrupting navigation rather than raising an error. This widens the loop counters and scan positions to 64-bit (`npy_intp`). The stored index arrays were *already* 64-bit, so the on-disk/in-memory footprint and per-operation runtime are unchanged; only the transient kernel arithmetic was widened.

## Testing

Both changes are verified bit-for-bit identical to the pre-PR backend:

- **52 `BPTree` unit tests** pass (`skbio/tree/tests/bp`), including the compiled `test_bp_cy.pyx` that exercises the `cimport`ed C vtable.
- **627,102 exhaustive equivalence checks** (`BPTree` vs `TreeNode` across 22 operations, balanced/caterpillar/random topologies × 4 seeds) — 0 failures.
- The rmM rewrite was additionally cross-checked against a brute-force O(span) oracle over millions of `(i, j)` pairs spanning varied block layouts and tail-block sizes — 0 disagreements.

## Benchmarks

Measured with the standalone `bptree-benchmarks` suite: each git ref built into an isolated `pip --target` install and timed in a fresh subprocess (`min` of repeated trials), across `balanced` / `caterpillar` / `random` topologies at 1K – 500K tips. Baseline is the pre-PR backend (`bp_tree` @ #2498).

**`lca` between far-apart nodes — old ns → new ns (speed-up):**

| topology | 1K | 100K | 500K |
|---|---|---|---|
| balanced | 7440 → 593 (**12.5×**) | 702K → 1020 (**688×**) | 3.67M → 1004 (**3650×**) |
| caterpillar | 5703 → 639 (**8.9×**) | 520K → 1055 (**493×**) | 2.72M → 1220 (**2233×**) |
| random | 7461 → 626 (**11.9×**) | 693K → 927 (**747×**) | 3.64M → 1072 (**3392×**) |

The old cost grows ~linearly with tree size on all three topologies; the new cost is a flat sub-microsecond regardless of size or shape. `rmq`/`rMq` themselves show the same pattern (up to ~5000–6700× at 500K).

**`height` / `deepest_node` (whole-tree range) — speed-up old → new:**

| topology | 1K | 100K | 500K |
|---|---|---|---|
| balanced | 1.7× | 1.9× | 1.5× |
| caterpillar | 18–21× | ~1300× | **~6000×** |
| random | 1.4–2.0× | 1.6–3.0× | 1.6× |

**Unchanged (controls), old → new ≈ 1.0× on every topology/size:**

- `lca` between *near* nodes: 0.24 µs → 0.23 µs (small range, already cheap).
- Navigation (`parent`, `first_child`, `next_sibling`, `close`, …): within noise.
- Construction (`parse_newick`, `from_treenode`): within noise.
- Memory footprint: unchanged (no stored array changed dtype; the 64-bit widening touches
  only transient kernel arithmetic).

## Scope / follow-up

- A follow-up PR rebalances the Cython/Python boundary of the same backend and stacks on top of this branch.

### DIFF
--- a/skbio/tree/bp/_bp.pyx
+++ b/skbio/tree/bp/_bp.pyx
@@ -550,35 +550,139 @@ cdef class BPTree:
             return self.bwdsearch(i - 1, -2) + 1
 
     cpdef SIZE_t rmq(self, SIZE_t i, SIZE_t j) nogil:
-        """The leftmost minimum excess in i -> j"""
-        cdef:
-            SIZE_t k, min_k
-            SIZE_t min_v, obs_v
+        """The leftmost minimum excess in i -> j.
 
-        min_k = i
-        min_v = self.excess(i)  # a value larger than what will be tested
-        for k in range(i, j + 1):
-            obs_v = self.excess(k)
-            if obs_v < min_v:
-                min_k = k
-                min_v = obs_v
-        return min_k
+        The minimum excess over [i, j] is found in O(log n): the two partial
+        end-blocks are scanned directly and the full interior blocks are
+        covered by a range query over the rmM tree. The leftmost position that
+        attains it is then recovered with a single ``fwdsearch``.
+        """
+        cdef:
+            SIZE_t d_star, p, blk_end, tree_v
+            SIZE_t bi, bj
+            int b
+
+        if i >= j:
+            return i
+
+        b = self._rmm.b
+        bi = i // b
+        bj = j // b
+
+        # smallest absolute excess over [i, j]
+        d_star = self._e_index[i]
+        if bi == bj:
+            for p in range(i + 1, j + 1):
+                if self._e_index[p] < d_star:
+                    d_star = self._e_index[p]
+        else:
+            # remainder of the block containing i (inclusive of i)
+            blk_end = min((bi + 1) * b, self.size)
+            for p in range(i + 1, blk_end):
+                if self._e_index[p] < d_star:
+                    d_star = self._e_index[p]
+            # full interior blocks (bi, bj) via the rmM tree
+            if bi + 1 <= bj - 1:
+                tree_v = self._rmq_tree_min(0, 0, self._rmm.n_internal,
+                                            <int>(bi + 1), <int>(bj - 1))
+                if tree_v < d_star:
+                    d_star = tree_v
+            # leading part of the block containing j (inclusive of j)
+            for p in range(bj * b, j + 1):
+                if self._e_index[p] < d_star:
+                    d_star = self._e_index[p]
+
+        # leftmost position in [i, j] whose excess equals the minimum
+        if self._e_index[i] == d_star:
+            return i
+        return self.fwdsearch(i, <int>(d_star - self._e_index[i]))
 
     cpdef SIZE_t rMq(self, SIZE_t i, SIZE_t j) nogil:
-        """The leftmost maximmum excess in i -> j."""
+        """The leftmost maximmum excess in i -> j.
+
+        Symmetric to :meth:`rmq`: an O(log n) range-maximum over the rmM tree
+        followed by a single ``fwdsearch`` for the leftmost attaining position.
+        """
         cdef:
-            SIZE_t k, max_k
-            SIZE_t max_v, obs_v
+            SIZE_t m_star, p, blk_end, tree_v
+            SIZE_t bi, bj
+            int b
 
-        max_k = i
-        max_v = self.excess(i)  # a value larger than what will be tested
-        for k in range(i, j + 1):
-            obs_v = self.excess(k)
-            if obs_v > max_v:
-                max_k = k
-                max_v = obs_v
+        if i >= j:
+            return i
 
-        return max_k
+        b = self._rmm.b
+        bi = i // b
+        bj = j // b
+
+        # largest absolute excess over [i, j]
+        m_star = self._e_index[i]
+        if bi == bj:
+            for p in range(i + 1, j + 1):
+                if self._e_index[p] > m_star:
+                    m_star = self._e_index[p]
+        else:
+            blk_end = min((bi + 1) * b, self.size)
+            for p in range(i + 1, blk_end):
+                if self._e_index[p] > m_star:
+                    m_star = self._e_index[p]
+            if bi + 1 <= bj - 1:
+                tree_v = self._rmq_tree_max(0, 0, self._rmm.n_internal,
+                                            <int>(bi + 1), <int>(bj - 1))
+                if tree_v > m_star:
+                    m_star = tree_v
+            for p in range(bj * b, j + 1):
+                if self._e_index[p] > m_star:
+                    m_star = self._e_index[p]
+
+        # leftmost position in [i, j] whose excess equals the maximum
+        if self._e_index[i] == m_star:
+            return i
+        return self.fwdsearch(i, <int>(m_star - self._e_index[i]))
+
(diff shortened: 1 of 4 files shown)


monkeytim19:
@qiyunzhu @mortonjt to review

skbio/tree/bp/_bp.pyx
```diff
@@ -33,44 +33,47 @@ BOOL = np.uint8
...
-        cdef int offset  # tip offset in binary tree for a given parenthesis
-        cdef int lower_limit  # the lower limit of the bucket a parenthesis is in
-        cdef int upper_limit  # the upper limit of the bucket a parenthesis is in
-        cdef int min_ = 0 # m, absolute minimum for a blokc
-        cdef int max_ = 0 # M, absolute maximum for a block
-        cdef int excess = 0 # e, absolute excess
+        # i/j and the block bounds index into B[0, 2n), so they are SIZE_t
+        # (64-bit); lvl/pos iterate the small rmM binary tree and stay int.
+        cdef SIZE_t i, j  # parenthesis positions
+        cdef int lvl, pos  # rmM-tree level / node within a level
```
wasade: Why is this not `SIZE_t` as well?

skbio/tree/bp/_bp.pyx
```diff
@@ -550,35 +553,139 @@ cdef class BPTree:
...
+
+        if i >= j:
+            return i
+
+        b = self._rmm.b
+        bi = i // b
+        bj = j // b
+
+        # smallest absolute excess over [i, j]
+        d_star = self._e_index[i]
```
wasade: We still are caching excess, which can be computed. It's separate in scope from the O(log n) work here, but relevant as if I remember right, it should be computable from the structures inside the RMM object.

wasade: `self._e_index[i]` is used multiple times, can this be placed into a local variable?

skbio/tree/bp/_bp.pyx
```diff
@@ -550,35 +553,139 @@ cdef class BPTree:
...
-            if obs_v > max_v:
-                max_k = k
-                max_v = obs_v
+        b = self._rmm.b
+        bi = i // b
+        bj = j // b
 
-        return max_k
+        # largest absolute excess over [i, j]
+        m_star = self._e_index[i]
```
wasade: Same here on reuse of `self._e_index[i]`

skbio/tree/bp/_bp.pyx
```diff
@@ -550,35 +553,139 @@ cdef class BPTree:
...
+            for p in range(bj * b, j + 1):
+                if self._e_index[p] > m_star:
+                    m_star = self._e_index[p]
+
+        # leftmost position in [i, j] whose excess equals the maximum
+        if self._e_index[i] == m_star:
+            return i
+        return self.fwdsearch(i, m_star - self._e_index[i])
+
+    cdef SIZE_t _rmq_tree_min(self, int node, int node_lo, int node_hi,
```
wasade: Is there a reason these need to be `int` rather than `SIZE_t`?

wasade (changes requested):
Just some minor change requests. This is very cool!

qiyunzhu:
Looks awesome! Very major improvement.

skbio/tree/bp/_bp.pyx
qiyunzhu: I agree that `SIZE_t` should be used here, since `i` and `j` are indices and cannot be negative. I am further recommending using `SIZE_t` in any places where we know the variable is never negative (e.g., `n_tip` and `height`).

Additionally, I think that the custom type `SIZE_t` itself is outdated and can be safely replaced with the built-in type `Py_ssize_t`, which matches `np.intp`. The word `SIZE_t` may raise confusion because C type `size_t` is unsigned whereas `Py_ssize_t` is signed.

monkeytim19:
# Added Changes

**Type consistency (signed-index widening)**
- Widened non-negative index/count variables from C `int` to the signed pointer-width type
- Dropped the now-redundant `<int>` casts at the `_rmq_tree_` call and converted the three `<int>(pow(2, height) − 1)` casts to the signed type (kept pow).
- Removed a dead `vbar` declaration.
- Replaced the `INT_MAX` with `PY_SSIZE_T_MAX`, so the "+∞" sentinel can't be exceeded by a real excess value now that the kernel is 64-bit.

**Local-variable reuse**
- Assigned `self._e_index[i]` into a local variable `e_i` in both `rmq` and `rMq`.

**`SIZE_t` → `Py_ssize_t` rename**
- Removed the custom `ctypedef cnp.npy_intp SIZE_t` and replaced every use with the built-in `Py_ssize_t` across _bp.pyx, _bp.pxd, _bp_binary_tree.pxd. 

_Performance was benchmarked the same way as it was done initially and no regression was observed._

wasade:
Good catch. Cython moved a while ago from SIZE_t to Py_ssize_t, forgot about that.

qiyunzhu:
@monkeytim19 and I discussed offline. We think that all comments have been addressed and this PR is good to be merged. We also identified some Cython code hygiene issues that can be fixed in the future, such that Cython won't need to fallback to Python and extra performance can be squeezed.

PR #2546, opened 2026-08-26
BUG: pass permdisp's seed down to pcoa

LarytheLord:
### Description

`permdisp` accepts a `seed` and its docstring states that the seed is there to make the output deterministic, but the internal `pcoa` call does not forward it.

With `method="eigh"` this is invisible, since that solver is deterministic and never touches the seed. With `method="fsvd"` the solver draws a random projection from an unseeded generator, so each call produces a different ordination and therefore a different statistic and p-value, even when the caller passes a seed. On a 60 by 60 matrix, repeated calls with the same seed disagree on every run. On larger inputs the movement is big enough to matter: on a 300 sample matrix I saw the p-value range from 0.01 to 0.70 across repeated calls with an identical seed.

The fix forwards the seed. I checked that this does not disturb the `eigh` path: `pcoa` only consumes the seed inside `_fsvd`, so passing a generator through leaves it untouched on the `eigh` path, and the statistic and p-value for an `eigh` call are identical before and after this change. The `fsvd` path is the only behavior that changes, and it had no reproducible behavior to preserve.

### Testing

The existing `fsvd` tests pass despite the bug because their distance matrix is nine by nine, which is small enough that every random projection recovers the same subspace, so the result looks stable. The new test uses a matrix large enough for the projections to actually disagree, and asserts that two calls with the same seed return the same statistic.

It runs with `permutations=0`, because the observed statistic alone depends on the ordination and the permutation loop only adds runtime. I confirmed the test is a real guard rather than a passing assertion: with the fix reverted it fails on all ten of ten runs, and it covers both the centroid and the median test.

```
pytest skbio/stats/distance/tests/test_permdisp.py    # 22 passed
```

The full suite shows no change against the baseline on this branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 ### Bug Fixes
 
+* Fixed `permdisp` not passing its `seed` down to `pcoa`, which left `method="fsvd"` unseeded. Because the FSVD solver draws a random projection, results were not reproducible even when a seed was given; on larger matrices repeated calls with the same seed could return materially different p-values. `method="eigh"` was unaffected, being deterministic.
 * Fixed an unexpected behavior in differential abundance tests (`ancombc` and `dirmult_lme`) where string columns in the metadata that can be cast into numbers (e.g., `["1", "2", "3"]`) were treated as numerical. Now they are treated as categories ([#2539](https://github.com/scikit-bio/scikit-bio/pull/2539)).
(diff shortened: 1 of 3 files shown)


qiyunzhu:
Oh this is indeed a simple but important bug fix!

sfiligoi (approved):
Simple and elegant fix.

PR #2552, opened 2026-08-28
BUG: raise ValueError for non-positive max_iter in optspace

l1joseph:
## Summary

Calling `optspace` with `max_iter=0` (or a negative value) raised `UnboundLocalError: cannot access local variable 'S' where it is not associated with a value` instead of a clear error. `S` is first assigned inside the main iteration loop, so a `max_iter` that produces zero iterations leaves it unbound by the time the function builds `X_hat` from `U @ S @ V.T`. `rpca` passes `max_iter` straight through to `optspace`, so the same crash was reachable from the public entry point. This was flagged during review of an earlier version of this branch and had survived a few rounds of revision.

`optspace` now validates `max_iter` alongside its other parameter checks (`method`, `dimensions`) and raises `ValueError` for `max_iter < 1`, with a matching docstring `Raises` entry added to both `optspace` and `rpca`.

## Tests

- `test_zero_max_iter_error` and `test_negative_max_iter_error` in `test_optspace.py`, asserting `ValueError` for `max_iter=0` and `max_iter=-1`.
- The same two cases added to `test_rpca.py` for the pass-through path.

## Checklist

### DIFF
--- a/skbio/stats/ordination/_optspace.py
+++ b/skbio/stats/ordination/_optspace.py
@@ -458,7 +458,7 @@ def optspace(X, dimensions=3, max_iter=10000, tol=1e-5, method="GD"):
     elif method not in ("GN", "GD"):
         raise ValueError("Method must be 'GN' or 'GD'")
 
-    elif max_iter < 1:
+    elif type(max_iter) is not int or max_iter < 1:
         raise ValueError("Max_iter must be a positive integer")
 
     # Create observed mask
--- a/skbio/stats/ordination/_optspace.py
+++ b/skbio/stats/ordination/_optspace.py
@@ -369,6 +369,10 @@ def optspace(X, dimensions=3, max_iter=10000, tol=1e-5, method="GD"):
         The rank of the matrix to recover. Default is 3.
     max_iter : int, optional
         Maximum number of iterations. Default is 10000.
+
+        .. versionchanged:: 0.7.4
+            A non-positive or non-integer value now raises ``ValueError``
+            instead of failing with an unrelated error partway through.
     tol : float, optional
         Convergence tolerance. Default is 1e-5.
     method : {'GD', 'GN'}, optional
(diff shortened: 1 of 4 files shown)


skbio/stats/ordination/_optspace.py
```diff
@@ -456,6 +458,9 @@ def optspace(X, dimensions=3, max_iter=10000, tol=1e-5, method="GD"):
     elif method not in ("GN", "GD"):
         raise ValueError("Method must be 'GN' or 'GD'")
 
+    elif type(max_iter) is not int or max_iter < 1:
```
qiyunzhu: This will compare False if `max_iter` is a NumPy integer etc. I think that NumPy can provide a more robust test. For example `np.issubdtype(type(max_iter), np.integer)`.

qiyunzhu (commented):
I think this is a very straightforward fix. But the current module (optspace) may not be the best place to host the fix. `max_iter` is a standard scikit-learn style parameter. Other scikit-bio functions may use it too. If we really need to validate it, we can add a helper function to somewhere like `skbio.util`. Furthermore, I feel that it is not necessary to fix it, letting alone having three unit tests in optspace explicitly for this matter. If the docstring already says `max_iter` is a positive integer, the user should be responsible for setting it correctly. Fixing it is defensive coding / over engineering.

l1joseph:
Fixed to use `np.issubdtype(type(max_iter), np.integer)` instead of `type() is not int`, so NumPy integer types are accepted now. Added `test_numpy_integer_max_iter_accepted`, 34/34 pass.

Should I move the `max_iter` validation to a shared helper instead, or leave it scoped to `optspace`?

qiyunzhu:
Hmmm, I think let's leave it in optspace and merge for now. I can think of the scope later, once there are multiple functions with `max_iter`.

PR #2553, opened 2026-08-28
MAINT: cache the center_distance_matrix Numba kernels and document the symmetry requirement

l1joseph:
## Summary

`e_matrix_means_nb` and `f_matrix_inplace_nb` (added in #2508) were the only Numba kernels in the codebase without `cache=True`; every other kernel (`_permanova.py`, `_mantel.py`, `_unifrac.py`, `_permdisp.py`, `_mmvec.py`) sets it. Neither kernel closes over anything but its own array and scalar arguments, so Numba's on-disk cache applies without complication. Added `cache=True` to both.

`f_matrix_inplace_nb` also has an undocumented constraint carried over from its Cython counterpart, `f_matrix_inplace_cy`, but dropped somewhere in the port: it approximates the column mean of column `col` with `row_means[col]`, the row mean of row `col`, which only equals the true column mean when the input is symmetric. For a non-symmetric matrix this does not raise, it silently returns an incorrectly centered result. `_cutils.pyx` already documents this constraint on the Cython side; I restored the equivalent note on the Numba kernels and on the public `center_distance_matrix_nb` entry point's `mat` parameter.

## Tests

Ran `test_util.py` and `test_principal_coordinate_analysis.py` (45 passed, 10 subtests) twice, including `test_center_distance_matrix_numba_matches_cython` and `test_engine_numba_matches_cython`, to exercise the on-disk cache path introduced by `cache=True` and confirm it doesn't change results. Re-verified against the current branch tip after `development_numba_gpu` was rewritten upstream.

## Checklist

### DIFF
--- a/skbio/stats/ordination/_center_distance_matrix_numba.py
+++ b/skbio/stats/ordination/_center_distance_matrix_numba.py
@@ -25,13 +25,18 @@ except ImportError:
 
 if NUMBA_AVAILABLE:
 
-    @njit(parallel=True)
+    @njit(parallel=True, cache=True)
     def e_matrix_means_nb(mat, centered, row_means):
         """Apply the E-matrix transform and collect row/global means in one pass.
 
         Writes ``-0.5 * d**2`` into ``centered`` and returns the global mean of
         that transformed matrix. Values are promoted to float64 before squaring
         to match the Cython ``double`` accumulator (matters for float32 input).
+
+        ``row_means`` is the mean of each *row* of ``centered``; it is not
+        also the column means unless ``mat`` is symmetric. This function does
+        not require symmetry on its own, but its output feeds directly into
+        ``f_matrix_inplace_nb``, which does.
         """
         n = mat.shape[0]
         global_sum = np.float64(0.0)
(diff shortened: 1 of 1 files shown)


skbio/stats/ordination/_center_distance_matrix_numba.py
```diff
@@ -25,13 +25,18 @@
 
 if NUMBA_AVAILABLE:
 
-    @njit(parallel=True)
+    @njit(parallel=True, cache=True)
```
sfiligoi, line 28: Numba cache is too risky for production code.
We should not force it. Certainly not on function-by-function manner.

skbio/stats/ordination/_center_distance_matrix_numba.py
```diff
@@ -50,12 +55,19 @@ def e_matrix_means_nb(mat, centered, row_means):
 
         return (global_sum / np.float64(n)) / np.float64(n)
 
-    @njit(parallel=True)
+    @njit(parallel=True, cache=True)
```
sfiligoi, line 58: Same here.
We don't want caching.

sfiligoi:
@l1joseph Thanks for the documentation clarification.

However, this PR cannot be merged for two reason:
a) Numba caching is potentially dangerous, so we do not want caching to be forced in a production library code.
b) `development_numba_gpu` has already been merged, an no additional PR against it will be considered.

Thus, please open a documentation-only PR against `main`.

PR #2554, opened 2026-08-28
BUG: make the Gauss-Newton Jacobian entry-indexed and its adjoint exact

l1joseph:
## Summary

#2512 added a projection onto the complement of the S tangent space by calling `_solve_S` from inside `jacobian_UV` and `jacobian_UV_adj`. `_solve_S` runs its own `scipy.sparse.linalg.lsmr` call whose matvec, `jacobian_S`, builds the dense m x n product `U @ S @ V.T`. So every outer-loop LSMR matvec in the Gauss-Newton step now triggers a nested LSMR solve, and each iteration of that nested solve reconstructs a dense array from a handful of observed entries. `np.where(observed_mask)` is also recomputed separately in `jacobian_UV`, `jacobian_UV_adj`, and `_solve_S`, three times per matvec.

The S-projection doesn't need an iterative solve. It's a projection onto `range(J_S)` complement, and `J_S` is only N x r^2 (N the number of observed entries, 9 columns at rank 3), so a one-time economy QR of `J_S` per outer Gauss-Newton iteration replaces the nested LSMR entirely: the projection becomes `w - Q @ (Q.T @ w)` and the solve becomes back-substitution against `R`. `jacobian_S`, `jacobian_S_adj`, `jacobian_UV`, and `jacobian_UV_adj` are rewritten to index directly into the observed rows and columns instead of materializing dense products, and `rows`, `cols`, `U @ S`, `S @ V.T`, and the QR factors are hoisted out of the per-matvec closures into `solve_gauss_newton_step` and `optspace`, computed once per outer iteration instead of once per matvec.

The rewrite also turned up a real correctness gap in the current code, not just a performance one: the outer LSMR call assumes `matvec` and `rmatvec` are an exact adjoint pair, and the nested-LSMR projection was breaking that assumption. On a random test pair the old code is off by 1.7e-4; the new exact projection is off by 3.6e-15. On a converged fixture the two solvers still agree (max abs difference 1.76e-6, relative Frobenius 1.04e-7), but on a non-converged HDLSS case (50 samples x 398 features, 10% observed, max_iter=20) they diverge by O(1), and the new trajectory is the one that doesn't diverge (masked SSE 4.30e3 vs 6.13e3, ||X_hat||_F 43 vs 329 for the old code).

A, the N x r^2 basis for the S-projection, can be rank-deficient. I used a pivoted economy QR and drop columns where |R_ii| is below 1e-12 times |R_00| before back-substituting, rather than falling back to a separate least-squares solve, since it reuses the same Q the projection already needs.

## Benchmarks

Rank 3, 5 outer iterations, before vs after, same machine:

| case | method | before | after | speedup |
|---|---|---|---|---|
| 50 x 200 @ 10% | GN | 2.26 s | 0.54 s | 4.2x |
| 50 x 1000 @ 5% | GN | 9.24 s | 1.82 s | 5.1x |
| 50 x 1000 @ 10% | GN | 5.08 s | 1.66 s | 3.1x |
| 100 x 2000 @ 5% | GN | 29.32 s | 3.01 s | 9.7x |
| 100 x 2000 @ 5% | GD | 0.15 s | 0.05 s | 3.0x |

GD benefits too, because `_compute_obj` calls `_solve_S` inside the line search, up to 16 times per call.

## Tests

- `TestJacobian`: `jacobian_S`/`jacobian_S_adj` against the dense reference (5.6e-17 / 3.3e-16), the adjoint dot-product identity for `jacobian_UV`/`jacobian_UV_adj` (new pair 3.6e-15, old pair 1.7e-4), and idempotence/orthogonality of the S-complement projection.
- `TestRankDeficient`: three cases that trigger the pivoted-QR fallback.
- `TestSolverAgreement`: end-to-end `method="GN"` before/after on the existing fixture.
- `test_gauss_newton`, `test_basic_completion`, and `test_gradient_descent` pass unchanged, no tolerances loosened.
- Full `skbio/stats/ordination/` suite: 193 passed, 5 skipped.

Based directly on the current tip of #2512 (`rpca_dev` at ea757faa), re-verified unchanged as of this PR.

## Checklist

### DIFF
--- a/skbio/stats/ordination/_optspace.py
+++ b/skbio/stats/ordination/_optspace.py
@@ -35,7 +35,7 @@ References
 from warnings import warn
 
 import numpy as np
-from scipy.linalg import svd
+from scipy.linalg import svd, qr, solve_triangular
 from scipy.sparse.linalg import svds, lsmr, LinearOperator
 
 
@@ -116,7 +116,83 @@ def _svd_init(X_trimmed, r):
     return U, V
 
 
-def jacobian_S(U, V, S, rows, cols):
+def _obs_basis(Ui, Vj, tol=1e-12):
+    """Factorize the S-Jacobian restricted to the observed entries.
+
+    ``J_S`` is the linear map ``dS -> P_\\Omega(U dS V^T)``. Written as a matrix
+    acting on ``dS.ravel()`` it has one row per observed entry ``k = (i, j)``,
+
+    A[k, a * r + b] = U[i, a] * V[j, b],
+
+    i.e. ``A[k] = kron(U[i], V[j])``, which matches the row-major
+    ``s.reshape(r, r)`` convention used throughout this module. ``A`` is
+    ``(n_observed, r ** 2)``, so it is thin (only 9 columns at the default rank
+    of 3) and can be factorized once per outer iteration and reused by every
+    matrix-vector product.
+
+    The economy QR factorization ``A[:, perm] = Q R`` yields both operations
+    needed downstream: ``Q`` is an orthonormal basis of ``range(J_S)``, and
+    ``R`` back-substitutes the least-squares solve for ``S``.
+
+    Column pivoting is used so that rank deficiency (possible when the observed
+    entries do not determine all ``r ** 2`` degrees of freedom of ``S``) is
+    detected from the magnitudes of the diagonal of ``R``, which pivoting orders
+    non-increasingly. Trailing columns whose pivot falls below ``tol`` relative
+    to the leading pivot are dropped, giving the basic (minimum-column-support)
+    least-squares solution rather than a division by a near-zero pivot.
+
+    Parameters
+    ----------
+    Ui : ndarray of shape (n_observed, r)
+        Rows of ``U`` gathered at the observed row indices.
+    Vj : ndarray of shape (n_observed, r)
+        Rows of ``V`` gathered at the observed column indices.
+    tol : float, optional
+        Relative pivot threshold for rank detection.
+
+    Returns
+    -------
+    Q : ndarray of shape (n_observed, rank)
+        Orthonormal basis of ``range(J_S)``.
+    R : ndarray of shape (rank, rank)
+        Upper triangular factor of the retained columns.
+    perm : ndarray of shape (rank,)
+        Indices of the retained columns of ``A``.
+    """
+
+    r = Ui.shape[1]
+
+    # A[k, a * r + b] = Ui[k, a] * Vj[k, b]
+    A = (Ui[:, :, None] * Vj[:, None, :]).reshape(-1, r**2)
+
+    Q, R, perm = qr(A, mode="economic", pivoting=True)
+
+    # Pivoting orders |diag(R)| non-increasingly, so the numerical rank is the
+    # length of the leading run of pivots above the relative threshold.
+    pivots = np.abs(np.diag(R))
+    if pivots.size == 0 or pivots[0] == 0.0:
+        rank = 0
+    else:
+        rank = int(np.count_nonzero(pivots > tol * pivots[0]))
+
+    return Q[:, :rank], R[:rank, :rank], perm[:rank]
+
+
+def _project_S_complement(Q, w):
+    """Project w onto the orthogonal complement of the S tangent space.
+
+    ``Q`` is an orthonormal basis of ``range(J_S)`` as returned by
+    :func:`_obs_basis`, so the projector onto the complement is
+
+    P_\\perp w = w - Q (Q^T w).
+
+    This is exact and costs ``O(n_observed * r ** 2)``.
+    """
+
+    return w - Q @ (Q.T @ w)
+
+
+def jacobian_S(Ui, Vj, S):
     """Compute J_S(dS).
 
     The Jacobian is
@@ -127,13 +203,18 @@ def jacobian_S(U, V, S, rows, cols):
     the reconstruction error over the observed entries.
 
     J_S(dS) = P_\\Omega(U dS V^T).
+
+    Evaluated only at the observed entries, this is
+
+    J_S(dS)[k] = U[i] dS V[j]^T,
+
+    which avoids forming the dense ``(m, n)`` product.
     """
 
-    W = U @ S @ V.T
-    return W[rows, cols]
+    return (Ui @ S * Vj).sum(axis=1)
 
 
-def jacobian_S_adj(U, V, w, observed_mask):
+def jacobian_S_adj(Ui, Vj, w):
     """Compute J_S*(W).
 
     The Jacobian adjoint is defined with respect to the inner product by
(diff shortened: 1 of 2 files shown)


qiyunzhu (commented):
Looks decent to me, though I cannot validate the maths. This PR replaces the iterative LSMR function with an exact solver. This is usually a good move. Calling @aj-pinto to comment on the math. I want to note two things:

1. In some cases an exact solver may cost more memory than an iterative approximaton method. It will be helpful to profile the memory. As I eyeball, multiple three matrix multiplications were replaced with simpler math, so the total memory footprint should be smaller already.

2. As you noted there are specific treatment of rank-deficient cases. Such challenges are common in omics data. Scaling up rank-deficient tests may be useful. For example, a feature with only zeros or a constant number; two features that are identical or proportional to each other (colinearity). My guess is that the current PR handles rank deficency better. Functional test is sufficient, no need to scale up the unit test.

l1joseph:
Renamed `R` to `residual` in `solve_gauss_newton_step`. `Q` is a genuine QR basis there, so `R` next to it for the residual was confusing.

Peak RSS for `optspace(method="GN")`, old (dense nested-LSMR, pre-PR) vs new (this PR), two runs each on the smaller two sizes:

| size | old | new |
|---|---|---|
| 50x1000 @10% | 4.10 / 5.30 MB | 5.12 / 6.14 MB |
| 200x5000 @5% | 113.40 / 112.81 MB | 124.31 / 119.38 MB |
| 500x10000 @2% (max_iter=1) | 446.21 MB | 427.12 MB |

Not what I expected. New is slightly higher at the two smaller sizes, only lower at the largest, by about 4%. My guess is the new version holds several basis arrays alive across the whole outer loop through closures, while the old dense reconstructions were transient, so cutting the dense reconstruction doesn't automatically win on peak RSS the way it wins on wall clock.

qiyunzhu:
@l1joseph Thanks! So the memory regression is marginal or negative on large datasets, which is quite acceptable I think.

aj-pinto:
@l1joseph 

These results look good! I made a similar attempt at computing the exact inverse of the S-Jacobian, but I ran into an issue with the Kronecker product where the einsum function could form potentially massive intermediate arrays. I'm not sure if this is actually true, and since your results don't seem excessively large, it may not be an issue. 

Have you stress-tested this with higher values of n_observed, e.g., a million? In your largest trial (500 x 10k with 2%), you had n_observed = 100k, which seems large enough for most reasonable applications, but I'm still curious if the memory will blow up with more observed entries. In the worst case, we don't want the algorithm to crash if someone tries to impute only a few entries of an otherwise large matrix.

l1joseph:
`jacobian_S_adj`'s einsum (`"k,ka,kb->ab"`) doesn't form a large intermediate. It's a two-index contraction over `k`, which resolves to a GEMM-like reduction, not a materialized `(n_observed, r, r)` array.

`_obs_basis`'s `A` does scale with `n_observed` though, since that's what the pivoted QR factors. Stress-tested `_obs_basis` directly up to 10M observed entries:

| n_observed | A (theoretical) | peak delta |
|---|---|---|
| 100k | 7.2 MB | 23.3 MB |
| 1M | 72.0 MB | 212.3 MB |
| 5M | 360.0 MB | 1057.9 MB |
| 10M | 720.0 MB | 2112.7 MB |

Scales linearly, no blowup, but real usage runs about 3x the theoretical size of `A` alone, most likely `scipy.linalg.qr`'s pivoted path making an internal Fortran-order copy for LAPACK. At 10M observed entries (a dense matrix around 3162x3162, or larger at moderate density), that's about 2GB for this step alone.

aj-pinto:
@l1joseph 

I see, so constructing A in _obs_basis does explicitly form an n_observed x r^2 matrix. This can greatly surpass the memory requirements of storing the original matrix, right? For example, consider a 5k x 5k matrix, 40% observed, with rank 5. It has 25M entries, but 10M are observed. If we explicitly construct A, that forms an array with 250M entries. This array is also reconstructed within every iteration. 

To avoid this, the code currently uses iterative LSMR solvers on the J_S operator. Each solve is actually very quick relative to the Gauss-Newton LSMR solve. However, it performs the same computations multiple times, which we would ideally want to avoid.

We might be able to avoid some unnecessary computation by assembling A^T A rather than A. This array is only r^2 x r^2, which is much better than n_observed x r^2. To compute the pseudoinverse (_solve_S) of some vector b of size n_observed, we can then take (A^T A)^-1 J_S* (b). In this case, pre-composing with J_S* may be a necessary evil to avoid forming the actual r^2 x n_observed pseudoinverse. The greatest benefit of this is being able to invert A^T A once, rather than performing LSMR on b multiple times per iteration as the code currently does. Still, forming A^T A without forming A could be a challenge.

l1joseph:
Kronecker guess turned out to be off by O(1). Here are two alternatives, which both build the r²×r² gram matrix without forming A:

- **Pairwise loop**: `O(n_observed)` memory, `O(n_observed·r⁴)` time.
- **Chunked accumulation**: process in blocks, accumulate `G += A_chunk.T @ A_chunk`. `O(B·r²)` memory (tunable), `O(n_observed·r²)` time, same as now. This one's better.

Both verified to 1e-13/1e-14 vs `lstsq` and the current QR solve.

Two caveats. `κ(A^TA) = κ(A)²`, so conditioning is worse than QR on poorly-conditioned inputs. And in the rank-deficient case it gives the minimum-norm solution instead of the current "basic" (zeroed-coefficient) one. Same residual, just a different point, and `TestRankDeficient` currently asserts the zeroed-coefficient behavior.

aj-pinto:
@l1joseph 

Nice! The chunked accumulation approach seems like the best. Have you numerically verified the theoretical memory and time scaling?

It is true that the conditioning would be worse in theory. This could have consequences for numerical stability, so we would want to stress-test this at some point.

A compromise could be to preliminarily assess the memory requirements of forming A based on the size of the original input. Then, we could decide which method to use based on a specified cutoff for the size of A. For smaller problems, the stability of the QR method may almost always be preferable. For larger problems, which are generally more stable, we may be more comfortable sacrificing the poorer conditioning of A^T A for the memory it saves. Although, we would want to characterize the stability of both methods before pursuing this.

l1joseph:
Verified both.

Scaling (n_observed vs peak memory / time, block=200k):

| n_observed | chunked mem | QR mem | chunked time | QR time |
|---|---|---|---|---|
| 100k | 7.2 MB | 17.3 MB | 0.03s | 0.16s |
| 1M | 27.6 MB | 193.0 MB | 0.09s | 0.60s |
| 5M | 28.3 MB | 1028.2 MB | 0.49s | 17.08s |
| 10M | 28.5 MB | 2081.5 MB | 0.91s | 15.34s |

Chunked memory stays flat regardless of n_observed, as expected. It's also faster at every size, 5x to 35x, not just cheaper.

Stability, controlling κ(A) directly:

| κ(A) | QR rel err | Gram rel err |
|---|---|---|
| 1 | 1.1e-15 | 2.0e-15 |
| 1e2 | 9.4e-16 | 8.7e-14 |
| 1e4 | 2.8e-15 | 9.2e-12 |
| 1e6 | 1.1e-15 | 1.2e-10 |
| 1e8 | 1.1e-15 | 1.0 (fails) |

QR stays at machine precision regardless of κ(A). Gram degrades as κ² exactly as predicted, and fails completely by κ(A)=1e8.

I think we should change what the hybrid approach keys on. A size cutoff picks the wrong variable: a large, well-conditioned problem is fine on Gram, and a small, poorly-conditioned one can already be failing. Chunked accumulation produces `G` (`r²×r²`) directly, so checking `cond(G)` is nearly free, `O(r⁶)`, and can be done every time instead of decided upfront by size. Proposal: always accumulate `G` via chunking, use it if `κ(A) < 1e3` or so (well before the 1e-12 error zone), fall back to full pivoted QR only on the rare ill-conditioned case.

PR #2555, opened 2026-08-29
DOC: document the symmetry requirement of the center_distance_matrix Numba kernels

l1joseph:
## Summary

f_matrix_inplace_nb approximates the column mean of column col with row_means[col], the row mean of row col, rather than computing it separately. That only equals the true column mean when the input is symmetric. For a non-symmetric matrix this does not raise, it silently returns an incorrectly centered result. The Cython counterpart, f_matrix_inplace_cy, already documents this constraint; the Numba port dropped it. This restores the equivalent note on both Numba kernels and on the public center_distance_matrix_nb entry point's mat parameter.

## Tests

Docstring change only, no test changes needed.

## Checklist
- [ ] I have documented all public-facing changes in the changelog. (docstring-only, no public API or behavior change)

@sfiligoi


sfiligoi (approved):
Looks great

PR #2557, opened 2026-08-30
PERF: extend the PERMANOVA row-tile Numba kernel to condensed matrices

l1joseph:
The numba engine for `permanova` only used the fast single-pass row-tile kernel for full, non-condensed distance matrices. Condensed input fell back to the slower per-permutation kernel instead. This adds a condensed form of the row-tile kernel, reading distances from the 1-D condensed array through the standard condensed-index mapping instead of a full 2-D matrix, and removes the check that was routing condensed matrices to the slow path. `engine="numba"` now takes the same fast path regardless of matrix form. The two row-tile kernels share a `_condensed_row_base` helper for the index mapping instead of duplicating it inline.

`_compute_f_stat`'s two `engine == "numba"` branches are no longer reachable from `permanova()` for either matrix form now that both go through the row-tile path. They're left in place, documented as such, since they're still directly unit-tested reference kernels (`test_sW_full_nb`/`test_sW_condensed_nb`).

Benchmark, condensed `DistanceMatrix`, `permutations=999`:

| n | cython | numba | speedup |
|---|---|---|---|
| 50 | 36.4ms | 2.8ms | 13.1x |
| 200 | 46.6ms | 5.3ms | 8.7x |
| 500 | 78.9ms | 11.8ms | 6.7x |

JIT warm up is a one time cost, under a second on first call, not included above.

Tests:
- `test_permanova_condensed_nb_matches_cython`, deterministic parity vs cython
- `test_permanova_rowtile_condensed_nb_stat_matches_cython`
- `test_permanova_rowtile_condensed_nb_matches_full_nb`, condensed and full agree under the same seed
- `test_permanova_rowtile_condensed_nb_crosses_chunk_boundary`, exercises the driver's internal chunking
- `test_permanova_rowtile_condensed_nb_odd_n`, odd sample size (n=51), no prior test covered this

Checklist:

### DIFF
--- a/skbio/stats/distance/_permanova.py
+++ b/skbio/stats/distance/_permanova.py
@@ -274,7 +274,17 @@ if NUMBA_AVAILABLE:
                         rsum[p] * inv_group_sizes[groupings_T[mirror_row, p]]
                     )
 
-    @njit(parallel=True, cache=True)
+    @njit(inline="always")
+    def _condensed_row_base(row, n):
+        """Flat offset such that condensed_matrix[base + col] is (row, col).
+
+        Only valid for col > row. Numba inlines this at both call sites in
+        ``_permanova_f_stat_sW_rowtile_condensed_nb`` below, so factoring it
+        out costs nothing at runtime.
+        """
+        return row * n - ((row + 2) * (row + 1)) // 2
+
+    @njit(parallel=True)
     def _permanova_f_stat_sW_rowtile_condensed_nb(
         condensed_matrix, groupings_T, inv_group_sizes, partials
     ):
@@ -299,8 +309,9 @@ if NUMBA_AVAILABLE:
         inv_group_sizes : np.ndarray, shape (num_groups,)
             Reciprocal group sizes (``1.0 / group_sizes``).
         partials : np.ndarray, shape (n // 2, C) with C >= n_perm
-            Reused scratch; each iteration zeros and fills the first
-            ``n_perm`` entries of its own row.
+            Reused scratch; each iteration fills the first ``n_perm`` entries
+            of its own row (no separate zero pass needed -- the first write
+            per row is an assignment, not an accumulate).
 
         """
         k = condensed_matrix.shape[0]
@@ -310,13 +321,11 @@ if NUMBA_AVAILABLE:
 
         for row_idx in prange(n_half):
             local_sW = partials[row_idx]
-            for p in range(n_perm):
-                local_sW[p] = 0.0
             rsum = np.empty(n_perm, np.float64)
 
             for p in range(n_perm):
                 rsum[p] = 0.0
-            base = row_idx * n - ((row_idx + 2) * (row_idx + 1)) // 2
+            base = _condensed_row_base(row_idx, n)
             for col in range(row_idx + 1, n):
                 val = np.float64(condensed_matrix[base + col])
                 val2 = val * val
@@ -325,13 +334,13 @@ if NUMBA_AVAILABLE:
                         groupings_T[col, p] == groupings_T[row_idx, p]
                     )
             for p in range(n_perm):
-                local_sW[p] += rsum[p] * inv_group_sizes[groupings_T[row_idx, p]]
+                local_sW[p] = rsum[p] * inv_group_sizes[groupings_T[row_idx, p]]
 
             mirror_row = n - row_idx - 2
             if mirror_row != row_idx:
                 for p in range(n_perm):
                     rsum[p] = 0.0
-                mbase = mirror_row * n - ((mirror_row + 2) * (mirror_row + 1)) // 2
+                mbase = _condensed_row_base(mirror_row, n)
                 for col in range(mirror_row + 1, n):
                     val = np.float64(condensed_matrix[mbase + col])
                     val2 = val * val
@@ -366,7 +375,8 @@ if NUMBA_AVAILABLE:
         Parameters
         ----------
         distmat : DistanceMatrix
-            Full (non-condensed) distance matrix.
+            Full or condensed distance matrix; dispatches to the matching
+            row-tile kernel based on ``distmat._flags["CONDENSED"]``.
         grouping : np.ndarray, shape (n,)
             Integer group labels (0-indexed).
         group_sizes : np.ndarray, shape (num_groups,)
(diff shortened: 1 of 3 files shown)


sfiligoi:
@l1joseph Please provide benchmarks for at least 5k^2 and 25^k DMs.

Please also disclose the CPU you tested this on (e.g. number of cores)

LarytheLord:
I checked this against a from scratch brute force reference that I wrote independently, not derived from the PR or its own tests, and against both existing trusted Cython kernels, the full matrix one and the condensed one. All four agree across fifty randomized trials with n from 3 to 79 and up to five groups, and I also ran it separately at n=500, 2000, 3001 (odd) and 5000, all matching to floating point noise. Ran the full test suite locally too, 44 out of 44 pass on the permanova file and no regressions against the baseline on the full suite.

Worth mentioning since it is a nice side effect of the change. The old condensed numba kernel had an inconsistency where the mirror row branch never promoted to float64 before squaring while the primary branch did, so a float32 input ended up accumulating half its sum in float32 precision and half in float64. This kernel promotes both branches the same way, and since it becomes the only reachable path for engine="numba" on condensed matrices, that inconsistency goes away with it.

Only open item from my side is the 5k and 25k benchmarks Igor already asked for above. Everything else checks out to me.

sfiligoi:
@LarytheLord Thanks for the review, and great catch about the cython implementation.

@l1joseph Do you want to fix the cython issue in this PR, too, or do we let Abid fix it as a separate PR?

l1joseph:
Benchmarks at larger sizes (condensed `DistanceMatrix`, `permutations=999`, run in isolation on a dedicated AMD MI300A node to avoid noise):

| n | cython | numba | speedup |
|---|---|---|---|
| 5000 | 3.68s | 0.47s | 7.8x |
| 25000 | 105.1s | 10.5s | 10.0x |

l1joseph:
@sfiligoi The actual Cython kernel (`permanova_f_stat_sW_condensed_cy` in `_cutils.pyx`) doesn't have this issue. `val` is declared `cdef double` once and reused identically in both the primary and mirror-row loops, so both branches upcast from the fused-type input to double on every read, consistently. The asymmetry Abid found was specific to the old, now-unreachable Numba reference kernel this PR retires, not the Cython source it was ported from.

skbio/stats/distance/_permanova.py
```diff
@@ -631,10 +716,21 @@ def permanova(
...
     sample_size, num_groups, distance_matrix, group_sizes, s_T, engine, grouping
 ):
-    """Compute PERMANOVA pseudo-F statistic."""
+    """Compute PERMANOVA pseudo-F statistic.
+
+    Only reachable with ``engine == "cython"``: ``permanova()`` routes every
+    ``engine == "numba"`` call through the row-tile path
+    (``_run_permanova_rowtile_nb``) instead, for both full and condensed
+    matrices, so the ``engine == "numba"`` branches below can never execute
+    through the public API. They're kept, along with
```
sfiligoi: Why do we care to keep these?
Isn't it enough to compare against plain cython implementations?

I am worrying that this will make it hard for this code to be ever removed.

l1joseph:
@sfiligoi Good catch, removed the dead branches and the now-unused `engine` param. The reference kernels are kept, since they're still directly tested and used to validate the new row-tile kernel.

skbio/stats/distance/_permanova.py
```diff
@@ -163,8 +164,7 @@ def _permanova_f_stat_sW_condensed_nb(condensed_matrix, group_sizes, grouping):
             for col_idx in range(row_idx + 1, n):
                 if grouping[col_idx] == group_idx:
                     condensed_idx = (
-                        row_idx * n + col_idx
```
sfiligoi, line 166: Is there a good reason for the change in formatting?

We should keep the PR diff minimal, unless there is a real semantic need for the change.

skbio/stats/distance/_permanova.py
```diff
@@ -360,9 +443,7 @@ def _run_permanova_rowtile_nb(
             # is a non-contiguous view, so this materialises the C-contiguous
             # copy the kernel needs.
             groupings_T = np.ascontiguousarray(buf[:chunk_size].T)
-            _permanova_f_stat_sW_rowtile_nb(
-                distmat.data, groupings_T, inv_group_sizes, partials
-            )
+            rowtile_kernel(distmat.data, groupings_T, inv_group_sizes, partials)
```
sfiligoi, line 450: Here is another example.

There are many others.
Please clean up the PR, so that there are no changes that do not have a clear semantic driver.

l1joseph:
@sfiligoi I accidentally ran ruff format on the whole file, pulling in unrelated pre-existing formatting drift instead of just the lines I changed. Will clean up the diff right now.

l1joseph:
@sfiligoi Cleaned up, rebuilt from `main`. All tests still pass.

Note: `ruff format` still flags this file, but that's pre-existing drift on `main` itself, not from this PR, except one line in our new condensed kernel that I deliberately wrote to match its sibling kernel's existing (also-flagged) style:

```python
# both the pre-existing sibling kernel and our new one have this,
# ruff would collapse both to one line:
local_sW[p] += (
    rsum[p] * inv_group_sizes[groupings_T[mirror_row, p]]
)
```

qiyunzhu:
Question to all: I see that the current code and the incoming code both hard coded to float64. What is the rationale here? The Cython code has a custom type `TReal`, which covers both float and double. Modern Cython has a fused type `floating` which does exactly that. Does the PERMANOVA function not support native float32 arithmetics?

sfiligoi:
@qiyunzhu The fp64 accumulator is a hard requirement.
Even if the input DM elements are fp32, for any non-trivial matrix size we are accumulating so many elements that you get significant loss of precision if you use anything less than fp64.

Everything else does preserve the input numerical precision.
But the accumulator needs to be high precision.

sfiligoi (approved):
Looks good to me

qiyunzhu:
@sfiligoi I see. But what if the backend / device doesn't support fp64? For example PyTorch defaults to fp32. There are devices which we cannot test like TPU that don't support fp64.

l1joseph:
@qiyunzhu Good question, but outside what this PR touches. The numba kernel is CPU/numpy-only, dispatch skips it for non-numpy backends entirely.

Separately, the existing array-API path forces `s_T` to float64 but leaves `s_W` at native precision.
```python
# _permanova_array_api
s_T = xp.sum(dm * dm, dtype=xp.float64) / sample_size / 2.0

# _permanova_sW_xp, no dtype specified
return xp.sum(dm * dm * mask * row_w[:, None])
```

sfiligoi:
@qiyunzhu I think we cannot really support any backend that does not support fp64, not even in software-emulation mode.
At least not in the general case. The loss of precision is real.

That said, there are use cases where fp32-only would be acceptable, e.g. trivially small DMs.

Or we could use workaround, like ordering the values before summing them.
But that's very expensive.

As @l1joseph pointed out, not a problem for this PR, as it is CPU only.
But it is a general issue for when we do want to support non-CPU backends.

PR #2559, opened 2026-08-30
ENH: add engine= selector with Numba geometric-median to permdisp

l1joseph:
Adds an optional Numba compute engine for the geometric-median (`test="median"`) computation used by `permdisp`, wired through the existing `engine=`/`_resolve_engine` dispatch pattern used elsewhere in the package. The Cython `geomedian_axis_one` kernel is unchanged; the Numba kernel is numerically verified against it to `atol=1e-10`. `engine` is only resolved when `test="median"`, since that's the only test type that reads it; the docstring documents `engine` as ignored when `test="centroid"`.

The Numba kernel runs the Weiszfeld iteration sequentially rather than with `prange`: the inner per-point/per-dimension loops are only tens to low hundreds of elements, too small to amortize numba's per-region parallel dispatch overhead, which would otherwise be paid on every outer iteration and on every one of the thousands of calls `permdisp` makes per invocation (`num_groups * (permutations + 1)`).

The Numba branch casts its input with `np.asarray(X, dtype=np.float64)` rather than `np.ascontiguousarray`, since the input is typically a transposed view (`group_data.T`); `asarray` only casts dtype without forcing a copy, and numba handles the resulting non-contiguous array directly.

Benchmark, `permdisp(test="median", engine=...)`, `permutations=10`:

| n | cython | numba | speedup |
|---|---|---|---|
| 50 | 5.2ms | 4.7ms | 1.12x |
| 200 | 7.8ms | 6.6ms | 1.19x |
| 500 | 16.9ms | 14.8ms | 1.15x |

The geometric-median kernel is a small fraction of `permdisp`'s per-call cost, so the end-to-end speedup is modest.

Tests:
- numba kernel matches Cython to `atol=1e-10`, on both a fixed fixture and randomized shapes
- full `permdisp()` call agrees across engines
- invalid engine raises
- `engine` is ignored (does not raise, even if invalid) when `test="centroid"`
- calling the numba engine directly without numba installed raises `ImportError`

Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * `TreeNode.prune` and `TreeNode.bifurcate` now accept an `inplace` parameter (default `True`, preserving the previous in-place behavior) and return the resulting tree. This makes them consistent with other whole-tree methods such as `shear` and `root_at_midpoint`. Set `inplace=False` to leave the original tree unchanged and operate on a copy ([#2495](https://github.com/scikit-bio/scikit-bio/pull/2495)).
 * Transition probability matrix computation has been introduced ([#2496](https://github.com/scikit-bio/scikit-bio/pull/2496)).
 * `pcoa` and `center_distance_matrix` can now use the Numba backend for distance-matrix centering via `engine="numba"` [#2508](https://github.com/scikit-bio/scikit-bio/pull/2508).
+* `permdisp` now supports the optional Numba backend via an `engine` parameter, providing a JIT-compiled alternative to the Cython geometric-median (`test="median"`) computation.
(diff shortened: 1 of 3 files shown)


sfiligoi:
This is likely in conflict with #2547 , which was both opened first and seems more complete.

@l1joseph Can you review that PR, and let us know if you still think that your implementation is superior?
(Much larger benchmarks would also be required)

l1joseph:
Agreed, closing in favor of #2547 — it's more complete (handles the exact-coincidence edge case and documents the f_oneway ordering issue, neither of which this PR addressed). Thanks for the pointer.

PR #2560, opened 2026-08-30
PERF: replace mmvec np.add.at scatter with fused Numba kernel

l1joseph:
`mmvec`'s Adam optimizer accumulates per-minibatch gradient contributions into the full parameter matrix with `np.add.at`, once for the main embedding gradient and once for the bias gradient, on every training step. This replaces both calls with a single fused Numba kernel that scatter-adds a batch of row-indexed contributions in one pass. The kernel is used automatically when Numba is installed and falls back to `np.add.at` otherwise, so results are unaffected when Numba isn't available. Out-of-range ids are silently skipped, matching the semantics of the kernel this replaces.

This is a sequential scan over the batch rather than a parallel one: the batch size, not the row count, is what's small here, so a single pass over the batch beats parallelizing over the output rows. `ids` (the batch's row indices) is cast to a contiguous `intp` array once in `loss_and_grad`, instead of being re-cast independently for both the main and bias gradient calls.

Benchmark, scatter-add of a `(batch, cols)` contribution block into a `(rows, cols)` accumulator, vs `np.add.at`:

| rows | cols | batch | np.add.at | numba | speedup |
|---|---|---|---|---|---|
| 500 | 32 | 64 | 23.0us | 4.1us | 5.6x |
| 2000 | 64 | 256 | 169.3us | 29.1us | 5.8x |
| 5000 | 128 | 1024 | 1236.4us | 146.8us | 8.4x |

JIT warm up is a one time cost, not included above.

Tests:
- bit-for-bit match for the general case
- single-column bias shape
- all contributions target the same row
- empty batch leaves the accumulator untouched
- out-of-range ids (negative and `>= n_rows`) are silently skipped
- full `loss_and_grad` agrees between the Numba and `np.add.at` paths

Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 * When using the Numba backend, Permanova is up to 8x faster [#2488](https://github.com/scikit-bio/scikit-bio/pull/2488).
 * Improved `TreeNode.copy` such that it can handle node cross-references correctly: If a node attribute refers to another node in the tree, the copied node attribute will be redirected to the corresponding node in the new tree ([#2497](https://github.com/scikit-bio/scikit-bio/pull/2497)).
 * On a GPU-resident matrix, the fused Numba kernel accelerates `permanova` and `mantel` on the device; on a datacenter GPU it is much faster than the CPU engines (for example, `permanova` at 25000 samples and 9999 permutations in about 7 s versus about 426 s for OpenMP Cython on an MI300X) [#2511](https://github.com/scikit-bio/scikit-bio/pull/2511).
-* Accelerated `mmvec`'s Adam optimizer by replacing the per-minibatch `np.add.at` scatter-accumulation in the gradient computation with a fused Numba kernel (used automatically when Numba is installed; falls back to `np.add.at` otherwise). Results are numerically identical.
+* Accelerated `mmvec`'s Adam optimizer by replacing the per-minibatch `np.add.at` scatter-accumulation in the gradient computation with a fused, sequential Numba kernel (used automatically when Numba is installed; falls back to `np.add.at` otherwise), 5.7-8.2x faster at representative batch sizes. Results are numerically identical.
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 * When using the Numba backend, Permanova is up to 8x faster [#2488](https://github.com/scikit-bio/scikit-bio/pull/2488).
 * Improved `TreeNode.copy` such that it can handle node cross-references correctly: If a node attribute refers to another node in the tree, the copied node attribute will be redirected to the corresponding node in the new tree ([#2497](https://github.com/scikit-bio/scikit-bio/pull/2497)).
 * On a GPU-resident matrix, the fused Numba kernel accelerates `permanova` and `mantel` on the device; on a datacenter GPU it is much faster than the CPU engines (for example, `permanova` at 25000 samples and 9999 permutations in about 7 s versus about 426 s for OpenMP Cython on an MI300X) [#2511](https://github.com/scikit-bio/scikit-bio/pull/2511).
-* Accelerated `mmvec`'s Adam optimizer by replacing the per-minibatch `np.add.at` scatter-accumulation in the gradient computation with a fused, sequential Numba kernel (used automatically when Numba is installed; falls back to `np.add.at` otherwise), 5.7-8.2x faster at representative batch sizes. Results are numerically identical.
+* Accelerated `mmvec`'s Adam optimizer by replacing the per-minibatch `np.add.at` scatter-accumulation in the gradient computation with a fused, sequential Numba kernel, 5.7-8.2x faster at representative batch sizes. Results are numerically identical. Opt in with the new `engine="numba"` parameter on `mmvec`/`MMvec`; the default `engine="numpy"` keeps using `np.add.at`.
 
 ### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


sfiligoi:
@l1joseph The current policy is that we do not implicity use numba, even if it is available.
A user must explicitly request it through the engine argument.

This may change in the future, but we insist on this in the current development round.

CC @qiyunzhu

l1joseph:
@sfiligoi Added an explicit `engine=` parameter to `mmvec`/`MMvec`, defaulting to `"numpy"` (`np.add.at`). Numba is only used with `engine="numba"`. Doesn't route through the global `set_config("engine")` default since that's `"cython"`, which mmvec has no equivalent of. Tests updated accordingly, 51/51 pass.

sfiligoi:
@l1joseph Please use engine="cython" instead of "numpy".

I.e., cython implies "not JIT compiled and part of the standard python environment"...
(And there is nothing more basic than Numpy for any sizable data structure in Python)

l1joseph:
`engine` default is now `"cython"`, still dispatching to `np.add.at` (no actual Cython code, per the library-wide convention you described). Also updated docstring. Tests/Changelog updated.

qiyunzhu:
@sfiligoi You are correct that currently `engine` can only be `cython` or `numba`. However `cython` is misleading since there is no Cython code involved here. In Pandas, the `engine` parameter can be set to `cython`, if there is Cython code, for example in [`DataFrameGroupBy.mean`](https://pandas.pydata.org/docs/reference/api/pandas.api.typing.DataFrameGroupBy.mean.html); or `python`, if there is no Cython code, for example in [`DataFrame.apply`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.apply.html). We may add the `python` option too.

qiyunzhu:
@l1joseph Was your benchmark for the `_scatter_add_grad` call alone or for the entire `mmvec` call? If the former, I would ask you to test the latter to see if the overall gain is significant. In this work we noticed that L-BFGS-B is much faster than Adam, so the function now defaults to L-BFGS-B, while retaining Adam as a fallback that might be useful in some cases. Accelerating a single step in Adam while adding a new compiled engine (and maintenance burden) needs to be justified with a large overall performance gain.

l1joseph:
@sfiligoi @qiyunzhu I wanted to look into whether `_scatter_add_grad` actually helps end-to-end. I found a couple things I wanted your guidance on.

Profiling a real `mmvec()` Adam run shows `_scatter_add_grad` is under 1% of total time. This confirms your concern. `loss_and_grad` itself is 77% of it, and two things in there are worth fixing regardless of Numba. `rng.choice(..., p=weights)` rebuilds and revalidates its sampling distribution on every single minibatch call even though `weights` never changes for the whole run, 29% of each call. `Y` also comes out of ingestion in Fortran order, making the per-batch row gather cache-hostile. Both fixes are bit-for-bit identical to current output.

On top of that, fusing the rest of `loss_and_grad` (matmul, softmax, gradient accumulation, with the scatter-add folded directly in) into a Numba kernel gives a further real gain. It needs `fastmath=True` to get any speedup at all. Without it it's a regression (0.87x). This means it's not bit-exact. Verified deviation is about 1e-15 on losses and 5e-13 on embeddings, checked element by element across a full 786,000-iteration training run and confirmed not to grow versus a 15,720-iteration check, same order of magnitude at both lengths.

Real full runs at `mmvec()` defaults (`max_iter=1000`, `batch_size=50`), not extrapolated except where noted:

**Mid scale (200x2000x200, 786,000 minibatch calls)**

| variant | time | speedup |
|---|---|---|
| current (HEAD) | 871.9s | 1.00x |
| + contiguous Y | 815.2s | 1.07x |
| + CDF hoist (new `engine="cython"`) | 579.2s | **1.51x, bit-identical** |
| + Numba fused (new `engine="numba"`) | 353.7s | **2.47x** |

**Large scale (500x5000x500, 4,832,000 minibatch calls)**

| variant | time | speedup |
|---|---|---|
| current (HEAD, extrapolated from a short run, validated to within 2.8% against seven other real full runs) | ~19,814s (5h30m) | 1.00x |
| + contiguous Y | 15,387.1s | 1.29x |
| + CDF hoist (new `engine="cython"`) | 6,886.4s | **2.88x, bit-identical** |
| + Numba fused (new `engine="numba"`) | 4,983.1s | **3.98x** |

Element-wise correctness at the full 786,000-iteration mid-scale run, current cython (with the two fixes) versus HEAD, zero differing elements across all four output arrays. New numba versus HEAD, deviation stays at 1e-15 to 5e-13 with no accumulation from 15,720 to 786,000 iterations.

Also checked the L-BFGS path since it's the actual default. `_train_lbfgs`'s setup computed `X.T @ Y` via a Python loop with `np.bincount`, when it's just a sparse matmul. `X_coo.T @ Y` is bit-identical and gives 1.63x-2.66x. It's also more memory efficient with it going from 18.7 GB down to 337 MB at large scale.

Didn't pursue fusing `loss_and_grad_fb` itself. L-BFGS is extremely sensitive there, a 1e-16 gradient perturbation moves the returned `ranks` by ~1%, so fusion would make `engine="numba"` visibly diverge from `engine="cython"` on the default optimizer to save a few seconds.

This makes the current `_scatter_add_grad_nb` in this PR dead code on the numba path. Separately, I found a pre-existing bug while in here. `engine` is documented as "ignored for optimizer='lbfgs'" but `_resolve_engine` runs before the optimizer branch, so `engine="numba"` with `optimizer="lbfgs"` raises `ImportError` on a machine without Numba.

Should I keep working on the numba backend? Should I keep working on the cython backend, for both optimizers? Is the `fastmath` tradeoff acceptable at all? And, do you want me to close this PR and open a new one for this?

l1joseph:
@qiyunzhu No rush, I just wanted to briefly follow up on this PR so that it doesn't get stuck. Would greatly appreciate any guidance you can provide.

qiyunzhu:
@l1joseph I thought a bit and would tentatively suggest that you move on from the current effort. As your benchmark shows, `_scatter_add_grad` isn't a significant component of the runtime, therefore adding a compiled kernel for it alone is not worth. Meanwhile, `loss_and_grad` looks to me already quite vectorized. I am not sure if there is a big room to improve using a compiled kernel (at the cost of maintenance burden). Some Python optimizations like memory space reuse might help modestly.

Although Numba with `fastmath=True` does help, but for Cython we can also add the `-ffast-math` compiler flag and expect certain amount of performance gain. The current scikit-bio avoids any aggressive compiler flags (like `-O3` or `-march=native`). I don't know the background of this historical design. Perhaps it is to ensure broad compatibility? This issue can be re-visited as a whole at some point.

l1joseph:
Agreed. Should I narrow this PR down to the `rng.choice` CDF hoist, the `Y` contiguity fix, and the L-BFGS sparse `X_coo.T @ Y` matmul, and drop numba, fastmath, and engine entirely?

PR #2561, opened 2026-08-30
PERF: vectorize Welch's t-test in dirmult_ttest

l1joseph:
`dirmult_ttest` builds a `statsmodels.stats.weightstats.CompareMeans` object on every posterior draw to compute Welch's t-test statistic, p-value, and confidence interval. This replaces that per-draw object construction with a closed-form computation of the same quantities (`t = diff / se`, Welch-Satterthwaite degrees of freedom, and the t-distribution survival function and quantile from `scipy.stats`). The per-draw loop computes only the sufficient statistics (mean difference, standard error, degrees of freedom); the t-statistic, p-value, and confidence interval are computed in one vectorized pass across all draws afterward. Results are numerically identical to the previous implementation.

`dirmult_ttest` also now validates `draws` explicitly: anything other than a positive integer raises `ValueError`, rather than the previous implementation's uncontrolled behavior on `draws=0` (empty arrays reaching `.mean()`/`.min()`/`.max()`, silently producing `NaN` with a warning).

Benchmark, full `dirmult_ttest` call, `draws=128`:

| n1 | n2 | features | before | after | speedup |
|---|---|---|---|---|---|
| 10 | 10 | 50 | 31.9ms | 14.0ms | 2.3x |
| 25 | 25 | 200 | 70.3ms | 55.4ms | 1.3x |
| 50 | 50 | 500 | 225.4ms | 210.7ms | 1.1x |

The speedup shrinks as the table grows because the per-draw Dirichlet-multinomial resampling step, unchanged by this PR, increasingly dominates the runtime.

Tests:
- closed-form output matches `statsmodels.stats.weightstats.CompareMeans` to `rtol=1e-12`
- invalid `draws` (zero, negative, non-integer) raises `ValueError`

Checklist:

### DIFF
--- a/skbio/stats/composition/_dirmult.py
+++ b/skbio/stats/composition/_dirmult.py
@@ -226,7 +226,7 @@ def dirmult_ttest(
     b7     7.600734  1.480232 -0.601277  4.043888  0.017077  0.068310   False
 
     """
-    from statsmodels.stats.weightstats import CompareMeans
+    from scipy.stats import t as t_dist
 
     rng = get_rng(seed)
 
@@ -241,15 +241,15 @@ def dirmult_ttest(
     groups, labels = _check_grouping(grouping, matrix, samples)
     trt_idx, ref_idx = _check_trt_ref_groups(treatment, reference, groups, labels)
 
-    cm_params = dict(alternative="two-sided", usevar="unequal")
-
     # initiate results
     m = matrix.shape[1]
-    delta = np.zeros(m)  # inter-group difference
-    tstat = np.zeros(m)  # t-test statistic
-    pval = np.zeros(m)  # t-test p-value
-    lower = np.full(m, np.inf)  # 2.5% percentile of distribution
-    upper = np.full(m, -np.inf)  # 97.5% percentile of distribution
+    n1 = len(trt_idx)
+    n2 = len(ref_idx)
+
+    # Per-draw sufficient statistics for Welch's t-test.
+    diff = np.empty((draws, m))  # mean(treatment) - mean(reference)
+    se = np.empty((draws, m))  # Welch standard error
+    dof = np.empty((draws, m))  # Welch-Satterthwaite degrees of freedom
 
     for i in range(draws):
         # Resample data in a Dirichlet-multinomial distribution.
@@ -259,31 +259,31 @@ def dirmult_ttest(
         trt_mat = dir_mat[trt_idx]
         ref_mat = dir_mat[ref_idx]
 
-        # Calculate the difference between the two means.
-        delta += trt_mat.mean(axis=0) - ref_mat.mean(axis=0)
+        # Welch's (unequal-variance) t-test sufficient statistics. This is the
+        # closed form of statsmodels' CompareMeans.ttest_ind / tconfint_diff with
+        # usevar="unequal"; sample variances use ddof=1.
+        vn1 = trt_mat.var(axis=0, ddof=1) / n1
+        vn2 = ref_mat.var(axis=0, ddof=1) / n2
+        diff[i] = trt_mat.mean(axis=0) - ref_mat.mean(axis=0)
+        se[i] = np.sqrt(vn1 + vn2)
+        dof[i] = (vn1 + vn2) ** 2 / (vn1**2 / (n1 - 1) + vn2**2 / (n2 - 1))
 
-        # Create a CompareMeans object for statistical testing.
-        # Welch's t-test is also available in SciPy's `ttest_ind` (with `equal_var=
-        # False`). The current code uses statsmodels' `CompareMeans` instead because
-        # it additionally returns confidence intervals.
-        cm = CompareMeans.from_data(trt_mat, ref_mat)
+    # Vectorized two-sided Welch's t-test across all draws at once.
+    tstat_ = diff / se
+    pval_ = 2.0 * t_dist.sf(np.abs(tstat_), dof)
 
-        # Perform Welch's t-test to assess the significance of difference.
-        tstat_, pval_, _ = cm.ttest_ind(value=0, **cm_params)
-        tstat += tstat_
-        pval += pval_
+    # 95% confidence intervals of the difference (alpha=0.05, two-sided).
+    tcrit = t_dist.ppf(0.975, dof)
+    lower_ = diff - tcrit * se
+    upper_ = diff + tcrit * se
 
-        # Calculate confidence intervals.
-        # The final lower and upper bounds are the minimum and maximum of all lower
-        # and upper bounds seen during sampling, respectively.
-        lower_, upper_ = cm.tconfint_diff(alpha=0.05, **cm_params)
-        np.minimum(lower, lower_, out=lower)
-        np.maximum(upper, upper_, out=upper)
-
-    # Normalize metrics to averages over all replicates.
-    delta /= draws
-    tstat /= draws
-    pval /= draws
+    # Aggregate across draws: averages for point estimates, and the widest
+    # interval (min lower / max upper) across draws, matching prior behavior.
+    delta = diff.mean(axis=0)
+    tstat = tstat_.mean(axis=0)
+    pval = pval_.mean(axis=0)
+    lower = lower_.min(axis=0)
+    upper = upper_.max(axis=0)
 
     # Correct p-values for multiple comparison.
     if p_adjust is not None:
--- a/skbio/stats/composition/_dirmult.py
+++ b/skbio/stats/composition/_dirmult.py
@@ -261,21 +261,30 @@ def dirmult_ttest(
 
         # Welch's (unequal-variance) t-test sufficient statistics. This is the
         # closed form of statsmodels' CompareMeans.ttest_ind / tconfint_diff with
-        # usevar="unequal"; sample variances use ddof=1.
+        # usevar="unequal" (and, checked separately, of scipy.stats.ttest_ind
+        # with equal_var=False); sample variances use ddof=1. Kept as this
+        # closed form rather than delegating to either library call: both were
+        # benchmarked and are markedly slower per draw than this vectorized
+        # arithmetic (scipy.stats.ttest_ind in particular measured 2.4-4x
+        # slower here, presumably from its more general-purpose per-call
+        # overhead), which would undercut the point of this PR.
         vn1 = trt_mat.var(axis=0, ddof=1) / n1
         vn2 = ref_mat.var(axis=0, ddof=1) / n2
+        pooled = vn1 + vn2
         diff[i] = trt_mat.mean(axis=0) - ref_mat.mean(axis=0)
-        se[i] = np.sqrt(vn1 + vn2)
-        dof[i] = (vn1 + vn2) ** 2 / (vn1**2 / (n1 - 1) + vn2**2 / (n2 - 1))
+        se[i] = np.sqrt(pooled)
+        dof[i] = pooled**2 / (vn1**2 / (n1 - 1) + vn2**2 / (n2 - 1))
 
     # Vectorized two-sided Welch's t-test across all draws at once.
     tstat_ = diff / se
-    pval_ = 2.0 * t_dist.sf(np.abs(tstat_), dof)
+    pval_ = t_dist.sf(np.abs(tstat_), dof)
+    pval_ *= 2.0
 
     # 95% confidence intervals of the difference (alpha=0.05, two-sided).
     tcrit = t_dist.ppf(0.975, dof)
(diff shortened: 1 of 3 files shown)


sfiligoi:
@l1joseph Can you provide benchmarks to at least 25k features?
(E.g. EMP input)
Please also disclose the CPU you tested this on (e.g. number of cores)

qiyunzhu:
This is the right direction. statsmodels is not designed for handling many features at once. Replacing statsmodels functions with vectorized operation will get a big performance boost. In scikit-bio there are only a few places where statsmodels is used. I think they can all be vectorized.

That being said, the performance gain in vectorizing `CompareMeans` in `dirmult_ttest` is moderate compared with another and true monster: `MixedLM` in `dirmult_lme`. Previous a volunteer did some constructive work on it (#2409), but we were too afraid to move forward due to consideration of statistical correctness beyond performance since linear mixed effect model is really tricky.

Another worth migration is FDR correction methods (`p_adjust`). I am currently taking care of it in another branch.

skbio/stats/composition/_dirmult.py
```diff
@@ -49,6 +49,36 @@ def _dirmult_draw(matrix, rng):
     return clr(rng.gamma(shape=matrix, scale=1.0, size=matrix.shape), validate=False)
 
 
+def _welch_draw_stats(trt_mat, ref_mat, n1, n2):
```
qiyunzhu: This can probably be further accelerated by reusing the memory space allocation of those matrices and vectors.

l1joseph:
@qiyunzhu Should I also take a look at `dirmult_lme`, as its own PR, or try to contribute to #2409?

l1joseph:
@qiyunzhu I updated `_welch_draw_stats` to write into pre-allocated buffers via `out=`/in-place ops instead of allocating 6 fresh arrays per draw. Here are the benchmarking results:

| m | old | new | speedup |
|---|---|---|---|
| 500 | 232.7ms | 177.1ms | 1.31x |
| 5000 | 1716.7ms | 1676.3ms | 1.02x |
| 25000 | 9215.3ms | 9041.3ms | 1.02x |

The win shrinks as m grows, since allocation overhead becomes a smaller fraction of the work once the variance/mean compute itself dominates at EMP scale.

l1joseph:
@sfiligoi Benchmarks including 25k features, run on a Cosmos node (MI300A, 192 cores), n1=50, n2=50, draws=128:

| m | statsmodels | closed-form | closed-form + buffer reuse |
|---|---|---|---|
| 500 | 282.4ms | 183.4ms | 182.4ms |
| 5,000 | 2,135.7ms | 1,704.8ms | 1,671.3ms |
| 25,000 | 24,973.2ms | 9,597.5ms | 9,388.7ms |

qiyunzhu:
@l1joseph Can I see your updated code? My intuition says there is a chance to get bigger improvement than 1.02x with the memory allocation trick. Also, if you don't mind please refrain from working on the `dirmult_lme` issue. There are further mathematical considerations beyond just optimization.

qiyunzhu:
@l1joseph I tested myself and found that buffer reuse does indeed only bring marginal performance gain, in agreement with your test result.

Additionally, another valuable optimization is that `np.var` can take a pre-computed `np.mean` to save compute. This has been supported since NumPy 2.0, and scikit-bio's minimum NumPy support version is 2.0, therefore it is safe. In my tests it has 10-20% performance gain for the function call alone (but still dwarfed by Gamma distribution draw). So the whole function becomes (note the first four lines):

```python
def _welch_draw_stats(trt_mat, ref_mat, n1, n2, diff, se, dof, work):
    np.mean(trt_mat, axis=0, out=diff)
    np.mean(ref_mat, axis=0, out=work)
    np.var(trt_mat, axis=0, ddof=1, mean=diff, out=se)
    np.var(ref_mat, axis=0, ddof=1, mean=work, out=dof)
    diff -= work
    se /= n1
    dof /= n2

    np.add(se, dof, out=work)
    np.square(se, out=se)
    se /= n1 - 1
    np.square(dof, out=dof)
    dof /= n2 - 1
    se += dof
    np.square(work, out=dof)
    dof /= se
    np.sqrt(work, out=se)
```

There are further micro-optimizations here (e.g., set `ddof=2 * n1 - n1 ** 2` in `np.var` call can let you save the `se /= n1` division) but they are ugly, therefore probably not worth.

Another idea I thought about is to reorder `matrix` such that `trt_idx` and `ref_idx` become two contiguous slices. Therefore `trt_mat` and `ref_mat` will also become two contiguous memory blocks, which will accelerate arithmetics. However tests showed that the performance gain is marginal, and it will break the current stochastic reproducibility (despite mathematical equivalence). Therefore it is probably not the way to go.

qiyunzhu:
Checked with 5.6-Sol. Here's its assessment. To summarize it identifies a memory regression issue, and pointed out that streaming the per-draw intermediate and breaking CLR call into in-place calculations plus a scipy ufunc adoption can effectively resolve the memory problem.

https://chatgpt.com/s/t_6a96614368048191844571f7a0d13596

l1joseph:
Code is pushed.

Took your `mean=` suggestion. Also fixed `_dirmult_draw`'s gamma sampling allocating a fresh `matrix.shape` array every draw, and (per the 5.6-Sol assessment) did the CLR log/center step in place on that disposable buffer instead of calling the shared `clr()`, and swapped `scipy.stats.t.sf`/`ppf` for the lower-level `scipy.special.stdtr`/`stdtrit` ufuncs. All verified numerically identical to the current output first.

Rerun on an isolated node, n1=50/n2=50/draws=128, median of 5:

| m | welch-only (posted) | + all of the above |
|---|---|---|
| 500 | 237.6ms | 214.1ms (1.11x) |
| 5000 | 2161.3ms | 1933.4ms (1.12x) |
| 25000 | 11199.8ms | 10056.9ms (1.11x) |

I agree with 5.6's suggestion for streaming aggregation, since my current code holds several `(draws, features)` arrays where `(features,)` would do, and that scales badly past 25k features. Is that within scope of this PR or should I open a separate one, since it touches both the loop and the post-processing section.

I also checked replacing the trt/ref fancy-indexing with np.take, and skipping gather entirely using np.mean/np.var directly on full matrix. However both measure slower, so didn't include them. Profiled per draw cost at m=25k: gamma draw ~86%, clr ~7%, gather ~2%, Welch stats ~4.5%.

qiyunzhu:
@l1joseph Thanks for the update. The finding of `standard_gamma` is very cool!! I think it would significantly reduce intermediate memory allocation. I would say that streaming aggregation is within the scope of this PR. I envision that it might change the numerical precision slightly, considering that adding numbers one at a time vs altogether may not have the same numerical process. But I think such price is worth. The final result may have only very minor discrepancy in extreme cases. Without streaming aggregation, the current code is still bottlenecked by memory.

skbio/stats/composition/_dirmult.py
```diff
@@ -43,10 +50,67 @@ def _dirmult_draw(matrix, rng):
...
+
+    The Gamma draw is disposable (nothing else reads it), so the CLR transform
+    is applied in place rather than via :func:`clr`, which is a shared,
+    array-API-general function that cannot assume it owns its input. This is
+    bit-identical to ``clr(draw, validate=False)``.
+
+    """
+    draw = rng.standard_gamma(matrix, size=matrix.shape, out=out)
+    np.log(draw, out=draw)
+    draw -= draw.mean(axis=-1, keepdims=True)
```
qiyunzhu: You can further re-use buffer for the `draw.mean` call. For example `np.mean(draw, axis=1, out=work)` then `draw -= work[None, :]`.

l1joseph:
Replaced the `(draws, features)` diff/se/dof/tstat/pval/lower/upper arrays with running sums and running min/max over `(features,)` buffers, updated per draw instead of after the full loop.

Rerun on an isolated node, n1=50/n2=50/draws=128, median of 5, peak RSS via `ru_maxrss`:

| m | time before | time streaming | RSS before | RSS streaming |
|---|---|---|---|---|
| 500 | 215.9ms | 214.5ms | 157.3MB | 152.3MB |
| 5000 | 1932.5ms | 2023.2ms | 217.2MB | 173.6MB |
| 25000 | 10150.2ms | 10072.9ms | 479.6MB | 252.9MB |

47% peak RSS reduction at 25k features, matching your structural argument. Speed is roughly neutral on this hardware though, not the 12.5% speedup in the 5.6 transcript, actually about 5% slower at m=5000. My guess is the 128 small `stdtr`/`stdtrit` calls per draw cost more in per-call dispatch than the one big batched call saves.

skbio/stats/composition/_dirmult.py
```diff
@@ -333,30 +347,39 @@ def dirmult_ttest(
...
+        np.maximum(upper, bound_i, out=upper)
+
+        delta_sum += diff_i
+        tstat_sum += tstat_i
+        pval_sum += pval_i
+
+    # Aggregate across draws: averages for point estimates (widest interval
+    # for the confidence bounds was already tracked above), matching prior
+    # behavior.
+    delta = delta_sum / draws
```
qiyunzhu: These three calculations can be done in-place. They are already floats so dividing them by an integer in-place is safe.

skbio/stats/composition/_dirmult.py
```diff
@@ -312,14 +312,28 @@ def dirmult_ttest(
...
+    # can differ from the prior behavior at floating-point noise level.
+    delta_sum = np.zeros(m)
+    tstat_sum = np.zeros(m)
+    pval_sum = np.zeros(m)
+    lower = np.full(m, np.inf)
+    upper = np.full(m, -np.inf)
+
+    # Per-draw scratch, reused across draws so the loop allocates no new
+    # (m,)-length arrays.
+    diff_i = np.empty(m)  # mean(treatment) - mean(reference)
```
qiyunzhu, line 335: Lots of vectors are allocated. At least some of them are redundant. For example `work` can be reused in several steps. Redundancy introduces explicity, which is good for understanding. I personally prefer conciseness though.

skbio/stats/composition/_dirmult.py
```diff
@@ -333,30 +347,39 @@ def dirmult_ttest(
...
+        # the ufuncs are faster than the corresponding scipy.stats.t methods.
+        np.abs(tstat_i, out=pval_i)
+        pval_i *= -1
+        stdtr(dof_i, pval_i, out=pval_i)
+        pval_i *= 2.0
+
+        # 95% confidence interval of the difference for this draw
+        # (alpha=0.05, two-sided).
+        stdtrit(dof_i, 0.975, out=tcrit_i)
+        np.multiply(tcrit_i, se_i, out=margin_i)
```
qiyunzhu: It will be helpful to comment such code blocks with the original math formula such that people will understand what is going on.

skbio/stats/composition/_dirmult.py
```diff
@@ -333,30 +347,39 @@ def dirmult_ttest(
...
+
+        # Two-sided Welch's t-test p-value for this draw. Uses the
+        # scipy.special ufuncs directly rather than scipy.stats.t.sf/ppf: same
+        # result (stdtr(dof, -x) is the upper-tail probability of the
+        # symmetric t distribution, i.e. sf(x)), and scipy's own docs note
+        # the ufuncs are faster than the corresponding scipy.stats.t methods.
+        np.abs(tstat_i, out=pval_i)
+        pval_i *= -1
+        stdtr(dof_i, pval_i, out=pval_i)
+        pval_i *= 2.0
```
qiyunzhu: This `*= 2.0` can be moved outside the loop to save a small bit of compute.

qiyunzhu:
Very good! This is a big save of memory.

l1joseph:
Cut the loop's scratch buffers from 9 down to 5, `work` now holds the t-statistic, then the p-value, then the t-critical/margin in sequence, each reused once its predecessor is summed, with a comment explaining each handoff. Moved `*= 2.0` outside the loop (accumulate the un-doubled tail probability, double the sum once at the end). Also did the final delta/tstat/pval divisions in place instead of allocating copies.

Speed and memory are both roughly flat versus the streaming version, which is expected.

qiyunzhu:
I think this PR can be merged. @l1joseph Thank you for making the dedicated and significant improvement of this function!

qiyunzhu:
> @qiyunzhu Should I also take a look at `dirmult_lme`, as its own PR, or try to contribute to #2409?

I thought further and now I think that if you want, feel free to give it a try. #2409 was about parallelizing the algorithm using `Parallel`, rather than rebuilding a vectorized `MixedLM`. It is probably cleaner to start from scratch.

PR #2336, opened 2025-11-16
Implemented multiple evolutionary distance metrics based on sequence alignments

qiyunzhu:
Please complete the following checklist:

* [x] I have read the [contribution guidelines](https://scikit.bio/contribute.html).

* [x] I have documented all public-facing changes in the [changelog](https://github.com/scikit-bio/scikit-bio/blob/main/CHANGELOG.md).

* [ ] **This pull request includes code, documentation, or other content derived from external source(s).** If this is the case, ensure the external source's license is compatible with scikit-bio's [license](https://github.com/scikit-bio/scikit-bio/blob/main/LICENSE.txt). Include the license in the `licenses` directory and add a comment in the code giving proper attribution. Ensure any other requirements set forth by the license and/or author are satisfied.

  - **It is your responsibility to disclose** code, documentation, or other content derived from external source(s). If you have questions about whether something can be included in the project or how to give proper attribution, include those questions in your pull request and a reviewer will assist you.

* [x] This pull request does not include code, documentation, or other content derived from external source(s).

Note: [This document](https://scikit.bio/devdoc/review.html) may also be helpful to see some of the things code reviewers will be verifying when reviewing your pull request.

PR #2562, opened 2026-09-01
BUG: fix _degen_nonca_hash for sequence types without non-canonical characters

dylanpulver:
`DNA.to_definites()` and `RNA.to_definites()` raise at the method's documented default (`noncanonical=True`):

```python
>>> from skbio import DNA
>>> DNA('ACGTRYN').to_definites()
TypeError: 'NoneType' object is not subscriptable
```

Reproduces on released 0.7.3 and on current `main`.

`skbio/sequence/_grammared_sequence.py:264`

```python
            else:
                cls.__canonical_hash = cls._degenerate_hash
        return cls.__degen_nonca_hash
```

When a type declares no non-canonical characters, the `else` branch assigns to `__canonical_hash` instead of `__degen_nonca_hash`. Two consequences:

1. `_degen_nonca_hash` stays `None` and is returned as such; `to_definites` subscripts it at line 870.
2. The cached canonical-character mask is overwritten with the degenerate mask. For `DNA`, `_canonical_hash` goes from `ACGT` to `BDHKMNRSVWY` after any access to `_degen_nonca_hash`. Nothing outside `_canonical_hash` itself reads that mask today, so this half is latent rather than user-visible.

`DNA`, `RNA` and any `GrammaredSequence` subclass that does not override `noncanonical_chars` take this branch. `Protein`, which declares `O`/`U`, does not — and `Protein('ACDEFX').to_definites()` works, which is the built-in control. The existing `test_to_definites` uses `ExampleGrammaredSequence`, whose `noncanonical_chars` is `"Q"`, so the `else` branch has never executed in the suite. Introduced in #2336.

The fix is one token: `__canonical_hash` → `__degen_nonca_hash`.

Disclosed and deliberately left alone: unlike the `if` branch two lines above, this branch stores `_degenerate_hash` itself rather than a copy, so the two cached masks alias. Nothing mutates `_degenerate_hash` after construction today, so this is safe as written — but say the word and I will make it `.copy()` for symmetry.

Four tests: `test_to_definites` for `DNA`/`RNA` across `wild`/`gap`/`del`/single-character with literal expected values, plus `test_degen_nonca_hash` (a type with non-canonical characters, `QWXYZ`, and one without, `XY`), `test_degen_nonca_hash_does_not_alter_canonical_hash` and `test_to_definites_without_noncanonical_chars`.

`make test` (`cd ci && python -m skbio.test`), same venv, Python 3.14, macOS:

```
clean main : 3812 passed, 78 skipped, 0 failed
patched    : 3816 passed, 78 skipped, 0 failed    (+4 = exactly the new tests)
```

Mutants, each verified in place before measuring:

- **A** revert the fix → all 4 new tests fail.
- **B** fix only the call site (`if noncanonical and self.noncanonical_chars:`), leaving line 264 → the two cache tests still fail, so the tests pin the mask and not just the crash.
- **C** right slot, wrong array (`cls._canonical_hash`) → the mask test and both `to_definites` tests fail.

Pins, stated rather than padded: B passes both `to_definites` tests, C passes the does-not-clobber test. Neither survives the set as a whole.

Not tested: macOS / Python 3.14 only, not the full CI matrix. Docs build and the array-api, numba, gpu and wheels workflows were not run. `ruff check` passes at the pinned 0.15.14. `ruff format` was **not** applied: both touched test files already fail `ruff format --check` on unmodified `main`, so formatting them here would be unrelated churn.

Prepared with AI assistance (Claude Opus 5, used as a coding agent in this repository).


qiyunzhu:
@dylanpulver Thanks for the contribution! This is a clean fix. Original bug was a miserable typo. Let me merge.

Issue #1702, opened 2020-04-28
Beta diversity computations for braycurtis metric produce NaNs and consequently issue with DistanceMatrix


ElenaVasylenko:
Good day , I'm using scikit-bio version "0.5.5" and experience issue while computing beta diversity for **braycurtis** metric when array have several (2 or more) zero hits. 
### **Steps to reproduce:**
1. For example, we have such input parameters:
```

counts = array([[0, 3, 0],[0, 3, 0], [2, 3, 0.622535],[0, 0, 0],[3, 3, 0.704623],
                          [0, 0, 0], [4, 3, 0.799652],[5, 1, 1.0],[5, 3, 1.0],[6, 1, 1.0]])

ids = ['Abelson murine leukemia virus GCF 000848265.1',
                   'Avian orthoreovirus AVS B GCF 000891595.1',
                   'Camelpox virus GCF 000839105.1',
                   'Canine morbillivirus GCF 000854065.1',
                   'Colorado tick fever virus GCF 000853265.1',
                   'Cowpox virus Brighton Red GCF 000839185.1',
                   'Crimean Congo hemorrhagic fever orthonairovirus GCF 000854165.1',
                   'Cynomolgus cytomegalovirus GCF 001963235.1',
                   'Hepacivirus C GCF 000861845.1',
                   'Human T lymphotropic virus 2 GCF 000847505.1',]
```
2. Lets call beta diversity calculation with 'braycurtis' metric.
`beta_diversity_distance_matrix = beta_diversity(metric='braycurtis', counts=counts, ids=ids, validate=True,pairwise_func=sklearn.metrics.pairwise_distances)`
3. Check result. I've receive the following error:
```
line 873, in _validate "Data must be symmetric and cannot contain NaNs.")
skbio.stats.distance._base.DistanceMatrixError: Data must be symmetric and cannot contain NaNs.

```
### **Root cause**
I've investigated the root cause of this problem and have the following notes:
1. I've looked through your source codes and found out that underhood you use scipy to compute pairwise distance for beta diversity. Especially this function: `  scipy.spatial.distance.pdist(X, metric, *args, **kwargs) `  (find documentation here  https://docs.scipy.org/doc/scipy-1.0.0/reference/generated/scipy.spatial.distance.pdist.html#scipy.spatial.distance.pdist). 
[...]

PR #2565, opened 2026-09-02
ENH: Warn when beta_diversity's result contains nan values

jissen706:

- [x] I have documented all public-facing changes in the [changelog](https://github.com/scikit-bio/scikit-bio/blob/main/CHANGELOG.md).
- [ ] **This pull request includes code, documentation, or other content derived from external source(s).**

Addresses #1702.

## Description

As reported in #1702: when `counts` contains two or more all-zero samples, some beta diversity metrics divide by the total count of each pair (e.g. Bray-Curtis is `sum(|u-v|) / sum(u+v)`), so the distance between two all-zero samples is `0/0 = nan`. This currently passes through `beta_diversity` silently:

```python
>>> import numpy as np
>>> from skbio.diversity import beta_diversity
>>> counts = [[0, 0, 0], [1, 2, 3], [0, 0, 0]]
>>> dm = beta_diversity('braycurtis', counts)  # no error, no warning
>>> dm[0, 2]
nan
```

A maintainer's response in the issue thread confirmed a warning/error here "would be a good feature addition" and invited a PR.

This adds a check right after the pairwise distances are computed: if the resulting array contains any `nan`, `beta_diversity` emits a `UserWarning` explaining the likely cause (all-zero samples + a metric that divides by total count) and suggesting the user either filter out empty samples beforehand or inspect the returned matrix. I went with a warning rather than an error since:
- `DistanceMatrix` already accepts `nan` values (no exception is raised constructing it), so this isn't changing any return values or breaking existing callers, only surfacing something that was already happening;
- `nan` isn't necessarily wrong for all use cases (some callers may want to filter it out downstream), so silently succeeding is preferable to raising, but silently succeeding *without any indication* is what prompted the original report.

The check is generic (`np.isnan(distances).any()` on the actual computed result) rather than metric-specific, so it only fires when `nan` is actually present — it won't false-positive for metrics unaffected by this (e.g. `'euclidean'`) or when only a single sample is all-zero (which most metrics handle fine against a non-zero sample).

Added `test_all_zero_samples_nan_warns` and `test_all_zero_samples_nan_does_not_warn_for_unaffected_metric` to `skbio/diversity/tests/test_driver.py`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@
 * `permanova` and `pcoa` now honor `engine="numba"` by not taking the scikit-bio-binaries path (the Cython-equivalent acceleration) [#2510](https://github.com/scikit-bio/scikit-bio/pull/2510).
 * Fixed `SymmetricMatrix.filter` and `SymmetricMatrix.permute` silently resetting a non-zero diagonal to `0` when the matrix was stored in condensed form. Both methods reconstructed the result from its condensed representation, which only carries off-diagonal values, without passing the diagonal through. The diagonal is now subset and reordered along with the rows and columns. `DistanceMatrix` is unaffected because it is always hollow ([#2516](https://github.com/scikit-bio/scikit-bio/issues/2516)). Thank @LarytheLord for reporting and diagnosing the root cause.
 
+### Miscellaneous
+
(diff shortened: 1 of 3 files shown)


qiyunzhu:
This is good but there is now a better way to address it. Just apply the helper function [`_check_composition`](https://github.com/scikit-bio/scikit-bio/blob/main/skbio/stats/composition/_base.py#L30C5-L30C23). It will raise an error if there are all-zero samples.

PR #2568, opened 2026-09-03
ENH: add a jplace format for BPTree and fix BPTree Newick grammar handling

### LINKED ISSUE #2349: Is it expected to replace underscores from tip names in the tree?

monkeytim19:
Builds on the `BPTree` Newick I/O added in #2528. Resolves #2349.

This PR does two related things for `BPTree`: it adds a `jplace` I/O format to the `skbio.io` registry, and it fixes several Newick grammar cases in the compiled backend that both the new format and the existing `newick` format depend on.

## 1. `jplace` format (new feature)

Registers the [jplace format](https://doi.org/10.1371/journal.pone.0031009) (Matsen et al. 2012) in the `skbio.io` registry, scoped to `BPTree`:

- **Reader** returns the reference tree as a `BPTree`, preserving its `{}` edge numbers. Placement records are **dropped** — use `skbio.tree.bp.parse_jplace` to get the placements table alongside the tree.
- **Writer** emits a `jplace` document containing the tree with an empty `placements` list.
- **Sniffer** disambiguates jplace (a JSON object with the mandatory top-level members) from every other scikit-bio format.
- New `JplaceFormatError` (a `FileFormatError` subclass) is raised on malformed input, per the registry convention.

Both directions run in the compiled (Cython) backend, reusing `parse_newick`/`write_newick` for the `{}` edge numbers, so throughput is preserved on the large reference trees typical of placement studies.

The scope is deliberately minimal: it migrates the existing jplace read path from [improved-octo-waddle](https://github.com/biocore/improved-octo-waddle) and adds the matching writer, without introducing a placement-aware container type. `parse_jplace(data) -> (DataFrame, BPTree)` remains the way to access placement records.

## 2. Newick grammar corrections (bug fix, #2349)

The compiled Newick reader/writer that backs `BPTree` mishandled several grammar cases that the pure-Python `TreeNode` reader already handled. This aligns the two (and is what lets the jplace tree round-trip losslessly):

- **Comments** — `[...]` comments (including nested ones) are stripped before parsing, instead of corrupting the balanced-parentheses topology or being misread as edge numbers.
- **Quoted labels** — single-quoted labels round-trip: an embedded apostrophe reads as a single `'` (a doubled `''` collapses) and is written back doubled; the name/length split is quote-aware.
- **Whitespace / special characters** — labels containing whitespace or any of `` ' _ ; , : ( ) [ ] `` are quoted on write so the output re-parses losslessly.
- **Edge numbers** — read from `{}` only; `[]` is now a standard Newick comment. `parse_newick` and the `newick` → `BPTree` reader gained an optional `convert_underscores` parameter (default `True`, matching `TreeNode`); the jplace reader opts out so underscores in accession labels are preserved.

## Changes

- `skbio/tree/bp/_bp_io.pyx` — `parse_jplace` (migrated from iow) and `write_jplace`; comment stripping, quote-aware name/length parsing, apostrophe (un)doubling, whitespace/special-character quoting, and `convert_underscores` plumbing for the Newick reader/writer.
- `skbio/tree/bp/__init__.py` — export `parse_jplace`, `write_jplace`.
- `skbio/io/format/jplace.py` — the new format (sniffer, `BPTree` reader/writer).
- `skbio/io/_exception.py`, `skbio/io/__init__.py` — `JplaceFormatError`; register the format module.
- `skbio/io/format/newick.py` — `newick` → `BPTree` reader accepts `convert_underscores`.
- Tests — `skbio/io/format/tests/test_jplace.py` (read/write, sniffer, `parse_jplace` helper), `skbio/tree/tests/bp/test_bp_io.py` (direct `write_jplace`), and `skbio/io/format/tests/test_newick.py` (grammar
  round-trips).
- `CHANGELOG.md` — Features entry (jplace) and Bug Fixes entry (grammar).

## Testing

`pytest skbio/io/format/tests/test_jplace.py skbio/io/format/tests/test_newick.py skbio/tree/tests/bp` passes locally; the new tests cover the added reader/writer and grammar paths.

## Checklist


monkeytim19:
@qiyunzhu @mortonjt

monkeytim19:
@qiyunzhu Have resolved CHANGELOG merge conflict. Ready for review.

skbio/io/format/jplace.py
```diff
@@ -0,0 +1,140 @@
...
+the tree.
+
+Writing a ``BPTree`` produces a jplace document containing the reference tree
+(with its ``{}`` edge numbers) and an empty ``placements`` list -- a ``BPTree``
+holds a reference tree but no placements. The tree parse and serialization run
+in scikit-bio's compiled backend, so throughput is preserved for large files.
+
+Caveats
+~~~~~~~
+- Multiplicities are not supported by :func:`skbio.tree.bp.parse_jplace`; any
```
qiyunzhu, line 50: When rendered, `skbio.tree.bp.parse_jplace` is displayed as plain text without a link. This is because the `bp` submodule has not been added to `skbio/tree/__init__.py`, therefore it does not exist in the document tree. Nevertheless, I don't think the `bp` submodule needs to be exposed to the user. It currently contains format parsing functions only. Instead I think `parse_jplace` doesn't need to be mentioned here.

skbio/io/format/jplace.py
```diff
@@ -0,0 +1,140 @@
...
+        raise JplaceFormatError("Could not parse file as jplace: %s" % e) from e
+    return tree
+
+
+@jplace.reader(BPTree)
+def _jplace_to_bp(fh, cls=None):
+    return _read_tree(fh)
+
+
+@jplace.writer(BPTree)
```
qiyunzhu, line 136: There aren't Newick reader/writer yet. Consider adding them?

qiyunzhu (commented):
This looks very good. I have manually tested reading / writing real jplace files and it works so far. The documentation page looks nice.

There is an issue though: the documentation pages of `skbio.tree.BPTree.read` and `write` don't display "Supported file formats include:..." like other classes do ([example](https://scikit.bio/docs/latest/generated/skbio.tree.TreeNode.read.html)). This is because the `read` and `write` methods in `_bp.pyx` are hand-written instead of using skbio's I/O system. Other classes simply do `read = Read()` and `write = Write()` to register the methods ([example](https://github.com/scikit-bio/scikit-bio/blob/a3beaa632077a532330343d5db0d8e30e58b6705/skbio/tree/_tree.py#L110)). This problem may be due to Cython submodules cannot mimick the Python behavior well, but I am not sure.

skbio/tree/tests/bp/test_bp_io.py
```diff
@@ -298,5 +299,66 @@ def test_parse_newick_linear_tree(self):
         self.assertEqual([n.name for n in skbio_tree.tips()], ["b"])
 
 
+class JPlaceWriteTests(TestCase):
+    def setUp(self):
+        # A reference BPTree carrying edge numbers, obtained from the fixture.
+        jplacedata = open(get_data_path('200/placement.jplace')).read()
+        _, self.bp = parse_jplace(jplacedata)
```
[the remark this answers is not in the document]
qiyunzhu: It will make more sense to place example jplace files in `skbio/io/formats/tests/data` rather than `skbio/tree/tests/bp/data/200`.

monkeytim19:
@qiyunzhu 
### Changes made to address the comments made

**Missing format table from the BPTree.read/write docs**
- Root cause confirmed: the stock Read()/Write() descriptors cache the generated method onto the class/instance, which a Cython cdef class can't allow (immutable type dict, no instance __dict__)
- Fix: BPTree-local _BPRead/_BPWrite subclasses in _bp.pyx reuse the base docstring machinery
- BPTree.read/write now render "Supported file formats include: newick, jplace" like TreeNode (see attached images)
- The Newick reader/writer work should also be enabled as part of PR [#2528] already

**jplace fixture location**
- Moved placement.jplace to skbio/io/format/tests/data/, consolidated the duplicated jplace tests into test_jplace.py (moved the one stray newick test back to NewickTests), and fixed the leaked file handles

**Copilot findings**
- _strip_comments: raises on an unterminated [ comment instead of duplicating text
- parse_jplace: validates the document is a JSON object and requires version
- write_jplace: rejects a string fields, enforces edge_num
- Whitespace before a branch-length : now yields name=None, not ""
- New tests cover each

Screenshot 2026-09-12 at 11 20 39 PM
Screenshot 2026-09-12 at 11 20 54 PM

monkeytim19:
@qiyunzhu After this merging this back into bp_tree branch, I would suggest to refresh `bp_tree` with the changes that have taken place in `main` by merging the latter into the former. This is to borrow the framework that has been set-up already in main for the GPU/Numba work with `BPTree`.

skbio/io/format/jplace.py
```diff
@@ -0,0 +1,137 @@
...
+
+def _read_tree(fh):
+    """Parse a jplace document and return its reference tree (drop placements)."""
+    from skbio.tree.bp._bp_io import parse_jplace
+
+    # The compiled parser scans the whole document at once; materialize the
+    # handle first. It raises ValueError/KeyError on malformed input; re-raise
+    # as the format's error type per the registry convention.
+    try:
+        _, tree = parse_jplace(fh.read())
```
mortonjt, line 122: Because this reader returns only the reference tree, routing through `parse_jplace` makes valid files depend on placement-table limitations and builds a potentially large DataFrame that is immediately discarded. The jplace specification permits each pquery to contain either `n` or `nm`; an otherwise valid `nm`-only document currently raises `JplaceFormatError` here. On large placement files this path also expands every placement/name before throwing the result away. Please parse/validate the top-level JSON and call `parse_newick(data["tree"], convert_underscores=False)` directly (or add a tree-only helper), so reading a tree neither rejects valid placement encodings nor scales with the discarded placement table.

skbio/io/format/jplace.py
```diff
@@ -0,0 +1,137 @@
...
+    except (ValueError, KeyError) as e:
```
mortonjt, line 123: Malformed JSON objects are not consistently converted to the advertised `JplaceFormatError`. For example, setting `fields`, `placements`, or `tree` to `null` currently leaks a `TypeError` (`'NoneType' object is not iterable`, `len(None)`, or `str.strip` on `None`). Please validate member/container types or normalize the expected structural exceptions here so malformed jplace input follows the format-error contract.

skbio/tree/bp/_bp_io.pyx
```diff
@@ -42,50 +39,131 @@ cdef inline cnp.int32_t number_from_edge(unicode token):
...
     else:
-        name = token.replace("'", "").replace('"', "").strip()
+        split_idx = _find_length_colon(token)
+        if split_idx != -1:
+            # name:length[{edge}]
+            name = _unquote_name(token[:split_idx], convert_underscores)
+            token_parsed = token[split_idx + 1:]
+            length = length_from_edge(token_parsed)
+            edge = number_from_edge(token_parsed)
+        elif u'{' in token:
```
mortonjt, line 154: This brace check is not quote-aware. `parse_newick("('a{3}',c)r;")` currently stores the name as `"'a"` and edge 3, while `parse_newick("('a{b}',c)r;")` raises converting `b` to an integer. Braces inside a single-quoted label are literal label content, just like colons handled by `_find_length_colon`. Please locate only an unquoted trailing edge annotation (and add a quoted-brace regression test).

skbio/tree/bp/_bp_io.pyx
```diff
@@ -435,9 +551,87 @@ def parse_jplace(object data):
...
+        field_names = ["edge_num"]
+    elif isinstance(fields, str):
+        raise TypeError("fields must be a list of field names, not a string")
+    else:
+        field_names = list(fields)
+        if "edge_num" not in field_names:
+            raise ValueError("fields must include 'edge_num'")
+
+    buf = StringIO()
+    write_newick(tree, buf, True)
```
mortonjt, line 626: An ordinary `BPTree` read from Newick has the default edge value 0 on every node, so this emits e.g. `((a:0.0{0},b:0.0{0})c:0.0{0},d:0.0{0})r:0.0{0};`. In jplace, edge numbers index placement locations; duplicate IDs make that mapping ambiguous, so this should not be presented as valid jplace output. Please either assign unique edge numbers when none are present or reject trees whose edge-number map is not valid/unique, rather than knowingly writing repeated `{0}` values.

mortonjt (changes requested):
Thanks for the substantial work here. I checked the current head and the focused test suites pass (68 jplace/newick/BP I/O tests, plus the full BP test directory). I found a few correctness issues in the new read/write paths that should be addressed before merging, especially acceptance of valid `nm` jplace records and emission of duplicate edge IDs.

qiyunzhu:
@monkeytim19 The updated documentation looks what it is supposed to be. I have no additional comments.

monkeytim19:
To address the additional four points raised by @mortonjt:

**Tree-only jplace reads (nm records + discarded placement table)**
- The reader no longer routes through `parse_jplace`. It validates the top-level JSON and calls `parse_newick(doc["tree"], convert_underscores=False)` directly.

**Malformed members leaking TypeError**
- The reader keeps the structural contract (all four members required, matching the sniffer) and type-checks tree/placements/fields, so a present-but-null member now raises `JplaceFormatError` instead of leaking `TypeError`. `parse_jplace` is hardened the same way for its own callers.
- The registry writer now surfaces backend `ValueError/TypeError` as `JplaceFormatError` too, mirroring the reader.

**Quote-unaware {} edge detection**
- Added `_find_edge_curly`, which locates the edge annotation only outside single quotes (mirroring `_find_length_colon` for the branch-length :). A { inside a quoted label is now literal content: `parse_newick("('a{3}',c)r;”)` yields name a{3} / edge 0, and 'a{b}' no longer raises. Unquoted name{3} is unchanged. Added a quoted-brace regression test.

**Duplicate edge IDs on write**
- `write_jplace` now requires unique edge numbers and rejects a tree that lacks them (e.g. one read from plain Newick, which has 0 on every edge), rather than emitting ambiguous duplicate {0}s. The root is excluded from the check, since jplace doesn't number the root edge and it defaults to 0 in a BPTree (and can legitimately coincide with a real edge 0).

qiyunzhu (approved):
@monkeytim19 Thanks for the update! Let's merge.

PR #2569, opened 2026-09-04
BUG: apply the permdisp Numba dtype check per test

LarytheLord:
### Description

Follow-up to #2547, as promised there.

`_run_permdisp_numba` rejects any ordination that is not float32 or float64, on
the grounds that the Cython path raises on anything else. That holds only for
`test="median"`, which reaches `geomedian_axis_one`, a fused function carrying
just those two signatures. `test="centroid"` never reaches that kernel: it uses
`group_data.mean(axis=0)`, which numpy is happy to compute for any numeric
dtype. So on an integer ordination the two engines disagree about whether the
call is even legal:

```
test="centroid", int64 ordination
  cython  0.5367018290895094
  numba   TypeError
```

This also contradicts the documented contract. `permdisp`'s own `Raises`
section already scopes the restriction to one test and names the centroid test
as the way around it:

> If, when using the spatial median test, the pcoa ordination is not of type
> np.float32 or np.float64, the spatial median function will fail and the
> centroid test should be used instead

Under `engine="numba"` that advice did not work, because the guard rejected the
input before the centroid test could run. `engine=` is meant to select an
implementation, not change which inputs are accepted, so the check is now made
per test.

The median test keeps the float32 or float64 requirement. The centroid test
accepts any numeric dtype and refuses the rest. That boundary is not arbitrary:
it is where numpy stops being able to do arithmetic, and it is where the Cython
path itself raises. Checking every dtype through both engines on the centroid
test now gives no case where one computes and the other raises:

| dtype | cython | numba |
|---|---|---|
| float64 | 0.5353619887121516 | 0.5353619887121513 |
| float32 | 0.535361647605896 | 0.535361832311309 |
| float16 | nan | 0.5372472502169272 |
| int64, int32 | 0.5367018290895094 | 0.5367018290895084 |
| bool | nan | nan |
| complex, timedelta64, longdouble | computes | computes |
| object, str, datetime64 | raises | raises |

Integers are the clean case because numpy's `mean` promotes them to float64
before the Cython path uses them, so both engines end up doing float64
arithmetic and agree to 1.9e-15 with identical p-values.

The engines are not bit-identical for every dtype, and the comment now says so
rather than implying otherwise. The Numba kernels accumulate in float64 whatever
comes in, while Cython computes in the input dtype, so a float32 or float16
ordination differs by that dtype's precision. That was already true for float32,
which the guard has always accepted.

Only a hand-built `OrdinationResults` can reach any of this: `pcoa` returns
float64 samples for both `eigh` and `fsvd`, even from a float32
`DistanceMatrix`.

The `Raises` section gains the matching entry. That condition is not new, since
the Cython path already raised for a non-numeric ordination on the centroid
test, but it was never written down.

### Testing

`test_non_float_ordination_raises_like_cython` only exercised `test="median"`,
which is why this was missed. It now covers the centroid test in both
directions: int64 and int32 are computed rather than refused and the two engines
return the same statistic and the same p-value, and an object ordination is
refused by both. Strings arrive as object too, since that is how pandas stores
them, so object is the only non-numeric dtype reachable through
`OrdinationResults`.

The integer ordination has to be scaled before casting: the pcoa coordinates on
this fixture are all smaller than one, so a direct `astype("int64")` truncates
every one of them to zero and the comparison degenerates into asserting that two
NaNs match. The test asserts the statistic is finite so that it cannot silently
become vacuous again.

Three checks that the assertions actually bite, each done by breaking one thing
and confirming the test fails: removing the guard entirely, removing only the
non-numeric clause, and removing the scaling. `pytest
skbio/stats/distance/tests/test_permdisp.py` passes, 38 tests, with and without
`NUMBA_DISABLE_JIT=1`, on arm64 macOS and on x86-64 Linux. `ruff check` is clean
on both files.

No changelog entry: `engine="numba"` for `permdisp` landed in #2547 and has
not been released, so this corrects unreleased behavior rather than something a
user could have depended on.

### DIFF
--- a/skbio/stats/distance/_permdisp.py
+++ b/skbio/stats/distance/_permdisp.py
@@ -376,6 +376,9 @@ def permdisp(
         If, when using the spatial median test, the pcoa ordination is not of
         type np.float32 or np.float64, the spatial median function will fail
         and the centroid test should be used instead
+    TypeError
+        If, when using the centroid test, the pcoa ordination is not of a
+        numeric type.
     ValueError
         If the test is not centroid or median, or if method is not eigh or fsvd.
     TypeError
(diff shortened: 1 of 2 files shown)


sfiligoi (approved):
Looks good to me

sfiligoi:
The doc failure is due to external issue... I am considering it passing the CI.

PR #2570, opened 2026-09-04
PERF: faster BPTree construction and SkbioObject registration

monkeytim19:
## Summary

Three small, independent, output-preserving improvements to `BPTree`, stacked on the BP succinct-tree backend (#2498), the range-excess / 64-bit work (#2541), and newick I/O (#2528). The main changes include:

1. Built the `select` index in O(n) instead of O(n log n)
2. Implemented a single compiled traversal for `TreeNode.from_bptree`
3. Registered `BPTree` as a `SkbioObject`

No query hot path is touched, and the two perf changes are bit-/byte-identical to prior behavior.

## Performance

Median across balanced / caterpillar / random topologies; per-commit isolated builds, min-of-N timing. Absolute times are machine-specific (macOS, CPython 3.10) — the relative change is the point.

| Change | Operation (100k tips) | Before → After | Speedup | Memory |
|---|---|---|---|---|
| O(n) `select` index build | `BPTree.from_treenode` | 137 → 128 ms | ~7% | unchanged¹ |
|  | `parse_newick` | 118 → 110 ms | ~7% | unchanged¹ |
| Compiled `TreeNode.from_bptree` | BP → `TreeNode` | 185 → 160 ms | ~13% | unchanged |
| `SkbioObject.register(BPTree)` | all query / nav / construction | — | Negligible | unchanged |

¹ bit-for-bit identical stored arrays; `BPTree.data` stays at 4 B/node. Speedups hold at 500k tips (select-index ~7–8%, `from_bptree` ~11%) and across all three topologies. The `SkbioObject` registration is import-time only: 99% of measured op-cells fall within ±10% of baseline (pure noise).

### 1. PERF: build the BPTree `select` index in O(n)
The select index (`_k_index_0` / `_k_index_1`, the position of the k-th close/open parenthesis) was built with `np.unique(np.cumsum(...))` — an O(n log n) sort over two cumulative-rank arrays. It is equivalently the set of positions where each bit's running count increments, i.e. a single O(n) `np.flatnonzero` pass. The new build is **bit-for-bit identical** to the `np.unique` result (verified across topologies, including the degenerate `B[0] == 0` case) and leaves the stored arrays — and thus the memory footprint — unchanged. The index build itself drops ~10× (≈9.5 ms → 0.95 ms at 100k tips); end to end this is the ~7% `from_treenode` / `parse_newick` speedup above.

### 2. PERF: single compiled traversal for `TreeNode.from_bptree`
`TreeNode.from_bptree` walked the balanced-parentheses tree from Python, crossing the Python/C boundary once per accessor per node. This moves the traversal into one compiled pass (`BPTree._to_node_arrays`, returning per-node name/length/edge/parent arrays in preorder); `from_bptree` is reduced to pure-Python `TreeNode` assembly over those arrays. `_bp.pxd` and the C vtable are unchanged (a single new `def`). Output is **byte-identical**; the irreducible `TreeNode` object-creation cost dominates at the largest sizes.

### 3. ENH: register BPTree as a SkbioObject
`BPTree` was the only user-facing tree class not in the `SkbioObject` hierarchy. A Cython `cdef class` cannot subclass a pure-Python ABC (`first base of 'BPTree' is not an extension type`), and a pure-Python wrapper would add a ~34 ns/call dispatch tax to every operation. Instead this uses **virtual registration** — `SkbioObject.register(BPTree)` plus the required `__str__` — so `isinstance(bp, SkbioObject)` is `True` and `BPTree` participates like `TreeNode`, while staying **fully compiled at zero runtime/memory cost** (import-time only).

## Correctness

All BPTree tests pass.

## Checklist


monkeytim19:
@qiyunzhu @mortonjt

qiyunzhu (commented):
I am bit wary of the approach of virtual registration of a Cython cdef class, as this mixes Python interface and Cython core. My envisioned structure is a Python class `BPTree` with interfacing tasks plus a Cython class `_BPTree` with heavy lifting tasks. However, you noted, "a pure-Python wrapper would add a ~34 ns/call dispatch tax to every operation." Let's discuss on this performance regression and its meaning to our design.

qiyunzhu:
I discussed with @monkeytim19 . We decided to merge it and move on. Future tasks will involve 1) downstream applications (like phylogenetic diversity), 2) an alternative Numba kernel, which will require re-designing of a dispatcher, 3) Numba GPU support.

Issue #1470, opened 2016-12-12
Feature request: Non-fatal error for reading corrupted fasta records (skbio.io.read)


jolespin:
I encountered an error (below) reading in corrupted UNITE fasta files (fungal ITS sequence database).  Most of the sequences were fine but a few of them had characters outside of the DNA alphabet, when a record contained a bad character, the parser threw a fatal error.

It could be useful to have an error message giving the header of the corrupted entry and then continue reading the file (i.e. a non-fatal error).  

The sequences I was using were from: https://unite.ut.ee/sh_files/UNITE_public_22.08.2016.fasta.zip
I gave them a list of which ones were corrupted and I think they fixed it in the newer release. 
When I first encountered this error, it took me a bit to figure out which entries were corrupted. 

```
path = "/Users/Mu/Downloads/UNITE_public_22.08.2016.fasta"
>>> for seq_record in skbio.io.read(path, constructor=skbio.DNA, format="fasta", lowercase=True):
...     pass
...
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/Mu/anaconda/lib/python3.5/site-packages/skbio/io/registry.py", line 506, in <genexpr>
    return (x for x in itertools.chain([next(gen)], gen))
  File "/Users/Mu/anaconda/lib/python3.5/site-packages/skbio/io/registry.py", line 531, in _read_gen
    yield from reader(file, **kwargs)
  File "/Users/Mu/anaconda/lib/python3.5/site-packages/skbio/io/registry.py", line 1008, in wrapped_reader
    yield from reader_function(fhs[-1], **kwargs)
  File "/Users/Mu/anaconda/lib/python3.5/site-packages/skbio/io/format/fasta.py", line 677, in _fasta_to_generator
    **kwargs)
  File "/Users/Mu/anaconda/lib/python3.5/site-packages/skbio/sequence/_grammared_sequence.py", line 336, in __init__
    self._validate()
  File "/Users/Mu/anaconda/lib/python3.5/site-packages/skbio/sequence/_grammared_sequence.py", line 360, in _validate
    list(self.alphabet)))
ValueError: Invalid character in sequence: b'U'.
Valid characters: ['R', 'H', 'V', 'B', 'N', 'W', 'S', 'C', 'M', '.', 'T', 'K', '-', 'D', 'G', 'Y', 'A']
Note: Use `lowercase` if your sequence contains lowercase characters not in the sequence's alphabet.
```

[...]

PR #2573, opened 2026-09-08
ENH: Name the offending record when a FASTA record fails to parse

jissen706:

- [x] I have documented all public-facing changes in the [changelog](https://github.com/scikit-bio/scikit-bio/blob/main/CHANGELOG.md).
- [ ] **This pull request includes code, documentation, or other content derived from external source(s).**

Addresses #1470.

## Description

As reported in #1470: when one record among many in a FASTA file fails validation (e.g., a character outside the sequence's alphabet), the raised `ValueError` gives no indication of *which* record was at fault:

```python
>>> import skbio
>>> for rec in skbio.io.read('big_file.fasta', constructor=skbio.DNA, format='fasta'):
...     pass
ValueError: Invalid character in sequence: b'U'.
Valid characters: [...]
```

In a file with many thousands of records, tracking down the bad one meant bisecting the file by hand. The issue thread has two suggestions: make it non-fatal by default (first commenter pushed back — "should stay a fatal error by default"), or keep it fatal but name the record in the message (second commenter's ask, unopposed). This PR does the latter, since it's the point of actual agreement in the thread and doesn't require introducing a new `errors=`-style API to decide on.

`_fasta_to_generator` now catches `ValueError` raised while constructing each record (in both the plain-FASTA and FASTA+QUAL code paths) and re-raises it with the record's ID prefixed, chaining the original exception via `from e` so the full underlying message is still there:

```
ValueError: Could not parse the FASTA record with ID 'bad_record': Invalid character in sequence: b'U'.
Valid characters: [...]
```

This is the shared reader used by all FASTA-backed constructors (`Sequence`, `DNA`, `RNA`, `Protein`, `TabularMSA`), so the fix applies uniformly.

Updated one existing test (`test_keep_spaces`) whose assertion depended on the old un-prefixed message starting with "Invalid char" — changed to an `assertIn` check, since the underlying validation message is unchanged, just no longer at the start of the string. Added two new regression tests covering both code paths (with and without a QUAL file).


qiyunzhu:
@jissen706 Thanks for addressing this 10-year-old feature request. As I eyeball the code, this solution may not be perfect -- scikit-bio's FASTA parser accepts sequences without an ID. Thus you could get an error message like `"Could not parse the FASTA record with ID ''.` But this is still better than nothing, and the cases of ID-less FASTA are very rare. So I think this PR is good to go as-is.

PR #2574, opened 2026-09-09
BUG: return the coincident point from geomedian_axis_one

LarytheLord:
### Description

Follow-up to #2547, and the separate `geomedian_axis_one` fix discussed there.

When every point handed to the geometric median sits on the same spot, every
inverse distance is zero, so their total is zero and the loop that normalizes
the weights by that total divides by zero. The test that would have stopped the
iteration, `nzeros == n`, sits after that division rather than before it, so the
kernel raises instead of returning the point:

```python
Dinvs = _sum(Dinv)
for i in range(n):
    W[i] = Dinv[i] / Dinvs     # ZeroDivisionError when every D[i] is zero
...
if nzeros == 0:
    y1 = T
elif nzeros == n:              # too late
    break
```

This is reachable from `permdisp` with `test="median"` on a distance matrix
holding mutually identical samples, which is what technical replicates look
like:

```
permdisp(dm, grouping, test="median", dimensions=3)
  engine="cython"  ZeroDivisionError
  engine="numba"   0.9999999999999999
```

The fix moves the `nzeros` count and its test above the division. Nothing else
changes: `Dinvs` and `W` are simply not computed on the iteration that breaks,
and neither is read afterwards.

The Numba port added in #2547 already ordered the test this way, which is why it
returned an answer where Cython raised. After this the two engines agree on that
input, to 1.1e-16 with the same p-value.

One piece of context worth recording. Upstream hdmedians, which this kernel was
ported from, begins with `# cython: cdivision=True`, so the same division there
yields a NaN rather than raising, and the NaN happens never to be read. The
vendored copy carries the per-function `boundscheck` and `wraparound`
decorators but not `cdivision`, which is the whole reason the behavior differs.
Reordering seemed better than adding the directive: it makes the degenerate case
explicit instead of leaving correctness resting on a NaN that nothing reads.

### Testing

Three new kernel-level and end-to-end tests: coincident input at several shapes,
the same on the float32 specialization since the fused kernel compiles a
separate branch for it, and a `permdisp` call on a matrix with identical samples
in each group. A fourth checks that the two `permdisp` engines now agree on that
input. All four fail against the unpatched kernel.

The change was also checked by compiling the original and patched kernel side by
side as two extension modules and comparing them directly: 400 of 400
randomized non-degenerate inputs bit-identical, across shapes, both dtypes and
three magnitude scales, with partial coincidence and the `n == 1` early return
unchanged. Run on arm64 macOS and again on x86-64 Linux.

`pytest skbio/stats/distance/tests/test_permdisp.py` passes. The wider
`skbio/stats/distance/` and `skbio/stats/ordination/` runs match the baseline on
an unpatched checkout.


sfiligoi (approved):
Looks good to me

qiyunzhu (commented):
Looks to me a clean fix. Good job!

PR #2575, opened 2026-09-09
ENH: add engine="fast" to the compute engine selector

LarytheLord:
### Description

The `engine` default is conservative, which is the right default but means a
user only benefits by naming an engine and by knowing which one is quicker for
the function they are calling. `engine="fast"` moves that decision into the
library: pick whichever engine is expected to be quickest here, with no promise
beyond that.

`_resolve_engine` gains an optional `fast` argument saying what `"fast"` means
for the calling function, since that depends on the function and on what is
installed:

```python
engine = _resolve_engine(
    engine,
    ("cython", "numba"),
    fast="numba" if NUMBA_AVAILABLE else "cython",
)
```

`permanova`, `mantel` and `permdisp` pass it. Falling back to `"cython"` when
Numba is not importable means the option degrades on a plain install rather
than raising. A function that does not pass `fast` resolves `"fast"` to the
conservative default engine, so it is a no-op there rather than an error.

One detail worth flagging about where the branch sits. It comes after the
global default is read rather than before it, and falls back to a fixed
constant rather than re-reading the option:

```python
if engine is None:
    engine = get_config("engine")
if engine == "fast":
    engine = fast if fast is not None else _DEFAULT_ENGINE
```

`"fast"` stays a per-call option: `set_config` validates the engine option
against `["cython", "numba"]` and is untouched here, so the global default
cannot be `"fast"`. This shape just means a single branch handles `"fast"`
wherever it came from, and that a function offering nothing faster degrades
instead of raising. Nothing has to move if that whitelist is ever widened.

The default is unchanged and remains `"cython"`. `"fast"` has to be asked for,
and the docstrings say plainly that its result may differ from the default in
the last bits.

One consequence worth naming. Because the branch lives in `_resolve_engine`,
every function that calls it now accepts `"fast"`, including the two I did not
wire: `beta_diversity` and `center_distance_matrix` take it as a no-op and
resolve to `"cython"`. Note that is the conservative default rather than the
global option, so with `set_config("engine", "numba")` in force those two still
get Cython for `engine="fast"`, which is the safe direction but worth knowing.
Their docstrings also still list only `{"cython", "numba"}`. Say the word if you
would rather I documented it there too, or wired them properly in the same PR.

### Testing

`_resolve_engine` carries the resolution tests, in `skbio/tests/test_config.py`:
`"fast"` taking the caller's target, falling through to the conservative
default when no target is given, being resolved after the global lookup, still
being checked against `supported`, and staying a no-op when the default is
itself `"fast"`. A further test pins the other half of the contract, that
`set_config("engine", "fast")` keeps raising, so the per-call-only rule is not
something a later change can drop silently. Breaking the behavior five
different ways kills them:

| mutation | tests that fail |
|---|---|
| remove the `"fast"` branch | all five |
| ignore the caller's target, always use the default | two |
| move the branch before the global lookup | two |
| fall back to the option instead of the fixed constant | one |
| add `"fast"` to the `set_config` whitelist | one |

Those run with Numba absent, which is the environment the default test job
uses. The one case that cannot, `fast="numba"` resolving to `"numba"`, is a
separate `numba_code` test so the Numba job still covers it. Proving the
caller's target is honored without Numba is done with a target the resolver
could not have produced by itself: `fast="julia"` has to surface as
`engine='julia' is not supported`.

At the call sites there are two tests each, one per branch of the `fast=`
argument. The `numba_code` one checks that `engine="fast"` is accepted and
returns the same answer as `engine="numba"`, and says only that: it does not
try to prove which engine ran, because `permanova` and `permdisp` reduce in
parallel so their last bits move with thread state, and `mantel`'s two engines
agree bit for bit at every size tried.

The other runs only where Numba is absent, gated on `NUMBA_AVAILABLE`, and
compares `engine="fast"` against `engine="cython"`. That one is an exact
equality rather than a tolerance, because without Numba `"fast"` resolves to
`"cython"` and runs the identical Cython call; only one engine is involved at
all. It is also the only thing covering the `fast=` argument in that lane,
since the `numba_code` tests skip there. Hardcoding `fast="numba"` at any of
the three call sites, so that the conditional is ignored, fails the
corresponding test with the `ImportError` a user without Numba would hit.

Run on arm64 macOS and x86-64 Linux, with and without Numba, using the same
`python -m skbio.test` the CI job runs and comparing against a clean checkout
of `main` in the same session. The branch adds 13 tests and no failures: on
x86-64 without Numba, `main` gives 3855 passed and 143 skipped against the
branch's 3864 and 147, the four extra skips being the `numba_code` tests that
the Numba job picks up instead. With Numba present it goes from 3904 passed
and 94 skipped to 3914 and 97, the three extra skips being the no-Numba tests
stepping aside. The `pytest -m numba` job goes from 49 to 53. No new failure
in any of them.


skbio/tests/test_config.py
```diff
@@ -82,6 +93,66 @@ def fake_import(name, *args, **kwargs):
             with self.assertRaisesRegex(ImportError, "requires the optional numba"):
                 _resolve_engine("numba", ("cython", "numba"))
 
+    def test_fast_resolves_to_what_the_caller_names(self):
+        # A target the resolver could not have arrived at on its own shows that
+        # the caller's value is what gets used, and needs no optional
+        # dependency to check.
+        with self.assertRaisesRegex(ValueError, "engine='julia' is not supported"):
```
sfiligoi, line 100: Why "julia"?

"julia" does not seem nonsensical to me.
Indeed, it is an existing programming language!

Please change it to something really unlikely.
E.g. "SantaGoesSkiing" ?

LarytheLord: Changing to "SantaGoesSkiing" and agreed. The value is doing sentinel duty here rather than standing in for a plausible engine: it is only there to show that the caller's `fast=` argument is what gets used, so it reads better if nobody can mistake it for something scikit-bio might one day support. Comment updated to say that.

For completeness on what I did not touch: "julia" is still the generic unsupported-engine value in the two older tests in this file and in test_permanova, test_mantel, test_permdisp and test_unifrac. Those only need a value that is not in the supported list, so I left them rather than widen this PR.

skbio/stats/distance/tests/test_mantel.py
```diff
@@ -243,6 +243,19 @@ def test_pearson_flat_engine_numba_matches_cython(self):
             self.assertAlmostEqual(obs[1], exp[1])
             npt.assert_allclose(obs[2], exp[2])
 
+    @numba_code
+    def test_engine_fast_is_accepted(self):
```
sfiligoi, line 247: Shouldn't we have the same test for cython, too?
I.e., for the tests where Numba is not present?

Same for the other tests like this.

sfiligoi (approved):
Looks good to me

PR #2578, opened 2026-09-10
PERF: fit dirmult_lme's random-intercept model in closed form for all features at once

l1joseph:
Hi Qiyun, I spent the long weekend working on this. After looking over the codebase and with the help of Claude, I found out that with `re_formula`/`vc_formula` unset (the default), `dirmult_lme`'s per-feature `MixedLM` fit reduces to a 1-D REML optimization over one variance ratio. This allows us to compute the profile likelihood in closed form for all features of a draw at once, instead of running `MixedLM.fit` per feature. The fallback is to the old per-feature path for any `re_formula`, `vc_formula`, `model_kwargs`, `fit_method`, `fit_kwargs`, or `fit_converge` setting. The new math lives in this new private module, `skbio/stats/composition/_lme.py`.

Benchmark, full `dirmult_lme` call, `draws=128`:

| samples | features | groups | before | after | speedup |
|---|---|---|---|---|---|
| 100 | 50 | 20 | 400.8s | 2.199s | 182x |
| 200 | 300 | 40 | 6137.6s | 11.530s | 532x |

Correctness:
- Matches `MixedLM`'s own `get_fe_params`/`get_scale`/`hessian`/`loglike` at matched parameters, worst relative error ~1e-13
- Matches a dense brute-force scan of the same profile likelihood across 1600 fits, worst disagreement 4.3e-8 log-likelihood units
- Against 1440 randomized real `MixedLM.fit` runs, `MixedLM` itself only converged 71-91% of the time depending on balance. Where both converged, agreement was 1e-6 to 1e-8. Where `MixedLM` under-converged, this implementation reached a higher log-likelihood in 100% of those cases
- End to end against a reference harness proven bit-identical to the released code: 5 realistic configs, 100% agreement on which features are flagged significant

Would be extremely helpful to get your advice on these two things:

1. Pinned output values shift in the 3rd-4th decimal. Every disagreeing draw is one where `MixedLM` stopped at a lower likelihood than this fit reaches.
2. Currently `Reps`/`NaN` accounting is set up to be more granular on boundary fits. The old code counted a failed draw toward `Reps` and NaN'd the whole feature; the new code excludes only the failed draws (`Reps=128, pvalue=NaN` becomes `Reps=41, pvalue=0.211` in one traced case).

In regards to #2409, my changes don't affect 2409's approach to the general non-random-intercept case.

I also noticed that statsmodels' `holm` correction hardcodes a `gc.collect()` per feature, and it uses up 53s at 2000 features, and affects both `dirmult_lme` and `dirmult_ttest`. Would it be useful to address that by vectorizing the correction across all features in one call instead of one call per feature, as its own PR?

Tests:
- matches `MixedLM`'s own internal functions at matched parameters
- matches a dense brute-force scan of the profile likelihood
- matches or exceeds `MixedLM.fit` across randomized designs
- fallback path is bit-identical to the old implementation
- edge cases: singleton groups, all-singleton groups, one dominant group, two groups, no-intercept formulas, zero counts

Checklist:


qiyunzhu:
@l1joseph Congratulations!! I think this is a really tough task. Finding a closed form solution to replace all mixed LM runs is a great leap, and it makes the entire function useful (previously it was just too slow). I know it isn't easy to get the faster solution that ~~match~~ beats the slow original. In a separate thread I was working on a similar task (#2572). Even with the help of AI, this wasn't easy.

Regarding `holm`: My AI friend suggested the same thing, and I have got it fixed. Now you can use this [`_adjust_pvalues`](https://github.com/scikit-bio/scikit-bio/blob/a76573f1129bc53c04bff239b8f64416cfee0fdc/skbio/stats/composition/_utils.py#L111) helper function, which basically re-implements Holm and other p-value correction methods from scratch, in a way that is not only faster than statsmodels, but also optimized for many runs. This already made `dirmult_ttest` even faster (from the much improved baseline you recently did). Please merge upstream main and resolve conflicts and you will get this code.

Note that this `_adjust_pvalues` can be further optimized by vectorizing over small batches of runs. But my AI friend and I discussed back-n-forth and we eventually decided that it is not worth for now.

l1joseph:
Re-ran the benchmark from the PR description on the current committed code, isolated node, same configs, single run for the old path, 5 runs for the new path's large config to check stability:

| samples | features | groups | old | new | speedup |
|---|---|---|---|---|---|
| 100 | 50 | 20 | 384.3s | 2.063s | 186x |
| 200 | 300 | 40 | 6497.2s | ~2.0s (median of 5, range 1.92-2.54s) | ~3200x |

The large-config number improved on the original 532x claim because this branch already picked up `_adjust_pvalues` from your PR, which removes statsmodels' per-feature `gc.collect()` from the holm correction.

qiyunzhu:
@l1joseph Thanks! This looks very cool. I see that the new math can recover more fitted features than the previous version. This is a behavior change, but I think it is a good one and it fixes suboptimal math. Meanwhile, I am amazed that the phrase "closed form" is so frequently used over the past few weeks. The current PR is way more "analytical" than the previous per-feature per-draw optimization, but it currently still has numerical optimization ("Batched golden-section search on log theta"), though in a much controllable manner. Overall I think it is a great improvement. Please sync with main and refine the changelog item (I just made them concise).

skbio/stats/composition/_lme.py
```diff
@@ -0,0 +1,351 @@
...
+
+from typing import NamedTuple
+
+import numpy as np
+
+# Search bounds and resolution for the maximization, in log(theta).
+_LOG_LO = -20.0
+_LOG_HI = 20.0
+_N_GRID = 65
+_N_REFINE = 45
```
qiyunzhu, line 47: How is this number of iterations determined? Following an established source? I am thinking that if we specify an early stopping criterion, e.g., `tol=1e-6`, in addition to `max_iter=45`, this may reduce unnecessary iterations and further improve performance. But this suggestion isn't a priority.

qiyunzhu (approved):

PR #2580, opened 2026-09-17
BUG: fix `to_indices` and `to_cigar` on single-segment alignment paths

### LINKED ISSUE #2579: BUG: `AlignPath.to_indices` and `PairAlignPath.to_cigar(seqs)` crash on a path with one segment
If two sequences align with no gaps, `pair_align` returns a path with a single segment. Calling `to_indices()` or `to_cigar(seqs=...)` on that path raises instead of returning a result.

```python
from skbio.alignment import pair_align
from skbio.sequence import DNA

seqs = [DNA("ACGT"), DNA("ACTT")]
path = pair_align(*seqs).paths[0]
print(path)
path.to_indices()
path.to_cigar(seqs=seqs)
```

```
<PairAlignPath, positions: 4, segments: 1, CIGAR: '4M'>
numpy.exceptions.AxisError: axis 1 is out of bounds for array of dimension 1
TypeError: iteration over a 0-d array
```

Expected, going by the docstrings (`to_indices` returns shape `(n_sequences, n_positions)`, `to_cigar` with `seqs` returns `=`/`X` codes):

```
[[0 1 2 3]
 [0 1 2 3]]
2=1X1=
```

The same paths work fine as soon as there are two or more segments. `to_cigar()` without `seqs` also works.

`to_indices()` fails the same way on a path with a single sequence, whatever the number of segments:

```python
from skbio.alignment import AlignPath
AlignPath(lengths=[3, 1, 2], states=[0, 1, 0], starts=[0]).to_indices()
# AxisError: axis 1 is out of bounds for array of dimension 1
```

**Cause**

Both methods call `np.squeeze` on an array that can have a size-1 axis that matters:

- `to_indices`: `bits = np.squeeze(self._to_bits())`. `_to_bits()` is already 2-D `(n_sequences, n_segments)`. With one segment (or one sequence) squeeze makes it 1-D, then `np.repeat(..., axis=1)` fails.
- `to_cigar`: `states = np.squeeze(self._states)`. `_states` is `(1, n_segments)` for a pairwise path. With one segment squeeze makes it 0-d, then `zip(self._lengths, states)` fails.

**Versions**

Reproduces on `main` (0ea899c8, 0.7.4-dev) and on the 0.7.3 release from PyPI. Python 3.13.13, numpy 2.5.3, macOS 26.6.2 (arm64).

I have a small fix with tests ready (drop the squeeze in `to_indices`, take `self._states[0]` in `to_cigar`). Happy to open a PR if this looks right to you.

haeganm:
Fixes #2579

### Description

A gap-free alignment is a path with a single segment, which is what `pair_align` returns for e.g. `ACGT` vs `ACTT`. On such a path `AlignPath.to_indices()` raised `AxisError` and `PairAlignPath.to_cigar(seqs=...)` raised `TypeError: iteration over a 0-d array`. `to_indices()` also raised on any path with a single sequence.

Both come from `np.squeeze` removing an axis that is size 1 but still meaningful:

- `to_indices`: `_to_bits()` already returns `(n_sequences, n_segments)`, so the squeeze was not needed. With one segment or one sequence it made the array 1-D and `np.repeat(..., axis=1)` failed. Now it uses `_to_bits()` directly.
- `to_cigar`: `_states` is `(1, n_segments)` for a pairwise path. With one segment squeeze made it 0-d and the `zip` over it failed. Now it takes `self._states[0]`.

No change in output for paths that already worked.

### Tests

- Added `test_to_indices_single_segment` and `test_to_cigar_single_segment`: one-segment paths (match and gap), non-zero starts, all three `gap` modes, a single-sequence path, and paths coming out of `pair_align`.
- Both new tests fail on `main` (`AxisError` / `TypeError`) and pass with the fix.
- As a sanity check I compared `to_indices()` and `to_cigar(seqs)` against a brute-force reference on 5000 random pairwise paths with 1 to 4 segments: 0 mismatches with the fix, 1293 on `main` (the one-segment paths).
- Full suite via `python -m skbio.test`: 3881 passed, 132 skipped.
- `ruff check` and `ruff format --check` (0.15.14) pass on `_path.py`.

### Checklist

* [ ] **This pull request includes code, documentation, or other content derived from external source(s).**


qiyunzhu:
@haeganm Thanks! This is a clean, well-tested and well documented fix. I will merge when tests pass.

PR #2581, opened 2026-09-17
ENH: accept engine="fast" for the unifrac beta_diversity metrics

LarytheLord:
## What this does

`beta_diversity`'s `unweighted_unifrac` and `weighted_unifrac` already accepted `engine=`, but `engine="fast"` fell through to the conservative default, because the two call sites never passed `fast=` to `_resolve_engine`. They now resolve it the same way `permanova`, `mantel` and `permdisp` have since #2575: Numba when Numba is installed, Cython otherwise. The default is unchanged.

## What it does not do

No GPU work here, and no new kernels. This only decides which of the two existing engines `"fast"` picks.

It also leaves one call site alone on purpose. `center_distance_matrix` in `skbio/stats/ordination/_utils.py:226` is now the only `_resolve_engine` call in the codebase without a `fast=`, and I measured it rather than assuming: Numba is only 1.05x to 1.12x quicker than Cython there once warm, against a fixed compilation cost, which puts break-even somewhere between 96 and 933 calls depending on size. Wiring `"fast"` to Numba there would make a single small call slower, which is the opposite of what `"fast"` is supposed to promise. Doing it properly needs a size threshold with the crossover measured, and that belongs in its own PR. Happy to open an issue for it if you would rather it were tracked.

## Why Numba is the right thing for "fast" to pick here

Measured on an x86 Linux node (AMD HPC Fund, 16 Numba threads), balanced rooted tree, sparse Poisson count table. Warm times, best of 3 to 5 runs. The speedups are computed from the unrounded timings, so recomputing them from the rounded millisecond columns below will be off by a fraction of a percent:

| metric | samples | taxa | pairs | cython | numba | speedup |
|---|---|---|---|---|---|---|
| unweighted_unifrac | 100 | 500 | 4,950 | 61.7 ms | 4.1 ms | 15.1x |
| unweighted_unifrac | 250 | 1,000 | 31,125 | 578.8 ms | 20.1 ms | 28.8x |
| unweighted_unifrac | 500 | 2,500 | 124,750 | 7.27 s | 161.6 ms | 45.0x |
| unweighted_unifrac | 1,000 | 5,000 | 499,500 | 73.6 s | 1.24 s | 59.5x |
| weighted_unifrac | 100 | 500 | 4,950 | 108.5 ms | 3.6 ms | 29.9x |
| weighted_unifrac | 250 | 1,000 | 31,125 | 968.5 ms | 13.9 ms | 69.8x |
| weighted_unifrac | 500 | 2,500 | 124,750 | 20.6 s | 102.3 ms | 201.1x |
| weighted_unifrac | 1,000 | 5,000 | 499,500 | 112.0 s | 673.6 ms | 166.2x |
| unweighted_unifrac | 2,000 | 10,000 | 1,999,000 | 615.0 s | 8.92 s | 69.0x |
| weighted_unifrac | 2,000 | 10,000 | 1,999,000 | 907.0 s | 4.42 s | 205.2x |

Numba wins at every size and on both metrics. The speedup grows with size for `unweighted_unifrac`. For `weighted_unifrac` it is not monotonic, 201x at 500 samples then 166x at 1,000 then 205x at 2,000, so the trend is upward but I am not claiming it simply widens with the table. The 2,000-sample point is the largest I could measure at all: Cython takes 10 to 15 minutes for a single call there, so a size beyond it is not practical to time on both arms.

The cost is the one-off compilation, and it is worth being concrete since it is the reason `"fast"` is opt-in rather than the default. Measured separately, one fresh process per case, so the first call really does include the compilation:

| metric | samples | taxa | cython first call | numba first call | |
|---|---|---|---|---|---|
| unweighted_unifrac | 100 | 500 | 114.1 ms | 3.62 s | numba 31.8x slower |
| unweighted_unifrac | 1,000 | 5,000 | 69.3 s | 4.38 s | numba 15.8x faster |
| weighted_unifrac | 100 | 500 | 155.6 ms | 3.33 s | numba 21.4x slower |
| weighted_unifrac | 1,000 | 5,000 | 125.4 s | 3.60 s | numba 34.8x faster |

So on a small table the first call is genuinely worse, by about 32x, and it takes roughly 61 further calls at that size to pay the compilation back, or about 30 for `weighted_unifrac`. On a 1,000-sample table the argument is over on the first call: compiling and running still beats Cython by 16x to 35x. That is the whole reason the docstring says a single small run can be slower, and the reason this is wired to `"fast"` rather than made the default.

## Results are not bit-identical between the engines

The docstring carries the same caveat the three call sites in #2575 use, that results may differ from the default in the last bits, and it is not hypothetical here. Measured on the same node, Cython and Numba disagree on 84,444 of 90,000 cells at 300 samples by 1,500 taxa, with a maximum absolute difference of 3.0e-15 for `unweighted_unifrac` and 2.0e-14 for `weighted_unifrac`. That is ordinary floating-point reassociation from the parallel pass, not a correctness problem, but it is why `"fast"` is opt-in and why the existing engine tests compare with a tolerance rather than for equality.

## The tests

Five, doing three different jobs.

`test_unifrac_fast_resolves_numba` and `test_unifrac_fast_resolves_cython` assert the resolution directly. Comparing results cannot show which engine `"fast"` picked, because the two unifrac engines agree well within any tolerance a test could use, so the decision is a named module constant and these assert it per lane. That is also why I hoisted it rather than leaving the conditional inline: it makes the decision testable.

`test_unifrac_engine_fast_is_accepted` and `test_unifrac_engine_fast_is_cython_without_numba` check that `"fast"` is plumbed through end to end and gives the right answer, in both lanes. The second is exact rather than a tolerance, since without Numba `"fast"` runs the identical Cython call.

`test_beta_diversity_engine_fast_with_pairwise_func_is_quiet` covers the case where a `pairwise_func` stops the Numba kernels being usable. `engine="numba"` warns there, because the caller asked for Numba and did not get it. `"fast"` stays quiet, because the caller asked scikit-bio to choose and nothing they asked for was overridden, which is also what `engine="cython"` and the default already do. That behavior was undocumented and untested before this. It has no no-Numba counterpart and the comment explains why: without Numba, `"fast"` resolves to Cython, which cannot reach the warning at all.

## One thing I noticed and did not change

`_numba_unifrac_fast_path_eligible` is given the raw `engine` argument rather than the resolved one, so its warning only fires for a literal `engine="numba"`. That also means `engine=None` with a global Numba default falls back silently today, which predates this PR. I left it as it is, since changing it would alter behavior for the default case in a PR that is only meant to wire `"fast"`. Say the word if you would rather I fixed it here or in a follow-up.

## Two small structural choices

Both call sites want the same answer for `"fast"`, so rather than repeating the conditional at each one I compute it once as a module-level `_UNIFRAC_FAST_ENGINE`. The three call sites in #2575 write the conditional inline, but they have one occurrence per file; this is the first place it would repeat within a file, which is why it differs.

`NUMBA_AVAILABLE` is imported from `skbio/diversity/beta/_unifrac.py` rather than recomputed with a local `try: import numba`. The three siblings each define their own, but `_driver.py` already imports the Numba unifrac kernels from that module, so reusing its flag keeps one source of truth instead of a second try/except that could drift from it.

## CHANGELOG

Not in this commit. The entry has to cite this PR's own number, so I will add it as a follow-up commit once the number exists, the same way #2575, #2547 and #2511 each did.

## Verification

Run on an x86 Linux node (AMD HPC Fund), not on my laptop, as a single job with nothing else touching the tree. The files were checksummed against my local copies first.

- `python -m skbio.test`, the full suite: **3917 passed, 99 skipped, 0 failed.**
- `pytest -m numba` over `skbio/diversity`: 12 passed.
- unifrac plus `test_config`, Numba present: 59 passed, 2 skipped.
- The same two files in the no-Numba lane, reproduced with an import blocker: 48 passed, 13 skipped.
- Mutation tested, five mutations, four killed. Forcing `"fast"` to Cython, inverting the condition, and flipping the no-Numba branch to Numba are each caught by the two resolution tests. Making `"fast"` warn is caught by the `pairwise_func` test. The harness aborts if a mutation's edit does not actually change the file, and the source is restored and diffed after each one.
- The one that survives is removing the `fast=` argument from both `_resolve_engine` calls while leaving the constant in place. Nothing catches that, because with `"fast"` falling back to the default the results are identical to Cython's, and Cython and Numba are already numerically indistinguishable here. I could only close it with a mock, which this project avoids, so I am flagging it rather than papering over it.


qiyunzhu:
@LarytheLord Great news. Thanks! Just one thought: I know AI's love to create verbose unit test function names, but `test_unifrac_fast_resolves_to_numba_when_numba_is_installed` etc. are just way too long...

LarytheLord:
Shortened, thanks. for the others which are already merged I will short them in next pr

qiyunzhu:
Thanks! Let me merge.

